PR #2, opened 2015-07-11
Simplify code

JASON0916:
**The change is mainly about Bot._requestUrl(self, url, method, data=None).** In sendPhoto(), sendAudio() and sendVideo i found that **data.keys()[1] represents the file type**, so i tried to simply the code.
For Chinese government's sake i can't access telegram, which disables my test. I'm sorry if my code brings you trouble.

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -561,36 +561,16 @@ class Bot(object):
         """
 
         if method == 'POST':
-            if 'photo' in data and isinstance(data['photo'], file):
+            file_types = ['photo', 'audio', 'document']
+            # In sendVideo(), sendAudio() and sendPhoto(), data.keys()[1] represents the file type
+            if data.keys()[1] in file_types and isinstance(data.items()[1], file):
                 try:
-                    photo = data.pop('photo')
+                    file_pop = data.pop(data.keys()[1])
 
                     return requests.post(
                         url,
                         data=data,
-                        files={'photo': photo}
-                    )
-                except requests.RequestException as e:
-                    raise TelegramError(str(e))
-            if 'audio' in data and isinstance(data['audio'], file):
-                try:
-                    audio = data.pop('audio')
-
-                    return requests.post(
-                        url,
-                        data=data,
-                        files={'audio': audio}
-                    )
-                except requests.RequestException as e:
-                    raise TelegramError(str(e))
-            if 'document' in data and isinstance(data['document'], file):
-                try:
-                    document = data.pop('document')
-
-                    return requests.post(
-                        url,
-                        data=data,
-                        files={'document': document}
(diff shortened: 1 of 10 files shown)


leandrotoledo:
Hey Jason, thanks for the pull request. I'll be testing/reviewing your code.
You mentioned about the Chinese firewall issue, how about a proxy feature? Would that help? If so, would you file a bug so we could start discussing more about it?

leandrotoledo:
Would you mind to push ignoring .idea folder though? Thanks!

JASON0916:
Actually i am going to buy a VPN latter. I noticed that there is a new version of your code,  i will work on the new version then, just ignore the pull request this time 0 -0. I'm sorry for the .idea, i will pay attention on that next time.

telegram/bot.py
```diff
@@ -561,36 +561,16 @@ def _requestUrl(self,
         """
 
         if method == 'POST':
-            if 'photo' in data and isinstance(data['photo'], file):
+            file_types = ['photo', 'audio', 'document']
+            # In sendVideo(), sendAudio() and sendPhoto(), data.keys()[1] represents the file type
+            if data.keys()[1] in file_types and isinstance(data.items()[1], file):
```
leandrotoledo: In case of data is None, data.keys() will raise an AttributeError.

telegram/bot.py
```diff
@@ -561,36 +561,16 @@ def _requestUrl(self,
         """
 
         if method == 'POST':
-            if 'photo' in data and isinstance(data['photo'], file):
+            file_types = ['photo', 'audio', 'document']
```
leandrotoledo: Missing 'video'

leandrotoledo:
No worries at all.
The way you simplified the code looks great! I made some comments and considerations. How about a is_inputfile function on InputFile to make the if even cleaner?

data['photo', 'video', ...] could be either:
- a file instance, like open('telegram.png', 'rb') or
- a str, like 'http://telegram.org/telegram.png (InputFile takes care to download and stream the content), as I'm writing this the HEAD tests if 'http' in data['video'], but I'm sure there are better ways to validate an URL.
if none of the options above, it falls on urlencode(data) simple post request.

Thanks!

JASON0916:
I found the discussion related on [discussion site](http://stackoverflow.com/questions/827557/how-do-you-validate-a-url-with-a-regular-expression-in-python). I think we may use urlparse in our case, even url.startwith('http') is OK for the flexible url rules. There seems to be little invalid url. How do you think about that?

leandrotoledo:
Uhm, after reading the post you sent, startswith seems to be enough for
now, we can handle URL errors when fetching content then.
Em 11/07/2015 2:31 AM, "CM" <notifications@github.com> escreveu:

PR #3, opened 2015-07-11
Simplify code

JASON0916:
Git is so hard for me. I add '.gitignore' and '.idea/' to .gitignore but it doesn't work, so i delete it directly orz. 
I found that most [send*() functions] like sendVideo(), sendsticker() ... are similar, can we simplify them by decorator ?


leandrotoledo:
Hey Jason, thanks for your pull request!
So, I've tested your code and sometimes it's crashing because dicts don't keep orders when insertions may occur (so data.keys()[1] won't be always what you want):

Testing sendAudio - File
> /home/leandrotoledo/workspace/python-telegram/telegram/bot.py(608)_requestUrl()
-> if len(data) >= 2 and data.keys()[1] in file_type and \
(Pdb) len(data)
2
(Pdb) data.keys()[1] in file_type
False
(Pdb) data.keys()
['audio', 'chat_id']
(Pdb) data.keys()[1]
'chat_id'

You can try something like:
    def _isfilerequest(self,
                       data=None):
        if data:
            file_types = ['audio', 'document', 'photo', 'video']
            file_type = [i for i in data.keys() if i in file_types]

            if file_type:
                file_content = data[file_type[0]]
                ...

        return False

leandrotoledo:
Great idea about decorators!

JASON0916:
Your method seems to be much better, i will try. I didn't imagine problem like that. T^T

leandrotoledo:
All tests passed right away! Yay. I'm merging your pull request.

Thank you very much! :)

PR #9, opened 2015-07-14
PEP8 related problems

JASON0916:
Hey, i used PyCharm to read your code and it reminded me of two problems.
**1st:**
    There are redundant parentheses in code **url = '%s/getMe' % (self.base_url)** and **return (url, data)**
**2nd:**
    Since message() and require_authentication() are in Class Bot but don't accept **self** as a       parameter, i think maybe we should make them **staticmethod**s?

Actually it doesn't matter a lot, just ignore this request if you want to keep your coding style.
And i have a problem about the **TelegramError**. Why don't we use **‘error_code’ field** to handle errors? Maybe we can update the code if the contents change.

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -65,13 +65,14 @@ class Bot(object):
           A telegram.User instance representing that bot if the
           credentials are valid, None otherwise.
         """
-        url = '%s/getMe' % (self.base_url)
+        url = '%s/getMe' % self.base_url
 
         json_data = self._requestUrl(url, 'GET')
         data = self._parseAndCheckTelegram(json_data)
 
         return User.de_json(data)
 
+    @staticmethod
     def message(func):
         """
         Returns:
@@ -102,6 +103,7 @@ class Bot(object):
             return Message.de_json(data)
         return wrap
 
+    @staticmethod
     def require_authentication(func):
         functools.wraps(func)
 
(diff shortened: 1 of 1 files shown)


leandrotoledo:
Hey Jason, that's an awesome work! Thank you for your pull request. Just one thing though:

message and require_authentication methods are decorators, you are right, they don't receive self as a parameter, but a function instead. But because decorators calls the function in an implicit way these methods can't be static.

See below:

  File "/usr/lib/python2.7/unittest/loader.py", line 254, in _find_tests
    module = self._get_module_from_name(name)
  File "/usr/lib/python2.7/unittest/loader.py", line 232, in _get_module_from_name
    __import__(name)
  File "/home/leandrotoledo/workspace/python-telegram-bot/tests/test_bot.py", line 5, in <module>
    import telegram
  File "/home/leandrotoledo/workspace/python-telegram-bot/telegram/__init__.py", line 28, in <module>
    from .bot import Bot
  File "/home/leandrotoledo/workspace/python-telegram-bot/telegram/bot.py", line 15, in <module>
    class Bot(object):
  File "/home/leandrotoledo/workspace/python-telegram-bot/telegram/bot.py", line 124, in Bot
    reply_markup=None):
TypeError: 'staticmethod' object is not callable

Everything else is working :) Just resend your pull request without staticmethods and I'll merge that.

JASON0916:
Oh, so that's how it is! You are right. I figured it out, thank you for the explanation.

leandrotoledo:
Still, if you want to send another pull request, just remove @staticmethod these functions. the redundant parentheses were a good catch.

JASON0916:
May be next time ^_^ .

PR #11, opened 2015-07-17
Base class

JASON0916:
For the reasons i said in [issue site](https://github.com/leandrotoledo/python-telegram-bot/issues/10#issuecomment-122149473), i made the base class without using **NotImplementedError**  orz...


leandrotoledo:
Testing...

PR #23, opened 2015-07-30
Bot._requestUrl: raise if method not in ('GET', 'POST')

wjt:
If this happens, there is a bug in the library. Better to raise the
error explicitly rather than return a value of a totally unexpected type
and crash later.


leandrotoledo:
Thank for the pull request, Will!

PR #24, opened 2015-08-06
Fix emojis on py3

franciscod:
The py2 string literals are interpreted as bytes on py3, so emojis on py3 aren't working.

Using `b'\xF0\x9F\x98\x81'` instead of `'\xF0\x9F\x98\x81'` fixes it: it now works in both py2 and py3 (tested with 2.7.10 and 3.4.3)


leandrotoledo:
Nice catch. Thank you!

franciscod:
seeing #80  and #85, i think we should revert this, and instead do something like 

```python
from __future__ import unicode_literals
```

or its alternatives from here http://python-future.org/unicode_literals.html

franciscod:
and we should really have unit tests on these things for py2 and py3

jh0ker:
from __future__ import unicode_literals
Well now I feel stupid about https://github.com/leandrotoledo/python-telegram-bot/pull/85 :grin:

franciscod:
Haha, it happens frequently; python has really clever solutions for some stuff :)

PR #43, opened 2015-08-20
confirm processed message in examples

rahiel:
You used that `if LAST_UPDATE_ID < update.update_id` because .getUpdates sends updates from and including `LAST_UPDATE_ID`, the last message that we already processed. Now it's one bigger so telegram considers the update as confirmed and doesn't resend it. (from the [offset description](https://core.telegram.org/bots/api#getupdates))


leandrotoledo:
Very well noted, thank you for your pull request.

PR #45, opened 2015-08-24
use long polling in examples

rahiel:
The benefit of long polling is that it will return the Update immediately while waiting, instead of completing the waiting period like sleep does.


leandrotoledo:
Thank you!

PR #46, opened 2015-08-25
add remaining chat actions

rahiel:
The ChatActions from the doc weren't there. The telegram API supports the "record" and "upload" messages, but the client sees the same message for both (at least now on Android).


leandrotoledo:
Thank you for your pull request!

PR #51, opened 2015-09-02
Testing

njittam:
I added a CommandHandler and an example for how to use it. I also added a unit test but i am new to unittesting so i dont know if I am doing it the right way.
I also am not sure if the run and run_once methods are fine. 
They do work but I don't know if there is a better way to make such methods.


leandrotoledo:
W00t!!! Nicely done! Thank you very much!

PR #57, opened 2015-09-08
Add ability for custom filename in send methods and InputFile

ergoz:
Methods sendAudio, sendVideo, sendVoice, sendDocument and other that uses real files updated to add ability to set custom filename that will show in telegram client.


telegram/inputfile.py
```diff
@@ -67,7 +67,10 @@ def __init__(self,
 
         if isinstance(self.input_file, file):
             self.input_file_content = self.input_file.read()
-            self.filename = os.path.basename(self.input_file.name)
+            if 'filename' in self.data and self.data['filename']:
```
leandrotoledo: ``and self.data['filename']`` is redundant of ``if 'filename' in self.data``, so you can keep only the first part.

telegram/inputfile.py
```diff
@@ -67,7 +67,10 @@ def __init__(self,
...
+                self.filename = self.data['filename']
```
leandrotoledo: Instead ``self.filename = self.data.pop('filename')`` otherwise I'll pass along Telegram API and it will raise an error because ``filename`` is not a known parameter.

leandrotoledo:
Nicely done, please just make the minor changes I've noted and we'll be good to merge in.
Thank you for this PR!

leandrotoledo:
Yay! Thanks. I'll take care of these Travis tests errors, it wasn't your PR but some TelegramError parse issues.

ergoz:
Telegram api server 502 bad gateway :) How to run travis building again?

leandrotoledo:
That's a good question! haha
Refresh doesn't seem to be working. I'll be sending tests shortly though. But I've been pulling this change into my workspace to run tests locally, so far so good.

PR #67, opened 2015-09-14
Feature/requests

peczony:
use requests instead of urllib if possible; add optional timeout and decorator to get and post


leandrotoledo:
Hey, thanks for this PR, I've tried to run it locally but I'm getting these errors (full trace here http://pastebin.com/Dnp2q7Sg).

Would be nice to disable urllib3 warnings as well

https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning

leandrotoledo:
I had to revert but looking forward to having a new one with these details sorted, I'm also working to try t figure that out.

PR #85, opened 2015-10-22
Decode Emoji byte strings into unicode strings if using Python 3

jh0ker:
I used a class decorator to call a function upon class creation, that decodes every class attribute of the type 'bytes' into 'str' if sys.version_info.major == 3

Fix for https://github.com/leandrotoledo/python-telegram-bot/issues/80


jh0ker:
Hey, I just realized I messed up programming languages in the docstrings of telegram/emoji.py. It says 'Class' instead of 'type'. Maybe you could fix it, so I don't have to create another pull request :)

franciscod:
My #24 fix was wrong: the emojis should always be strings, not byte arrays... ,maybe we should dig deeper on py2 and py3 unicode compatibility, many bugs come from these things..

PR #87, opened 2015-10-23
Remove `.encode('utf-8')` from echobot

### LINKED ISSUE #86: @ character
If I type @ in a chat I got this. (using https://github.com/leandrotoledo/python-telegram-bot/blob/master/examples/echobot.py)

TypeError: b'hola @honguitobot' is not JSON serializable

franciscod:
This was making the bot throw TypeError on py3.

Closes #86


franciscod:
this is similar to #85, and aims to fix #24 in a better way

all tests fail as now, yeahhhh

jh0ker:
Thank you :)

Issue #16, opened 2015-07-20
Make tests more generic


leandrotoledo:
The way it is right now it will fail if not ToledosPalaceBot, if more generic Travis could run all the tests when pull requesting, etc.

Jzarecta:
I want to learn how the module handles the declared commands assigned by botfather and be able to route them to the desired action.

leandrotoledo:
Hi @Jzarecta, it doesn't. Bot API doesn't provide methods to get declared commands, you have to create your own handlers parsing text messages when polling updates from their servers. Have a look at examples folder from this repository, hopefully it will be a good start.

Feel free to reach out me for any further questions.

Jzarecta:
Actually the example folder didn't gave me much insight. Perhaps I didnt read it right, but I was able to declare updates on the api, but that was it, the rest was a lonely def that supposely was the command to use, but it didnt work as expected. I had the following code:
```python
def p(value):
    #Buscamos el API
    bitapeso=urllib2.urlopen("http://bitapeso.com/json/").read()
    jobject = json.loads(mxn, 'utf-8')
    val=jobject['mxn']
    if val == None:
        peso = float(val)
    else:
        peso = float(val) * value
    return peso.strip()
```

Maybe I miss something, or a lot.

franciscod:
We could get the token from an env-var

franciscod:
are these tests https://travis-ci.org/leandrotoledo/python-telegram-bot/jobs/87105084 from #87 failing due to a testing token issue?

franciscod:
there are a lot of hardcoded things, right? message/file ids, message contents

leandrotoledo:
The way it's writen atm I'm the only one who can see the bot interation (as it sends messages to chat_id corresponding my telegram user). @jh0ker has suggested a public channel so anyone planning to submit a feature can see the results for their thats. In order to accomplish that we would need to change some tests and avoid hardcoding as @franciscod mentioned. I'll spend some time on it this week.

leandrotoledo:
Seems like the amount of efforts to have the same result does justify this rework right now.
The token and the chat_id are now shared on `tests/base.py` so anyone can run tests. Until we have a better (and easier way) to emulate client requests or generic tests I'd say to keep the way we have.

PR #88, opened 2015-10-26
make tests more independent from specific bot and its token

rvetrov:
- merge some tests 
- for those tests that are strong related to PythonTelegramBot and its resources: do not perform them when LOCAL_TEST environmental variable is set

Now every contributor can run tests locally using their bots.
May be it will be usefull to add some tips about running tests to the help.

Fix for  #16


rvetrov:
Are Travis' settings ok? It seems there is not TOKEN variable in environment

leandrotoledo:
Nice work @rvetrov! I'm checking Travis...

jh0ker:
This is outdated

Issue #89, opened 2015-11-01
Pending update process without error (after network crash)


kingkeyx:
Hi,

I'm a beginner developer, so I'm not sure if this is really a bug (maybe the error is somewhere else).

Description:
On Windows 7 + Spyder IDE, using "bot.getUpdates(offset=id+1,timeout=20)" will search for updates for 20 seconds and return an empty list (--> [ ] ). This is the desired behaviour since there are no new updates. Performing the same operation without network will result in an urllib2.URLError. However, performing the operation with network and **turning WiFi off after some seconds** leaves the process unfinished, even after switching WiFi back on.

This behaviour is surprising to me, I expected that the update-process would crash with an error. This would allow me to handle the error somehow and go on with my code. In the moment, **the unfinished job blocks all following lines of code**.

I'm pretty sure that I observe the same problem on my Raspberry Pi, however, I don't have the skills to give any better bug report

Thanks.

rahiel:
Hello, I am having the same issue: the long-polling doesn't return and the bot gets stuck at bot.getUpdates. 

`urllib2.urlopen` has an optional [timeout parameter](https://docs.python.org/2/library/urllib2.html#urllib2.urlopen), maybe it would help passing it to urlopen [here](https://github.com/leandrotoledo/python-telegram-bot/blob/688d6af541a4b0effc0160ae92239b5a8611ef50/telegram/utils/request.py#L93).

jh0ker:
Why did you add 1.0 to the timeout?
 `timeout = data["timeout"] + 1.`

rahiel:
> Why did you add 1.0 to the timeout?
> timeout = data["timeout"] + 1.

Telegram will wait `timeout` seconds and then return, this returning takes some time (network latency etc.). From testing I saw that Telegram usually returns within 0.3 seconds, so waiting for a full second should be safe.

jh0ker:
I see. Do you think there is a better solution to this? It could be a bit confusing. Or at least document the reason.

rahiel:
I don't know how to do this differently, this is probably the best way. Maybe we should wait a bit longer in the future but a second should be more than enough.

Would a comment like `# wait an extra second for the data to return` help? If you wait for exactly the timeout time, python will raise the exception just before the data returns.

jh0ker:
Yes, I think that's the best solution. Please also add this to the timeout parameter docstring in getUpdates, so people won't be confused about why getting updates takes one second extra.

[...]

PR #90, opened 2015-11-02
raise exception if telegram times out on long-polling

rahiel:
This is for issue #89. What it does is enforce the long-polling in the library, so if telegram doesn't return within the agreed time, we raise an exception.

I think the library shouldn't raise an exception, but it should just ignore this error and continue. I am not sure how to do that, need some help on that!


zubchick:
need  this

jh0ker:

PR #91, opened 2015-11-03
Make user_from optional to work with channels channels

jh0ker:
Fix for https://github.com/leandrotoledo/python-telegram-bot/issues/84


leandrotoledo:
Awesome! Thanks @jh0ker

PR #95, opened 2015-11-09
fix LGPLv3 badge

rahiel:
This fixes the license image badge.


leandrotoledo:
Thank you! Nice catch!

PR #97, opened 2015-11-10
Fallback strategy for mimetype discovery

avanheuvelen:
I ran into an image that didn't return a mimetype with `imghdr`. I first quickfixed it with the http content-type header, but decided that `mimetypes` would be a better alternative.

I've now added them both, but perhaps something like pillow would be better?

`mimetypes` ofcourse tries to guess the mimetype according to the file extension, which might not be entirely trustworthy. The Content-Type header is probably even less trustable. But it works for me, so I thought I'd make a pull request.

### DIFF
--- a/telegram/inputfile.py
+++ b/telegram/inputfile.py
@@ -79,7 +79,7 @@ class InputFile(object):
 
         if 'http' in self.input_file:
             self.input_file_content = urlopen(self.input_file).read()
-            self.mimetype = InputFile.is_image(self.input_file_content)
+            self.mimetype = InputFile.is_image(self.input_file_content, self.input_file)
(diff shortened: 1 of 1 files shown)


jh0ker:
Thank you for your contribution!
Since this could be a security risk, we're not including it in the library. Also, calling `urlopen` again will create more overhead, which is probably superfluous for most users.

avanheuvelen:
Yes, I was thinking about the second urlopen. It's not necessary and could easily be removed by not making is_image a static method.

Why do you see this a security risk though?

jh0ker:
Any an executable file with a .jpg extension would be accepted, no?

Issue #36, opened 2015-08-13
Better documentation of best practices for error handling


waltonjones:
It would be useful to add a section to the documentation dealing with error handling. There are certain issues that can lead to errors that may not be immediately obvious.

Example: What happens when a user deletes a chat before the bot has responded? I think it produces a Telegram server error, but what is the best way to deal with this situation safely without crashing the bot.

leandrotoledo:
Thank you for reminding of that!

franciscod:
Yeah, same thing with bot being removed from a group. It might be useful to have an additional higher level bot with error handling, queueing, command handlers...

jh0ker:
@franciscod We are actually working on something like that. Check out the example bot on branch boteventhandler. It's still work in progress though ;)

leandrotoledo:
Related to #135

PR #99, opened 2015-11-12
improve echobot with cleaner code and basic exception handling

rahiel:
This removes the global variable, removes getting the latest update_id in main (because if you use None, Telegram will send you all unconfirmed messages by default), and implements some basic exception handling (helps a bit for #36). Tested with python 2 and 3.


jh0ker:
Thank you very much, Rahiel :)

PR #109, opened 2015-11-24
Indrucing "Updater" and "Dispatcher"

jh0ker:
Updater is a class that builds on top of `Bot` and pulls updates from Telegram for you. To deliver them to you, it uses the `Dispatcher`. You can register handler-functions for all kinds of messages in the `Dispatcher` and they will receive them if an update of their kind is received. For handlers that take a long time for IO-operations, you'll be able to let them run asynchronously in a thread pool. 

Both services run each in a speperate thread, so your application remains responsible, if you want. Else, the `Updater.idle()`-method provides an interruptable waiting loop, that can subscribe to system signals like `SIGTERM`, if you want to run the bot as a daemon. 

The `Updater` offers getting updates via polling as well as via webhook and you can switch between them with the change of one line!

Check the examples in the examples folder!

A lot of the code from the first commit is from @naveenvhegde, so big thank you for that!
Also thanks to @leandrotoledo for your help with all that stuff.


leandrotoledo:
Nicely done! - Sorry took me a while to review all the code.
Really good job @jh0ker, kudos on you!
Let's add a one paragraph or two about this on README/docs.

Thank you again! I'll be releasing this right away.

PR #114, opened 2015-12-01
Separate start_webhook and setWebhook

Balduro:
I think starting and setting a webhook should be separated tasks.
It adds more flexibility when you want to access through apache etc.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -9,6 +9,7 @@ Contributors
 The following wonderful people contributed directly or indirectly to this project:
 
 - `Avanatiker <https://github.com/Avanatiker>`_
+- `Balduro <https://github.com/Balduro>`_
 - `bimmlerd <https://github.com/bimmlerd>`_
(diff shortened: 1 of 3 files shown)


jh0ker:
I see, i forgot that the webhook needs to be specified by the user when using apache. We could just add a parameter to `start_webhook`, like `custom_webhook_url` or something. Wouldn't that be easier?

Balduro:
why do two diffrent tasks in one funktion?
is:
start_webhook(network, port, path)
bot.setWebhook("https://server.tld/path/as/seen/from/outside", "path/to/cert")

realy harder than one bloated start_webhook call?
It gets cleaner, additionaly: when you restart your program (after crash or something like this)
you dont have to set the webhook again and again.
I think telegramm would always appreciate less calls.

jh0ker:
You're right :+1:

jh0ker:
I updated the code in the example bot and merged directly into master. Thanks a lot for your contribution :)

PR #127, opened 2015-12-16
API updates Oct and Nov

jh0ker:
The class `GroupChat` has been refactored to `Chat` and now represents all chats. Also, `Message` now contains the optional fields `supergroup_chat_created`, `migrate_to_chat_id`, `migrate_from_chat_id` and `channel_chat_created.`


leandrotoledo:
Suggestion: Eventually we could have a ChatType class, listing off all the types, as we do in ChatAction.

jh0ker:
@leandrotoledo I thought about that, but at the time of implementation, there was no use for that. There was no situation where you would assign the chat type, you'd only decode from json

PR #138, opened 2015-12-27
Fix file.download with custom_path

Gems:


leandrotoledo:
Nice catch!

jh0ker:
Thank you for your contribution :)

PR #141, opened 2015-12-31
Change a print statement to use logger instead

rbpasker:


jh0ker:
Hey, good catch! We don't need the import statement from `traceback` then, I guess.

rbpasker:
yes, i'm pushing that

jh0ker:
Great :+1:

PR #151, opened 2016-01-10
enable sending by URL for all sendX methods

jh0ker:


telegram/inputfile.py
```diff
@@ -69,19 +70,30 @@ def __init__(self,
...
-            self.mimetype = mimetypes.guess_type(self.filename)[0] or \
-                DEFAULT_MIME_TYPE
-
-        if 'http' in self.input_file:
-            self.input_file_content = urlopen(self.input_file).read()
-            self.mimetype = InputFile.is_image(self.input_file_content)
-            self.filename = self.mimetype.replace('/', '.')
+            elif from_url:
+                self.filename = \
+                    self.input_file.url.split('/')[-1]\
```
leandrotoledo: os.path.basename also works with URIs, like
`basename('http://abc.com/test.mp3')`
`=> 'test.mp3'`

telegram/inputfile.py
```diff
@@ -69,19 +70,30 @@ def __init__(self,
...
+                        .split('?')[0].split('&')[0]
```
leandrotoledo: These splits assume the last argument of a GET query will be a file name, right?

leandrotoledo:
Did some minor comments, but code looks good.
Would be nice to have tests for audio, video and voice as well. We could even point to this repo' data folder to get the files.

jh0ker:
@leandrotoledo Nice! I'll get on that :)

telegram/inputfile.py
jh0ker: Good catch!

telegram/inputfile.py
jh0ker: Yes, indeed. I should add some checks to that, weather that's actually the case.

leandrotoledo:
Adding some tests for send(Audio,Video,Voice) by URL, although, tests are failing because the file_size deosn't match (from client perspective I receive the file, but it's unplayable or an empty image).

Needs further investigation.

leandrotoledo:
@leandrotoledo LGTM, good to be merged/released.

PR #161, opened 2016-01-23
better exception handling

tsnoam:


jh0ker:
Looks good to me! As always, thanks a lot, Noam :)

PR #164, opened 2016-01-27
Parse mode for HTML added

jlmadurga:
Just new HTML parse mode added.


jh0ker:
Awesome! Thank you for your contribution :)

PR #169, opened 2016-02-01
Little document extend

trollknurr:
It was a quite tricky for me to find how send a image file from disk via `sendPhoto`.
I found answer in closed issues. In that answer moderator advice to read README carefully, i go to README, but didn't found any note about sending file from disk(
This few strings can save a little time for just arrived users


ratijas:
for just arrived, `git add --patch` and also `git diff` before commit.  i mean white-spaces removal.

leandrotoledo:
Nice catch, thanks for your contribution!

Issue #146, opened 2016-01-04
Implement inline bots


jh0ker:
- `answerInlineQuery` as a new Method
- `InlineQuery` as a new type, which can be accessed in the new field `inline_query` in `Update`
- `InlineQueryResult[Article|Photo|Gif|Mpeg4Gif|Video]` as the reply types for answers

jh0ker:
Implementation can now be found on branch inlinebots. Unstable and missing unit tests.

carlosmuller:
Using InlineQueryResultPhoto Doesnt show properly all images, most of time only show blank results. All images I'm using are jpegs  less them 5mb

jh0ker:
@carlosmuller Hmm... Could you show us some of the code?

On another note, the previously documented (and therefore implemented by me) `ChosenInlineResult` has been quietly removed from the API. Your problems may also be related to Telegrams problems with the API.

carlosmuller:
http://pastebin.com/MNv2ReyB
def inlinequery (bot, update):
	if update.inline_query is not None and update.inline_query.query:
		api = "https://api.deckbrew.com/mtg/cards/typeahead?q=<NAME>"
		query = update.inline_query.query.split("//")[0].strip()
		logger.debug("Query is :{0}".format(query))
		cards = requests.get(api.replace("<NAME>",query.lower().replace(",","")).replace("'","")).json()
		images = list()
		i = 0
		if(cards == []):
			return
		if (len(cards) == 1):
			images.append(InlineQueryResultPhoto(id=hex(getrandbits(64))[2:],
												photo_url = cards[0]['editions'][0]['image_url'],
												thumb_url = cards[0]['editions'][0]['image_url']))
			logger.debug("Card {0} : {1}".format(cards[0]['name'],cards[0]['editions'][0]['image_url']))
			bot.answerInlineQuery(update.inline_query.id, results=images)
			return
		for card in cards:
			img =''
			if(card['editions'][0]['multiverse_id']==0):
				img = card['editions'][1]['image_url']
			else:
				img = card['editions'][0]['image_url']
			images.append(InlineQueryResultPhoto(id=i,caption=card['name'],
													title=card['name'],
													photo_url = img,
													thumb_url = img))
			logger.debug("Card {0} {1} : {2}".format(i,card['name'],img))
			i = i +1
		bot.answerInlineQuery(update.inline_query.id, results=images)

jh0ker:
[...]

PR #174, opened 2016-02-07
Implement Inlinebots

jh0ker:
Implement Telegram Bot API changes from January 4, 2016 and January 14, 2016
Related to #146


telegram/bot.py
```diff
@@ -586,6 +586,56 @@ def sendChatAction(self,
...
+        data = {'inline_query_id': inline_query_id,
+                'results': results}
+
+        if cache_time is not None:
+            data['cache_time'] = cache_time
+        if is_personal is not None:
+            data['is_personal'] = is_personal
+        if next_offset is not None:
+            data['next_offset'] = next_offset
+
```
tsnoam: who may call this method? what will happen if `cache_time` is not int? if `is_personal` is not bool and `next_offset` is not str?
can it cause an exception raised at a later time? (will cause confusion on the real origin of the problem)

telegram/choseninlineresult.py
```diff
@@ -0,0 +1,82 @@
...
+        from_user (:class:`telegram.User`):
+        query (str):
+
+    """
+
+    def __init__(self,
+                 result_id,
+                 from_user,
+                 query,
+                 **kwargs):
```
tsnoam: The line feed after each argument... Is it from the google style guide?

telegram/dispatcher.py
```diff
@@ -254,21 +257,23 @@ def processUpdate(self, update, context=None):
...
 
+            # Telegram update (message)
+            else:
+                self.dispatchTelegramMessage(update, context)
+                handled = True
+        elif isinstance(update, Update) and \
+                (update.inline_query is not None or
+                 update.chosen_inline_result is not None):
+                self.dispatchTelegramInline(update, context)
+                handled = True
```
tsnoam: I think that this method (`processUpdate()`)  became too complex and hard to follow the flow (fwiw, it wasn't simple before neither :-\ )
I suggest refactoring it and possibly splitting it into sub-methods to handle the internal flows.

telegram/inlinequeryresult.py
```diff
@@ -0,0 +1,416 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser Public License for more details.
+#
+# You should have received a copy of the GNU Lesser Public License
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+
+"""
+This module contains the classes that represent Telegram InlineQueryResults
```
tsnoam: can you add a link to the relevant section in the api documentation?

telegram/inlinequeryresult.py
```diff
@@ -0,0 +1,416 @@
...
+        thumb_url (Optional[str]):
+        thumb_width (Optional[int]):
+        thumb_height (Optional[int]):
+    """
+
+    def __init__(self,
+                 id,
+                 title,
+                 message_text,
+                 **kwargs):
```
tsnoam: why **kwargs and not arguments with default value as None?

telegram/inlinequeryresult.py
```diff
@@ -0,0 +1,416 @@
...
+        # Required
+        super(InlineQueryResultArticle, self).__init__('article', id)
+        self.title = title
+        self.message_text = message_text
+
+        # Optional
+        self.parse_mode = kwargs.get('parse_mode', '')
+        self.disable_web_page_preview = kwargs.get('disable_web_page_preview',
+                                                   False)
+        self.url = kwargs.get('url', '')
```
tsnoam: What will happen if `url` is not a string? (similar question to the other kwargs)

telegram/inlinequeryresult.py
```diff
@@ -0,0 +1,416 @@
...
+        id (str): Unique identifier for this result, 1-64 Bytes
+
+    """
+
+    def __init__(self,
+                 type,
+                 id):
+        # Required
+        self.type = str(type)
+        self.id = str(id)
```
tsnoam: nitpick: `type` & `id` are python bulitins. you might want to consider this instead:
```
def __init__(self,
             type_,
             id_):
    self.type = type_
    self.id = id_
```

telegram/inlinequeryresult.py
```diff
@@ -0,0 +1,416 @@
...
+
+        Returns:
+            telegram.InlineQueryResultArticle:
+        """
+        if not data:
+            return None
+
+        return InlineQueryResultArticle(**data)
+
+
```
tsnoam: all above comments applies for the following classes

tests/test_choseninlineresult.py
```diff
@@ -0,0 +1,80 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+
+"""This module contains a object that represents Tests for Telegram
+ChosenInlineResult"""
+
+import os
```
tsnoam: `os` is not used

tests/test_choseninlineresult.py
```diff
@@ -0,0 +1,80 @@
...
+import unittest
+import sys
+sys.path.append('.')
```
tsnoam: are you sure that this is necessary? afaik, current working directory is in `sys.path` by definition. am i missing something?

tests/test_choseninlineresult.py
```diff
@@ -0,0 +1,80 @@
...
+
+        self.json_dict = {
+            'result_id': self.result_id,
+            'from': self.from_user.to_dict(),
+            'query': self.query
+        }
+
+    def test_choseninlineresult_de_json(self):
+        """Test ChosenInlineResult.de_json() method"""
+        print('Testing ChosenInlineResult.de_json()')
```
tsnoam: you don't really need the docstring and print here.
when running `nosetests -v` it will print the method name by default and imho this is the most informative for the developer.

tests/test_choseninlineresult.py
```diff
@@ -0,0 +1,80 @@
...
+
+        result = telegram.ChosenInlineResult.de_json(self.json_dict)
+
+        self.assertEqual(result.result_id, self.result_id)
+        self.assertEqual(result.from_user.to_dict(), self.from_user.to_dict())
```
tsnoam: use `self.assertDictEqual` instead

tests/test_inlinequery.py
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+    def test_inlinequery_de_json(self):
+        """Test InlineQuery.de_json() method"""
+        print('Testing InlineQuery.de_json()')
+
+        inlinequery = telegram.InlineQuery.de_json(self.json_dict)
+
+        self.assertEqual(inlinequery.id, self.id)
+        self.assertEqual(inlinequery.from_user.to_dict(), self.from_user.to_dict())
```
tsnoam: 1. use `self.assertDictEqual` instead
1. nitpick: line is longer than 79 chars

tests/test_inlineresult.py
```diff
@@ -0,0 +1,442 @@
...
+        self.disable_web_page_preview = True
+        self.url = 'url'
+        self.hide_url = True
+        self.description = 'description'
+        self.thumb_url = 'thumb url'
+        self.thumb_height = 10
+        self.thumb_width = 15
+
+        self.json_dict = {
+            'type': 'article',
```
tsnoam: consider using a variable for `'article'` (i.e.:
```
self.type = 'article'
...
self.json_dict = {
    'type': self.type,
...
```
then later on you can call `assertEqual` with `self.type`

tests/test_inlineresult.py
```diff
@@ -0,0 +1,442 @@
...
+        self.assertEqual(article.message_text, self.message_text)
+        self.assertEqual(article.parse_mode, self.parse_mode)
+        self.assertEqual(article.disable_web_page_preview,
+                         self.disable_web_page_preview)
+        self.assertEqual(article.url, self.url)
+        self.assertEqual(article.hide_url, self.hide_url)
+        self.assertEqual(article.description, self.description)
+        self.assertEqual(article.thumb_url, self.thumb_url)
+        self.assertEqual(article.thumb_height, self.thumb_height)
+        self.assertEqual(article.thumb_width, self.thumb_width)
```
tsnoam: can `self.assertDictEqual` be used here instead of all the `assertEqual`

tsnoam:
i see that the coverage had decreased. i couldn't see in travis why (it claims that the files don't exist). on the next CI check lets make sure that the important parts are covered properly.

telegram/dispatcher.py
jh0ker: I agree with you on that ^^

tests/test_choseninlineresult.py
jh0ker: I am actually not sure about that either, but this has been done in all other tests as well, so I just assumed it would be necessary.

tests/test_choseninlineresult.py
jh0ker: As above, this is done in other test packages, so I went with it.

tests/test_choseninlineresult.py
jh0ker: I did not know about this. It's done in other tests like that as well, maybe we can change it in other places as well.

tests/test_inlinequery.py
jh0ker: Weird that travis does not notice it

tests/test_inlineresult.py
jh0ker: agreed

jh0ker:
I think the coverage decreases because there is no test for `answerInlineQuery`, since I was unsure about how to test it automatically

telegram/inlinequeryresult.py
jh0ker: I noticed, but I did it like this cause it is done in other places of the code (eg User.py), for consistency.

telegram/inlinequeryresult.py
jh0ker: I could say that it would make the code unnecessarily big, but I guess I was also kind of lazy about it. As with a lot of other comments, this is done throughout the library, so I went with it.

telegram/bot.py
jh0ker: It will either cause the JSON library to raise an error (`<type> is not JSON-serializable`) or an API error, if it goes through to telegram servers.

telegram/inlinequeryresult.py
jh0ker: (Similar answer)

telegram/choseninlineresult.py
jh0ker: Seemed to be library convention as it is done in most other places, too.

tests/test_inlineresult.py
jh0ker: Yes

jh0ker:
Travis is failing due to network errors

telegram/bot.py
tsnoam: In that case I think that enforcing the proper types is required. Helps identifying errors in the proper place and avoids user questions later.
For `int` and `bool` it can be done just as simple as `int(cache_time)` & `bool(is_personal)`.
About `str` there's the unicode compatibility in python 2.x. So checking `isinstance(next_offset, basestring)` and raising `ValueError` is in order.

telegram/inlinequeryresult.py
tsnoam: imho, using *args & **kwargs must be used with care. i find the most useful when you wrap another function (and even then it can be misleading and a pitfall for typos). when it's for "your own" function you should be explicit and specify each argument in the function prototype.

telegram/choseninlineresult.py
tsnoam: oh well.

telegram/inlinequeryresult.py
tsnoam: (similar answer as well :) )

telegram/inlinequeryresult.py
tsnoam: ok

tests/test_choseninlineresult.py
tsnoam: you don't have to keep going with that then :-\
after all, these are unitests and they should be as most convenient for the developer.

PR #176, opened 2016-02-09
Conversation

njittam:
Hey,
I added support for what I call conversations.
This is like when you create a new bot the botfather ask you first for a name and then for a for a username.
I also added an example.
In this example the bot will wait for a /start.
the will ask the user to enter some words.
When the user enters a '.' bot wil say the sentance and the conversation will be ended.

### DIFF
--- a/examples/coversationEchobot.py
+++ b/examples/coversationEchobot.py
@@ -1,0 +1,102 @@
+
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Simple Bot to reply to Telegram messages
+# This program is dedicated to the public domain under the CC0 license.
+
+"""
+Based on Echobot
+This Bot uses the Updater class to handle the bot.
+
+First, a few handler functions and class are defined. Then, those functions are passed to
+the Dispatcher and registered at their respective places.
+Then, the bot is started and runs until we press Ctrl-C on the command line.
+
+Usage:
+Basic Conversation Echobot example, asks for words and says the whole sentence
+when the users says '.'.
+Press Ctrl-C on the command line or send a signal to the process to stop the
+bot.
+"""
+
+from telegram import Updater
+import logging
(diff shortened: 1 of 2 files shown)


naveenvhegde:
Hi,

Saw your implementation. Can you explain more about how to handle conversation. 

How to handle states with in conversation ?

njittam:
Hi, 
Well the dispatcher will initiate one object for one chat. So when in chat 1 someone says /start and in chat 2 someone else says /start in the example. There will be two objects initiated. One for every Chat. 
This way you can keep track of states within your class.

The only two states the dispatcher cares about is 'running' and 'not running'.
to say to the dispatcher hey I am not done yet you let the __call__ method return False
and when the conversation is over you return True.

tsnoam:
Hi @njittam 

It seems like there were CI failures on the code. Have you checked it prior to push?

Anyway, the idea of a "Conversation" bot seems like a very good idea to us & we would be happy to add it. However, we were wondering whether it really requires an infrastructural change and can't be implemented using the current dispatching mechanism?
Do you think the "Conversation" bot can be modified to be implemented as an example (and added to the `examples` dir)?

 - Noam

njittam:
Hi, 
I am going to close this pull request because it failed the tests.
It might not be necessary to but in my opinion is this a really easy way to implement a conversation to your bot.
I already created an example.

naveenvhegde:
@njittam, having conversation integrated with library is great, but it requires a lot of thoughts as it has state associated with it.

I suggest you write up a wiki on the project page about your thoughts and once the idea gets concrete, it will be easy to implement.

njittam:
will do.

jh0ker:
@njittam Please check out our latest 4.0rc1 release. Your proposal should integrate very nicely with it, due to structural changes in the Dispatcher.

PR #177, opened 2016-02-09
join() threads for a cleaner stop procedure

tsnoam:
plus some minor fixes / code cleanups

### DIFF
--- a/telegram/dispatcher.py
+++ b/telegram/dispatcher.py
@@ -22,7 +22,7 @@
 import logging
 from functools import wraps
 from inspect import getargspec
-from threading import Thread, BoundedSemaphore, Lock, Event
+from threading import Thread, BoundedSemaphore, Lock, Event, current_thread
 from re import match
 from time import sleep
 
@@ -33,7 +33,8 @@ H = NullHandler()
 logging.getLogger(__name__).addHandler(H)
 
 semaphore = None
-running_async = 0
+async_threads = set()
+""":type: set[Thread]"""
 async_lock = Lock()
 
 
@@ -58,11 +59,10 @@ def run_async(func):
         """
         A wrapper to run a thread in a thread pool
         """
-        global running_async, async_lock
         result = func(*pargs, **kwargs)
         semaphore.release()
         with async_lock:
-            running_async -= 1
+            async_threads.remove(current_thread())
         return result
 
     @wraps(func)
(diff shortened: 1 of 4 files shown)


telegram/utils/webhookhandler.py
```diff
@@ -80,6 +84,23 @@ def do_POST(self):
                              update.update_id)
             self.server.update_queue.put(update)
 
-        else:
-            self.send_error(403)
+        except _InvalidPost as e:
+            self.send_error(e.http_code)
             self.end_headers()
+
+    def _validate_post(self):
```
jh0ker: I may be missing something, but this is not called, is it?

telegram/updater.py
```diff
@@ -276,26 +266,49 @@ def stop(self):
...
 
-                self.logger.debug("Requesting Dispatcher to stop...")
-                self.dispatcher.stop()
-                while dispatcher.running_async > 0:
-                    sleep(1)
+                self._stop_httpd()
+                self._stop_dispatcher()
+                self._join_threads()
 
                 self.logger.debug("Dispatcher stopped.")
```
jh0ker: please move this line into `_stop_dispatcher`

telegram/utils/webhookhandler.py
tsnoam: your eyes are working properly :)
i forgot to call `self._validate_post()` instead of the `if` statement in `do_POST()`

telegram/dispatcher.py
```diff
@@ -33,7 +34,8 @@
 logging.getLogger(__name__).addHandler(H)
 
 semaphore = None
-running_async = 0
+async_threads = dict()
```
tsnoam: will check if using `set()` can work (and then no need for UUID)

telegram/dispatcher.py
```diff
@@ -70,11 +71,11 @@ def async_func(*pargs, **kwargs):
         """
         A wrapper to run a function in a thread
         """
-        global running_async, async_lock
-        thread = Thread(target=pooled, args=pargs, kwargs=kwargs)
+        uu = str(uuid.uuid1())
```
jh0ker: Do we still need this, or could we just go with the default thread name?

telegram/dispatcher.py
```diff
@@ -33,7 +34,8 @@
 logging.getLogger(__name__).addHandler(H)
 
 semaphore = None
-running_async = 0
+async_threads = set()
+""":type: set[Thread]"""
```
jh0ker: Does pycharm recognize `"""asyc_threads (set[Thread]): """`? This format would be preferrable, since it's a google style docstring.

tsnoam: in order not to leave this "in the air"
it was agreed that we can use the Sphinx docstring comment style for this kind of case (i.e. "standalone" variable)
this was the most reliable form we found and the most friendly to pycharm

PR #190, opened 2016-03-01
Silent Messages

msangalli:


tsnoam:
@msangalli 
Thanks for the good work, we'd be happy to merge this.
However, we'd like a few enhancements, if you're up to it:

1. The silent message is not only for sendMessage method, but for others as well. Can you please add modify the other methods as well?
1. Can you add unitests for this code?
1. Make sure flake8 tests pass.
1. Please squash the three commits into one commit, to keep history clean & tidy.

leandrotoledo:
Hi @msangalli, thanks for your contribution, #195 has implemented this solution for this methods and the ones mentioned by @tsnoam so I'll mark this one as closed.

Thank you.

Issue #184, opened 2016-02-22
Overextending the API


taemery:
This API tries to implement everything: polling, web socket HTTP server, etc. This can make it hard to use in circumstances (GAE) where only the raw API is required. I propose separating the servers out and focusing on just a stateless API. 

(The specific reason GAE doesn't work is referenced in the GAE incompatibility issue)

rahiel:
> I propose separating the servers out and focusing on just a stateless API.

Hey, they are sort off separated already. The raw [API](https://github.com/python-telegram-bot/python-telegram-bot#api) is in the `Bot` class. The `Updater` class (the one that does "everything") is built upon that one. I am only using `Bot` without problems.

Maybe we can solve this in the documentation: the readme treats the `Bot` class as legacy and old while there are still good reasons to prefer `Bot` over `Updater`. Instead we can present them as two different methods to make bots, recommending `Updater` over `Bot` when possible because it is easier. Still `Bot` should probably be mentioned first because the library is "A Python wrapper around the Telegram Bot API".

Thoughts?

taemery:
Specifically for GAE and other hosted code providers, the Updater class has dependencies that cannot be satisfied. I feel like the whole bot should be a dependency of Updater because at compile time Updater cannot be present.

rahiel:
I get it now, thanks. We can solve this by putting `Updater` in a separate submodule that must be explicitly imported. I will look into it.

leandrotoledo:
Maybe not only ``Updater``, but all the others extended classes like ``JobQueue`` and so on...
It has a price though, this may break backward-compatibility (which is not good for users not aware of these changes) and at the moment I feel like more people is using ``Updater`` than GAE, so just let keep that in mind.

jh0ker:
@leandrotoledo I think the change should be made in v4.0, along with some other interface changes

leandrotoledo:
Thanks to @rahiel #198 has been merged and will be released on v4.0.

PR #198, opened 2016-03-11
lazily load Updater & move extended classes to submodule

rahiel:
I've added a `Updater` function to the original `__init__.py` that will lazily load the Updater class and return an Updater instance. 

This should not break existing code that uses `from telegram import Updater`, while fixing #184.


rahiel:
Tests pass :smile:, @taemery could you help test this on GAE?

jh0ker:
Very nice @rahiel :)

leandrotoledo:
@jh0ker +1 @rahiel. Maybe we should add an INFO log mentioning this is a deprecated function and will be removed in the future?

leandrotoledo:
Also, let's move the extended classes to a different submodule, here follows some suggestions:

- ext
- extensions
- contrib

Looking forward to hearing some input from y'all.

taemery:
I can test this tomorrow, don't merge yet

taemery:
GAE works! I can put my stamp of approval on this pull request

rahiel:
Thanks @taemery :blush:.

> +1 @rahiel. Maybe we should add an INFO log mentioning this is a deprecated function and will be removed in the future?

@leandrotoledo there isn't an alternate interface to `Updater` yet. I also hoped we could avoid changing the library's API. 

> Also, let's move the extended classes to a different submodule, here follows some suggestions:

That is probably cleaner, but I think the `Updater` class is now a core part of `python-telegram-bot`, aren't more people using that to make Bots now? The library now has part that is just a wrapper around the Telegram bot api, and a more elaborate bot framework. If it's just these two classes, they can stay in the same module. (My original PR moved the extended classes over to a new submodule, but I thought it was unnecessary.)

And `Updater` will be lazily loaded after this PR, so people not using it shouldn't be bothered by it.

leandrotoledo:
> That is probably cleaner, but I think the Updater class is now a core part of python-telegram-bot, aren't more people using that to make Bots now?

Agree. I didn't mean though we should make people use ``Updater`` less and less. These are great helpers for this library and people, indeed, are using it more. Although IMHO we should, somehow, distinguish between core classes - as in Objects and Methods described on Telegram Bot API webservice, and the extra layers created by the developers of ``python-telegram-bot``, these are on different layers when we talk about the wrapper vs helper classes. 

> And Updater will be lazily loaded after this PR, so people not using it shouldn't be bothered by it.
 Show all checks

Until people get used to it, lazying importing it from a submodule shouldn't bother either.

My 2cents.

rahiel:
@leandrotoledo I agree, the separation is good. If we decide on a name for the submodule, I can move the extended classes over. (I wasn't happy having a telegram.updater.Updater instance before, that's why I rolled it back :sweat_smile:)

leandrotoledo:
I'm down for telegram.ext.*, who is with (or against) it?

rahiel:
> I'm down for telegram.ext.*, who is with (or against) it?

I also like the sound of telegram.ext. I changed some of the tests to import from telegram.ext, but kept an import of `Updater` from `telegram`to test for backwards compatibility.

leandrotoledo:
Great work @rahiel, just a minor comment/suggestion:

`import warnings`
`...`
`warnings.warn("Call to deprecated function, please refer to telegram.ext when using Updater from now on.", category=DeprecationWarning)`

Everything else does look good to be merged!

rahiel:
@leandrotoledo the [docs](https://docs.python.org/2/library/warnings.html#warning-categories) mention that `DeprecationWarning`s are ignored by default, I couldn't see it when using `telegram.Updater`, I put a normal warning instead.

leandrotoledo:
@rahiel good catch.
All good, merging. Cheers!

Issue #200, opened 2016-03-13
Dispatсher process command with no-break space like unknow command.


shelomentsevd:
Occurs only if message sent from web version of telegram.

**How to reproduce**
1. In web version of telegram write a command to bot: '/command  '(two or more spaces) or press 'TAB' for auto-complete(will add no-break space to the end of message).
2. Bot will recognize command like 'command\xc2\xa0'('\xc2\xa0' is no-break space)

leandrotoledo:
Woah! Nice catch.

leandrotoledo:
#201 has been merged. Thanks!

PR #201, opened 2016-03-13
Dispather command processing fix(#200).

shelomentsevd:
See description in issue #200

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -23,6 +23,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Noam Meltzer <https://github.com/tsnoam>`_
 - `Oleg Shlyazhko <https://github.com/ollmer>`_
 - `Rahiel Kasim <https://github.com/rahiel>`_
+- `Shelomentsev D <https://github.com/shelomentsevd>`_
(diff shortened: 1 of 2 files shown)


leandrotoledo:
Restarting build, seems like Travis wants attention, to be confirmed.

leandrotoledo:
So, this is the reason it's failing:
> telegram/dispatcher.py:551:80: E501 line too long (82 > 79 characters)

So, my only concern with that is casting the string to UTF-8 (assuming it will be always it). We got rid of any decodes/encodes from our codebase long ago to avoid such conflicts, one approach would be expanding an if-block to check if this conditions applies and then, finally, cast/split/cast back to unicode. But that would be a workaround for something that doesn't necessarily is on us. Let's hear what @jh0ker and @njittam have to say about it.

I think it's also worth to report to Telegram Web this bug as well.

shelomentsevd:
> I think it's also worth to report to Telegram Web this bug as well.
Not sure that is a bug.
I made new commit, because found solution without casting to UTF-8.

leandrotoledo:
Nice! I liked this one better! Thanks for your contribution @shelomentsevd!
Merging, cheers!

PR #202, opened 2016-03-13
Updater webhook bootstrapping improvements

tsnoam:


telegram/updater.py
```diff
@@ -273,6 +316,13 @@ def _start_webhook(self, listen, port, url_path, cert, key):
 
         self.httpd.serve_forever(poll_interval=1)
 
+    def _gen_webhook_url(self, listen, port, url_path, use_ssl):
+        return '{proto}://{listen}:{port}{path}'.format(
+            proto='https' if use_ssl else 'http',
+            listen=listen,
+            port=port,
+            path=url_path if url_path else '/')
```
jh0ker: Currently, the `url_path` would not have to start with a `/`, so it would make sense to change the pattern to `{port}/ {path}` and in this line, put an empty string by default

jh0ker:
One minor comment, everything else look good :)

telegram/updater.py
tsnoam: actually at this point of the flow, `url_path` is already sanitized and it is guaranteed it starts with a `/`, so i'll just do:
```
return '{proto}:// ...
...
path=url_path)
```

PR #208, opened 2016-03-17
declare support for python 3.5, drop old py 3.2 reference

rahiel:
We weren't testing python 3.2, and it was declared end-of-life on [2016-02-20](https://docs.python.org/devguide/#status-of-python-branches). This adds testing for 3.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
+  - "3.5"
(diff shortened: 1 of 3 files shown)


rahiel:
I also thought about adding the [python versions](https://img.shields.io/pypi/pyversions/python-telegram-bot.svg) badge to replace that big table in the readme, but there's no room to have all badges on one line.

Any badge we can (re)move?

PR #211, opened 2016-03-19
Adjust sphinx docs to telegram.ext submodule

jh0ker:


leandrotoledo:
@jh0ker Please also update README

leandrotoledo:
Besides that, everything does look good!

jh0ker:
Should I also update the examples to import from `telegram.ext`?

leandrotoledo:
Good catch! I think you should.

rahiel:
@jh0ker Looks good, there's only something weird with the formatting, search for `__Note:__` in the [readme](https://github.com/python-telegram-bot/python-telegram-bot/blob/ext-docs/README.rst#telegram-ext).

PR #213, opened 2016-03-22
Prepare Release of v3.4

jh0ker:
I moved the `UpdateQueue` to `telegram.util` as it is not used directly. I also added lazy loading to the other two classes moved to `telegram.ext`


leandrotoledo:
@jh0ker makes sense, +1

leandrotoledo:
LGTM, merging!

Cheers!

PR #217, opened 2016-03-23
Only set webhook if using SSL and also send certificate

jh0ker:


telegram/ext/updater.py
```diff
@@ -282,43 +283,47 @@ def _start_webhook(self, listen, port, url_path, cert, key,
                        bootstrap_retries, webhook_url):
         self.logger.debug('Updater thread started')
         use_ssl = cert is not None and key is not None
-        url_path = "/%s" % url_path
+        if not url_path.startswith('/'):
+            url_path = '/' + url_path
```
tsnoam: i think it's better to make sure url_path is string:
```
url_path = '/{0}'.format(url_path)
```

tsnoam:
@jh0ker 
looks ok. one comment on code though

Issue #234, opened 2016-04-15
Change some classes due to consistency 


kiddick:
I think it is good to keep consistency everywhere.

There are three classes without explicitly inheritance from `object`:

```python
python-telegram-bot\telegram\ext\dispatcher.py:
   80  
   81  
   82: class Dispatcher:
   83      """
   84      This class dispatches all kinds of updates to its registered handlers.

python-telegram-bot\telegram\ext\updater.py:
   38  
   39  
   40: class Updater:
   41      """
   42      This class, which employs the Dispatcher class, provides a frontend to

python-telegram-bot\tests\test_updater.py:
  633  
  634  
  635: class MockBot:
  636  
  637      def __init__(self, text, messages=1, raise_error=False,
```

tsnoam:
@kiddick @kiddick this sounds right. would you like to make the change and send a PR?

PR #235, opened 2016-04-15
Fix class declaration according to consistency.

kiddick:
See https://github.com/python-telegram-bot/python-telegram-bot/issues/234


leandrotoledo:
Re-running Travis.

tsnoam:
@kiddick Thanks for your help

Issue #224, opened 2016-03-31
Pressing Ctrl-C several times in a row doesn't stop the bot gracefully


jtpio:
Hi,

I was trying out the `echobot2.py` example program and wanted to stop the bot by sending a SIGINT.

Pressing Crtl-C **only once** makes the bot stop gracefully, that's good.

However, pressing Ctrl-C several times in a row makes the bot hang, and it never stops. The logs:
```
2016-03-31 20:24:30,364 - telegram.ext.updater - DEBUG - dispatcher - started
2016-03-31 20:24:30,365 - telegram.ext.dispatcher - DEBUG - Dispatcher started
2016-03-31 20:24:30,365 - telegram.ext.updater - DEBUG - updater - started
2016-03-31 20:24:30,365 - telegram.ext.updater - DEBUG - Updater thread started
2016-03-31 20:24:30,365 - telegram.bot - DEBUG - Entering: setWebhook
2016-03-31 20:24:30,522 - telegram.bot - DEBUG - True
2016-03-31 20:24:30,522 - telegram.bot - DEBUG - Exiting: setWebhook
2016-03-31 20:24:30,522 - telegram.bot - DEBUG - Entering: getUpdates
^C2016-03-31 20:24:33,545 - telegram.ext.updater - DEBUG - Stopping Updater and Dispatcher...
2016-03-31 20:24:33,545 - telegram.ext.updater - DEBUG - Requesting Dispatcher to stop...
2016-03-31 20:24:34,366 - telegram.ext.dispatcher - DEBUG - orderly stopping
2016-03-31 20:24:34,366 - telegram.ext.dispatcher - DEBUG - Dispatcher thread stopped
2016-03-31 20:24:34,366 - telegram.ext.updater - DEBUG - dispatcher - ended
2016-03-31 20:24:34,447 - telegram.ext.updater - DEBUG - Waiting for dispatcher thread to end
2016-03-31 20:24:34,447 - telegram.ext.updater - DEBUG - dispatcher thread has ended
2016-03-31 20:24:34,448 - telegram.ext.updater - DEBUG - Waiting for updater thread to end
^C^C^C^C^C^C2016-03-31 20:24:40,670 - telegram.bot - DEBUG - No new updates found.
2016-03-31 20:24:40,670 - telegram.bot - DEBUG - []
2016-03-31 20:24:40,670 - telegram.bot - DEBUG - Exiting: getUpdates
[...]

PR #240, opened 2016-04-18
exit immediately if receiving second interrupt #224

jh0ker:
`os._exit` is a bit dirty, but it does the job.


leandrotoledo:
Reseting tests for 2.6.

leandrotoledo:
LGTM, merging.

PR #247, opened 2016-04-23
InputFile to support any objects with .read attribute

leandrotoledo:
Assigning to @tsnoam as reviewer.


leandrotoledo:
Ops, not sure if we are using this branch anymore @jh0ker?

PR #248, opened 2016-04-23
InputFile to support any objects with .read attribute

leandrotoledo:
Changing PR #247 to master instead.
Assigning to @jh0ker


tests/test_bot.py
```diff
@@ -154,6 +154,16 @@ def testSendGIFURLPhoto(self):
 
     @flaky(3, 1)
     @timeout(10)
+    def testSendBufferedReaderPhoto(self):
```
jh0ker: How is this method different to testSendPhoto?

leandrotoledo:
Oops, test is wrong, should be a BufferedReader object instead a file.

PR #251, opened 2016-04-24
release semaphore on exceptions

jh0ker:
If an exception was raised in a handler running with `@run_async`, it would not release the semaphore.


leandrotoledo:
LGMT!

PR #258, opened 2016-04-25
CR for 4.0

tsnoam:
fixes as of CR.
(this is the first bunch of fixes... more might come)


jh0ker:
Sweet!

leandrotoledo:
Please also apply `if not None` to:
- bot.py: `if next_offset or next_offset == '':` and `if webhook_url or webhook_url == '':`

telegram/ext/messagehandler.py
```diff
@@ -22,7 +22,71 @@
 from .handler import Handler
 from telegram import Update
 
-from .filters import *  # flake8: noqa
+
+class Filters(object):
```
leandrotoledo: Can we keep filters on filter.py? Or is there advantage of having it here?

telegram/ext/jobqueue.py
```diff
@@ -22,11 +22,7 @@
 import logging
 import time
 from threading import Thread, Lock
-
-try:
-    from queue import PriorityQueue
-except ImportError:
-    from Queue import PriorityQueue
+from queue import PriorityQueue
```
leandrotoledo: Wouldn't that break Py2?

telegram/ext/updater.py
```diff
@@ -28,12 +28,7 @@
 from time import sleep
 import subprocess
 from signal import signal, SIGINT, SIGTERM, SIGABRT
-
-# Adjust for differences in Python versions
-try:
-    from queue import Queue  # flake8: noqa
-except ImportError:
-    from Queue import Queue  # flake8: noqa
+from queue import Queue
```
leandrotoledo: Wouldn't that break Py2?

telegram/ext/messagehandler.py
tsnoam: we discussed it on the telegram group. no reason to keep the `filters` module.

telegram/ext/jobqueue.py
tsnoam: @leandrotoledo 
no, because `future` brings the `queue` package compatibility to py2

telegram/ext/updater.py
tsnoam: @leandrotoledo 
same answer

leandrotoledo:
Please also update:
https://github.com/python-telegram-bot/python-telegram-bot/wiki/Transistion-guide-to-Version-4.0

Other than that it does look good to me.

PR #263, opened 2016-04-27
Cr4.0 api2

tsnoam:


rahiel:
Leaving a note that this fixes long polling which is currently broken in master and release 4.0.1.

Issue #268, opened 2016-04-29
replykeyboardmarkup doesn't work with string keyboard buttons


rahiel:
According to the [docs](https://core.telegram.org/bots/api#keyboardbutton): 
> This object represents one button of the reply keyboard. For simple text buttons String can be used instead of this object to specify text of the button. Optional fields are mutually exclusive.

So the buttons can be types `str` or `KeyboardButton`.

But currently `replykeyboardmarkup.py` assumes they are of type `KeyboardButton` and fails with `str` buttons, see [here ](https://github.com/python-telegram-bot/python-telegram-bot/blob/998040da924313355f5c27e9b8a175591bca6cf5/telegram/replykeyboardmarkup.py#L77). (`AttributeError: 'str' object has no attribute 'to_dict'`)

Code to reproduce:
```python
import telegram

def main():
    bot = telegram.Bot(token=TOKEN)
    update_id = None

    while True:
        update_id = serve(bot, update_id)

def serve(bot, update_id):
    reply_markup = telegram.ReplyKeyboardMarkup([["Yes", "No"]])
    for update in bot.get_updates(offset=update_id, timeout=3.):
        chat_id = update.message.chat_id
        update_id = update.update_id + 1
        message = update.message.text.encode('utf-8')

        if (message):
            bot.send_message(chat_id=chat_id, text="Want to crash?", reply_markup=reply_markup)

    return update_id

if __name__ == '__main__':
    main()
```

PR #269, opened 2016-04-29
support str and KeyboardButton for reply_markup

rahiel:
This is a start for fixing #268. We should probably also add a test for this.


jh0ker:
LGTM, merging for 4.0.2

PR #271, opened 2016-05-01
Fix Bot.getUpdates() docstring

Djaler:


jh0ker:
Good catch! Thank you for your contribution :)

Issue #275, opened 2016-05-02
timeout when start_polling with clean=True


shouya:
code to reproduce the problem:

```python
updater = Updater(token=xxx)
updater.start_polling(clean=True)
```

error output:
```
Traceback (most recent call last):
  File "main.py", line 43, in <module>
    main()
  File "main.py", line 34, in main
    updates = bot.getUpdates()
  File "/usr/local/lib/python3.5/site-packages/telegram/bot.py", line 114, in decorator
    result = func(self, *args, **kwargs)
  File "/usr/local/lib/python3.5/site-packages/telegram/bot.py", line 1315, in getUpdates
    result = request.post(url, data, timeout=urlopen_timeout)
  File "/usr/local/lib/python3.5/site-packages/telegram/utils/request.py", line 83, in decorator
    raise NetworkError('URLError: {0}'.format(error.reason))
telegram.error.NetworkError: URLError: _ssl.c:629: The handshake operation timed out
```

```
$ pip3 list | grep telegram
python-telegram-bot (4.0.3)
```

I'll fire a PR to fix this later.

jh0ker:
#282 hopefully fixed this issue - if the issue persists in the next release (or current github master), please re-open :)

PR #276, opened 2016-05-02
Update updater.py

shouya:
fixing python-telegram-bot/python-telegram-bot#275

### DIFF
--- a/telegram/ext/updater.py
+++ b/telegram/ext/updater.py
@@ -124,7 +124,7 @@ class Updater(object):
             if not self.running:
                 self.running = True
                 if clean:
-                    self._clean_updates()
+                    self._clean_updates(timeout=10)
 
                 # Create & start threads
                 self._init_thread(self.dispatcher.start, "dispatcher")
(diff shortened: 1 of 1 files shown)


jh0ker:
Travis is failing because of `flake8`. We moved to 99 characters per line instead of 79 and didn't update the configuration. I updated the config to reflect the changes and restarted travis tests.

jh0ker:
A timeout is not really what we want here - if there are no new updates, we don't want to wait 10 seconds but return immediately. I think what we should do here is increase `network_delay`, which is not sent to the telegram servers but instead only used for the `urlopen` timeout. Still, it might happen that the `urlopen` call times out, so we should catch that exception and move the `_clean_updates` call to the bootstrap phase, like `_set_webhook`.

shouya:
If there is no new updates, definitely it shouldn't wait for 10 secs. An expected way is to return `[]` or `None` from `getUpdates`. Not sure how python-telegram-bot deal with this. Probably tuning `network_delay` is a more proper way to do. Feel free to close my PR if it doesn't meet the requirement. I just wanted to point out the bug over there.

jh0ker:
An empty list would indeed be the expected result. I'll close this PR and send a new one tomorrow. Thank you for pointing out the bug :)

PR #278, opened 2016-05-03
add possibility for only one callback function

motom001:


jh0ker:
This will break existing code

Issue #284, opened 2016-05-08
Recent changes in API


leandrotoledo:
May 6, 2016

- Added the field emoji to the Sticker object. Your bot can now know the emoji a sticker corresponds to.
- Added the field forward_from_chat to the Message object for messages forwarded from channels.

rahiel:
This was fixed in #291.

PR #291, opened 2016-05-11
Api additions 2016-05-06

jh0ker:
referencing #284

### DIFF
--- a/telegram/sticker.py
+++ b/telegram/sticker.py
@@ -30,6 +30,7 @@ class Sticker(TelegramObject):
         width (int):
         height (int):
         thumb (:class:`telegram.PhotoSize`):
+        emoji (str):
         file_size (int):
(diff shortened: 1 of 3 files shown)


tsnoam:
@jh0ker lgtm

telegram/sticker.py
```diff
@@ -54,6 +56,7 @@ def __init__(self,
         self.height = int(height)
         # Optionals
         self.thumb = kwargs.get('thumb')
+        self.emoji = str(kwargs.get('emoji', ''))
```
leandrotoledo: str() breaks Python 2. Let the user handle the unicode / bytes casting.

leandrotoledo:
Python 2 didn't like it. @rahiel likes when Python 2 dislikes.

PR #299, opened 2016-05-19
Remove some debug prints. IMHO there is still enough debug printing a…

springjools:
…nd if there is too much of it, the relevant parts are harder to find.

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -105,10 +105,10 @@ class Bot(TelegramObject):
 
         @functools.wraps(func)
         def decorator(self, *args, **kwargs):
-            logger.debug('Entering: %s', func.__name__)
+            #logger.debug('Entering: %s', func.__name__)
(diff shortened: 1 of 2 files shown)


leandrotoledo:
I agree with removing the DEBUG log from telegram/bot.py message decorator.
About the telegram/ext/jobqueue.py, I'll leave to @jh0ker to provide some input as he did some rework there.

jh0ker:
Yes, please take a look at #307 
The whole ticking thing has been removed, the job queue now sleeps when there are no jobs due. This should reduce spam by a lot. If you still think it's too much, please leave a comment on that PR

leandrotoledo:
Hi @springjools, so please rollback telegram/ext/jobqueue.py and instead a comment, remove lines 108 and 111 from bot.py and this will be good to be merged.

Thank you!

leandrotoledo:
Marking as obsolete.

PR #300, opened 2016-05-19
added flags for re.compile

artinnok:
I have added flags for re.compile function. You can see docs [here](https://docs.python.org/2/library/re.html#re.I). I want to use an re.I and can't find how to provide them, and created an this pull request.


jh0ker:
There are two ways to use flags without this:

1. Compile the pattern yourself with the flags you want, then pass the compiled pattern to the handler
2. Put the flag in front of the regex, as described [here](https://docs.python.org/2/library/re.html#regular-expression-syntax) under `(?iLmsux)`

As you can see in the code, the pattern is checked and only compiled if it's a string type.
If this does not work for you, feel free to re-open :)

PR #303, opened 2016-05-22
stripping token of whitespaces before starting bot

Eldinnie:
If tokens get read in with trailing or leading whitespaces bot will not start.
Added code to check and warn the user if token contains whitespaces and strip them.
Moved logger setup for this purpose


leandrotoledo:
Please consider to add a test for this case. something like:

```
    def testInvalidToken4(self):
        self._test_invalid_token('1234:abcd1234\n')
```

Eldinnie:
@leandrotoledo I see you already did that?

leandrotoledo:
Hi @Eldinnie, thanks for your good work here, but we've decided to reflect API errors and raise InvalidToken, as described on #306. So I'll close this one.

Eldinnie:
NP

PR #307, opened 2016-05-25
Make job queue API similar to the dispatcher, add new functionality

jh0ker:
The `JobQueue` API has been changed to look similar to the 4.x `Dispatcher` API

Instead of simply passing a callback function to the `JobQueue.put` method, you now have to pass a `Job` object, which `__init__` takes now the arguments `callback`, `interval` and `repeat` (these were passed to `put` before).

The `Job` class exposes the boolean property `enabled`, which you can use to pause execution of the job temporarily, and the method `schedule_removal`, which marks the job for complete removal from the queue. *Note:* Because the `JobQueue` internally uses a `PriorityQueue`, the job will remain in the queue until it is at the front of the queue, but it will not be executed again.
In addition to the `bot` argument, callback functions now receive their respective `Job` object in an argument called `job`.

The `JobQueue` class now has a `jobs` method, which returns a `tuple` of all jobs currently in the queue.

*Under the hood*
Instead of continuously checking if any jobs should run, the `JobQueue` now waits until the next job is due. If a new job is added in the meantime, the waiting time is re-calculated as necessary. As a result, the `ticking_interval` argument has been removed from both `JobQueue` and `Updater`.


leandrotoledo:
Assigning to @tsnoam

telegram/ext/jobqueue.py
```diff
@@ -141,14 +162,69 @@ def stop(self):
...
+        self.context = context
+
+        self.name = callback.__name__
+        self._remove = Event()
+        self._enabled = Event()
+        self._enabled.set()
+
+    def run(self):
+        """Executes the callback function"""
+        self.callback(self.job_queue.bot, self)
```
shouya: will it be better if it's capable to run a job without being added to a job queue?

jh0ker: This method is used by the job queue, though of course you can also run it from outside.

shouya: IMHO, `Job.run` method shouldn't depend on a job queue. Rather, I think `bot` should become a field in Job class. One possible solution is to specify `bot` as argument in `__init__` method. Or it can simply be a public field that's automatically set when added to a job queue.

At least this makes the tests easier to write, since one does no longer need to add a job to a queue in order to check if this job is valid to run.

telegram/ext/callbackqueryhandler.py
```diff
@@ -34,10 +34,15 @@ class CallbackQueryHandler(Handler):
         pass_update_queue (optional[bool]): If the handler should be passed the
             update queue as a keyword argument called ``update_queue``. It can
             be used to insert updates. Default is ``False``
+        pass_job_queue (optional[bool]): If the callback should be passed the job queue as a
+            keyword argument called ``job_queue``. It can be used to schedule new jobs.
+            Default is ``False``
     """
```
tsnoam: @jh0ker 
I find the whole docstring very confusing. Maybe this need rewrite as well?
example:
```When set to true, the callback (what callback? the function just called? soemthing else?) will be passed to the job queue as a keyword argument called ``job_queue``. It can be used to schedule new jobs.```

File: telegram/ext/commandhandler.py [hunk 3]
```diff
@@ -41,10 +41,20 @@ class CommandHandler(Handler):
         pass_update_queue (optional[bool]): If the handler should be passed the
             update queue as a keyword argument called ``update_queue``. It can
             be used to insert updates. Default is ``False``
+        pass_job_queue (optional[bool]): If the callback should be passed the job queue as a
+            keyword argument called ``job_queue``. It can be used to schedule new jobs.
+            Default is ``False``
     """
```
tsnoam: Same comment as above

telegram/ext/dispatcher.py
```diff
@@ -90,11 +90,16 @@ class Dispatcher(object):
...
         update_queue (Queue): The synchronized queue that will contain the
             updates.
+        job_queue (Optional[telegram.ext.JobQueue]): The ``JobQueue`` instance to pass onto handler
+            callbacks
+        workers (Optional[int]): Number of maximum concurrent worker threads for the ``@run_async``
+            decorator
     """
 
-    def __init__(self, bot, update_queue, workers=4, exception_event=None):
+    def __init__(self, bot, update_queue, job_queue=None, workers=4, exception_event=None):
```
tsnoam: Please add `job_queue=None` to the end of the keyword args for backward compatibility with users using unnamed args.

telegram/ext/choseninlineresulthandler.py
```diff
@@ -35,10 +35,15 @@ class ChosenInlineResultHandler(Handler):
         pass_update_queue (optional[bool]): If the handler should be passed the
             update queue as a keyword argument called ``update_queue``. It can
             be used to insert updates. Default is ``False``
+        pass_job_queue (optional[bool]): If the callback should be passed the job queue as a
+            keyword argument called ``job_queue``. It can be used to schedule new jobs.
+            Default is ``False``
     """
```
tsnoam: same comment as for `callbackqueryhandler.py`

telegram/ext/handler.py
```diff
@@ -34,11 +34,15 @@ class Handler(object):
         pass_update_queue (optional[bool]): If the callback should be passed
             the update queue as a keyword argument called ``update_queue``. It
             can be used to insert updates. Default is ``False``
+        pass_job_queue (optional[bool]): If the callback should be passed the job queue as a
+            keyword argument called ``job_queue``. It can be used to schedule new jobs.
+            Default is ``False``
     """
```
tsnoam: same comment as for `callbackqueryhandler.py`

telegram/ext/inlinequeryhandler.py
```diff
@@ -34,10 +34,15 @@ class InlineQueryHandler(Handler):
         pass_update_queue (optional[bool]): If the handler should be passed the
             update queue as a keyword argument called ``update_queue``. It can
             be used to insert updates. Default is ``False``
+        pass_job_queue (optional[bool]): If the callback should be passed the job queue as a
+            keyword argument called ``job_queue``. It can be used to schedule new jobs.
+            Default is ``False``
     """
```
tsnoam: same comment as for `callbackqueryhandler.py`

telegram/ext/messagehandler.py
```diff
@@ -102,10 +102,15 @@ class MessageHandler(Handler):
         pass_update_queue (optional[bool]): If the handler should be passed the
             update queue as a keyword argument called ``update_queue``. It can
             be used to insert updates. Default is ``False``
+        pass_job_queue (optional[bool]): If the callback should be passed the job queue as a
+            keyword argument called ``job_queue``. It can be used to schedule new jobs.
+            Default is ``False``
     """
```
tsnoam: same comment as for `callbackqueryhandler.py`

telegram/ext/regexhandler.py
```diff
@@ -48,15 +48,21 @@ class RegexHandler(Handler):
         pass_update_queue (optional[bool]): If the handler should be passed the
             update queue as a keyword argument called ``update_queue``. It can
             be used to insert updates. Default is ``False``
+        pass_job_queue (optional[bool]): If the callback should be passed the job queue as a
+            keyword argument called ``job_queue``. It can be used to schedule new jobs.
+            Default is ``False``
     """
```
tsnoam: same comment as for `callbackqueryhandler.py`

telegram/ext/stringcommandhandler.py
```diff
@@ -39,10 +39,20 @@ class StringCommandHandler(Handler):
         pass_update_queue (optional[bool]): If the handler should be passed the
             update queue as a keyword argument called ``update_queue``. It can
             be used to insert updates. Default is ``False``
+        pass_job_queue (optional[bool]): If the callback should be passed the job queue as a
+            keyword argument called ``job_queue``. It can be used to schedule new jobs.
+            Default is ``False``
     """
```
tsnoam: same comment as for `callbackqueryhandler.py`

telegram/ext/stringregexhandler.py
```diff
@@ -47,15 +47,21 @@ class StringRegexHandler(Handler):
         pass_update_queue (optional[bool]): If the handler should be passed the
             update queue as a keyword argument called ``update_queue``. It can
             be used to insert updates. Default is ``False``
+        pass_job_queue (optional[bool]): If the callback should be passed the job queue as a
+            keyword argument called ``job_queue``. It can be used to schedule new jobs.
+            Default is ``False``
     """
```
tsnoam: same comment as for `callbackqueryhandler.py`

telegram/ext/typehandler.py
```diff
@@ -37,10 +37,20 @@ class TypeHandler(Handler):
         pass_update_queue (optional[bool]): If the handler should be passed the
             update queue as a keyword argument called ``update_queue``. It can
             be used to insert updates. Default is ``False``
+        pass_job_queue (optional[bool]): If the callback should be passed the job queue as a
+            keyword argument called ``job_queue``. It can be used to schedule new jobs.
+            Default is ``False``
     """
```
tsnoam: same comment as for `callbackqueryhandler.py`

telegram/ext/jobqueue.py
jh0ker: Hmm, you could always write a mock JobQueue, but I see your point. I'll get back to you about that.

telegram/ext/callbackqueryhandler.py
jh0ker: `callback` is the first argument to `__init__`, also you added a `to` there. It's actually the other way: `job_queue´ is passed to the `callback` function.

telegram/ext/dispatcher.py
jh0ker: As the dispatcher is rarely used on its own, I figured it would be worth it, but if you think we should be safe here, I'll change it :)

telegram/ext/callbackqueryhandler.py
tsnoam: @jh0ker
 just shows you how much I got confused 😉
another thought:
always pass job_queue and update_queue as keyword args 
then no documentation is needed here

telegram/ext/dispatcher.py
tsnoam: @jh0ker
unless there's a good reason to break backward compatibility here I think it's better be the last keyword arg

telegram/ext/callbackqueryhandler.py
jh0ker: I'll change the wording ;)
Always passing it means that all callback functions have to at least accept `**kwargs`, which they currently don't have to (and I'd like to keep it that way)

telegram/ext/jobqueue.py
jh0ker: @shouya I changed it so it takes the used `bot` arg as parameter and forwards it to the callback

Issue #302, opened 2016-05-22
Bot API 2.1


leandrotoledo:
May 22, 2016

- Bot API 2.1. Added more tools for group administrator bots. Your bot can now get a list of administrators and members count in a group, check a user's current status (administrator, creator, left the group, kicked from the group), and leave a group.
- Added new methods: getChat, leaveChat, getChatAdministrators, getChatMember, getChatMembersCount.
- Added support for edited messages and new mentions from Telegram v.3.9. New fields: edited_message in Update, edit_date in Message, user in MessageEntity. New value text_mention for the type field in MessageEntity.

leandrotoledo:
Some changes should come within this class:

- [ ] Add `**kwargs` to all API classes

leandrotoledo:
- [ ] New method: `telegram.Bot.getChat`
- [ ] New method: `telegram.Bot.leaveChat`
- [ ] New method: `telegram.Bot.getChatAdministrators`
- [ ] New method: `telegram.Bot.getChatMember`
- [ ] New method: `telegram.Bot.getChatMembersCount`
- [ ] New argument: `telegram.Update::edited_message`
- [ ] New argument: `telegram.Message::edit_date`
- [ ] New argument: `telegram.MessageEntity::user`
- [ ] New type: `telegram.MessageEntity::type(text_mention)`

leandrotoledo:
For those how want to give it a shot before the final release:

```
$ pip install https://github.com/python-telegram-bot/python-telegram-bot/archive/bot2.1.zip

or

$ git clone -b bot2.1 --single-branch https://github.com/python-telegram-bot/python-telegram-bot
$ cd python-telegram-bot
$ python setup.py install
```

Pending:
* Tests
* Docstrings

Not ready for production!

leandrotoledo:
I've been thinking, maybe we should move some @staticmethods to model-level functions in `telegram/utils/__init__.py`

Suggestions:

- [ ] telegram.utils.validate_token
- [ ] telegram.utils.deprecate
- [ ] telegram.message._fromtimestamp
- [ ] telegram.message._totimestamp
- [ ] telegram.inputfile.is_image

leandrotoledo:
PR #308 has been submitted for review.

PR #308, opened 2016-05-26
New methods for Bot 2.1 API

leandrotoledo:
List of new methods and details about 2.1 API here #302 

Main bullet points as follows:
- pre-commit doesn't check yapf style in examples folder
- added `**kwargs` to all API calls
- new error `BadRequest` (inherits `NetworkError` for backwards compatibility)
- changing order in factory `InputMessageContent` (because `**kwargs` may create a `InputLocationContent` instead of `InputVenueContent`)
- tests (by @rahiel - kudos!)
- test if fail for `testLeaveChat` (due the nature of the test)

Assigning to @jh0ker


telegram/bot.py
```diff
@@ -1302,3 +1462,8 @@ def __reduce__(self):
     edit_message_reply_markup = editMessageReplyMarkup
     get_updates = getUpdates
     set_webhook = setWebhook
+    leave_chat = leaveChat
+    get_chat = getChat
+    get_chat_administrators = getChatAdministrators
+    get_chat_member = getChatMember
+    get_chat_member_count = getChatMembersCount
```
jh0ker: should be `get_chat_members_count`

jh0ker:
LGTM!

jh0ker:
However, we need a new Handler class for Updates with `edited_message`

leandrotoledo:
@jh0ker can't we?

```
    @staticmethod
    def text(update):
+       message = update.message or update.edited_message
        return message.text and not message.text.startswith('/')

    @staticmethod
    def command(update):
+       message = update.message or update.edited_message
        return message.text and message.text.startswith('/')

...

        if isinstance(update, Update) and (update.message +or update.edited_message):
```

Only text seems to be editable.

leandrotoledo:
Uhm, or maybe edited_text or edited_command, thus it won't break any Handlers parsing update.message currently.

jh0ker:
Yes that's good! But then we still have that problem for the command handler...

leandrotoledo:
a0bb573 LGTM.

I think this is good to be merged.

leandrotoledo:
Tests were failing because test_filters still had Update as argument for filters, should be fixed on 80fbe98

PR #310, opened 2016-05-26
getUpdates: fix timeout default and semantics

### LINKED ISSUE #309: Network is unreachable
```
Python 2.7.9 (default, Mar  8 2015, 00:52:26) 
[GCC 4.9.2] on linux2
Type "copyright", "credits" or "license()" for more information.
>>> import telegram
>>> bot = telegram.Bot(token='123123:My_token')
>>> print(bot.getMe())
{'username': u'My_bot', 'first_name': u'Flexget', 'last_name': '', 'type': '', 'id': 123123}
>>> updates = bot.getUpdates()

  File "<pyshell#3>", line 1, in <module>
    updates = bot.getUpdates()
  File "/usr/local/lib/python2.7/dist-packages/telegram/bot.py", line 109, in decorator
  File "/usr/local/lib/python2.7/dist-packages/telegram/bot.py", line 1215, in getUpdates
    result = request.post(url, data, timeout=urlopen_timeout)
  File "/usr/local/lib/python2.7/dist-packages/telegram/utils/request.py", line 83, in decorator
    raise NetworkError('URLError: {0}'.format(error.reason))
NetworkError: URLError: [Errno 101] Network is unreachable
>>> chat_id = bot.getUpdates()[-1].message.chat_id

  File "<pyshell#4>", line 1, in <module>
    chat_id = bot.getUpdates()[-1].message.chat_id
  File "/usr/local/lib/python2.7/dist-packages/telegram/bot.py", line 109, in decorator
  File "/usr/local/lib/python2.7/dist-packages/telegram/bot.py", line 1215, in getUpdates
    result = request.post(url, data, timeout=urlopen_timeout)
  File "/usr/local/lib/python2.7/dist-packages/telegram/utils/request.py", line 83, in decorator
    raise NetworkError('URLError: {0}'.format(error.reason))
NetworkError: URLError: [Errno 101] Network is unreachable
>>> 
```

My network is okay.

tsnoam:
getUpdates() is a special method and REST RPC. Telegram servers allow a
special keyword 'timeout' to specify a timeout for the call in their
servers.
However, a bot developer should also take into account more parameters
like the network latency, and momentary slow SSL handshake with the
telegram server.

In addition, we have (unfortunately) confused the naive user by hiding
these parameters from him, trying unsuccesfully to make magic under the
hood.

So this commit brings back control and simplicity to the user by keeping
a much cleaner & closer semantics to the underlying layers:
1. timeout == telegram timeout on the REST RPC
2. total_timeout == timeout passed to urlopen. should include telegram
                    timeout and all other variables

fixes #309

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -26,6 +26,7 @@ from telegram import (User, Message, Update, UserProfilePhotos, File, ReplyMarku
                       NullHandler)
 from telegram.error import InvalidToken
 from telegram.utils import request
+from telegram.utils.deprecate import deprecate_network_delay
 
 logging.getLogger(__name__).addHandler(NullHandler())
 
@@ -982,10 +983,6 @@ class Bot(TelegramObject):
         Keyword Args:
             timeout (Optional[float]): If this value is specified, use it as
                 the definitive timeout (in seconds) for urlopen() operations.
-            network_delay (Optional[float]): If using the timeout (which is
-                a `timeout` for the Telegram servers operation),
-                then `network_delay` as an extra delay (in seconds) to
-                compensate for network latency. Defaults to 2.
 
         Returns:
             bool: On success, `True` is returned.
@@ -1182,36 +1179,37 @@ class Bot(TelegramObject):
         return url, data
 
     @log
-    def getUpdates(self, offset=None, limit=100, timeout=0, network_delay=.2):
+    def getUpdates(self, offset=None, limit=100, timeout=0, total_timeout=None, **kwargs):
         """Use this method to receive incoming updates using long polling.
 
         Args:
           offset:
-            Identifier of the first update to be returned. Must be greater by
-            one than the highest among the identifiers of previously received
-            updates. By default, updates starting with the earliest unconfirmed
-            update are returned. An update is considered confirmed as soon as
-            getUpdates is called with an offset higher than its update_id.
+            Identifier of the first update to be returned. Must be greater by one than the highest
+            among the identifiers of previously received updates. By default, updates starting with
+            the earliest unconfirmed update are returned. An update is considered confirmed as soon
+            as `getUpdates` is called with an offset higher than its update_id.
+
           limit:
-            Limits the number of updates to be retrieved. Values between 1-100
-            are accepted. Defaults to 100.
+            Limits the number of updates to be retrieved. Values between 1-100 are accepted.
+            Defaults to 100.
+
           timeout:
-            Timeout in seconds for long polling. Defaults to 0, i.e. usual
-            short polling.
-          network_delay:
-            Additional timeout in seconds to allow the response from Telegram
-            to take some time when using long polling. Defaults to 2, which
-            should be enough for most connections. Increase it if it takes very
-            long for data to be transmitted from and to the Telegram servers.
+            Timeout in seconds for long polling. Defaults to 0, i.e. usual short polling.
+
+          total_timeout (Optional[float]):
+            Timeout in seconds for the whole operation (`timeout` on telegram servers + network
+            latency). `total_timeout` will only be respected if when
+            `total_timeout` > `timeout` > 0
 
         Returns:
-            list[:class:`telegram.Update`]: A list of :class:`telegram.Update`
-            objects are returned.
+            list[:class:`telegram.Update`]: A list of :class:`telegram.Update` objects are
+            returned.
 
         Raises:
             :class:`telegram.TelegramError`
 
         """
+        total_timeout = deprecate_network_delay(timeout, total_timeout, **kwargs)
 
         url = '{0}/getUpdates'.format(self.base_url)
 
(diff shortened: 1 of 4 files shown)


telegram/bot.py
```diff
@@ -1170,31 +1166,31 @@ def editMessageReplyMarkup(
         return url, data
 
     @log
-    def getUpdates(self, offset=None, limit=100, timeout=0, network_delay=.2):
+    def getUpdates(self, offset=None, limit=100, timeout=0, total_timeout=None):
```
leandrotoledo: This will break for those who are using network_delay, but #310 adds **kwargs to all API calls, so I think we'll be good.

telegram/bot.py
```diff
@@ -1170,31 +1166,31 @@ def editMessageReplyMarkup(
...
-            Additional timeout in seconds to allow the response from Telegram
-            to take some time when using long polling. Defaults to 2, which
-            should be enough for most connections. Increase it if it takes very
-            long for data to be transmitted from and to the Telegram servers.
+            Timeout in seconds for long polling. Defaults to 0, i.e. usual short polling.
+
+          total_timeout (Optional[float]):
+            Timeout in seconds for the whole operation (`timeout` on telegram servers + network
+            latency). `total_timeout` will only be respected if when
+            `timeout` < 0 < `total_timeout`.
```
rahiel: >  `timeout` < 0 < `total_timeout`

Timeout is negative? It's different in the code `total_timeout > timeout > 0`

telegram/bot.py
```diff
@@ -1210,7 +1206,10 @@ def getUpdates(self, offset=None, limit=100, timeout=0, network_delay=.2):
         if limit:
             data['limit'] = limit
 
-        urlopen_timeout = timeout + network_delay
+        if total_timeout is not None and total_timeout > timeout > 0:
+            urlopen_timeout = timeout + total_timeout
+        else:
+            urlopen_timeout=None
```
leandrotoledo: `urlopen_timeout = None`

telegram/ext/updater.py
```diff
@@ -112,7 +112,7 @@ def _thread_wrapper(self, target, *args, **kwargs):
     def start_polling(self,
                       poll_interval=0.0,
                       timeout=10,
-                      network_delay=2,
+                      total_timeout=15,
```
leandrotoledo: Maybe we should add a **kwargs to this method.

leandrotoledo:
Minor comments made by @rahiel and me, other than that, it does look good to me (and also works from my desktop which has high latency to Telegram endpoints).

telegram/bot.py
tsnoam: @leandrotoledo obviously this must be added to release notes

telegram/bot.py
tsnoam: @rahiel  good catch, thanks

telegram/ext/updater.py
tsnoam: @leandrotoledo i'm not so sure. i prefer failures rather than silent errors

tsnoam: as agreed on the dev group, we'll do `warning.warn` to inform the user the `network_delay` had been ignored.

telegram/ext/updater.py
```diff
@@ -137,14 +139,16 @@ def start_polling(self,
...
+
         with self.__lock:
             if not self.running:
                 self.running = True
 
                 # Create & start threads
                 self._init_thread(self.dispatcher.start, "dispatcher")
                 self._init_thread(self._start_polling, "updater", poll_interval, timeout,
-                                  network_delay, bootstrap_retries, clean)
+                                  total_timeout, bootstrap_retries, clean)
```
jh0ker: This won't be backwards compatible - why would we continue to support network_delay on getUpdates and not here?

jh0ker:
LGTM

rahiel:
I think this may reintroduce issue #89, fixed in #90. Users using the `timeout` keyword but without a `total_timeout` will get that issue again, because then urlopen [here](https://github.com/tsnoam/python-telegram-bot/blob/fb0d778eb4e79476bbbe61bb0268a55d36ab2b57/telegram/utils/request.py#L148) will not get any timeout, and last year that caused `getUpdates` sometimes to never return. It was a "rare" issue.

Maybe we can make `total_timeout` an obligatory keyword if you use `timeout`? Or make it into a tuple of `(telegram_timeout, total_timeout)`?

tsnoam:
after long discussion on this subject on the developers groups we decided that this is eventually not the right approach. so closing this PR (new one will come instead soon enough)
the agreed approach was to retain current semantics and only change the `network_delay` to be 5 seconds (instead of 0.2 as of today)
reasons:
* backward compatibility (maintaining the current semantics)
* the problem itself is in fact that we've used too short network_delay. this delay was calculated using a bot running in Germany, literally next door to the Telegram bot servers farm and when the servers were much less under load (so SSL handshakes for instances were much quicker)

PR #314, opened 2016-05-28
Move large parts from README to Wiki

jh0ker:
The README was getting too big and hard to manage.
I moved the sections API, Extensions and JobQueue to different Wiki articles and added more explanations, now that space is not an issue.


rahiel:
I am happy with the wiki, things look more professional now :smile:.

PR #325, opened 2016-06-12
more robust echobot, let roboed go

rahiel:
I had a weird crash where I received an update without a message, so the call to `update.message.text` crashed it. This takes care of that and the case where an update leads to an `Unauthorized` exception when `len(bot.getUpdates()) > 1`.

We also say goodbye to `roboed.py`, it does not teach more than echobot and I am too lazy to update its exception handling code.


examples/echobot.py
```diff
@@ -25,29 +28,29 @@ def main():
...
-def echo(bot, update_id):
-
+def echo(bot):
+    global update_id
     # Request updates after the last update_id
     for update in bot.getUpdates(offset=update_id, timeout=10):
         # chat_id is required to reply to any message
         chat_id = update.message.chat_id
         update_id = update.update_id + 1
+        if not update.message:  # we ignore updates without messages
```
leandrotoledo: Updates without messages could be edited messages, nice catch.

examples/echobot.py
```diff
@@ -25,29 +28,29 @@ def main():
...
+            continue
         message = update.message.text
 
         if message:
```
leandrotoledo: Maybe for simplicity we could go with (and remove the ``if not update.message``):
```
if update.message:
   bot.sendMessage(chat_id=chat_id, text=update.message.text)
```

But it's your call.

leandrotoledo:
Minor comments, other than that looks good to me.

``So long and thanks for all the fish roboed.py``

examples/echobot.py
rahiel: That looks much better!

leandrotoledo:
Nailed it! Good to be merged.

PR #329, opened 2016-06-21
Remove annoying http.server logging

dev-dsp:


jh0ker:
Hello and thank you for your contribution!
However, maybe we should not completely remove that logging. Instead, why not log everything to the `logging` module instead, at DEBUG level? That way, you can still see incoming requests for debugging purposes

jh0ker:
@dev-dsp Are you planning to fix this or should I close it?

I'd say we should just take the original code, remove the timestamp (as it's included in logging anyways) and log it to `logging`

```python
    def log_message(self, format, *args):
        sys.stderr.write("%s - - [%s] %s\n" %
                         (self.address_string(),
                          self.log_date_time_string(),
                          format%args))
```
to
```python
    def log_message(self, format, *args):
        self.logger.debug("%s - - %s" %
                                    (self.address_string(),
                                     format % args))
```

jh0ker:
closing, I'll implement the solution mentioned and open a new PR.

PR #344, opened 2016-07-11
Move webhook handler logs to logging at DEBUG level

jh0ker:
As pomised in #329

### DIFF
--- a/telegram/utils/webhookhandler.py
+++ b/telegram/utils/webhookhandler.py
@@ -114,7 +114,7 @@ class WebhookHandler(BaseHTTPServer.BaseHTTPRequestHandler, object):
         string contains any % escapes requiring parameters, they should be specified as subsequent
         arguments (it's just like printf!).
 
-        The client ip and current date/time are prefixed to every message.
+        The client ip is prefixed to every message.
         """
 
         self.logger.debug("%s - - %s" % (self.address_string(), format % args))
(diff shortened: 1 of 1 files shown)


telegram/utils/webhookhandler.py
```diff
@@ -102,3 +102,19 @@ def _get_content_len(self):
...
+
+        This is used by all other logging functions.
+
+        It overrides ``BaseHTTPRequestHandler.log_message``, which logs to ``sys.stderr``.
+
+        The first argument, FORMAT, is a format string for the message to be logged.  If the format
+        string contains any % escapes requiring parameters, they should be specified as subsequent
+        arguments (it's just like printf!).
+
+        The client ip and current date/time are prefixed to every message.
```
MWeesenaar: Date/time is only prefixed when the user's logging configuration. 
As discussed in a private chat, this line is probably copy pasted. Therefore a fix suggestion is to fix the 'and current date/time' part.

telegram/utils/webhookhandler.py
```diff
@@ -102,3 +102,19 @@ def _get_content_len(self):
...
+        """
+
+        self.logger.debug("%s - - %s" % (self.address_string(), format % args))
```
MWeesenaar: PEP8 recommends to use logging in the form of:
`self.logger.debug("%s - - %s", self.address_string(), format % args)`
See: https://docs.python.org/2/howto/logging.html#logging-variable-data

MWeesenaar:
Added two lines of comment on the PR. 
- Documentation is not entirely correct (expected output);
- Added a suggestion for different string formatting in logging.

MWeesenaar:
LGTM!

PR #353, opened 2016-07-18
ChatMigrated exception

overquota:
When converting telegram group (OLD_CHAT_ID) to supergroup (NEW_CHAT_ID), chat id changes
After this, when trying to post message to OLD_CHAT_ID, Telegram API returns error with message:
{u'error_code': 400, u'ok': False, u'description': u'Bad Request: group chat is migrated to a supergroup chat', u'parameters': {u'migrate_to_chat_id': NEW_CHAT_ID}}

So this pull requests creates new exception ChatMigrated which returns NEW_CHAT_ID
With this functionality we can postpone fix our inconsistency data in database when user converts group to supergroup and we cannot update this info in database at the same time (eg datacenter failed or network lag)

Sorry for my English :)

### DIFF
--- a/telegram/error.py
+++ b/telegram/error.py
@@ -97,5 +97,5 @@ class ChatMigrated(TelegramError):
         Returns:
 
         """
-        super(ChatMigrated, self).__init__('Chat migrated to {0}'.format(new_chat_id))
+        super(ChatMigrated, self).__init__('Chat migrated')
         self.new_chat_id = new_chat_id
--- a/telegram/error.py
+++ b/telegram/error.py
@@ -97,5 +97,5 @@ class ChatMigrated(TelegramError):
         Returns:
 
         """
-        super(ChatMigrated, self).__init__('Chat migrated to {}'.format(new_chat_id))
+        super(ChatMigrated, self).__init__('Chat migrated to {0}'.format(new_chat_id))
         self.new_chat_id = new_chat_id
--- a/telegram/error.py
+++ b/telegram/error.py
@@ -39,13 +39,10 @@ def _lstrip_str(in_s, lstr):
 class TelegramError(Exception):
     """This object represents a Telegram Error."""
 
-    def __init__(self, message, **kwargs):
+    def __init__(self, message):
         """
         Args:
             message (str):
-
-        Kwargs:
-            chat_id (int):
 
         Returns:
 
@@ -59,7 +56,6 @@ class TelegramError(Exception):
             # api_error - capitalize the msg...
             msg = msg.capitalize()
         self.message = msg
-        self.chat_id = kwargs.get('chat_id')
 
     def __str__(self):
         return '%s' % (self.message)
@@ -92,4 +88,14 @@ class TimedOut(NetworkError):
 
 
 class ChatMigrated(TelegramError):
-    pass
+
+    def __init__(self, new_chat_id):
+        """
+        Args:
+            new_chat_id (int):
+
+        Returns:
+
+        """
+        super(ChatMigrated, self).__init__('Chat migrated to {}'.format(new_chat_id))
+        self.new_chat_id = new_chat_id
--- a/telegram/error.py
+++ b/telegram/error.py
@@ -39,10 +39,13 @@ def _lstrip_str(in_s, lstr):
 class TelegramError(Exception):
     """This object represents a Telegram Error."""
 
-    def __init__(self, message):
+    def __init__(self, message, **kwargs):
         """
         Args:
             message (str):
+
+        Kwargs:
+            chat_id (int):
 
         Returns:
 
@@ -56,6 +59,7 @@ class TelegramError(Exception):
             # api_error - capitalize the msg...
             msg = msg.capitalize()
         self.message = msg
+        self.chat_id = kwargs.get('chat_id')
 
     def __str__(self):
         return '%s' % (self.message)
@@ -85,6 +89,7 @@ class TimedOut(NetworkError):
 
     def __init__(self):
         super(TimedOut, self).__init__('Timed out')
+
 
 class ChatMigrated(TelegramError):
     pass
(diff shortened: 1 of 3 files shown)


telegram/utils/request.py
```diff
@@ -135,12 +135,14 @@ def _parse(json_data):
     except ValueError:
         raise TelegramError('Invalid server response')
 
-    if not data.get('ok') and data.get('description'):
-        return data['description']
+    if not data.get('ok'):
+        if data.get('parameters') and data['parameters'].get('migrate_to_chat_id'):
```
tsnoam: please save `data.get('parameters')` into a variable, so you won't need to get it 3 times.

telegram/utils/request.py
```diff
@@ -135,12 +135,14 @@ def _parse(json_data):
...
+            raise ChatMigrated(str(data['parameters']['migrate_to_chat_id']))
+        if data.get('description'):
+            return data['description']
```
tsnoam: while you're at it, can you save `data.get('description')` to a variable as well?

telegram/utils/request.py
```diff
@@ -135,12 +135,14 @@ def _parse(json_data):
...
 
     return data['result']
 
-
```
tsnoam: this extra line is required for pep8 conformance. please bring it back.

telegram/utils/request.py
```diff
@@ -135,12 +135,14 @@ def _parse(json_data):
     except ValueError:
         raise TelegramError('Invalid server response')
 
-    if not data.get('ok') and data.get('description'):
-        return data['description']
+    if not data.get('ok'):
+        if data.get('parameters') and data['parameters'].get('migrate_to_chat_id'):
+            raise ChatMigrated(str(data['parameters']['migrate_to_chat_id']))
```
tsnoam: afaik the chat_id is of type `int`, so I suggest you replace `str()` with `int()`.

telegram/error.py
```diff
@@ -85,3 +85,6 @@ class TimedOut(NetworkError):
 
     def __init__(self):
         super(TimedOut, self).__init__('Timed out')
+
```
tsnoam: you need an additional line for pep8 conformance

tsnoam:
@overquota 
Thank you for your contribution. Generally it looks good and we'd be happy to merge it. However:

1 Please see the comments I've put on the code.
2. The CI failed due to yapf test (an improved pep8 and style enforcer). Please check the "contribution guide" for more info about that.
3. You should also add yourself to the AUTHORS.rst file as described in the "contribution guide". :-)

overquota:
Hi! Thank you for comments!
the TelegramError exception waits only string in args, so i decided to do str(chat_id)
I rewrited TelegramError exception, so now it know how to process optional args (kwargs) and chat_id in it
Unfortunately I don't know is it a good decision

telegram/utils/request.py
```diff
@@ -136,13 +136,19 @@ def _parse(json_data):
...
-            return data['description']
+        description = data.get('description')
+        parameters = data.get('parameters')
+        if description:
+            if parameters:
+                migrate_to_chat_id = parameters.get('migrate_to_chat_id')
+                if migrate_to_chat_id:
+                    raise ChatMigrated(description, chat_id=int(migrate_to_chat_id))
+            else:
+                return description
```
tsnoam: The above code is correct with the JSON reply.
However, it is better to be more liberal with the way we interpret the replies from Telegram servers, especially that they like making changes without any prior notification.
So... please do something more similar to your original commit:
```
if parameters:
    if migrate... :
        raise
if description:
    return description
```

telegram/error.py
```diff
@@ -39,11 +39,14 @@ def _lstrip_str(in_s, lstr):
 class TelegramError(Exception):
     """This object represents a Telegram Error."""
 
-    def __init__(self, message):
+    def __init__(self, message, **kwargs):
```
tsnoam: as you suggested, this is not a good idea :-(
see suggestion on the `ChatMigrated` code below

telegram/error.py
```diff
@@ -86,5 +90,6 @@ class TimedOut(NetworkError):
     def __init__(self):
         super(TimedOut, self).__init__('Timed out')
 
+
 class ChatMigrated(TelegramError):
     pass
```
tsnoam: you can do something like this:
```
def __init__(self, new_chat_id):
    super(ChatMigrated, self).__init__('Chat migrated')
    self.new_chat_id = new_chat_id
```

tsnoam:
Hi @overquota ,
This looks much better, though there are now new comments on the code.

overquota:
Hmm, last change breaked CI, but I can't reproduce it:
`python-telegram-bot   master  $ cat test.py
#!/usr/bin/env python
from telegram import Bot
import sys

print sys.version
bot = Bot('1234:abcd1234')
bot.getMe()`

`python-telegram-bot   master  $ ./test.py
2.7.11 (default, Jan 22 2016, 08:29:18)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)]
  File "./test.py", line 7, in <module>
    bot.getMe()
  File "/Users/aarseniev/tmp/python-telegram-bot/telegram/bot.py", line 121, in decorator
  File "/Users/aarseniev/tmp/python-telegram-bot/telegram/bot.py", line 172, in getMe
    result = request.get(url)
  File "/Users/aarseniev/tmp/python-telegram-bot/telegram/utils/request.py", line 204, in get
    result = _request_wrapper('GET', url)
  File "/Users/aarseniev/tmp/python-telegram-bot/telegram/utils/request.py", line 185, in _request_wrapper
    raise Unauthorized()
telegram.error.Unauthorized: Unauthorized`

overquota:
Seems like CI failing on some temp reasons (timeout?)

rahiel:
> Seems like CI failing on some temp reasons (timeout?)

@overquota thanks for your work! Our Travis tests sometimes fail for no good reason, but they're all good now.

@tsnoam, time to merge?

PR #362, opened 2016-07-30
Class methods

jh0ker:
PR to keep track of development.

The idea of this PR is to add shortcut-methods to the classes defined by the API.

**Example:**
To reply to a message with another text message, you would have to do 
```python
bot.sendMessage(update.message.chat_id, text="Reply")
```
Now, you can also do
```python
update.message.reply_text("Reply")
```
which is nothing more than a shortcut for `bot.sendMessage(update.message.chat_id, *args, **kwargs)`.

**List of methods planned:**

- `Message`
 - [x] `reply_text`
 - [x] `forward`
 - [x] `reply_photo`
 - [x] `reply_audio`
 - [x] `reply_document`
 - [x] `reply_sticker`
 - [x] `reply_video`
 - [x] `reply_voice`
 - [x] `reply_location`
 - [x] `reply_venue`
 - [x] `reply_contact`
- `Chat`
 - [x] `send_action`
 - [x] `kick_member`
 - [x] `leave`
 - [x] `unban_member`
 - [x] `get` (static?)
 - [x] `get_administrators`
 - [x] `get_members_count`
 - [x] `get_member`
- `User`
 - [x] `get_profile_photos`
- `File`
 - [x] `get` (static?)
- `InlineQuery`
 - [x] `answer`
- `CallbackQuery`
 - [x] `answer`


bomjacob:
Any chance we could add `send_chat_action` to `Message` and `Chat` too?

jh0ker:
It's already on Chat, I think that is enough, since they are both look linked

bomjacob:
Alright yeah, that seems good. Just wasn't sure if you added it since it's not in the original list :)

jh0ker:
screenshot_20160812-214938

jh0ker:
Sorry, just got back to my laptop and saw the huge screenshot. But I'll leave it that way :sunglasses:

jh0ker:
I think we should leave the static `get` methods, they seem a bit confusing and not really worth it

KamranMackey:
@jh0ker Just do them. They aren't confusing.

jh0ker:
@KamranMackey The problem is that you need an instance of the class before you could use them, because only the instance can have a reference to the `telegram.Bot` instance which is needed to make the API call.

So it couldn't be a static method after all.

KamranMackey:
@jh0ker Oh, okay. Just leave them then.

telegram/bot.py
```diff
@@ -1179,6 +1181,26 @@ def editMessageReplyMarkup(self,
 
         return url, data
 
+    def create_references(self, d):
+        """
+        Recursively reates a reference to this Bot instance in the dict and all contained dicts and
```
tsnoam: Recursively *iterates*?

telegram/bot.py
```diff
@@ -1222,10 +1244,14 @@ def getUpdates(self, offset=None, limit=100, timeout=0, network_delay=5., **kwar
 
         if result:
             self.logger.debug('Getting updates: %s', [u['update_id'] for u in result])
+
```
tsnoam: dangling space

telegram/bot.py
```diff
@@ -1179,6 +1181,26 @@ def editMessageReplyMarkup(self,
 
         return url, data
 
+    def create_references(self, d):
```
tsnoam: pep8 requires 3 letters variables

telegram/bot.py
```diff
@@ -154,6 +154,8 @@ def decorator(self, *args, **kwargs):
             if result is True:
                 return result
 
+            self.create_references(result)
+
```
tsnoam: why is it needed in the `message` decorator? isn't it only for sending messages? (after you already have a bot instance...)

telegram/bot.py
```diff
@@ -1222,10 +1244,14 @@ def getUpdates(self, offset=None, limit=100, timeout=0, network_delay=5., **kwar
...
             self.logger.debug('Getting updates: %s', [u['update_id'] for u in result])
+
         else:
             self.logger.debug('No new updates found.')
 
-        return [Update.de_json(x) for x in result]
+        for u in result:
+            self.create_references(u)
+
+        return [Update.de_json(u) for u in result]
```
tsnoam: @jh0ker 
I'm conflicted here.
I know it's pretty the way you currently implemented it (explicit and self explanatory). However, on busy bots, you actually run twice on 100 updates each time this code is called. so it can actually have performance impact.

Suggestion for discussion:
Instead of having `create_references` method, the various `de_json` methods will accept `bot` as a second parameter (similar to what I've done with `File.de_json` for the `request` object) 

then you can do:
`return [Update.de_json(u, self) for u in result]`

A similar comment goes for the Webhook below

telegram/utils/webhookhandler.py
```diff
@@ -85,7 +86,10 @@ def do_POST(self):
 
             self.logger.debug('Webhook received data: ' + json_string)
 
-            update = Update.de_json(json.loads(json_string))
+            update_dict = json.loads(json_string)
+            self.server.bot.create_references(update_dict)
+            update = Update.de_json(update_dict)
```
tsnoam: See comment about performance above (in bot.py)

tsnoam:
I left some comments on the diff.
most of them were just nitpicks (pep8 and stuff) but there was one which was regarding performance and requires discussion.

telegram/bot.py
jh0ker: I would tend to agree, I added it for the tests because I used the return value of a sent message to "reply" to it.

tsnoam: I think we should look for another way. This change actually changes the behaviour of a critical path in "production" code and only for tests.

jh0ker:
@whipermr5 Yes, that is safe.

PR #372, opened 2016-08-10
added url shortener to botan

njittam:
Botan supports a url shortener to get more information about your users.

### DIFF
--- a/telegram/contrib/botan.py
+++ b/telegram/contrib/botan.py
@@ -16,6 +16,8 @@ class Botan(object):
     token = ''
     url_template = 'https://api.botan.io/track?token={token}' \
                    '&uid={uid}&name={name}&src=python-telegram-bot'
+    shorten_template = 'https://api.botan.io/s?token={token}' \
+                       '&user_ids={user_ids}&url={url}&src=python-telegram-bot'
 
     def __init__(self, token):
         self.token = token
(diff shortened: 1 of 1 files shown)


telegram/contrib/botan.py
```diff
@@ -44,3 +46,23 @@ def track(self, message, event_name='event'):
...
             self.logger.warn('Botan track error ' + str(error.reason))
             return False
+
+    def shorten(self, url, user_id):
+        if not (url.startswith('http://') or url.startswith('https://')):
+            url = 'http://' + url
+        try:
+            url2 = self.shorten_template.format(token=str(self.token),
+                                                user_ids=str(user_id),
+                                                url=str(url))
```
tsnoam: `str()` is not needed here. It is implied by the string formatting.

telegram/contrib/botan.py
```diff
@@ -44,3 +46,23 @@ def track(self, message, event_name='event'):
...
+            request = Request(url2,
+                              headers={'Content-Type': 'application/json'})
```
tsnoam: we allow 99 chars in a line (per pep8 extension). can you please "join" this lines? (that probably what cause yapf in travis to fail)

telegram/contrib/botan.py
```diff
@@ -44,3 +46,23 @@ def track(self, message, event_name='event'):
         except URLError as error:
             self.logger.warn('Botan track error ' + str(error.reason))
             return False
+
+    def shorten(self, url, user_id):
+        if not (url.startswith('http://') or url.startswith('https://')):
```
tsnoam: Can you please change to a shorter form:
`if not url.startswith(('http://', 'https://'))`

telegram/contrib/botan.py
```diff
@@ -44,3 +46,23 @@ def track(self, message, event_name='event'):
...
+                                                user_ids=str(user_id),
+                                                url=str(url))
+            request = Request(url2,
+                              headers={'Content-Type': 'application/json'})
+            with urlopen(request) as response:
+                shortened_url = response.read()
+            return shortened_url.decode('utf-8')
+        except HTTPError as error:
+            self.logger.warn('Botan track error ' + str(error.code) + ':' + error.read().decode(
+                'utf-8'))
```
tsnoam: please don't use string concat here, but rather the `logging` dict formatting:
`self.logger.warn('some text %s:%s', error.code, error.read().decode('utf-8')`

telegram/contrib/botan.py
```diff
@@ -44,3 +46,23 @@ def track(self, message, event_name='event'):
...
+            return None
```
tsnoam: why not raise `TelegramError` here?

telegram/contrib/botan.py
```diff
@@ -44,3 +46,23 @@ def track(self, message, event_name='event'):
...
+        except URLError as error:
+            self.logger.warn('Botan track error ' + str(error.reason))
+            return None
```
tsnoam: same two comments as above

tsnoam:
closing as of lack of response.

PR #375, opened 2016-08-10
entities add

CapitainCrunch:
Telegram's update message has new entities, described here https://core.telegram.org/bots/API#messageentity
I've added some, and fix command, according to docs.
If bot receives smth like test@test.com /start, update has such tuple:
'entities': [{'length': 12, 'offset': 0, 'type': 'email'}, {'length': 5, 'offset': 0, 'type': 'bot_command'}]
which is neither a bot_command nor an email

Still, devs can parse an entity, using indexes of the dicts and a length in an entity list

### DIFF
--- a/telegram/ext/messagehandler.py
+++ b/telegram/ext/messagehandler.py
@@ -35,7 +35,10 @@ class Filters(object):
 
     @staticmethod
     def command(message):
-        return message.text and message.text.startswith('/')
+        if len(message.entities) == 1:
+            return bool(message.entities[0].type == 'bot_command')
+        else:
+            return False
 
     @staticmethod
     def audio(message):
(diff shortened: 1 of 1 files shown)


jh0ker:
Hey, thank you for your contribution!

I'm wondering, you only accept the message if there is only one entity in the message... that could be confusing (at least to me it is). Wouldn't it be more intuitive to check if any of the entities is the one we are searching for?

Another thing: Could we then perhaps have some kind of higher-order function that you can use like this_

```python
def entities(list_of_entities):
    # return a new function that checks the message for those entities
```
and then

```python
MessageHandler([entities(MENTION, TEXT_MENTION)])
```

What do you think?

bomjacob:
Superseeded by #409

PR #378, opened 2016-08-11
CommandHandler: don't handle commands meant to other bots

### LINKED ISSUE #319: Bot is responding to commands for other bots in group chats when privacy option is turned off
### Steps to reproduce
1. Get a bot, turn off privacy option in `@botfather`, and drag it in a group with other bots (which should respond to `/start` command in a group).
2. Set up the bot with the following code:

``` python
import telegram, telegram.ext
tb = telegram.ext.Updater("TOKEN")
def a(bot, update):
    bot.sendMessage(update.message.chat_id, "responding to /start cmd")

tb.dispatcher.add_handler(telegram.ext.CommandHandler("start", b))
tb.start_polling()
```
1. Send command `/start@some_other_bot` to the group, and your bot will respond to it.
2. Send command `/start` to the group and your bot (and in fact many other bots) will respond to it too.
### Expected behaviour

Bot should not respond to cmds not pointing to it, and better give an option not to respond to cmd with no target in a group.
### Configuration

**Operating System:**
Mac OS X El Capitan

**Version of Python:**

`$ python3 -V`

`Python 3.5.1`

**Version of python-telegram-bot:**

`$ python3 -c 'import telegram; print(telegram.__version__)'`

`4.1.1`
### Logs

### Where the issue happens

``` python
update.message.text[1:].split(' ')[0].split('@')[0] == self.command
```

on line 45 of `telegram/ext/commandhandler.py`
It is not checking for the target of command issued by user.
### Suggested changes
1. [ ] Check for commands with `@` for whether it's pointed to the bot.
2. [ ] Add a parameter to `CommandHandler`: toggle whether if the bot should respond to simple commands (command without `@name_of_the_bot`) in a group, and default it to True. (for backward compatibility, and behaviour consistancy)

franciscod:
Fixes #319

### DIFF
--- a/telegram/ext/commandhandler.py
+++ b/telegram/ext/commandhandler.py
@@ -80,6 +80,14 @@ class CommandHandler(Handler):
 
         message = update.message or update.edited_message
 
+        command = message.text[1:].split(' ')[0]
+        if '@' in command:
+            split_command = command.split('@')
+            command, target_bot_name = split_command[0], split_command[1]
+            calling_me = (target_bot_name.lower() == dispatcher.bot.username.lower())
+            if not calling_me:
+                return # don't reply, we aren't being talked to
+
         if self.pass_args:
             optional_args['args'] = message.text.split(' ')[1:]
 
--- a/telegram/ext/commandhandler.py
+++ b/telegram/ext/commandhandler.py
@@ -57,9 +57,8 @@ class CommandHandler(Handler):
                  pass_args=False,
                  pass_update_queue=False,
                  pass_job_queue=False):
-        super(CommandHandler, self).__init__(callback,
-                                             pass_update_queue=pass_update_queue,
-                                             pass_job_queue=pass_job_queue)
+        super(CommandHandler, self).__init__(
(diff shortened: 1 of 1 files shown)


franciscod:
I'm not sure where does that `username` come from... but seems to work well here: https://github.com/franciscod/telegram-twitter-forwarder-bot/blob/master/basebot.py#L98

franciscod:
Ok it get's filled here when getMe is called https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/bot.py#L174. But I don't see where get_me is called (since I don't do that in my twitterforwarder bot, but the username is there!)

franciscod:
Oh perfect. When you check the property "username" since it's decorated with `info` it will issue the getMe call if it is not there!

franciscod:
I should add some tests. Don't merge this yet!

jh0ker:
Please read https://github.com/python-telegram-bot/python-telegram-bot/issues/319#issuecomment-226533591

jh0ker:
I will close this for now as discussed, since we need a better solution for this.

franciscod:
Cool :)

Issue #373, opened 2016-08-10
Weird KeyError with ConversationHandler and CallbackQueryHandler


bomjacob:
### Steps to reproduce
1. Run code example below:
```python
import logging
import os
import sys

from telegram import InlineKeyboardButton, InlineKeyboardMarkup
from telegram.ext import (Updater, ConversationHandler, CallbackQueryHandler, CommandHandler)
from telegram.ext.dispatcher import run_async


def error(bot, update, err):
    logging.warning('Update "%s" caused error "%s"' % (update, err))


def start(bot, update):
    buttons = [[InlineKeyboardButton(text='Click me!', callback_data='Something')]]
    bot.send_message(chat_id=update.message.chat.id,
                     text='Click the button below and look at log.',
                     reply_markup=InlineKeyboardMarkup(buttons))


# If you add a @run_async here the error changes to be on line 216 of ConversationHandler
def test(bot, update):
    # Doesn't matter what's in this func
    return ConversationHandler.END


logging.basicConfig(level=logging.DEBUG, format='%(asctime)s - %(name)s - %(levelname)s - %(message)s')

token = os.getenv('TEST-TOKEN')
if not token:
    logging.critical('NO TOKEN FOUND!')
    sys.exit()

updater = Updater(token)

dp = updater.dispatcher

test_handler = ConversationHandler(
    entry_points=[
        CallbackQueryHandler(test)
    ],
    states={},
    fallbacks=[]
)

start_handler = CommandHandler('start', start)

dp.add_handler(test_handler)
dp.add_handler(start_handler)

dp.add_error_handler(error)

updater.start_polling()

updater.idle()
```

2. Send a `/start` to bot

3. Click the `Click me` inline button

### Expected behaviour
No error in log.

### Actual behaviour
A KeyError is shown in log (see logs below).
Note: It doesn't actually crash.

### Configuration
**Operating System:**
Windows 10 Education

**Version of Python, python-telegram-bot & dependencies:**
``$ python -m telegram``
python-telegram-bot 5.0.0 (newest master branch)
urllib3 1.16
certifi 2016.08.02
future 0.15.2
Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  5 2016, 11:41:13) [MSC v.1900 64 bit (AMD64)]

### Logs
https://gist.github.com/bomjacob/d9ae92f414a5c796eb86ac36b78b56d0

MWeesenaar:
[...]

PR #389, opened 2016-08-21
Fix "key not found" exception if the very first message handler in a …

LiaungYip:
…ConversationHandler returns the state ConversationHandler.END.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -18,6 +18,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `jh0ker <https://github.com/jh0ker>`_
 - `JRoot3D <https://github.com/JRoot3D>`_
 - `jlmadurga <https://github.com/jlmadurga>`_
+- `Li-aung Yip <https://github.com/LiaungYip>`_
 - `macrojames <https://github.com/macrojames>`_
 - `naveenvhegde <https://github.com/naveenvhegde>`_
 - `njittam <https://github.com/njittam>`_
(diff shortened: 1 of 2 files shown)


jh0ker:
Thank you for your contribution! Did you already add yourself to AUTHORS.rst?

LiaungYip:
@jh0ker I have added myself to AUTHORS.rst.

jh0ker:
Ah, yapf is failing. Probably because you could change this
```python
if new_state == self.END: 
    if key in self.conversations:
        del self.conversations[key]
```
to this
```python
if new_state == self.END and key in self.conversations:
    del self.conversations[key]
```

LiaungYip:
Ah, I hadn't expected to fall afoul of the build system for a single line change.

I'll fix that formatting so it's as `yapf` expects and see if the build passes... if build still fails, I'll actually follow the process in CONTRIBUTING.rst.

jh0ker:
Could you include a fix for https://github.com/python-telegram-bot/python-telegram-bot/issues/373#issuecomment-238839786 too, if it's not too much work?

jh0ker:
friendly ping @LiaungYip

jh0ker:
Thank you for your contribution @LiaungYip :)

I included a fix for the comment I linked earlier and added unit tests for these two cases. By the way, your take on the `if` turned out to be correct, I only added an explicit `else` clause:

```python
if new_state == self.END:
    if key in self.conversations:
        del self.conversations[key]
    else:
        pass
```

PR #402, opened 2016-09-01
Reusable dispatcher

### LINKED ISSUE #401: raise RuntimeError('Connection Pool already initialized') after restarting update instances
### Steps to reproduce
1.  Using multiple telegram bot update instance in gevent based server 
2. Start them with updater.start_polling()
2. Stop all updater.stop()
3. Restart them with  updater.start_polling()
 
### Expected behaviour
Make possible to run mutliple updater instance with either shared urlib3 pools or seperate pools
 
### Actual behaviour
Failed to allocate new dispatcher workers due global constraint in urllib3 connection pool

### Configuration
Linux

```
urllib3 1.10 has no metadata
python-telegram-bot 5.0.0 has no metadata
```

### Logs
```
  File "/opt/devcores/env/local/lib/python2.7/site-packages/telegram/ext/updater.py", line 85, in __init__
    exception_event=self.__exception_event)
  File "/opt/devcores/env/local/lib/python2.7/site-packages/telegram/ext/dispatcher.py", line 124, in __init__
    raise RuntimeError('Connection Pool already initialized')
```

tsnoam:
Fixes #401

### DIFF
--- a/telegram/ext/updater.py
+++ b/telegram/ext/updater.py
@@ -60,7 +60,7 @@ class Updater(object):
             functions decorated with @run_async
         bot (Optional[Bot]): A pre-initialized bot instance. If a pre-initizlied bot is used, it is
             the user's responsibility to create it using a `Request` instance with a large enough
-            connection pool
+            connection pool.
         job_queue_tick_interval(Optional[float]): The interval the queue should
             be checked for new tasks. Defaults to 1.0
 
--- a/telegram/ext/updater.py
+++ b/telegram/ext/updater.py
@@ -345,14 +345,14 @@ class Updater(object):
         while updates:
             updates = self.bot.getUpdates(updates[-1].update_id + 1)
 
-    def stop(self):
+    def stop(self, force=False):
         """
         Stops the polling/webhook thread, the dispatcher and the job queue
         """
 
         self.job_queue.stop()
         with self.__lock:
-            if self.running or self.dispatcher.has_running_threads:
+            if self.running or self.dispatcher.has_running_threads or force:
                 self.logger.debug('Stopping Updater and Dispatcher...')
 
                 self.running = False
@@ -360,6 +360,7 @@ class Updater(object):
                 self._stop_httpd()
                 self._stop_dispatcher()
                 self._join_threads()
+                self.bot.request.stop()
 
     def _stop_httpd(self):
         if self.httpd:
--- a/telegram/ext/updater.py
+++ b/telegram/ext/updater.py
@@ -352,7 +352,7 @@ class Updater(object):
 
         self.job_queue.stop()
         with self.__lock:
-            if self.running or dispatcher.ASYNC_THREADS:
+            if self.running or self.dispatcher.has_running_threads:
                 self.logger.debug('Stopping Updater and Dispatcher...')
 
                 self.running = False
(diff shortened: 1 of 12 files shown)


telegram/ext/updater.py
```diff
@@ -74,7 +79,13 @@ def __init__(self, token=None, base_url=None, workers=4, bot=None):
         if bot is not None:
             self.bot = bot
         else:
-            self.bot = Bot(token, base_url)
+            # we need a connection pool the size of:
+            # * for each of the workers
+            # * 1 for Dispatcher
+            # * 1 for polling Updater (even if webhook is used, we can spare a connection)
+            # * 1 for JobQueue
+            self._request = Request(con_pool_size=workers + 3)
```
jh0ker: This would be a good occasion to add the connection for the main thread (so +4)

tests/test_updater.py
```diff
@@ -710,6 +753,8 @@ def __init__(self,
         self.bootstrap_attempts = 0
         self.bootstrap_err = bootstrap_err
         self.edited = edited
+        self.request = MockyMock()
```
jh0ker: Is there a reason we can't use `None` here?

jh0ker:
(Edited and extended) copy from the dev chat:
Honestly, big props that you actually managed to make the singleton thread pool used for the `run_async` decorator optional.

Maybe we should remove the singleton completely in the next major version and go full class based. The downside would be that the dispatcher had to be declared globally then,  which is not nice.

Still, here is a suggestion.

Right now, we have `Dispatcher.__async_queue` and `Dispatcher.__async_threads` with this behaviour:

- In case of singleton-mode, `_ASYNC_QUEUE` and `_ASYNC_THREADS`, which are declared on module level, are assigned to the (only one) `Dispatcher` instance
- In case of multiple instances (with `no_singleton` set to `True`), a new `Queue` and `set` instance are created for each instance

I'm wondering if the following behaviour would be preferrable, especially for future development:
- Remove the module-level declarations and implement a "classic" singleton pattern by having a module-level variable called `INSTANCE` and a `threading.Event` instance called `SINGLETON_MODE` that is `set` by default

- When the `Dispatcher` is instantiated first, set the `INSTANCE` to `self`
- If a second `Dispatcher` is instantiated, `clear` the `SINGLETON_MODE` event
- In `run_async`, check if the `SINGLETON_MODE` event is set. If it is, proceed as normal and use the `INSTANCE` variable to enqueue the function. If not, raise an exception.

In my opinion, that is more explicit and obvious than the current implementation and I think will be more maintainable. 
The current implementation also allows to use the `run_async` decorator even if multiple `Dispatcher` instances are present (correct me if I'm wrong here). To me, it also smells of hotfix a bit too much ;)

Let me know what you think, and I'm sure you have some great ideas on how to improve my proposal in case you agree with the general idea.

tests/test_updater.py
tsnoam: @jh0ker 
This is actually leftovers from a stage in the development. I just removed it now...

telegram/ext/updater.py
tsnoam: @jh0ker 
I don't think we need it... but it probably doesn't hurt to add another connection.

tsnoam:
@jh0ker 
I like your approach for this. I'll put it into code and lets see what we'll get :)

jh0ker:
I love it! Really good take on the idea, this is exactly what I had in mind, just better :laughing:

jh0ker:
LGTM :+1:

PR #404, opened 2016-09-03
Add get_entity to Message. (Fixes #400)

### LINKED ISSUE #400: Handling of MessageEntity length and offset
### Steps to reproduce
1. Send a message like "👩‍👩‍👧‍👧http://google.com" (that is, a family emoji (or really any complicated unicode codepoint) followed by a link)

2. The link should appear in a MessageEntity in `update.message.entities` with the type 'url'.

3. Try to use the `length` or `offset` attribute for basically any purpose... Fx. 
```python
entity = update.message.entities[0]
link = update.message.text[entity.offset:entity.offset + entity.length]
```

### Expected behaviour
In the example above you'd expect to have "http://google.com" in the `link` variable.

### Actual behaviour
`link` contains "://google.com".

### Why
This happens because telegram servers calculate lengths in 'UTF-16' codepoints, while python clearly doesn't (and really that's good since utf-16 is bad...). This means that python sees the family emoji as 8 characters while telegram sees it as 12 characters...

### Solution
Either:

1. Patch the length and offset inside MessageEntity's such that it matches what python thinks...
2. Add some sort of util function to convert to and from utf-8 and utf-16
3. Add some sort of custom slicing util function that does the conversion internally

### Configuration
**Operating System:**
Windows 10 Education

python-telegram-bot 5.0.0
urllib3 1.16
certifi 2016.08.08
future 0.15.2
Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  5 2016, 11:41:13) [MSC v.1900 64 bit (AMD64)]

**Note**
The behaviour of wide strings (weird unicode stuff) apparently changed as of python 3.3 (it now has a certain behaviour always where before it was a build-flag or something like that), so to support all python versions, we'd probably have to do some sort of wizardry... (See discussion at https://stackoverflow.com/questions/30775689/python-length-of-unicode-string-confusion for more info (kinda))

bomjacob:
Not sure if the docstring is correct...
Also this PR should be squashed, since it has some derpy commits.


rahiel:
Good work @bomjacob! 

I think at this point it makes sense to add a `get_entities`  function that takes as argument `entity_type` (string) and returns a list of all the  matching entities, and when `entity_type` is None, it returns all of them. With that this PR is good to go.

jh0ker:
Nice work @bomjacob and @rahiel! 

I also like the idea @rahiel, and it would be best to add all possible values for `MessageEntity.type` as constants in `MessageEntity` too. 

By the way, this PR is somewhat related to #362 because both add methods to Telegram classes.

bomjacob:
Danm.. I wrote that before I saw what you wrote @jh0ker. I'll pull the possible values out and add them as a constant instead.
Also... this might be an issue (depending on what we want): `text_link` and `text_mention` has extra data than just some text, which currently wouldn't get returned by `get_entities`... Should we _somehow_ add that?

bomjacob:
Was that what you had in mind in terms of constant @jh0ker ?

jh0ker:
Actually I had something in mind like [this](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/chataction.py) or [this](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/parsemode.py).

On your other question, why not just return a list of `MessageEntity` instead of a list of strings?

bomjacob:
Want them to be in the `MessageEntity` class, or in a seperate class?
And if in the same class, is there any easy way to get them all?

jh0ker:
Yeah in the `MessageEntity` class is fine IMO. You can still have the list of all types, just add a constant for each of them too.

jh0ker:
You renamed it to `ALL_TYPES` in `messageentity.py` but forgot to change it in `message.py` ;)

bomjacob:
Forgot to push :P

bomjacob:
Don't merge yet. I wanna quickly write a test.

bomjacob:
Should this be changed?
```
Args:
     types (Optional[list]): Only extract text of MessageEntities if its type
     in in this list. Defaults to all types.
```
Like do I need to specify that it's a list of MessageEntity types somehow, or is this okay?

jh0ker:
I think most docstrings have to be updated since they are referring to text instead of entities. Additionally, it might be a good idea to document when and why you should not use `Message.entities` directly and instead use these helper methods. 

Also, maybe it would be a good idea to rename `get_entity` to `get_entity_text`, just to make clear what it does. And can't you re-use `get_entity_text` in `get_entities`? I think there is some duplicate code there.

jh0ker:
While discussing with @bomjacob, we were re-thinking the `MessageEntity.text` attribute. Right now, it's not there by default and only "magically appears" if the entity object is processed by `Message.get_entities`.

Possible courses of action:

- Leave it like it is
- Define it with `None` as default
- Make it a property and raise a meaningful exception when queried before set
- Remove it completely and instead return a `dict` from `get_entities` that maps a `MessageEntity` to their `text`

bomjacob:
Alright. It's been decited to go with: "Remove it completely and instead return a dict from get_entities that maps a MessageEntity to their text"
Latest commit should make the necessary changes and also rename the getters to something more appropriate.
Note: currently missing proper docstrings.. hold on

rahiel:
We might need an extra snippet at the end of https://github.com/python-telegram-bot/python-telegram-bot/wiki/Code-snippets for this.

PR #409, opened 2016-09-13
Add entities filter

bomjacob:
Should ideally superseed #375.
We could also add individual filters for each Entity (which use this function)...

Not entirely happy with the docstring at line [98-99](https://github.com/python-telegram-bot/python-telegram-bot/compare/entities-filter?expand=1#diff-e1f433e10f5cf4b4c310490777ac95d2R98), but I think it's the best we can do.

This is the first of two revisions to the filter system (the second one will change the way to AND/OR them).


jh0ker:
I just realized that I was the one to say we should use a list here... To be fair though, the #411 PR was not planned at that point. And I still like the higher-order function.

PR #411, opened 2016-09-14
Make filters and/or-able using bitwise operators.

bomjacob:
Would allow filters to be and'ed/or'ed.
Using syntax like this:
```python
handler = MessageHandler((Filters.text | Filters.venue) & Filters.forwarded, callback)
```
Currently this isn't _too_ useful since the only filter that can be and'ed to any other is `forwarded`, but that should all change with #409.

The code currently requires all filters to be created as a class and then initiated before they can be used, which could be an inconvenience to the user if they want to create their own filters. To remedy this we could add a decorator like this:
```python
def Filter(f):
    def wrapper(self, message):
        return f(message)
    return type(f.__name__, (BaseFilter,), {'__call__': wrapper})()
```
That would allow users (and us) to make filters like this:
```python
@Filter
def Text(self, message):
    return bool(message.text and not message.text.startswith('/'))
```
The reason I didn't immediately go with this, is that it might be confusing if someone (especially someone new to python) reading the code stumbles upon a dynamic class creator.  
Thoughts?

- [x] Figure out implementation
- [x] Write the code
- [x] Write the docs
- [x] Add snippet to wiki
- [x] Wait for #409 to get merged?

Edit: After talking in dev group, the decorator has been declared too gross to look at :P

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -16,11 +16,31 @@
 #
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
-""" This module contains the MessageHandler class """
+""" This module contains the Filters for use with the MessageHandler class """
 
 
 class BaseFilter(object):
-    """Base class for all Message Filters"""
+    """Base class for all Message Filters
+
+    Subclassing from this class filters to be combined using bitwise operators:
+
+    And:
+
+        >>> (Filters.text & Filters.entity(MENTION)
+
+    Or:
+
+        >>> (Filters.audio | Filters.video)
+
+    Also works with more than two filters:
+
+        >>> (Filters.text & (Filters.entity(URL |Filters.entity(TEXT_LINK))))
+
+    If you want to create your own filters create a class inheriting from this class and implement
+    a `filter` method that returns a boolean: `True` if the message should be handled, `False`
+    otherwise. Note that the filters work only as class instances, not actual class objects
+    (so remember to initialize your filter classes).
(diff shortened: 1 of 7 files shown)


telegram/ext/filters.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class BaseFilter(object):
+    """Base class for all Message Filters
+
+    Subclassing from this class filters to be combined using bitwise operators:
+
+    And:
+
+        >>> (Filters.text & Filters.entity(MENTION)
```
jh0ker: missed a `)` here

telegram/ext/filters.py
```diff
@@ -0,0 +1,196 @@
...
+
+    Or:
+
+        >>> (Filters.audio | Filters.video)
+
+    Also works with more than two filters:
+
+        >>> (Filters.text & (Filters.entity(URL |Filters.entity(TEXT_LINK))))
```
jh0ker: this also looks wrong to me

telegram/ext/filters.py
```diff
@@ -0,0 +1,196 @@
...
+        return MergedFilter(self, or_filter=other)
+
+    def filter(self, message):
+        raise NotImplementedError
+
+
+class MergedFilter(BaseFilter):
+    """Represents a filter consisting of two other filters."""
+
+    def __init__(self, base_filter, and_filter=None, or_filter=None):
```
jh0ker: Please add a docstring indicating that `and_filter` and `or_filter` are mutually exclusive

PR #414, opened 2016-09-18
Properly split and handle arguments in CommandHandler

eligao:
Assume we got a command with more than one continuous white-space characters (which does happen) :

> /func[space][space]arg1[space]arg2[space][space]arg3[space][space][space]arg4

after `split(' ')`, we have
`['/func', '', 'arg1', 'arg2', '', 'arg3', '', '', 'arg4']`
which includes many unnecessary empty strings that leads to improper command handling.
This is also different from how the system handles standard `argv[]`.
However with `split()`, we have the proper parameters:
`['/func', 'arg1', 'arg2', 'arg3',  'arg4']`
Also it eliminates other unnecessary white-spaces like `\r`,`\n`, etc.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -11,6 +11,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Avanatiker <https://github.com/Avanatiker>`_
 - `Balduro <https://github.com/Balduro>`_
 - `bimmlerd <https://github.com/bimmlerd>`_
+- `Eli Gao <https://github.com/eligao>`_
 - `ErgoZ Riftbit Vaper <https://github.com/ergoz>`_
 - `franciscod <https://github.com/franciscod>`_
 - `Jacob Bom <https://github.com/bomjacob>`_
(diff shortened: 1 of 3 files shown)


eligao:
But after this commit, `' '.join(args)` would give a "regularized" version of the original arguments, which

eligao:
But after this commit, one could get a "regularized" version of the original argument.
Also they could probably use `update.message.text` to get the whole original argument.
```
>>> s='/func   1 2  3 4'
>>> ' '.join(s.split(' ')[1:])
'   1 2  3 4'
>>> ' '.join(s.split()[1:])
'1 2 3 4'
>>>
```

bomjacob:
Yeah that's true. Please address the docstring change I reviewed above and it looks good to go as far as I'm concerned.

eligao:
Noted :)

eligao:
I've updated the docstring in CommandHandler.
Also I did the same change in StringCommandHandler because they look similar, but I'm not quite sure if it looks good because I never used this class.

PR #415, opened 2016-09-21
Start additional threads only when necessary

wronglink:
I have several troubles with `JobQueue` and `Dispatcher` automatically start new threads during it's initialisation. I guess it's quite unexpected behaviour to have a branch of threads just by creating new `Updater`, without starting it. Instead of it, I propose the `Updater` instance should start all necessary threads on inside it's start methods: `start_polling` and `start_webhook`.

### DIFF
--- a/telegram/ext/dispatcher.py
+++ b/telegram/ext/dispatcher.py
@@ -84,6 +84,7 @@ class Dispatcher(object):
         self.bot = bot
         self.update_queue = update_queue
         self.job_queue = job_queue
+        self.workers = workers
 
         self.handlers = {}
         """:type: dict[int, list[Handler]"""
@@ -104,8 +105,6 @@ class Dispatcher(object):
                 self._set_singleton(self)
             else:
                 self._set_singleton(None)
-
-        self._init_async_threads(uuid4(), workers)
 
     @classmethod
     def _reset_singleton(cls):
@@ -193,6 +192,7 @@ class Dispatcher(object):
             self.logger.error(msg)
             raise TelegramError(msg)
 
+        self._init_async_threads(uuid4(), self.workers)
         self.running = True
         self.logger.debug('Dispatcher started')
 
--- a/telegram/ext/jobqueue.py
+++ b/telegram/ext/jobqueue.py
@@ -20,6 +20,7 @@
 
 import logging
 import time
+import warnings
 from threading import Thread, Lock, Event
 from queue import PriorityQueue, Empty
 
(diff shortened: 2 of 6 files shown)


telegram/ext/dispatcher.py
```diff
@@ -193,6 +192,7 @@ def start(self):
             self.logger.error(msg)
             raise TelegramError(msg)
 
+        self._init_async_threads(uuid4(), self.workers)
```
jh0ker: I think I agree with this change, but I'd like @tsnoam to give it a look as well.

telegram/ext/jobqueue.py
```diff
@@ -51,10 +49,6 @@ def __init__(self, bot, prevent_autostart=False):
         """:type: float"""
         self._running = False
```
jh0ker: Same here

telegram/ext/jobqueue.py
```diff
@@ -51,10 +49,6 @@ def __init__(self, bot, prevent_autostart=False):
...
 
-        if not prevent_autostart:
-            self.logger.debug('Auto-starting %s', self.__class__.__name__)
-            self.start()
-
     def put(self, job, next_t=None):
         """Queue a new job. If the JobQueue is not running, it will be started.
```
jh0ker: This docstring is wrong (and was wrong before your PR). The job queue is never started in this method. Please remove that sentence.

telegram/ext/updater.py
```diff
@@ -157,6 +157,7 @@ def start_polling(self,
                 self.running = True
 
                 # Create & start threads
+                self._init_thread(self.job_queue.start, "job_queue")
```
jh0ker: Here you can simply call `self.job_queue.start()`, I believe. The method is already spawning a new thread.

telegram/ext/updater.py
```diff
@@ -208,6 +209,7 @@ def start_webhook(self,
...
```
jh0ker: See above comment

jh0ker:
Hey, thanks! I did a short code review. As mentioned in the review, I am hereby pinging @tsnoam to check out the code as well ;)

telegram/ext/updater.py
tsnoam: +1 @jh0ker  comment here.
@wronglink can you please fix?

wronglink:
@jh0ker @tsnoam I've updated PR and added deprecation warning about `prevent_autostart`, so I guess it could be merged in 5.2.

tsnoam:
for some reason, i couldn't see the sources with decreased coverage. i'm attempting rebuild with travis, maybe it'll help.

wronglink:
@tsnoam I've merged master branch, so forced TravisCI to recalculate coverage. Now files with decreased coverage are showed on coveralls.

tsnoam:
@wronglink 
thanks for that.
seems like the decrease in coverage is due to removed lines, changing stats in 0.1%. so i'm going merge now.
thanks again for your contribution.

Issue #418, opened 2016-09-23
PyPI out of date | echobot2 example not working with PyPI version


sixhobbits:
After installing python-telegram-bot through pip, the example code doesn't run.  Specifically, the examples now use 
```
update.message.reply_text('Hi!')
```

to send messages, while the version in PyPI still needs
```
bot.sendMessage(update.message.chat_id, text='Hi!')
```

I suggest that

1. the PyPI package should be updated to reflect the latest version of this library
2. instructions for installing from source should be added to README.md

### Steps to reproduce
1. pip install python-telegram-bot

2. create echobot2 and insert code from examples folder

3. python echobot2.py

4. send message to bot 

### Expected behaviour
echobot should reply with same message

### Actual behaviour
echo bot throws error 
```
(test) nsbot@nstravel:~/git/NSGroupTickets/NSBot$ python echobot2.py
2016-09-23 14:56:00,713 - telegram.ext.dispatcher - ERROR - An uncaught error was raised while processing the update
Traceback (most recent call last):
  File "/home/nsbot/VEs/test/lib/python3.5/site-packages/telegram/ext/dispatcher.py", line 203, in process_update
    handler.handle_update(update, self)
  File "/home/nsbot/VEs/test/lib/python3.5/site-packages/telegram/ext/messagehandler.py", line 139, in handle_update
    return self.callback(dispatcher.bot, update, **optional_args)
  File "echobot2.py", line 41, in echo
    update.message.reply_text(update.message.text)
AttributeError: 'Message' object has no attribute 'reply_text'
```

### Configuration
Ubuntu 16.04

**Version of Python, python-telegram-bot & dependencies:**

``$ python -m telegram``
```
(test) nsbot@nstravel:~/git/NSGroupTickets/NSBot$ python -m telegram
python-telegram-bot 5.0.0
urllib3 1.17
certifi 2016.08.31
future 0.15.2
Python 3.5.2 (default, Sep 10 2016, 08:21:44)  [GCC 5.4.0 20160609]
```

### Logs
Insert logs here (if necessary)

jh0ker:
Yes, you are correct. When I merged the PR, I forgot that I had already modified the one example. We are closing in on milestone 5.1 (https://github.com/python-telegram-bot/python-telegram-bot/milestone/7), and release the new PyPI package ASAP ;)

Until then, I will add a note in the examples README with a link to this issue.

jh0ker:
[...]

PR #419, opened 2016-09-23
Add install from source instructions to readme

sixhobbits:
Adds instructions to install from source to the readme. If this is the preferred way to use python-telegram-bot (as is suggested by the fact that in the new issues template you state that you only support the master branch) then this should also be specified. Currently the examples don't work with the PyPI version -- see [https://github.com/python-telegram-bot/python-telegram-bot/issues/418](https://github.com/python-telegram-bot/python-telegram-bot/issues/418)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -96,6 +96,13 @@ You can install or upgrade python-telegram-bot with:
 
     $ pip install python-telegram-bot --upgrade
 
+Or you can install from source with:
+
+.. code:: shell
+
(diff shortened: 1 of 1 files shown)


jh0ker:
Thank you :) 

The issue template, by the way, says
> Please note, we only support the latest version of python-telegram-bot and
master branch.

Also, only the `echobot2.py` example has been changed, all others should be working fine with v5.0 :+1:

PR #420, opened 2016-09-23
Add quote keyword argument to Message.reply_x methods

jh0ker:
Because it makes sense

Edit: dankest pull request ever


tsnoam:
@jh0ker LGTM

PR #426, opened 2016-09-27
add ReplyKeyboardHide

MarkusHackspacher:
This is the first thing I´m look for, to hide the Keyboard Symbol after the first question.

### DIFF
--- a/examples/conversationbot.py
+++ b/examples/conversationbot.py
@@ -17,7 +17,7 @@ Press Ctrl-C on the command line or send a signal to the process to stop the
 bot.
 """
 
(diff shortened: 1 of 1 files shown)


jh0ker:
It's already using `one_time_keyboard=True`, are you sure this makes a difference?

MarkusHackspacher:
With one_time_keyboard=True is the keyboard at the second question off, but you can see the symbol for keyboard, with ReplyKeyboardHide() are this symbol off, too.

MarkusHackspacher:
That PR is better: https://github.com/python-telegram-bot/python-telegram-bot/pull/439

PR #433, opened 2016-10-06
small fix for game related parts

PH89:
* bot.py fix copy paste error in url
* callbackquery.py make data field optional
* message.py introduce optional game field


jh0ker:
Awesome! Thank you for helping out. You should also add yourself to `AUTHORS.rst` if you're not already in there.

PR #435, opened 2016-10-09
fixes broken test cases with PhotoSize, Game and Animation classes

PH89:
However **testSendGame** and **test_set_game_score** both produce:

**BadRequest: u'Wrong file identifier/HTTP URL specified** 

for the official **@pythontelegrambottests** test bot.

This is very probably the case because the test bot it not correctly configured to be a game.

### DIFF
--- a/telegram/base.py
+++ b/telegram/base.py
@@ -74,17 +74,9 @@ class TelegramObject(object):
 
             value = self.__dict__[key]
             if value is not None:
-                if type(value) == list:
-                    l = list()
-                    for e in value:
-                        if hasattr(e, 'to_dict'):
-                            l.append(e.to_dict())
-                        else:
-                            l.append(e)
+                if hasattr(value, 'to_dict'):
+                    data[key] = value.to_dict()
                 else:
-                    if hasattr(value, 'to_dict'):
-                        data[key] = value.to_dict()
-                    else:
-                        data[key] = value
+                    data[key] = value
 
         return data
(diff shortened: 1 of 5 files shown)


bomjacob:
The @pythontelegrambottests is indeed not set up with games properly since I just tested locally with my own one. However, I'm hereby pinging @leandrotoledo about it.

telegram/base.py
```diff
@@ -74,9 +74,17 @@ def to_dict(self):
 
             value = self.__dict__[key]
             if value is not None:
-                if hasattr(value, 'to_dict'):
-                    data[key] = value.to_dict()
+                if type(value) == list:
```
jh0ker: If I understand this part correctly, you added this to decode attributes that are marked as "Array of ..." in the API docs? 

In case I assume correctly, we usually have a `de_list` method on classes like `PhotoSize` ([Implementation](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/photosize.py#L73) and [usage](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/message.py#L176) examples) when needed. 

However, I do like your solution for this. Maybe it could be implemented in a separate PR, for all cases where it is needed.

jh0ker:
@PH89 I'd like to get in touch with you via Telegram. If you're okay with that, please shoot me a message at [@jh0ker](https://telegram.me/jh0ker)

telegram/base.py
jh0ker: Update from a private chat: I was wrongfully thinking of `de_json` instead of `to_dict`, but the outcome is the same.

jh0ker:
Awesome! LGTM, merging

PR #439, opened 2016-10-12
hide inline keyboard in conversationbot example

bluehood:
It was just a minor thing that bugged me while going through the examples: as per Telegram [docs](https://core.telegram.org/bots/api/#replykeyboardmarkup), using `one_time_keyboard=True` only hides the inline keyboard, but does not disable it completely, leaving a button that the user can press to get the keyboard back.

This little change to the `conversationbot.py` example script makes it so that the inline keyboard is properly hidden after usage or when the conversation is halted.

### DIFF
--- a/examples/conversationbot.py
+++ b/examples/conversationbot.py
@@ -17,7 +17,7 @@ Press Ctrl-C on the command line or send a signal to the process to stop the
 bot.
 """
 
-from telegram import (ReplyKeyboardMarkup)
+from telegram import (ReplyKeyboardMarkup, ReplyKeyboardHide)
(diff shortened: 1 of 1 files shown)


MarkusHackspacher:
That is the same like my PR: https://github.com/python-telegram-bot/python-telegram-bot/pull/426/commits/ab30a580c7e12f723acad7b034e3d06531523b34
In my PR is not the line 109, but this is necessary for canel command. I´ll close my PR.

jh0ker:
Why not ^^ Thank you for your contribution :)

PR #431, opened 2016-10-03
Add api changes as of october

bomjacob:
Add changes as of October 3, 2016.
Mostly games: https://core.telegram.org/bots/api#games
Other changes: https://core.telegram.org/bots/api#recent-changes

- [x] Games
  - [ ] Unitests
- [x] `getWebhookInfo` and `WebhookInfo`
  - [x] Unitests
- [x] Allow HTTP url in all methods that accept `InputFile` or `file_id` (this requires disabling our own handling of this)
  - [x] Unitests
- [x] Url parameter in `answerCallbackQuery`
  - [ ] Unitests
- [x] `switch_inline_query_current_chat`
  - [ ] Unitests
- [x] Add caption fields to `send_audio`, `send_voice`, `InlineQueryResultAudio`, `InlineQueryResultVoice`, ``InlineQueryResultCachedAudio` and `InlineQueryResultCachedVoice`
  - [x] Unitests
- [x] `all_members_are_admins` in `Chat`
  - [x] Unitests
- [x] Handle the errors of `ResponseParameters` (no need unitests for the moment. we already had one response param and we didn't unitest it)

PR #440, opened 2016-10-15
Add switch_inline_query_current_chat

bobuk:
Add really useful option which is happened at last (3 Oct) update https://core.telegram.org/bots/api-changelog

So, now if you use switch_inline_query_current_chat="test" as an InlineKeyboardButton option, telegram automatically insert text "test" to current chat input box.


bomjacob:
This change is already getting implemented as part of #431.

PR #444, opened 2016-10-22
Remove deprecated argument from Updater docstring

runekri3:
Argument (job_queue_tick_interval) was removed with commit 3aedd78 but the docstring in updater.py wasn't.


tsnoam:
thanks for the contribution.
merging.

PR #452, opened 2016-11-02
Job queue time units

voider1:
This enables users to add intervals in seconds (default), minutes and hours for their Jobs.
It also enables users to define on which days the Job should be run (all by default).
It's backwards compatible!


voider1:
Thank you, @KamranMackey!

telegram/ext/jobqueue.py
```diff
@@ -21,10 +21,22 @@
 import logging
 import time
 import warnings
+import datetime
+from enum import Enum
 from threading import Thread, Lock, Event
 from queue import PriorityQueue, Empty
 
 
+class Days(Enum):
```
bomjacob: Weeks unfortunately don't always start on a monday. In some locales it's sundays.

telegram/ext/jobqueue.py
```diff
@@ -56,6 +68,12 @@ def __init__(self, bot, prevent_autostart=None):
         """:type: float"""
         self._running = False
 
+    def dt_time_to_seconds(self, dt_time):
```
bomjacob: Why not use `.total_seconds()` here?

telegram/ext/jobqueue.py
```diff
@@ -126,7 +156,9 @@ def tick(self):
                 self.logger.debug('Running job %s', job.name)
```
bomjacob: This debug would wrongly say that it's running a job even though it's not, if it's on an excluded day.

tests/test_jobqueue.py
```diff
@@ -169,6 +178,26 @@ def test_inUpdater(self):
         finally:
             u.stop()
 
+    def test_time_units(self):
```
bomjacob: Please split this into multiple test that only test one thing each.
E.g. have one that tests if seconds (remember to test floats too btw) and another test which tests if datetime objects work.

telegram/ext/jobqueue.py
```diff
@@ -21,10 +21,22 @@
 import logging
 import time
 import warnings
+import datetime
+from enum import Enum
```
bomjacob: Enums are a new feature that appeared in 3.4. Note that this library also supports python 2.7.
While they have been backported and are available on pypi, I would rather not add a whole dependency for something that is used so very little.

telegram/ext/jobqueue.py
```diff
@@ -208,22 +241,39 @@ class Job(object):
...
+                 repeat=True,
+                 context=None,
+                 days=tuple(day for day in Days)):
         self.callback = callback
         self.interval = interval
         self.repeat = repeat
         self.context = context
 
+        if not isinstance(days, tuple):
+            err_msg = "The 'days' argument should be of type 'tuple'"
```
bomjacob: Please just write the msg directly in the exception, it simply looks a bit cleaner.

voider1:
Thank you! I will address all the changes.

telegram/ext/jobqueue.py
KamranMackey: @bomjacob I don't think one extra dependency will hurt people. And Enums are not used so "very little". And there's no way to track how often Enums are being used either. But I know that they're used more than your "so very little" stat.

telegram/ext/jobqueue.py
KamranMackey: @bomjacob Does this actually matter all that much? I know this might come off as rude, but I don't see why this matters. It's a really tiny thing.

KamranMackey: And as far as I know, it would require too much code to be able to do an if statement to where sunday is day 0, as we don't even know how many locales have Sunday as the first day of the week.

telegram/ext/jobqueue.py
bomjacob: Granted, we use enums elsewhere in the project... but a enum in python could just as well be a class with some attributes.... There's a reason it wasn't added until 3.4... it's pretty much unnecessary. And of cause if there was a library that would make our library easier to maintain, easier to read or easier to develop on, we wouldn't hesitate to add it as a dependency - but Enums simply aren't that.

telegram/ext/jobqueue.py
bomjacob: Everywhere preferably :)

telegram/ext/jobqueue.py
bomjacob: Well I'm not sure, but I'd say that the entirety of North America is a pretty big demographic...
Question is if it even matters... as long as the code below correctly knows that a monday is a monday (and not necesarily a 0) then everything is fine... I'm not entirely sure that python is that smart though :/

telegram/ext/jobqueue.py
voider1: @bomjacob I'd personally like to keep the enum library, it's supported and a very clean way of doing it, it isn't a very big dependency, and could be used more throughout the project as it grows.

An alternative would look like:

```
>>> def enum(**named_values):
...     return type('Enum', (), named_values)
...
>>> Days = enum(mon=0, tue=1, wed=2, thu=3, fri=4, sat=5, sun=6)
>>> Days.mon
0
>>> class Days(object):
...     def __init__(self, day):
...         if day not in (Days.mon , Days.tue, Days.wed, Days.thu, Days.fri, Days.sat, Days.sun):
...             raise ValueError('gender not valid')
...         self.day = day
...
```

I don't like this pattern because it relies upon the integers too much and has way more overhead.
In my original example the Enum doesn't relie on the integers too much, only to check if one of the values of the Days which were supplied with the Job is today. The Enum subclass returns a special Enumeration object instead of the literal integer value. This is a lot nicer for error handling because people should use the supplied Days Enum subclass. If you don't do this, people could just pass in integers from 0-6 and it'd accept. As far as I know, we can't check if a value comes from a certain class if we do this with a plain new-style class. With an Enum we can check if it comes from the class it should come from and enforce people to use the Days Enum subclass.

telegram/ext/jobqueue.py
voider1: @bomjacob I've already done my research on this, [Python2](https://docs.python.org/2/library/datetime.html ) and [Python 3](https://docs.python.org/3/library/datetime.html) both return the .weekday() from 0-6 where monday is 0 and sunday is 6.

telegram/ext/jobqueue.py
voider1: What is the max line length for the project? I use 84 as my personal standard, so when a message is too long, I use a variable.

telegram/ext/jobqueue.py
bomjacob: If that's the case across all locales (that is, python doesn't automatically change that weekday depending on where it's run) then it shouldn't matter and this should be fine...

telegram/ext/jobqueue.py
bomjacob: Why do you want to prevent users from using integers? It's not really very python like to require inputs to be of a special type if you don't have to. My recommendation would be as follows:
Make the Days a simple class with the weekdays as class variables (attributes), this is what we use for enums elsewhere in the project and it's worked fine (if you believe that it absolutely wouldn't work here, please do correct me).
Just raise a ValueError if 0 < day > 7 inside ``Job`'s `__init__` or `JobQueue`'s `put`.

In the future we might upgrade to use Enums everywhere, but for now it's simply not necessary.

voider1: Code is more often read than written. Preventing users from using plain integers makes it clearer what someone's trying to accomplish when looking over the code.

For example:
```python
import datetime

test_job = Job(callback, datetime.time(hour=3), repeat=True, days=(0, 4, 6))
```

If I make my bot Open Source and someone reads this, it's very unclear what the original developer tries to do without first reading the documentation of this repo. It's also very confusing if you have different locals (like you pointed out) and you assume that 0 is sunday instead of monday. Of course, we would like to make the developer use our Days class. But he isn't enforced to do it, and I think we should assume people wouldn't if they don't really have to.

This code:
```python
import datetime

from telegram.ext.jobqueue import Days

test_job = Job(callback, datetime.time(hour=3), repeat=True, days=(Days.mon, Days.fri, Days.sun))
```

This code is a little longer, but it's entirely self-documenting! A developer who is unknown with the codebase will instantly know how it works. The syntax is clear, the functionality is clear, and we're enforcing them to use the Enum from the project instead of integers like we saw in the first example. I think this is more Pythonic, because "there should be one-- and preferably only one --obvious way to do it", except for string formatting.

telegram/ext/jobqueue.py
voider1: According to the official Datetime docs that is the case, so I think we're good on the mapping of the days.

telegram/ext/jobqueue.py
voider1: I've fixed this.

telegram/ext/jobqueue.py
voider1: Do you mean as method name, because `datetime.time` doesn't have a method called `total_seconds`. Am I missing something?

tests/test_jobqueue.py
voider1: I cannot find a reliable way to test the `float`, one time it succeeds, one time it fails. Do you know a way in which I can reliably test floats? I know for certain that floats work, is not testing them an option, because if the test starts on the wrong millisecond and rounds the wrong way, the test will fail.

telegram/ext/jobqueue.py
catman155: Python 2 is going to drop its support, so i agree with voider1 to implement enum.

TheUberCatman: I agree to use enum because Python 2 is going to drop its support soon. Python 3 is the new standard.

bomjacob:
Here's some comments from @tsnoam from our dev team group:
all from phone so I don't want to comment in github so I write here:

1. import numbers # or something like that

isinstance(x, number)

2. raise ValueError : don't use \ at eol

3. days=tuple(Days) 
no need: day for day in Days

4. race condition around midnight with line 156

5. was too lazy to read the unitests

6. all in all looks good. the only thing is that they use enum and unless it's part of "future" we already require as a dependency it can be implemented without enum which is a trivial change to the code.

telegram/ext/jobqueue.py
bomjacob: Please see @tsnoam comment below

voider1:
@bomjacob 

Point 2 doesn't work if it isn't a subclass of Enum. If the bottom-line is not using Enum, I'd have to manually get the ints from the class or implement an iterator in the class.

```python
>>> class Days(object):
...     mon = 0
...     tue = 1
...
...
>>> tuple(Days)
  File "<input>", line 1, in <module>
    tuple(Days)
TypeError: 'type' object is not iterable
```

About point 3, the only thing I could think of was that maybe if you said sunday and you made it so that the next_t would execute at 00:00, it wouldn't because it'd be monday again, however:

```python
>>> midnight = datetime.datetime(year=2016, month=11, day=6, hour=0, minute=0, second=0).weekday()
>>> midnight == datetime.datetime.now().weekday()
True
```

Which means that there's no problem, right?

Point 5 those are a lot of small changes, sadly, not entirely relying on integers, making sure users use the Days Enum. You guys should really think about making Enums part of the project. They're a lot nicer to work with then plain new-style classes, and do not add overhead in the code with implementing iterators or manually getting al the values from the class (which I'll have to do now). I'll make them new-style classes for now. But please, really just consider adding this to the whole project. It's a small dependency which makes these things a lot easier.

voider1:
@bomjacob I've addressed all the issues you mentioned. Could you review it again, please?

telegram/ext/jobqueue.py
voider1: I'm using a separate `get_days` function here, because if I put it in the class, it won't be filtered out by the list comprehension, and I won't be able to easily get the values from the attributes defined in the class.

telegram/ext/jobqueue.py
```diff
@@ -21,10 +21,26 @@
...
+    mon = 0
+    tue = 1
+    wed = 2
+    thu = 3
+    fri = 4
+    sat = 5
+    sun = 6
+
+
+def get_days():
```
jh0ker: I know what this is supposed to do, but it's still unreadable. I think the best solution would be to do something like we did in [`MessageEntity`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/messageentity.py#L83). Other ideas?

telegram/ext/jobqueue.py
```diff
@@ -21,10 +21,26 @@
...
 import time
 import warnings
+import datetime
+from numbers import Number
 from threading import Thread, Lock, Event
 from queue import PriorityQueue, Empty
 
 
+class Days(object):
+    mon = 0
```
jh0ker: Since we switched to a class with constants, the correct naming convention would be `UPPER_CASE` if I'm not mistaken.

telegram/ext/jobqueue.py
```diff
@@ -56,6 +72,12 @@ def __init__(self, bot, prevent_autostart=None):
         """:type: float"""
         self._running = False
 
+    def timedelta_to_seconds(self, tdelta):
```
jh0ker: This method seems to be redundant, especially since this looks like a purely internal method and you already check `isinstance(var, datetime.timedelta)`

telegram/ext/jobqueue.py
```diff
@@ -56,6 +72,12 @@ def __init__(self, bot, prevent_autostart=None):
...
+        try:
+            return tdelta.total_seconds()
+        except AttributeError:
+            raise AttributeError("timedelta_to_seconds only takes a datetime.timedelta object")
+
     def put(self, job, next_t=None):
         """Queue a new job.
```
jh0ker: Please update the docstring so it mentions that `next_t` can also be of type `timedelta`

telegram/ext/jobqueue.py
```diff
@@ -208,22 +242,36 @@ class Job(object):
...
     job_queue = None
 
-    def __init__(self, callback, interval, repeat=True, context=None):
+    def __init__(self, callback, interval, repeat=True, context=None, days=get_days()):
         self.callback = callback
         self.interval = interval
         self.repeat = repeat
         self.context = context
 
+        print(days)
```
jh0ker: Debug print

telegram/ext/jobqueue.py
```diff
@@ -208,22 +242,36 @@ class Job(object):
...
             (``True``) or only once (``False``). Defaults to ``True``
         context (Optional[object]): Additional data needed for the callback function. Can be
             accessed through ``job.context`` in the callback. Defaults to ``None``
+        days (Tuple): Defines on which days the job should be ran.
 
     """
     job_queue = None
 
-    def __init__(self, callback, interval, repeat=True, context=None):
+    def __init__(self, callback, interval, repeat=True, context=None, days=get_days()):
```
jh0ker: According to my earlier comment, this would change to `days=Days.ALL_DAYS` (or maybe even `days=Days.EVERY_DAY`? Your choice ^^)

telegram/ext/jobqueue.py
```diff
@@ -208,22 +242,36 @@ class Job(object):
         callback (function): The callback function that should be executed by the Job. It should
             take two parameters ``bot`` and ``job``, where ``job`` is the ``Job`` instance. It
             can be used to terminate the job or modify its interval.
-        interval (float): The interval in which this job should execute its callback function in
-            seconds.
+        interval (float or datetime.time): The interval in which this job should execute its
```
jh0ker: I believe you switched to `timedelta` instead of `time`, please update this docstring to reflect that change.

telegram/ext/jobqueue.py
```diff
@@ -123,10 +155,11 @@ def tick(self):
                 continue
 
             if job.enabled:
-                self.logger.debug('Running job %s', job.name)
-
                 try:
-                    job.run(self.bot)
+                    for day in job.days:
```
jh0ker: If someone would, for whatever reason, put a day twice (like `Job(..., days=(Days.mon, Days.mon))`) this would run the job twice on those days. I'm not sure this is a bug or a feature, but it could be easily changed by changing the type of `Job.days` from `tuple` to `set`. Your choice.

voider1:
@jh0ker I adressed all the changes and fixed them.

voider1:
@jh0ker About the idea of running them at a certain time, I'll maybe do that after this PR gets merged.

jh0ker:
LGTM :)

Issue #357, opened 2016-07-22
de_json() method messes with request


mssalnikov:
Hey! Your `de_json()` methods mess with actual data in the request object. 

Here: 

```
def de_json(data):
        """
        Args:
            data (dict):
        Returns:
            telegram.Update:
        """
        if not data:
            return None

        data['message'] = Message.de_json(data.get('message'))
        data['edited_message'] = Message.de_json(data.get('edited_message'))
        data['inline_query'] = InlineQuery.de_json(data.get('inline_query'))
        data['chosen_inline_result'] = ChosenInlineResult.de_json(data.get('chosen_inline_result'))
        data['callback_query'] = CallbackQuery.de_json(data.get('callback_query'))

        return Update(**data)
```

You see, you change `data` dictionary, and since it's one of python's mutable data structures, it changes too. Here is an example:

```
print('REQUEST', request.get_json(force=True))

update = telegram.update.Update.de_json(request.get_json(force=True))

print('REQUEST', request.get_json(force=True))
```

the result is:

> 2016-07-22T12:52:33.626071+00:00 app[web.1]: REQUEST {'update_id': 938970460, 'inline_query': {'query': '567', 'offset': '', 'from': {'first_name': 'Mike', 'username': 'zuldare', 'last_name': 'Salnikov', 'id': 97081120}, 'id': '416960237449083811'}}
> 2016-07-22T12:52:33.626213+00:00 app[web.1]: REQUEST {'inline_query': <telegram.inlinequery.InlineQuery object at 0x7f84fecdcf60>, 'message': None, 'update_id': 938970460, 'chosen_inline_result': None, 'callback_query': None, 'edited_message': None}

which should be not that big a deal in most cases, but in some could be very bad

leandrotoledo:
@mssalnikov Good catch, we should be calling super(TelegramObject, TelegramObject).de_json(data) which contains data = data.copy() like we do for most of the new models (recently implemented InlineQuery* etc).

PR #454, opened 2016-11-04
Prevented modifications to the request object's original data

HakimusGIT:
Currently the `de_json()` methods of some subclasses of `TelegramObject` make modifications to the data of the request object. Usually, this shouldn't be a problem, but it might be one if other parts of the code try to read the data and expect something else to be there. By calling `TelegramObject`'s `de_json()` method, which makes a copy of the data that can be worked with, this problem should be solved.

See #357 for an issue concerning this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,6 @@ telegram.webp
 
 # original files from merges
 *.orig
+
+# Exclude .exrc file for Vim
+.exrc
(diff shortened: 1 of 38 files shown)


HakimusGIT:
Looks like the tests are failing because too many requests were sent, like in #459.

tsnoam:
@HakimusGIT 
Thank you for your contribution. I reviewed the PR and it's good to be merged. Though:
1. I've pushed a commit undoing some style changes (probably yapf fault) which are unrelated to the PR and, ehm, yapf was wrong there.
2. I'm waiting to see one last time the travis build pass before merging.

PR #456, opened 2016-11-06
Download as bytesIO object

OctoNezd:
Now all tests passed.
I think.

### DIFF
--- a/telegram/file.py
+++ b/telegram/file.py
@@ -79,3 +79,7 @@ class File(TelegramObject):
             filename = basename(url)
 
         self.bot.request.download(url, filename)
+
+    def downloadbytes(self):
+        url = self.file_path
+        return self.bot.request.downbytes(url)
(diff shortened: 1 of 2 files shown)


amamaenko:
This code won't pass the PEP8 checks. It's a good practice to run pylint before doing pull requests.

bomjacob:
I assume you're `OctoNerd` on telegram too?
@jh0ker suggested yesterday (when you said you'd make this PR) that we instead "change the existing `download` function to accept file-like objects in addition to filenames?" You even said that you like that way, so why make this PR?
(Also the reason your code wasn't passing the yapf test is probably the inline import... remember to read (and follow) a repository's [CONTRIBUTING](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/.github/CONTRIBUTING.rst) file before submitting PRs.)

PR #458, opened 2016-11-06
File.downbyte added

OctoNezd:
Please notice:I haven`t changed File.download() behaviour, because it will ruin backwards compability.

### DIFF
--- a/telegram/file.py
+++ b/telegram/file.py
@@ -79,3 +79,7 @@ class File(TelegramObject):
             filename = basename(url)
 
         self.bot.request.download(url, filename)
+
+    def downbyte(self):
+        url = self.file_path
+        return self.bot.request.downbyte(url)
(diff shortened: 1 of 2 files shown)


bomjacob:
Please explain how making `file.download()` also accept file like objects and changing behaviour depending on it in any way breaks backwards compatibility.

OctoNezd:
Because I think many of bots use code like this:
`File.download(id, "blah-blah-blah")`
And when suddendly, args change to id, file, cpath, their code will break. Am I right?/

bomjacob:
Not really no.
Since python is not a statically typed language, a method can accept a parameter that could be different types. Then inside the function you check (using `isinstance`) the type of the parameter and act accordingly.

OctoNezd:
https://travis-ci.org/python-telegram-bot/python-telegram-bot/jobs/173651523
Line 983: I think its not mine code fault

OctoNezd:
Oh. I will update code and make new pull req
P.S. I hope I haven`t annoyed you with mine PR`s

bomjacob:
There's really no need to make a separate PR every time something needs to be changed, but alright :)

PR #461, opened 2016-11-08
Fix telegram API change, returning '404 Not found' 

lisitsky:
For bots with incorrect credentials telegram [returns](https://api.telegram.org/bot12) 
```
HTTP/1.1 404 Not Found
... headers skipped ...

{
    "description": "Not Found",
    "error_code": 404,
    "ok": false
}
```

Because of 404 response status http library raises naive NotFound exception, which is not meant to be caught by a request wrapper. As a result test is broken. This patch catches the exception and raises an internal `telegram.TelegramError('Invalid server response')` as it was conceived before in test [testInvalidSrvResp](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/tests/test_bot.py#L229).

P.S. I'm new to python-telegram-bot, so comments are welcome ;)

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -13,6 +13,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `bimmlerd <https://github.com/bimmlerd>`_
 - `Eli Gao <https://github.com/eligao>`_
 - `ErgoZ Riftbit Vaper <https://github.com/ergoz>`_
+- `Eugene Lisitsky <https://github.com/lisitsky>`_
 - `franciscod <https://github.com/franciscod>`_
 - `Jacob Bom <https://github.com/bomjacob>`_
 - `JASON0916 <https://github.com/JASON0916>`_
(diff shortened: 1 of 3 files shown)


jh0ker:
Sweet, thank you! Perhaps it would make sense to add an exception class to `telegram/error.py`?

lisitsky:
Thank you for feedback! :)
Yep, it look through the ``telegram/error.py`` and think ``InvalidToken`` could suite well here. 
In what else situations can we get 404? If we will raise this exception in all of them will we mislead a user? How do you think? Or it will be better to coin another one?

jh0ker:
Oh yes, raising `InvalidToken` sounds good to me. If we encounter other situations where a 404 is returned, we can adjust as necessary.

lisitsky:
Ok. I've updated.

jh0ker:
Neat! LGTM
I think you aren't in our `AUTHORS.rst` yet. If you want, please add yourself there and leave a comment. Then I'll merge.

PR #467, opened 2016-11-21
fix "TypeError: 'function' object is not iterable" in echobot2

bvanrijn:


bomjacob:
Please update your library. This should not happen with the newest version (which is what we target with our examples).

PR #471, opened 2016-11-30
Botapi2.3

### LINKED ISSUE #468: botapi 2.3
Bot API 2.3 was announced on 2016-11-21, need to support new changes.

- [x] New fields `channel_post` and `edited_channel_post` for `Update`.
    - It is not possible to unitest it with the current framework. Tested by hand.
- [x] New `force` parameter in `setGameScore`.
    - [x] Unitests.
- [x] Changed behaviour: messages with high scores will be update with new high scores by default. (documentation fix)
- [x] Use (new) `disable_edit_message` in `setGameScore` to disable the above new behaviour.
    - [x] Unitests.
- [x] The `edit_message` parameter from `setGameScore` is no longer in use. For backward compatibility, it will be taken into account for a while, unless `disable_edit_message` is passed explicitly.
    - [x] Need deprecation warning and maybe some sanity checks on input.
    - [x] Unitests.
- [x] New field `forward_from_message_id` for `Message`.
    - [x] Unitests.
- [x] New parameter `cache_time` for `answerCallbackQuery`.
    - It is not possible to unitest it with the current framework. Tested by hand.
- [x] Renamed `hide_keyboard` to `remove_keyboard` in `ReplyKeyboardRemove` for clarity. `hide_keyboard` will still work for a while for backward compatibility.
    - [x] Need deprecation warning on the old ReplyKeyboardHide.
    - [x] Unitests.
- [x] Need to "remember" for future releases to remove the deprecated code. (`ReplyKeyboardHide`, `edit_message` in `setGameScore`)

tsnoam:
Closes #468


telegram/ext/messagehandler.py
```diff
@@ -51,6 +51,11 @@ class MessageHandler(Handler):
             ``chat_data`` will be passed to the callback function. It will be a ``dict`` you
             can use to keep any data related to the chat that the update was sent in.
             For each update in the same chat, it will be the same ``dict``. Default is ``False``.
+        message_updates (Optional[bool]): Should "normal" message updates be handled? Default is
```
bomjacob: I wonder if we could use a filter for this instead? Now that it is actually possible to use more than one filter. Or is it more clear/explicit and nice using this method?

PR #482, opened 2016-12-13
jobqueue: add support for specifying next_t in datetime.datetime or datetime.time

jh0ker:
Title says it all, I guess. Many people asked for it, so here we go. You can now specify the `next_t` parameter of `JobQueue.put` as `datetime.datetime` or `datetime.time`, so the job will run at exactly the specified (date and) time.

```python
>>> from datetime import datetime, time, timedelta
>>> job_queue.put(happy_new_year_job, next_t=datetime(2017, 1, 1))
>>> job_queue.put(alarm_clock_job, next_t=time(8, 30))
>>> midnight_job = Job(midnight_callback, timedelta(days=1))
>>> job_queue.put(midnight_job, next_t=time(0, 0))
```

If you pass a `datetime.time` object for `next_t` and it's already later than the specified time (eg it's 10:00 and `next_t` is `time(8, 30)`), the job will execute the next day.

I made the `interval` parameter for `Job` optional, so you can just leave it for non-repeating jobs (however, you must explicitly pass `repeat=False`)

### DIFF
--- a/telegram/ext/jobqueue.py
+++ b/telegram/ext/jobqueue.py
@@ -23,8 +23,10 @@ import time
 import warnings
 import datetime
 from numbers import Number
-from threading import Thread, Lock, Event
+from threading import Thread, Lock, RLock, Event
 from queue import PriorityQueue, Empty
+
+from telegram.utils.promise import Promise
 
 
 class Days(object):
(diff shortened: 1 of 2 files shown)


jh0ker:
Related to #451

jh0ker:
closing in favour of #484

PR #483, opened 2016-12-13
API changes of December 04 (Bot API 2.3.1)

jh0ker:
- Use the new field max_connections in setWebhook to optimize your bot's server load
- Use allowed_updates in setWebhook and getUpdates to selectively subscribe to updates of a certain type. Among other things, this allows you to stop getting updates about new posts in channels where your bot is an admin.
- deleteWebhook moved out of setWebhook to get a whole separate method for itself.

I also renamed the `webhook_url` parameter of `setWebhook` to `url` like in the official API. The old name is still supported through `kwargs`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,7 +84,7 @@ make the development of bots easy and straightforward. These classes are contain
 Telegram API support
 ====================
 
-As of **3. Oct 2016**, all types and methods of the Telegram Bot API are supported.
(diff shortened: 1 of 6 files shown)


tsnoam:
Generally, looks ok. Two things come to mind:
1. Need to update `Updater._bootstrap()` so it will use `deleteWebHook` where applicable.
2. Since we don't unitest that part, make sure to manually test it.

PR #487, opened 2016-12-22
ConversationHandler breaks when bot is also used in Channels

Rostgnom:
When using a `ConversationHandler` for private and group chats in conjunction to having the bot added to a Channel, it will throw a None type exception because it cannot retrieve a chat/user id from the update.

I suggest completely ignoring all channel_posts, because the interactivity that a ConversationHandler provides is not useful in channels anyway.

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -116,6 +116,10 @@ class ConversationHandler(Handler):
         if not isinstance(update, Update):
             return False
 
+        # Ignore messages in channels
+        if update.channel_post:
+            return False
+
         chat, user = extract_chat_and_user(update)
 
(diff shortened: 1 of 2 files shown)


telegram/ext/conversationhandler.py
```diff
@@ -116,6 +116,10 @@ def check_update(self, update):
         if not isinstance(update, Update):
             return False
 
+        # Ignore messages in channels
+        if update.channel_post:
```
tsnoam: Can you please "join" this with the `if` above?

Rostgnom:
Changed the file as requested. I also added a notion that the ConversationHandler currently does not work in a group chat with multiple users, in that the states do only switch for the user performing the transition/action. Is that clear from what I added to the class doc?

PR #494, opened 2016-12-30
No empty strings in to_dict

### LINKED ISSUE #460: User.to_dict() returns absent fields with empty strings '' values
### Steps to reproduce
1. Register any bot in Telegram. Here I will use one with smudged sensitive information.
Get its own information using `getMe` telegram api call.
For example (output pretty-printed):
```
$ http https://api.telegram.org/bot17XXXXX3:AXXcXnXXXXXXXXQgcXXXXc/getMe
HTTP/1.1 200 OK
Access-Control-Allow-Methods: GET, POST, OPTIONS
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: Content-Length,Content-Type,Date,Server,Connection
Connection: keep-alive
Content-Length: 89
Content-Type: application/json
Date: Sun, 06 Nov 2016 22:00:54 GMT
Server: nginx/1.10.1
Strict-Transport-Security: max-age=31536000; includeSubdomains

{
    "ok": true,
    "result": {
        "first_name": "test",
        "id": 17XXXXX3,
        "username": "test_bot"
    }
}
```

2. Try to get the same information using Python Telegram Bot.
Version 3.4.0 (prior to 4.0.0):
```
$ bpython
bpython version 0.16 on top of Python 3.5.1 /Users/el/.pyenv/versions/3.5.1/bin/python3.5
>>> import telegram
>>> bot=telegram.Bot('bot17XXXXX3:AXXcXnXXXXXXXXQgcXXXXc')
>>> user=bot.getMe()
>>> user.to_dict()
{'first_name': 'test', 'id': 17XXXXX3, 'username': 'test_bot'}
```

Version after 4+ (for example 5.2.0):
```
$ bpython
bpython version 0.16 on top of Python 3.5.1 /Users/el/.pyenv/versions/3.5.1/bin/python3.5
>>> import telegram
>>> bot=telegram.Bot('bot17XXXXX3:AXXcXnXXXXXXXXQgcXXXXc')
>>> user=bot.getMe()
>>> user.to_dict()
{'first_name': 'test', 'last_name': '', 'type': '', 'id': 17XXXXX3, 'username': 'test_bot'}```
```

3. 

### Expected behaviour
Should return only fields got with Telegram API (id, first_name, username), do not add empty optional fields (last_name, type).

### Actual behaviour
Returned dict is populated with absent optional fields (last_name, type) with empty string values.

### Configuration
**Operating System:**
MaxOS 10.12.1 
Also tested on Linux boxes in different CI platforms (like Travis-CI, Codeship)

Reproduced on 3.4.5, 3.5.1, 3.5.2, 3.6b4+

```
$ python -m telegram
/Users/el/.pyenv/versions/3.5.1/bin/python: No module named telegram.__main__; 'telegram' is a package and cannot be directly executed
```
### Logs

Issue was found after upgrading python-telegram-bot from 3.2 to 5.2.0 used as an intermediate dependency in my project.

### Reason
One of the reasons to emerge the issue can be a [changes](https://github.com/python-telegram-bot/python-telegram-bot/compare/v3.2...v5.2.0?diff=split&name=v5.2.0#diff-4a13231a9fa379318457405e5df4712bL63) between versions. While dumping user data to dict, old version ignored empty field checking them (``if field: ...``) thus 0, false and some other values could be lost. New version checks them (``if field is not None: ...``) in this case we ignore only real empty (``None``) fields, but method ``User.__init__`` [uses empty string as default values](https://github.com/python-telegram-bot/python-telegram-bot/compare/v3.2...v5.2.0?diff=split&name=v5.2.0#diff-c9df4cc0aa3986f1f87042b987ec5bd4R47) for absent parameters. So User object gets ``''`` as values.

## Possible impact
Can break some other products dependant on returned values of ``getMe`` method.

## Proposition
Change signature of ``User`` constructor to use ``None`` as default value for absent optional parameters.

If you wish I can make a PR with tests.

bomjacob:
Should remove all cases of having empty strings returned from to_dict (if I didn't miss any).

Closes #460


telegram/contact.py
```diff
@@ -39,7 +39,7 @@ class Contact(TelegramObject):
 
     """
 
-    def __init__(self, phone_number, first_name, last_name='', user_id=0, **kwargs):
+    def __init__(self, phone_number, first_name, last_name=None, user_id=0, **kwargs):
```
tsnoam: I think `user_id` should be default to `None`, same reasons like with `Chat` above.

telegram/document.py
```diff
@@ -41,13 +41,14 @@ class Document(TelegramObject):
...
-    def __init__(self, file_id, thumb=None, file_name='', mime_type='', file_size=0, **kwargs):
+    def __init__(self, file_id, thumb=None, file_name=None, mime_type=None, file_size=0, **kwargs):
         # Required
         self.file_id = str(file_id)
         # Optionals
         self.thumb = thumb
         self.file_name = file_name
-        self.mime_type = str(mime_type)
+        if mime_type:
+            self.mime_type = str(mime_type)
```
tsnoam: Please make sure `self.mime_type` is always available. I think that this should be sufficient (no need to use `str()` here):
`self.mime_type = mime_type` 
If you still think `str()` is required, you can do:
`self.mime_type = str(mime_type) if mime_type is not None else None`

telegram/audio.py
```diff
@@ -46,9 +46,9 @@ class Audio(TelegramObject):
     def __init__(self,
                  file_id,
                  duration,
-                 performer='',
-                 title='',
-                 mime_type='',
+                 performer=None,
+                 title=None,
+                 mime_type=None,
                  file_size=0,
```
bomjacob: We have that in a few places... I don't really know...
It's the same on with the booleans on Message.

telegram/audio.py
```diff
@@ -57,7 +57,8 @@ def __init__(self,
         # Optionals
         self.performer = performer
         self.title = title
-        self.mime_type = str(mime_type)
+        if mime_type:
```
bomjacob: I'mma fix it anyways :)

bomjacob: Actually nvm, I saw your comment further down ^^

telegram/chat.py
```diff
@@ -53,11 +53,11 @@ class Chat(TelegramObject):
...
-                 title='',
-                 username='',
-                 first_name='',
-                 last_name='',
-                 all_members_are_admins=False,
+                 title=None,
+                 username=None,
+                 first_name=None,
+                 last_name=None,
+                 all_members_are_admins=None,
```
tsnoam: The API of Chat calls this var as `all_members_are_administrators`, here we use `admins`. Is there anything that translates `admins`. I think we actually have a bug here, or telegram might have silently changed the API.

telegram/file.py
```diff
@@ -40,13 +40,14 @@ class File(TelegramObject):
...
+    def __init__(self, file_id, bot, file_size=None, file_path=None, **kwargs):
         # Required
         self.file_id = str(file_id)
 
         # Optionals
-        self.file_size = int(file_size)
-        self.file_path = str(file_path)
+        self.file_size = file_size
+        if file_path:
+            self.file_path = str(file_path)
```
tsnoam: Just do
`self.file_path = file_path`
The `if` statement should not be there.

telegram/user.py
```diff
@@ -44,7 +44,14 @@ class User(TelegramObject):
         bot (Optional[Bot]): The Bot to use for instance methods
     """
 
-    def __init__(self, id, first_name, type='', last_name='', username='', bot=None, **kwargs):
+    def __init__(self,
+                 id,
+                 first_name,
+                 type=None,
```
tsnoam: I don't see the `type` attribute in documentation. I think we should remove it too.
I can only guess it's something Telegram had obsoleted.
You should leave it at the prototype until ver. 7 I guess, and put a deprecation warning if it specified.

telegram/webhookinfo.py
```diff
@@ -40,13 +40,20 @@ class WebhookInfo(TelegramObject):
 
     """
 
-    def __init__(self, url, has_custom_certificate, pending_update_count, **kwargs):
+    def __init__(self,
+                 url,
+                 has_custom_certificate,
+                 pending_update_count,
+                 last_error_date=None,
+                 last_error_message=None,
```
tsnoam: We're missing optionals: `max_connections` (`int`) & `allowed_updates` (`List[str]`)

telegram/chat.py
bomjacob: I'm about 90% sure they changed their API... so should we change it, or like keep both, or?

telegram/file.py
bomjacob: Right, whoops. I removed the other ones. Seems like I missed these.

telegram/user.py
bomjacob: Hmm.. I don't think it's ever been there tbh.. I might be us who made an error and confused User for Chat or something.

telegram/webhookinfo.py
bomjacob: I noticed but I thought @jh0ker was working on that?
Now that I look at it, it doesn't seem to be in our PR list... (I think I confused it for #483)

telegram/chat.py
tsnoam: We should follow their official API. 
We *can* do bkwd-compat for a while using **kwargs.
@jh0ker What do you think?

telegram/user.py
tsnoam: I think Telegram removed it.
@jh0ker what do you think?

telegram/webhookinfo.py
tsnoam: I thought that too, so I checked the official December 4th release notes and it wasn't there, so lets fix it here while we're at it.

telegram/chat.py
jh0ker: I'd say we keep bkwd-compat

telegram/user.py
jh0ker: I think I remember it from the official docs as well, but I could be wrong...

telegram/webhookinfo.py
bomjacob: Wait no... it totally is part of #483 isn't it, @jh0ker?

PR #499, opened 2017-01-06
Urllib3 stability

tsnoam:
Following the increasing reports about timeout errors and lost connections, this is an attempt to stabilize the infrastructure. This may or may not be related to changes in Telegram servers, but the plain fact of network programming is that you have to be smart with managing your TCP connections.

This is a copy/paste of a summary of the problem in the developers telegram group and the suggested solutions and what was finally chosen:

1. Q: What is the real problem?
A: The real problem is that a HTTP/TCP connection terminates and we don't know it in time.
Why? Too many reasons. But basically that's how the internet works.

2. Q: Where does this hurt us the most?
A: With getUpdates() long polling.
Why? We have 10 seconds (by default, user configurable) that we just wait and "trust" the server to return to us by that time. If the connection drops during that time, we just don't know it.

3. Q: If that so, why does sendMessage() (or other send methods) suffers similar problems?
A: Just like getUpdates() connection may drop, same can happen for sendMessage(). 
I suspect that another problem with sendMessage() is longer timeouts. However, I can't prove it at this time, because I failed of recreating it myself. I'll also need a tcpdump alongside it, checking the TCP flags and retransmits.

4. Q: We never had this before urllib3. What is this shit?!
A: Almost true. Before urllib3 we had many timeout errors during creating connections with Telegram servers. The entire TCP 3way handshake + SSL handshake (espeiclly the SSL handshake) was just too costy and caused lot of pain.
urllib3 allowed us to reuse existing connections and thus saving the expensive time of creating new connection each time. (we don't see this happen anymore) However, reusing connections opened us new problems like connections which we consider alive but they are actually not.
So here we are back to Q1.

Suggested solutions:
1. Add the HTTP header 'connection: keep-alive'. It's a hint to the remote server (Telegram server) that my intention to keep this connection for other uses, so please don't drop it.

2. Use agressive timeouts where possible. urllib3 (and also the standard-python urllib) allows to explicitly specify connect timeout and read timeout.
If urllib3 will decide to reuse existing connection, then the connection timeout will be ignored.
Now, what do we do with the read timeout: 
for getUpdates, I'll set the read timeout to 0.5 second longer than the long polling timeout. This will allow quicker recovery from dropped connection than what we have today. I believe that this will fix most of the problems.
about 'send' methods, I still need to think how to be agressive but not too much.

3. Fine tune the TCP keepalive.
Using SO_KEEPALIVE (like we do) is not enough. The defaults (if I recall correctly) is to send keepalives only after 5 minutes.
There is a way to control the frequency of TCP keepalives, however it is OS specific, so it requires some delicacy. (it's important to mention that I'm a Linux developer without access to MS-Windows, so that's an obstacle)

Eventually solutions 1 & 2 were implemented.
Solution 3 (TCP keepalive tunings)  were left aside for the moment as it requires per OS code (i.e. not portable) and we hope it won't be needed eventually.

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -158,8 +158,13 @@ class Bot(TelegramObject):
         return decorator
 
     @log
-    def getMe(self, **kwargs):
+    def getMe(self, timeout=None, **kwargs):
         """A simple method for testing your bot's auth token.
+
+        Args:
+            timeout (Optional[int|float]): If this value is specified, use it as the read timeout
+                from the server (instead of the one specified during creation of the connection
+                pool).
 
         Returns:
             :class:`telegram.User`: A :class:`telegram.User` instance representing that bot if the
@@ -171,7 +176,7 @@ class Bot(TelegramObject):
         """
         url = '{0}/getMe'.format(self.base_url)
 
-        result = self._request.get(url)
+        result = self._request.get(url, timeout=timeout)
 
         self.bot = User.de_json(result, self)
 
@@ -211,8 +216,9 @@ class Bot(TelegramObject):
                 interface options. A JSON-serialized object for an inline
                 keyboard, custom reply keyboard, instructions to remove reply
                 keyboard or to force a reply from the user.
-            timeout (Optional[float]): If this value is specified, use it as
-                the definitive timeout (in seconds) for urlopen() operations.
+            timeout (Optional[int|float]): If this value is specified, use it as the read timeout
+                from the server (instead of the one specified during creation of the connection
+                pool).
             **kwargs (dict): Arbitrary keyword arguments.
 
         Returns:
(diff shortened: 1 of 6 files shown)


telegram/bot.py
```diff
@@ -1430,7 +1493,7 @@ def getWebhookInfo(self, **kwargs):
 
         data = {}
 
-        result = self._request.post(url, data, timeout=kwargs.get('timeout'))
+        result = self._request.post(url, data, timeout=timeout)
```
MWeesenaar: `timeout` can be `None` here, where the previous timeout where either ints or floats.
Is the behaviour the same for 0 and None?

MWeesenaar:
Reviewed the code, had some low-impact questions with regards to @message in combination of an explicit `timeout` argument. This is, however, no show-stopper for this PR.
LGTM!

PR #513, opened 2017-02-04
Add CommandFilterHandler

jossalgon:
Add a handler that mix CommandHandler with MessageHandler filter.

Now you can filter commands, for example to separate a command from group and private or any filter you want.

```python
updater = Updater(token=TG_TOKEN)
dp = updater.dispatcher

dp.add_handler(CommandFilterHandler('start', filter_example, start))
```

 where filter_example is:
```python
def filter_example(msg):
    return msg.chat.type == 'private'
```

or
```python
updater = Updater(token=TG_TOKEN)
dp = updater.dispatcher

dp.add_handler(CommandFilterHandler('start', lambda msg: msg.chat.type == 'private', start))
```

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -21,7 +21,8 @@
 import logging
 
 from telegram import Update
-from telegram.ext import Handler
+from telegram.ext import (Handler, CallbackQueryHandler, InlineQueryHandler,
+                          ChosenInlineResultHandler)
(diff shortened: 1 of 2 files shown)


jh0ker:
I like the idea, but perhaps it would make more sense to add it to the existing `CommandHandler` instead of adding a new handler class. And it would be good to prevent duplicate code in `CommandHandler` and `MessageHandler` so maybe we should move it to some kind of common superclass.

PR #537, opened 2017-03-02
Introduce MessageQueue

thodnev:
This commit is 2nd step on bringing output messages throughput-limiting mechanisms to the lib.
I've briefly mentioned the ideas behind `MessageQueue` in dev chat before. Here are the main points of the commit:
- New `telegram.ext.messagequeue` module, containing `DelayQueue`, `MessageQueue` classes and `queuedmessage` decorator;
- `DelayQueue` is the core of throughput-limiting mechanism. It runs `Promise` callbacks, dispatched by queue in a thread, with respect to provided limits;
- `MessageQueue` is a class, which embeds two instances of `DelayQueue` in sequential-interconnection way (one for group messages, another for all messages). The figure below illustrates how it works: flow chart
- `queuedmessage` decorator is meant to be used with `telegram.bot.Bot` output conversation methods to make them accept additional kwargs. Messages are passed through `MessageQueue` depending on kwargs set;
- `telegram.ext.messagequeue` module is properly documented for good usability;
- Test of the core (`DelayQueue`) is added.

So, the next step in the process would be to couple `MessageQueue` with `telegram.bot.Bot` in best way ;)

### DIFF
--- a/telegram/ext/messagequeue.py
+++ b/telegram/ext/messagequeue.py
@@ -1,4 +1,25 @@
-'''A throughput-limiting message dispatcher for Telegram bots'''
+#!/usr/bin/env python
+#
+# Module author:
+# Tymofii A. Khodniev (thodnev) <thodnev@mail.ru>
+#
+# A library that provides a Python interface to the Telegram Bot API
+# Copyright (C) 2015-2017
+# Leandro Toledo de Souza <devs@python-telegram-bot.org>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser Public License for more details.
+#
+# You should have received a copy of the GNU Lesser Public License
+# along with this program.  If not, see [http://www.gnu.org/licenses/]
+'''A throughput-limiting message processor for Telegram bots'''
 from telegram.utils import promise
 
 import functools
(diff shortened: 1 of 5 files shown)


Rostgnom:
@thodnev Fucking awesome!!

jh0ker:
Fucking awesome!

telegram/ext/messagequeue.py
```diff
@@ -0,0 +1,309 @@
...
+            The actual `MessageQueue` used to delay outbond messages according
+            to specified time-limits.
+
+    Wrapped method starts accepting the next kwargs:
+
+    Args:
+        queued (:obj:`bool`, optional): if set to ``True``, the `MessageQueue`
+            is used to dispatch output messages.
+            Defaults to `self._is_queued_out`.
+        isgroup (:obj:`bool`, optional): if set to ``True``, the message is
```
jh0ker: We could at least guess this by using the chat id (negative chat_id usually means it's a group). I know it's an implementation detail, but I'd still consider it a sane default in case this is not specified.

telegram/ext/messagequeue.py
```diff
@@ -0,0 +1,309 @@
...
+            Defaults to ``False``.
+
+    Returns:
+        Either :obj:`telegram.utils.promise.Promise` in case call is queued,
+        or original method's return value if it's not.
+    '''
+
+    @functools.wraps(method)
+    def wrapped(self, *args, **kwargs):
+        queued = kwargs.pop('queued', self._is_queued_out)
```
jh0ker: Just to make sure I understand this correctly, `_is_queued_out` will be an attribute of `telegram.Bot`? In that case, I'd prefer a name that includes the word `default`, like `_queue_messages_default` (or maybe something more concise?)

telegram/ext/messagequeue.py
```diff
@@ -0,0 +1,309 @@
...
+                # used to impr. perf. in long-interval calls case
+                times = [now]
+            else:
+                # collect last in current limit time-window
+                times = [t for t in times if t >= t_delta]
+                times.append(now)
+            if len(times) >= self.burst_limit:  # if throughput limit was hit
+                time.sleep(times[1] - t_delta)
+            # finally dispatch one
+            try:
```
jh0ker: See my comment about promises earlier: `item` would be a `telegram.util.promise.Promise` object, right?

telegram/ext/messagequeue.py
```diff
@@ -0,0 +1,309 @@
+'''A throughput-limiting message dispatcher for Telegram bots'''
```
thodnev: Done

telegram/ext/messagequeue.py
```diff
@@ -0,0 +1,309 @@
...
+
+class DelayQueue(threading.Thread):
+    '''Dispatches callbacks from queue with specified throughput limits.
+    Creates a separate thread to dispatch callbacks with delays.
+
+    Args:
+        queue (:obj:`queue.Queue`, optional): used to pass callbacks to
+            thread.
+            Creates `queue.Queue` implicitly if not provided.
+        burst_limit (:obj:`int`, optional): numer of maximum callbacks to
```
thodnev: Done

telegram/ext/messagequeue.py
thodnev: Rejected. An accurate approach (a cached chat_id:type mapping) would be introduced further

telegram/ext/messagequeue.py
thodnev: Done. Changed to `_is_messages_queued_default`. In my opinion, `is` should be left because it's common practice if attribute or variable is boolean

telegram/ext/messagequeue.py
```diff
@@ -0,0 +1,309 @@
...
+            dispatch per time-window defined by `time_limit_ms`.
+            Defaults to 30.
+        time_limit_ms (:obj:`int`, optional): defines width of time-window
+            used when each dispatching limit is calculated.
+            Defaults to 1000.
+        exc_route (:obj:`callable`, optional): a callable, accepting 1
```
thodnev: Promises are used, but `DelayQueue` is left generic because it may be useful outside of the throughput-limiting scope

telegram/ext/messagequeue.py
thodnev: item is a tuple of form `(func, args, kwargs)` which becomes `(promise, (), {})` and `(delayqueue.__call__, (promise, ), {})` for corresponding processors

jh0ker:
Travis shows errors with `yapf`, did you install the pre-commit hook? If you did, and it shows as passing for you, I can try applying the changes manually.

thodnev:
As it may be seen, Travis shows yapf error only on Python 3.5.
Yes, I've used pre-commit hooks, done the 'make yapf' check and also checked coding style with flake8 by hand.
Seems that it's error of build scripts seen on 3.5 only (a Heisenbug ?? cause I see no realistic reasons why it occurs only on this ver)

jh0ker:
@thodnev We only run pre-commit hooks on python3.5 build, so we can save some time and easily see if the tests fail or if it's only the hooks that fail.

PR #545, opened 2017-03-19
Added link to @roolsbot

Rostgnom:


bomjacob:
I personally do not believe there is a need to have this in the readme file. I think it's better suited for the example pages on the wiki :)

tsnoam:
I agree with @bomjacob. Rejecting PR

PR #553, opened 2017-03-28
Update constants.py

revolter:
Add maximum inline query results constant as per [this page][1].

  [1]: https://core.telegram.org/bots/api#answerinlinequery


rahiel:
Thanks for your contribution, this is good to merge.

Issue #556, opened 2017-04-03
ValueError: If 'per_message=False', 'CallbackQueryHandler' doesn't work


mvasilenko:
Hello,

after upgrading python-telegram-bot library, got such error

> ValueError: If 'per_message=False', 'CallbackQueryHandler' doesn't work
> 

in following code 

```
    # Add conversation handler for weather
    conv_handler = ConversationHandler(
        entry_points=[CommandHandler('weather', weather)],
        states={
            LOCATION: [MessageHandler(Filters.location, location),
                       CallbackQueryHandler(button),
                       CommandHandler('skip', skip_location)]
        },
        fallbacks=[CommandHandler('cancel', cancel)],
    )

```

### Configuration
**Operating System:**
Linux

**Version of Python, python-telegram-bot & dependencies:**

``python-telegram-bot 5.3.0
urllib3 1.13.1
certifi 2017.01.23
future 0.16.0
Python 3.5.2 (default, Nov 17 2016, 17:05:23)  [GCC 5.4.0 20160609]
``

I'm new to python-telegram-bot, and looking for advice, how to fix it.
Thank you.

PR #561, opened 2017-04-12
Allow CallbackQueryHandler in ConversationHandler with per_mess…

jh0ker:
…age=False but show a warning 
ref #556


jh0ker:
I'm also happy about other ideas on how to solve this

Issue #571, opened 2017-04-22
Bot crashes, when user clicks on inline button of an inline query result


manorom:
### Steps to reproduce
1. Have a bot that
    a. Has a CallbackQueryHandler.
    b. Has a ConversationHandler registered before the CallbackQueryHandler.
    c. Displays inline buttons on an inline query result.
2. Get a query result with inline buttons and click on them.

### Expected behaviour
The bot handles the callback query.

### Actual behaviour
An `AttributeError: 'NoneType' object has no attribute 'id'` is raised by the ConversationHandler

### Configuration
**Operating System:** Win7 

**Version of Python, python-telegram-bot & dependencies:** 
python-telegram-bot 5.3.0
urllib3 1.20
certifi 2017.04.17
future 0.16.0
Python 3.6.1 (v3.6.1:69c0db5, Mar 21 2017, 18:41:36) [MSC v.1900 64 bit (AMD64)]

**I'm using the current master branch. Not the version from PyPI**

### Logs
```
2017-04-22 17:07:45,210 - telegram.ext.dispatcher - ERROR - An uncaught error wa
s raised while processing the update
Traceback (most recent call last):
  File "C:\ptb-callback-bug\env\lib\site-packages\telegram\ext\dispatcher.py", l
ine 265, in process_update
    if handler.check_update(update):
  File "C:\ptb-callback-bug\env\lib\site-packages\telegram\ext\conversationhandl
er.py", line 173, in check_update
    key = self._get_key(update)
  File "C:\ptb-callback-bug\env\lib\site-packages\telegram\ext\conversationhandl
er.py", line 154, in _get_key
    key.append(chat.id)
AttributeError: 'NoneType' object has no attribute 'id'
2017-04-22 17:08:22,658 - telegram.ext.updater - WARNING - Exiting immediately!
```

### Example Code
```python
from telegram import  (InlineKeyboardMarkup, InlineKeyboardButton,
                      InlineQueryResultArticle, InputTextMessageContent)
from telegram.ext import (Updater, CommandHandler, MessageHandler, Filters,
                          CallbackQueryHandler, ConversationHandler, InlineQueryHandler)

import logging

# Enable logging
logging.basicConfig(format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
                    level=logging.INFO)

logger = logging.getLogger(__name__)

def error(bot, update, error):
    logger.warn('Update "%s" caused error "%s"' % (update, error))


[...]

PR #574, opened 2017-04-23
Fix Bug #571 (second try)

manorom:
ConversationHandler will not process CallbackQuery if per_chat=True and
the CallbackQuery has no message attached to it (as is the case with
buttons on inline results)

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -167,7 +167,8 @@ class ConversationHandler(Handler):
         # Ignore messages in channels
         if (not isinstance(update, Update) or update.channel_post or self.per_chat
                 and (update.inline_query or update.chosen_inline_result) or self.per_message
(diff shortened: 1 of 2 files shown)


jh0ker:
Thank you for your contribution! I would appreciate if you could add a test case for this behaviour in [`/tests/test_conversationhandler.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/tests/test_conversationhandler.py)
Also, feel free to add yourself to [`AUTHORS.rst`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/AUTHORS.rst) :)

manorom:
I'm not 100% sure, but i think 8182dea0f8dd2cc23b9a4231ad767fa6b2b944d2 should catch the error

jh0ker:
Thank you :)

PR #575, opened 2017-04-23
Group filters

Eldinnie:
Add filters:
 - Filters.private (for messages in private chats)
 - Filters.group   (for messages in group chats)


jh0ker:
Looks great :) II would like to request one change: Instead of string literals, could you use `Chat.PRIVATE` and similar constants?

Eldinnie:
yes will do

Eldinnie:
done

jh0ker:
Excellent!

PR #577, opened 2017-04-23
Issue 566

### LINKED ISSUE #566: CommandHandler - Check if update meant to be sent to the bot - Case-sensitivity
When the bot registered its username as `@ThisIsMyBot`, it should be case insensitive in any usage.
But now when commands meant to be sent to the bot with different cases of that from `getMe()`, it gets omitted.

Reference: #515 

### Steps to reproduce
1. Put multiple bots in a group, here for example `@NotMyBot` and `@ThisIsMyBot` (assume all bots have access to all messages)

2. Then use different kinds of command patterns:

- (a) `/command`
- (b) `/command@NotMyBot`
- (c) `/command@ThisIsMyBot`
- (d) `/command@thisismybot`

### Expected behaviour
- (a) Both bots should receive `/command`
- (b) The command was not meant for my bot, so omitted.
- (c) The command was meant for my bot, so receiving the update.
- (d) The command was meant for my bot (even with different cases), so receiving the update.

### Actual behaviour
- (a) Received, correct
- (b) Omiited, correct
- (c) Received, correct
- (d) **The command got omitted**

### Configuration
**Operating System:**
Ubuntu 16.04 LTS

```
python-telegram-bot 5.3.0
urllib3 1.20
certifi 2017.01.23
future 0.16.0
Python 3.5.2 (default, Nov 17 2016, 17:05:23)  [GCC 5.4.0 20160609]
```

Eldinnie:
fixes #566


jh0ker:
Thanks :)

jeffffc:
Thank you for the fix ;)

Issue #579, opened 2017-04-25
urllib3 problems!


persian-telegram:
### Steps to reproduce
1. Just trying to run a simple echobot

2.

3.

### Expected behaviour
Tell us what should happen
- telegram bot should work as normal
- I tried everything and it worked on PythonAnywhere but i can't run it on my VPS.
- I have also tried installing virtualenv but didn't work 

$ virtualenv  telegram-bot
$ source telegram-bot/bin/activate
$ pip install python-telegram-bot

### Actual behaviour
Tell us what happens instead
error comes up
### Configuration
**Operating System:**
ubuntu 14.04

**Version of Python, python-telegram-bot & dependencies:**
python 2.7, latest python-telegram-bot 
``$ python -m telegram``

### Logs
Insert logs here (if necessary)
root@persian-telegram-bot:~# screen -S myTelegramBot

  File "/root/telegram-bot/local/lib/python2.7/site-packages/urllib3/poolmanager.py", line 303, in urlopen
    conn = self.connection_from_host(u.host, port=u.port, scheme=u.scheme)
  File "/root/telegram-bot/local/lib/python2.7/site-packages/urllib3/poolmanager.py", line 219, in connection_from_host
    return self.connection_from_context(request_context)
  File "/root/telegram-bot/local/lib/python2.7/site-packages/urllib3/poolmanager.py", line 232, in connection_from_context
    return self.connection_from_pool_key(pool_key, request_context=request_context)
  File "/root/telegram-bot/local/lib/python2.7/site-packages/urllib3/poolmanager.py", line 245, in connection_from_pool_key
    pool = self.pools.get(pool_key)
  File "/root/telegram-bot/lib/python2.7/_abcoll.py", line 363, in get
    return self[key]
  File "/root/telegram-bot/local/lib/python2.7/site-packages/urllib3/_collections.py", line 53, in __getitem__
    item = self._container.pop(key)
  File "/usr/lib/python2.7/collections.py", line 138, in pop
    if key in self:
TypeError: unhashable type: 'list'
Exception in thread updater:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/root/telegram-bot/local/lib/python2.7/site-packages/telegram/ext/updater.py", line 115, in _thread_wrapper
[...]

PR #580, opened 2017-04-25
enforce urllib3==1.20

jh0ker:
#579


jh0ker:
@persian-telegram I'll get back to you as soon as we know more.

PR #606, opened 2017-05-14
Update copyright notice to include 2017

bomjacob:
Not strictly needed, but it helps show that the project is being actively
developed which I find important.

On a side note we should really figure out what's happening to yapf on windows... I had to skip it to prevent removing the indents of a ton of comments all over the codebase.


tsnoam:
Thanks.

PR #608, opened 2017-05-15
Allow edited as seperate input

### LINKED ISSUE #607: add ability to handle only edited messages updates
Actually if you want to develop a bot where when a user edits his message it replies "you edited this message" or a bot that in groups kicks users that edit their messages (let's imagine a strange rule), there is not a handler to handle only edited messages.

messagehandler just handles any messages and not only those edited. So a handler or a filter to handle only edited messages is needed. Thanks

Eldinnie:
In short made it possible to tune messagehandler more to your wishes. and choose exactly what updates to receive. messages, edited_message or channel_post or a combination.

- Added the edited_updates argument to MessageHandler
- Added DeprecationWarning when using allow_edited
- replaced _is_allowed_message and _is_allowed_channel_post with _is_allowed_update
- Modified tests to reflect new way

fixes #607

### DIFF
--- a/telegram/ext/messagehandler.py
+++ b/telegram/ext/messagehandler.py
@@ -75,6 +75,7 @@ class MessageHandler(Handler):
                 'message_updates, channel_post_updates and edited_updates are all False')
         if allow_edited:
             warnings.warn('allow_edited is getting deprecated, please use edited_updates instead')
+            edited_updates = allow_edited
 
         super(MessageHandler, self).__init__(
             callback,
@@ -83,7 +84,6 @@ class MessageHandler(Handler):
             pass_user_data=pass_user_data,
             pass_chat_data=pass_chat_data)
         self.filters = filters
-        self.allow_edited = allow_edited
         self.message_updates = message_updates
         self.channel_post_updates = channel_post_updates
         self.edited_updates = edited_updates
@@ -97,7 +97,6 @@ class MessageHandler(Handler):
 
     def _is_allowed_update(self, update):
         return any([(self.message_updates and update.message),
-                    (self.allow_edited and update.edited_message),
                     (self.edited_updates and update.edited_message),
                     (self.channel_post_updates and update.channel_post)])
 
--- a/telegram/ext/messagehandler.py
+++ b/telegram/ext/messagehandler.py
@@ -37,8 +37,6 @@ class MessageHandler(Handler):
         callback (function): A function that takes ``bot, update`` as
             positional arguments. It will be called when the ``check_update``
             has determined that an update should be processed by this handler.
-        allow_edited (Optional[bool]): If the handler should also accept edited messages.
-            Default is ``False``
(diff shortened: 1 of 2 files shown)


telegram/ext/messagehandler.py
```diff
@@ -66,9 +68,13 @@ def __init__(self,
...
-                 channel_post_updates=True):
-        if not message_updates and not channel_post_updates:
-            raise ValueError('Both message_updates & channel_post_updates are False')
+                 channel_post_updates=True,
+                 edited_updates=False):
+        if not message_updates and not channel_post_updates and not edited_updates:
+            raise ValueError(
+                'message_updates, channel_post_updates and edited_updates are all False')
+        if allow_edited:
+            warnings.warn('allow_edited is getting deprecated, please use edited_updates instead')
```
evgfilim1: I think you should put a line below such as this:
```python
edited_updates = allow_edited
```

telegram/ext/messagehandler.py
```diff
@@ -80,6 +86,7 @@ def __init__(self,
         self.allow_edited = allow_edited
```
evgfilim1: ... and you can remove this line

telegram/ext/messagehandler.py
```diff
@@ -88,17 +95,14 @@ def __init__(self,
...
-    def _is_allowed_message(self, update):
-        return (self.message_updates
-                and (update.message or (update.edited_message and self.allow_edited)))
-
-    def _is_allowed_channel_post(self, update):
-        return (self.channel_post_updates
-                and (update.channel_post or (update.edited_channel_post and self.allow_edited)))
+    def _is_allowed_update(self, update):
+        return any([(self.message_updates and update.message),
+                    (self.allow_edited and update.edited_message),
```
evgfilim1: ... and this

Eldinnie:
This makes sense. will do

telegram/ext/messagehandler.py
bomjacob: Please don't... if you rename an argument, then it is sooo much nicer to also rename the attribute in the class while you are at it. Then in the future when we wanna remove the deprecated stuff, we don't have to change a ton of stuff.

Eldinnie: I'm not renaming an argument, I'm adding behavior and warning the old behavior is going to be deprecated.
What you suggest would be valid if `allow_edited` would reflect the change but it doesn't.
 If I do it this way, it will implement the new functionality and still work with the old way. If you want to remove the deprecated way in the future you need to remove three lines in the code instead of removing thee and modifying two.

bomjacob: You are indeed right... I'm sorry, I had the arguments flipped around in my head

evgfilim1:
Maybe you should make `allow_edited` and `edited_updates` mutually exclusive?

bomjacob:
They don't really need to be, if one of them is deprecated already

PR #613, opened 2017-05-17
Improved CommandHandler

evgfilim1:
Now you can pass list of commands instead of one command
It will be useful when you define some aliases for command or if commands (such as `/start` and `/help`) has the same behaviour

### DIFF
--- a/telegram/ext/commandhandler.py
+++ b/telegram/ext/commandhandler.py
@@ -30,7 +30,7 @@ class CommandHandler(Handler):
     name and/or some additional text.
 
     Args:
-        commands (str|list): The name of the command or list of commands this handler should
+        command (str|list): The name of the command or list of command this handler should
             listen for.
         callback (function): A function that takes ``bot, update`` as
             positional arguments. It will be called when the ``check_update``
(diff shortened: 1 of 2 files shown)


telegram/ext/commandhandler.py
```diff
@@ -79,7 +80,10 @@ def __init__(self,
             pass_job_queue=pass_job_queue,
             pass_user_data=pass_user_data,
             pass_chat_data=pass_chat_data)
-        self.command = command
+        if isinstance(command, str):
```
jh0ker: this could fail on python 2, if `command` is a `unicode` string

evgfilim1: Fixed

evgfilim1:
@jh0ker Fixed

jh0ker:
Thanks :) I'll merge once the Travis tests are green

evgfilim1:
Oh, I think I've broken this at all
```python
>>> _str = basestring
>>> _str
<type 'basestring'>
>>> '' is _str
False
>>> type('') == _str
False
>>> isinstance(type(''), _str)
False
>>> isinstance(type(u''), _str)
False
>>> type('')
<type 'str'>
>>> type(u'')
<type 'unicode'>
```

Eldinnie:
you should try `isinstance("", _str)` and `isinstance(u"", _str)`

evgfilim1:
@Eldinnie yes, that was my mistake, so I deleted message

Issue #617, opened 2017-05-18
May 18 changes


bomjacob:
### Payment:
- [x] New updates: shipping_query and pre_checkout_query
- [x] New message content fields: invoice and successful_payment
- [x] New bot methods: sendInvoice, answerShippingQuery, and answerPreCheckoutQuery
- [x] New button type: pay (new boolean argument for InlineKeyBoardButton)

### Weird rounded video messages
- [x] New bot method: sendVideoNote
- [x] New Message field: video_note
- [x] New ChatActions: record_video_note or upload_video_note
- [x] VideoNote class

### i18n
- [x] New User field: language_code
Note that the documentation uses "may" wording here... we should do some research as to when.

### Admin power???
- [x] Remove warning from delete_message

### Minor stuff
- [x] Rename new_chat_member to new_chat_member**s**
- [x] New fields gif_duration (InlineQueryResultGif) and mpeg4_duration (InlineQueryResultMpeg4Gif)

Eldinnie:
In Minor stuff `new_chat_members` is an array of `User` and not a single one

bomjacob:
@Eldinnie suggested we coordinate the work here so:
I'm doing the piliminary work on payments rn... not sure I'm up for writing tests this time around though...

PR #625, opened 2017-05-20
Add VideoNote

bomjacob:
refs #617 
This is where it gets weird....
According to telegram length is Video width and height.. but how that works with one parameter I couldn't tell you.. it also seems to be required but is not marked as such in the documentation....

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -687,9 +687,9 @@ class Bot(TelegramObject):
 
         Args:
             chat_id (int|str): Unique identifier for the message recipient - Chat id.
-            voice: Video note to send. Pass a file_id as String to send a video note that exists
-                on the Telegram servers (recommended) or upload a new video. Sending video notes
-                by a URL is currently unsupported
+            video_note (InputFile|str): Video note to send. Pass a file_id as String to send a
+                video note that exists on the Telegram servers (recommended) or upload a new video.
+                Sending video notes by a URL is currently unsupported
             duration (Optional[int]): Duration of sent audio in seconds.
             length (Optional[int]): Video width and height
             disable_notification (Optional[bool]): Sends the message silently. iOS users will not
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -672,6 +672,66 @@ class Bot(TelegramObject):
             **kwargs)
 
     @log
+    def send_video_note(self,
+                        chat_id,
+                        video_note,
+                        duration=None,
+                        length=None,
+                        disable_notification=False,
+                        reply_to_message_id=None,
+                        reply_markup=None,
+                        timeout=20.,
+                        **kwargs):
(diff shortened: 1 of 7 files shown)


telegram/bot.py
```diff
@@ -672,6 +672,66 @@ def send_voice(self,
...
+                        reply_to_message_id=None,
+                        reply_markup=None,
+                        timeout=20.,
+                        **kwargs):
+        """As of v.4.0, Telegram clients support rounded square mp4 videos of up to 1 minute
+        long. Use this method to send video messages
+
+        Args:
+            chat_id (int|str): Unique identifier for the message recipient - Chat id.
+            voice: Video note to send. Pass a file_id as String to send a video note that exists
```
tsnoam: s/voice/vide_note/
type: InputFile|str

telegram/bot.py
```diff
@@ -672,6 +672,66 @@ def send_voice(self,
...
+
+        Raises:
+            :class:`telegram.TelegramError`
+
+        """
+        url = '{0}/sendVideoNote'.format(self.base_url)
+
+        data = {'chat_id': chat_id, 'video_note': video_note}
+
+        if duration:
```
tsnoam: if duration `is not None`
(and same for `length` below)

telegram/inputfile.py
```diff
@@ -66,6 +67,9 @@ def __init__(self, data):
         elif 'certificate' in data:
             self.input_name = 'certificate'
             self.input_file = data.pop('certificate')
+        elif 'video_note' in data:
+            self.input_name = 'video_note'
+            self.input_file = data.pop('video_note')
```
tsnoam: I wonder - we have a lot of branching here. Should we move to a dict here? It's much more pythonic.

telegram/inputfile.py
```diff
@@ -117,7 +121,7 @@ def to_form(self):
                 form_boundary, 'Content-Disposition: form-data; name="%s"' % name, '', str(value)
             ])
 
-        # Add input_file to upload
+# Add input_file to upload
```
tsnoam: Please return the leading indentation

telegram/message.py
```diff
@@ -125,6 +129,7 @@ def __init__(self,
                  sticker=None,
                  video=None,
                  voice=None,
+                 video_note=None,
```
tsnoam: Same note like I left on #628 - should new parameters be added in the middle or in the end of the function?

telegram/inputfile.py
bomjacob: Right, danm yapf

telegram/bot.py
bomjacob: Sure? We don't do that anywhere else in bot.py?

telegram/inputfile.py
bomjacob: Not sure what you meant.. does this work?

PR #628, opened 2017-05-21
May 18 minor changes

bomjacob:
refs #617

### DIFF
--- a/telegram/inlinequeryresultgif.py
+++ b/telegram/inlinequeryresultgif.py
@@ -61,11 +61,11 @@ class InlineQueryResultGif(InlineQueryResult):
                  thumb_url,
                  gif_width=None,
                  gif_height=None,
-                 gif_duration=None,
                  title=None,
                  caption=None,
                  reply_markup=None,
                  input_message_content=None,
+                 gif_duration=None,
                  **kwargs):
 
         # Required
--- a/telegram/inlinequeryresultgif.py
+++ b/telegram/inlinequeryresultgif.py
@@ -32,6 +32,7 @@ class InlineQueryResultGif(InlineQueryResult):
         thumb_url (str): URL of the static thumbnail for the result (jpeg or gif).
         gif_width (Optional[int]): Width of the GIF.
         gif_height (Optional[int]): Height of the GIF.
+        gif_duration (Optional[int]): Duration of the GIF.
         title (Optional[str]): Title for the result.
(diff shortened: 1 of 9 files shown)


telegram/inlinequeryresultmpeg4gif.py
```diff
@@ -60,6 +62,7 @@ def __init__(self,
                  thumb_url,
                  mpeg4_width=None,
                  mpeg4_height=None,
+                 mpeg4_duration=None,
```
tsnoam: We had discussions about this in the past. Should new parameters be added in the location where Telegram documentation added them or in the end of the parameters list (before kwargs) to avoid backward compatibility issues.
I prefer the latter.
@jh0ker what do you think?

telegram/message.py
```diff
@@ -168,6 +169,7 @@ def __init__(self,
         self.location = location
         self.venue = venue
         self.new_chat_member = new_chat_member
```
tsnoam: I think we should help our users by  storing the value on `self._new_chat_member` and make `new_chat_member` a property + deprecation warning.

telegram/user.py
```diff
@@ -99,3 +99,22 @@ def get_profile_photos(self, *args, **kwargs):
         Shortcut for ``bot.getUserProfilePhotos(update.message.from_user.id, *args, **kwargs)``
         """
         return self.bot.getUserProfilePhotos(self.id, *args, **kwargs)
+
+    @staticmethod
+    def de_list(data, bot):
```
tsnoam: Just saying (don't think anything should be changed atm):
JSON can also be a list, so our notion of `de_json` & `to_json` should be fixed. (future todo?)

telegram/inlinequeryresultgif.py
```diff
@@ -59,6 +61,7 @@ def __init__(self,
                  thumb_url,
                  gif_width=None,
                  gif_height=None,
+                 gif_duration=None,
```
tsnoam: Same comment like `mpeg4_duration` below.

telegram/user.py
bomjacob: Hmm, so you're saying we shouldn't both have a de_list and a de_json, but that de_json should do both?

tsnoam: yes. but not now. we need to open an issue so we won't forgot.

PR #630, opened 2017-05-22
Add both handlers for queries from new Payment API

jeffffc:
Added PreCheckoutQueryHandler and ShippingQueryHandler

PR #622, opened 2017-05-19
Payment stuff

bomjacob:
Would like a `paymentbot.py` example script before we merge... It would also serve as a way to test everything, as we're unable to do it automatically.

#617

PR #631, opened 2017-05-22
Add both handlers for queries from new Payment API

jeffffc:
Added PreCheckoutQueryHandler and ShippingQueryHandler

(Re-open after merging #630 to beta branch)


bomjacob:
Oh I'd already handed adding all the needed filters in #622, but no worries, we'll just fix it in the merge I guess. Also it would probably have been better if you'd opened a new PR with all this extra stuff that did not get merged to the `beta` branch, instead of just piling it on, but we'll figure something out :)

tsnoam:
@jeffffc 
1. Would you mind merging from master?
2. When (& after) you merge please note the convention we use at functions like `answer_shipping_query()`:
When checking if an optional parameter was supplied we compare it to `None` like this:
```if ok is not None:
    do_something()
```
It's an important semantic difference as you might want to know that `ok` is actually False in contrary to never supplied.

telegram/ext/filters.py
```diff
@@ -293,6 +293,13 @@ def filter(self, message):
 
     game = _Game()
 
+    class _SuccessfulPayment(BaseFilter):
```
bomjacob: This filter seems to be duplicated as it's been added in another commit too ( see further down)

telegram/message.py
```diff
@@ -107,6 +107,43 @@ class Message(TelegramObject):
         left_chat_participant  (:class:`telegram.User`): Use `left_chat_member`
         instead.
 
+<<<<<<<<< Temporary merge branch 1
```
bomjacob: Something weird happened here....

tests/test_filters.py
```diff
@@ -118,6 +118,12 @@ def test_filters_game(self):
         self.message.game = None
         self.assertFalse(Filters.game(self.message))
 
+    def test_filters_successful_payment(self):
+        self.message.successful_payment = 'test'
+        self.assertTrue(Filters.successful_payment(self.message))
+        self.message.successful_payment = None
+        self.assertFalse(Filters.successful_payment(self.message))
+
```
bomjacob: Could you add a test for invoice too? I know you didn't actually add the filter, but i just saw that it was missing.

jeffffc: sure, meanwhile please check with the updated codes again, thanks a lot!

telegram/bot.py
```diff
@@ -1915,6 +1918,8 @@ def send_invoice(self,
             data['need_name'] = need_name
         if need_phone_number is not None:
             data['need_phone_number'] = need_phone_number
+        if need_email:
```
tsnoam: Please fix to: `if need_email is not None`

telegram/__init__.py
```diff
@@ -92,7 +93,6 @@
 from .shippingquery import ShippingQuery
 from .webhookinfo import WebhookInfo
 from .gamehighscore import GameHighScore
-from .videonote import VideoNote
```
tsnoam: nitpick:
why remove here and add above? (I prefer the minimal change in diff...)

telegram/bot.py
```diff
@@ -1926,7 +1931,9 @@ def answer_shipping_query(self,
                               shipping_query_id,
                               ok,
                               shipping_options=None,
-                              error_message=None):
+                              error_message=None,
+                              timeout=None,
```
tsnoam: timeout should not be explicit here. please remove it. (it should be part of kwargs)

telegram/bot.py
```diff
@@ -1951,18 +1959,32 @@ def answer_shipping_query(self,
...
         data = {'shipping_query_id': shipping_query_id, 'ok': ok}
 
-        if shipping_options is not None:
+        if ok is True:
             data['shipping_options'] = shipping_options
         if error_message is not None:
             data['error_message'] = error_message
 
-        return url, data
+        result = self._request.post(url_, data, timeout=timeout)
```
tsnoam: please change to
```
result = self._request.post(url_, data, **kwargs)
```

telegram/bot.py
```diff
@@ -1951,18 +1959,32 @@ def answer_shipping_query(self,
...
 
-    def answer_pre_checkout_query(self, pre_checkout_query_id, ok, error_message=None):
+        return result
+
+    def answer_pre_checkout_query(self, pre_checkout_query_id, ok,
+                                  error_message=None, timeout=None, **kwargs):
```
tsnoam: please remove the `timeout` argument. (like above)

telegram/bot.py
```diff
@@ -1985,14 +2008,24 @@ def answer_pre_checkout_query(self, pre_checkout_query_id, ok, error_message=Non
             :class:`telegram.TelegramError`
 
         """
-        url = '{0]/answerPreCheckoutQuery'.format(self.base_url)
+
+        if ((ok is not True and error_message is None) or
+                (ok is True and error_message is not None)):
```
tsnoam: In here, for readability I would prefer a shorter form like this:
```
if not (ok ^ (error_message is None)):
```

telegram/bot.py
```diff
@@ -1985,14 +2008,24 @@ def answer_pre_checkout_query(self, pre_checkout_query_id, ok, error_message=Non
...
+            raise TelegramError(
+                'answerPreCheckoutQuery: If ok is True, there should '
+                'not be error_message; if ok is False, error_message '
+                'should not be empty')
```
tsnoam: Open for discussion:
I don't think that we should raise an exception here. If user wants to do unorthodox things, he might know better than us. At most I would issue a warning here.

telegram/bot.py
```diff
@@ -1985,14 +2008,24 @@ def answer_pre_checkout_query(self, pre_checkout_query_id, ok, error_message=Non
...
+
+        url_ = '{0}/answerPreCheckoutQuery'.format(self.base_url)
 
         data = {'pre_checkout_query_id': pre_checkout_query_id, 'ok': ok}
 
         if error_message is not None:
             data['error_message'] = error_message
 
-        return url, data
+        result = self._request.post(url_, data, timeout=timeout)
```
tsnoam: should be:
```
result = self._request.post(url_, data, **kwargs)
```

telegram/ext/precheckoutqueryhandler.py
```diff
@@ -0,0 +1,70 @@
...
+        super(PreCheckoutQueryHandler, self).__init__(
+            callback,
+            pass_update_queue=pass_update_queue,
+            pass_job_queue=pass_job_queue,
+            pass_user_data=pass_user_data,
+            pass_chat_data=pass_chat_data)
+
+    def check_update(self, update):
+        if isinstance(update, Update) and update.pre_checkout_query:
+            return True
```
tsnoam: too long.... this would be sufficient:
```
return isinstance(update, Update) and update.pre_checkout_query
```

telegram/ext/shippingqueryhandler.py
```diff
@@ -0,0 +1,70 @@
...
+        super(ShippingQueryHandler, self).__init__(
+            callback,
+            pass_update_queue=pass_update_queue,
+            pass_job_queue=pass_job_queue,
+            pass_user_data=pass_user_data,
+            pass_chat_data=pass_chat_data)
+
+    def check_update(self, update):
+        if isinstance(update, Update) and update.shipping_query:
+            return True
```
tsnoam: same comments like above (too long....)

telegram/message.py
```diff
@@ -106,7 +106,6 @@ class Message(TelegramObject):
 
         left_chat_participant  (:class:`telegram.User`): Use `left_chat_member`
         instead.
-
```
tsnoam: please don't remove that line.

telegram/precheckoutquery.py
```diff
@@ -33,6 +33,8 @@ class PreCheckoutQuery(TelegramObject):
         currency (str): Three-letter ISO 4217 currency code
         total_amount (int): Total price in the smallest units of the currency (integer)
         invoice_payload (str): Bot specified invoice payload
+
+    Keyword Args:
```
tsnoam: These are not `Keyword Args` in the sense of how documentation is generated. Please remove.
Also, `bot` is missing from the docstring, please add it..

telegram/update.py
```diff
@@ -59,6 +64,8 @@ def __init__(self,
                  inline_query=None,
                  chosen_inline_result=None,
                  callback_query=None,
+                 shipping_query=None,
+                 pre_checkout_query=None,
```
tsnoam: For backward compatibility (if some1 for any reason, decided to use `Update` directly) please add the new parameters to the end (just before **kwargs).

telegram/__init__.py
bomjacob: That's on me... I moved it long ago to fix a circular import issue I had... not sure why it's showing up in this diff though...

telegram/bot.py
tsnoam: I take back this part. please ignore.

telegram/bot.py
bomjacob: Warning seems better... especially since we don't raise errors anywhere else if arguments are wrong, so it would be a bit strange to do it here imo

tsnoam: I'm ok with a warning.

jeffffc: okay for this one i believe i checked with the other existing methods beforehand about these mutual exclusive parameters
i raise exception there was because i was following `edit_message_caption` and similar ones with either inline messages or normal messages
what do you guys think?

telegram/bot.py
jeffffc: sure i could not think of such short version, thank you

telegram/bot.py
jeffffc: @tsnoam @bomjacob what do you guys think about this?

bomjacob: Hmm.. you seen to be correct... I don't really have a preference either way then... It's your call @tsnoam

telegram/__init__.py
tsnoam: well, what every git voodoo happened here, I prefer it not to be part of this PR

telegram/bot.py
tsnoam: @jeffffc 
lol... `edit_message_caption` exception is my doing. lets keep it that way.

tsnoam:
@jeffffc 
I'm sorry for the delay in my response, I had several busy days.
I responded to your question above.

jeffffc:
@tsnoam It's alright!
and @bomjacob I pushed the fixes already, please have a brief check, thanks!

tsnoam:
LGTM. Waiting for travis before merging.

tsnoam:
@jeffffc 
Thank you for tyour contribution.

jeffffc:
great, thank you all ;)

PR #634, opened 2017-05-22
Avoid confusion with user's urllib3 by renaming vendored urllib3

tsnoam:


Eldinnie:
Seems good to me if Travis approves.

Issue #635, opened 2017-05-24
Build via setuptools fails


evgfilim1:
### Steps to reproduce
1. `git clone https://github.com/python-telegram-bot/python-telegram-bot && cd python-telegram-bot`
2. `git submodule update --init --recursive`
3. `python setup.py`

### Expected behaviour
Successful build

### Actual behaviour
Build fails: `error: package directory 'telegram/vendor/urllib3/urllib3' does not exist`

### Configuration
**Operating System:**
`ArchLinux x86_64 linux-ck 4.11.2-1-ck`

**Version of Python, python-telegram-bot & dependencies:**
```bash
$ python -m telegram
python-telegram-bot 6.0.1
urllib3 1.21.1
certifi 2017.04.17
future 0.16.0
Python 3.6.1 (default, Mar 27 2017, 00:27:06)  [GCC 6.3.1 20170306]
```
Using latest commit from `master` branch:
`ff897ce (HEAD -> master, origin/master, origin/HEAD) Avoid confusion with user's urllib3 by renaming vendored urllib3`

### Logs
```
running build
running build_py
creating build
creating build/lib
creating build/lib/telegram
copying telegram/inlinequeryresultcachedvideo.py -> build/lib/telegram
copying telegram/emoji.py -> build/lib/telegram
copying telegram/user.py -> build/lib/telegram
copying telegram/voice.py -> build/lib/telegram
copying telegram/inlinekeyboardmarkup.py -> build/lib/telegram
copying telegram/userprofilephotos.py -> build/lib/telegram
copying telegram/photosize.py -> build/lib/telegram
copying telegram/__main__.py -> build/lib/telegram
copying telegram/document.py -> build/lib/telegram
copying telegram/inlinequeryresultcachedphoto.py -> build/lib/telegram
copying telegram/audio.py -> build/lib/telegram
copying telegram/inlinequeryresultmpeg4gif.py -> build/lib/telegram
copying telegram/inlinequeryresultcachedmpeg4gif.py -> build/lib/telegram
copying telegram/inlinequeryresultgame.py -> build/lib/telegram
copying telegram/replymarkup.py -> build/lib/telegram
copying telegram/replykeyboardmarkup.py -> build/lib/telegram
copying telegram/callbackgame.py -> build/lib/telegram
copying telegram/chatmember.py -> build/lib/telegram
copying telegram/inlinequeryresultaudio.py -> build/lib/telegram
[...]

PR #636, opened 2017-05-26
Travis: add test to make sure setup.py doesn't break

tsnoam:
Refs #635


bomjacob:
Should we maybe only do this once for py2 and once for py3 as to not make the tests take that much more time? (note didn't actually time it, but I'd assume it takes a bit of time to compile)

Issue #589, opened 2017-05-05
Problem parsing image


pattyland:
Hi there,

I have problems parsing a jpg image. I already postet this issue at https://github.com/home-assistant/home-assistant/issues/7413 but I'm not sure wether this a problem of python-telegram-bot parsing the image wrong or hass creating a corrupt image!

Greets, Sören

rahiel:
Could you share an example jpg image that fails to send?

pattyland:
Sure! Should have done this in the first place. I zipped it to be sure that GitHub doesn't tamper with it: [camera.fang.jpg.zip](https://github.com/python-telegram-bot/python-telegram-bot/files/981909/camera.fang.jpg.zip)

rahiel:
I can send the image just fine with `telegram.Bot.send_photo`, so I think the issue is with home-assistant.

pattyland:
@rahiel Could test this again while serving the image from any webserver? It's working fine when I send the image using the file option from Telegram, but not the URL

rahiel:
@pattyland posting your images here: 
![camera fang](https://cloud.githubusercontent.com/assets/6839756/26528267/0df8c88e-43a8-11e7-9e7a-d3820533e437.jpg)

This image hosted by github has the same sha256 hash as the image from your zip, so they're completely the same. I tested with
``` python
bot.send_photo(chat_id=chat_id, photo="https://cloud.githubusercontent.com/assets/6839756/26528267/0df8c88e-43a8-11e7-9e7a-d3820533e437.jpg")
```
and it works.

pattyland:
Thank you again for your help! I'm still searching the cause for this problem :/

tsnoam:
@pattyland 
Please note that when sending a photo using URL you are bound to the behaviour of Telegram servers in regards to identifying that the file type is what it should be. So for example, make sure that the URL contains '.jpg' in the end (if it's a jpeg, obviously).

pattyland:
@tsnoam Thanks for the tip! So the GitHub CDN URL should work, but it's not:
```bash
May 28 11:29:21 jarvis hass[19081]: INFO:homeassistant.core:Bus:Handling <Event call_service[L]: service_data=url=https://cloud.githubusercontent.com/assets/6839756/26528267/0df8c88e-43a8-11e7-9e7a-d3820533e437.jpg, domain=telegram_bot, service=send_photo, service_call_id=1977728880-22>
[...]

PR #641, opened 2017-05-29
fix setting default mimetype of inputfile

azogue:
## Description

When **sending a photo as a bytesio stream** and the mimetype isn't detected with the stream header, don't try to guess the type with `self.filename` if filename is `None`. In that case, set the default `application/octet-stream` mimetype.

This fixes this [problem](https://github.com/home-assistant/home-assistant/issues/7413#issuecomment-304508296), because `imghdr.what()` fails when analysing the pic in discuss (https://cloud.githubusercontent.com/assets/6839756/26528267/0df8c88e-43a8-11e7-9e7a-d3820533e437.jpg), so it's related to #589, closed today by @tsnoam.


tsnoam:
@azogue 
Would you be so kind and add a unitest for that scenario?

azogue:
Testing the pic in discuss and comparing the resulting mimetype? in `test_photo.py`? Do I include the pic with the non standard header in `tests/data/`?

tsnoam:
@azogue Exactly like you decribed, and also try to actually send the photo.
Thank you.

jh0ker:
Thanks :) Travis fails only because of video notes, and the change looks good. Would you like to add yourself to https://github.com/python-telegram-bot/python-telegram-bot/blob/master/AUTHORS.rst before I merge?

tsnoam:
@azogue Thanks for your contribution. Merging now.

PR #642, opened 2017-06-01
New filters

evgfilim1:
* Added new `status_update` filters
* Fixed tests to use new filters


bomjacob:
Please also leave the old tests in place.
Also is there a particular reason for using bitwise or instead of actual or? The or keyword returns as soon as one is true, while the bitwise has to check everything (so or keyword should be faster in theory.

evgfilim1:
@bomjacob Done, thanks

PR #649, opened 2017-06-07
Fix some documentation stuff

bomjacob:
- Adds favicon
- Fixes most warnings. The only remaining ones are things where I couldn't figure out why it was even a warning, or it was complaining that `float`s didn't exist or something...
- Automatically excludes all functions (methods inside of classes are seen as functions too in this case) that have any capital letters in them. This makes the `Bot` documentation about 50% nicer to read by removing e.g. sendAudio but keeping send_audio.


tsnoam:
LGTM

PR #653, opened 2017-06-08
fix tests for pycharm windows

Eldinnie:
Refractored self.id in new tests to self._id (as I did before) so testrunner in pycharm can run under Windows.


tsnoam:
LGTM. Just waiting for travis to finish...

PR #657, opened 2017-06-10
Decrease test collisions

bomjacob:
Randomize test order in an attempt to not have test collisions using [nose-randomly](https://pypi.python.org/pypi/nose-randomly).

We'll see if how big a benefit it gives us :)

We might also wanna have multiple test bots in the future so we don't get rate limited as much.

### DIFF
--- a/tests/test_audio.py
+++ b/tests/test_audio.py
@@ -28,6 +28,7 @@ sys.path.append('.')
 
 import telegram
 from tests.base import BaseTest, timeout
+from tests.reupload import get_file_id
 
 
 class AudioTest(BaseTest, unittest.TestCase):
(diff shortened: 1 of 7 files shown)


Eldinnie:
I restarted the build, I guess some errors were because i was simultaniously testing on appveyor and travis aswell.

Eldinnie:
It's no good... Don't know why exactly but stuff is failing

tests/test_audio.py
```diff
@@ -28,14 +28,15 @@
...
+from tests.reupload import get_file_id
 
 
 class AudioTest(BaseTest, unittest.TestCase):
     """This object represents Tests for Telegram Audio."""
 
     def setUp(self):
         self.audio_file = open('tests/data/telegram.mp3', 'rb')
-        self.audio_file_id = 'CQADAQADDwADHyP1B6PSPq2HjX8kAg'
+        self.audio_file_id = get_file_id('audio')
```
Eldinnie: This would upload the file for every test in test_audio. setUp() is called before every test. I would go for [setUpClass](https://docs.python.org/2/library/unittest.html#unittest.TestCase.setUpClass) in this case

bomjacob: Oh right... Whoops. Originally had it only do it when the class was imported (which would make it only run once in total over the entire lifetime of the tests). But along the way I realised that in some cases there's no need to upload all files if you for example are only running audio tests. But since it would have to way of figuring out which tests are gonna be run, I'll just switch it back to uploading on import I think. Unless you have a better idea? :)

bomjacob: Actually.... are you sure? In get_file_id there's a test to see if it's already been uploaded... wouldn't that be global across all tests in theory?

bomjacob:
Superseeded by #681

PR #658, opened 2017-06-10
Add PaymentBot Example, Bugfixes on payment methods

jeffffc:
Added Payment Example with both With-Shipping and Without-Shipping commands
Fixed error in answer_shipping_query where the shipping options are not populated from a list
Fixed error in answer_precheckout_query that the boolean checks was wrong

### DIFF
--- a/examples/paymentbot.py
+++ b/examples/paymentbot.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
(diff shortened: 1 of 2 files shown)


bomjacob:
Could you add a link to https://core.telegram.org/bots/payments#supported-currencies in places where you prefer to needed to *100?

bomjacob:
Thanks :)
LGTM

jeffffc:
my pleasure ;)

PR #662, opened 2017-06-11
Simplification of boolean checks

saschalalala:
While I worked on the other pull request, I saw that the boolean checks for the payment functions are somewhat unpythonic (See https://www.python.org/dev/peps/pep-0008/#programming-recommendations the last bullet point for reference).

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -1960,6 +1960,8 @@ class Bot(TelegramObject):
 
         """
 
+        ok = bool(ok)
+
         if ok and (shipping_options is None or error_message is not None):
             raise TelegramError(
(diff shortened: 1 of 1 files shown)


Eldinnie:
Look good, let's wait if Travis aggrees

jh0ker:
It was discussed in the developer group that while this change is good, we still should enforce `ok` to be a `bool` because else, we'll run into problems with the JSON encoding. So it would be nice if you added `ok = bool(ok)` to make sure of that.

saschalalala:
Will do.

saschalalala:
This should do it, I guess. Or should I add a comment for clarification?

jh0ker:
Thanks, it looks good :)

Issue #666, opened 2017-06-13
Add stop propagation logic


s-chizhik:
Hi! I see that `dispatcher` can take handlers to different groups:

```python
for group in self.groups:
    for handler in self.handlers[group]:
        try:
            if handler.check_update(update):
                handler.handle_update(update, self)
                break
        # Dispatch any errors
        except TelegramError as te:

```

Such realization doesn't allow us to cleanly stop handling from a group with a lower number. How do you think about some logic, which will make it? Ex, action handler can raise some exception, like `StopPropagationExeption`:
```python
def custom_handler(bot: Bot, update: Update):
    # handling stuff...
    raise StopPropagationException # or maybe wrap it to bot.updater.stop_propagation()

dispatcher.add(custom_handler)
```

```python
for group in self.groups:
    for handler in self.handlers[group]:
        try:
            if handler.check_update(update):
                handler.handle_update(update, self)
                break
        # Dispatch any errors
        except TelegramError as te:
    except StopPropagationExeption:
        pass
```

I've just noticed, that group mechanism allows us to create something like middleware, to handle such things as authorization or smth else. I suppose this logic would be useful.

jh0ker:
Hello,
This is indeed a duplicate of #559, but I didn't answer it on github yet, so:

Yes, adding an Exception like you proposed would be a very straightforward way to solve this. However, it has one issue: The return value of callbacks is used by the `ConversationHandler` to determine the new state. So, you couldn't have both, obviously. Perhaps this is not much of a problem, because you wouldn't want to use the two in combination. 
Using a method like `bot.dispatcher.stop_propagation()` would be an option, but it would be a bit strange, especially since there is no real reason for `Bot` to have a reference to `Dispatcher`. What do you think?

s-chizhik:
> This is indeed a duplicate of #559

Oh, I see it. Just haven't noticed that issue :(

> Yes, adding an Exception like you proposed would be a very straightforward way to solve this. However, it has one issue: The return value of callbacks is used by the ConversationHandler to determine the new state. So, you couldn't have both, obviously. Perhaps this is not much of a problem, because you wouldn't want to use the two in combination.

Probably you are right. I guess exception is better than return value. In case when user need to change a state of `ConversationHandler` – exception can include some field with needed state, and be handled by `ConversationHandler` which will set state and throw an exception to the dispatcher.

[...]

PR #667, opened 2017-06-14
Add `stop propagation` logic

s-chizhik:
Described in #666


jh0ker:
This looks pretty good, thank you :) I think we can simply ignore the `ConversationHandler` for this and say that you can either change state OR stop propagation. However, I would like to see some unit tests (similar to what you had in #666), in `test_updater.py` I guess.

tsnoam:
@s-chizhik 
We are aware that your PR was here first, however #738 covers an additional scenario and has unitests.
In general you've both came to a similar solution which kinda suggests that this is going in the right direction.
If you are still interested in seeing this integrated, I suggest that you join forces with @ihoru and work on top of #738.
Therefor I am closing this PR.

PR #671, opened 2017-06-15
Make error output more clear for too big uploads.

alateas:
I have faced unclear error output('Invalid server response') when tried to send big audio ~ 200mb via bot.send_audio. It was hard to understand that now you can upload files up to 50 MB using telegram api. So my small fix makes error output more clear by doing two things:
 - moving data pasing block after error checking
 - adding 413(Request Entity Too Large) status to check

### DIFF
--- a/telegram/utils/request.py
+++ b/telegram/utils/request.py
@@ -183,6 +183,11 @@ class Request(object):
             # 200-299 range are HTTP success statuses
             return resp.data
 
+        try:
+            message = self._parse(resp.data)
+        except ValueError:
+            message = 'Unknown HTTPError'
+
         if resp.status in (401, 403):
             raise Unauthorized(message)
         elif resp.status == 400:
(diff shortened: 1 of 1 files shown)


jh0ker:
Please see the travis log

alateas:
Ok, I see, there is a lot of connections with this error handling part. Client code and unit tests expect a certain behavior.  So i will think ho I can eject 413 error message gentle way.

jh0ker:
@alateas That's not the point. The point is that you moved the part where message is parsed further down, and now other parts of the code that need that variable don't work

jh0ker:
Oh I'm sorry, you already fixed that :smile:

alateas:
Reason of moving block of http error handling( https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/utils/request.py#L191-L200 ) up is that this block will never be executed. Because if we, for example, get a 502 http error, any way data will parsed in 187 line and json parsing will raise ValueError in 134 line.

Any way my last fix was wrong too :)

Eldinnie:
Can't you insert the check between the 200-209 check and the message parsing in the original file? It doesn't need the message.

alateas:
I can, but this check the similar with other http error checs at lines 191-200, will be strange to separate 413 http error check.

tsnoam:
@alateas I have a fix for your fix :-)
however, it seems that you haven't allowed edits to your branch by maintainers. Can you please allow it and I'll push? See:
https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

alateas:
@tsnoam Good news! Actually I have allowed maintainers edits, I rechecked it right now.

tsnoam:
I failed to push my commit. I wonder if I do something wrong.
Anyhow, I just edited the code using the github web interface. I hope I haven't made any typo.
Now we'll wait for travis to see if it works.

Eldinnie:
I I see this correctly. The way it is in the latest commit it will never raise an error if there is an error but not in one of the error codes. Shouldn't we add an
```
else:
  raise NetworkError(message)
```
at the bottom?

Eldinnie:
Nevermind, it just fell off the screen. It looks good to me.

tsnoam:
@alateas thanks for your contribution.

PR #674, opened 2017-06-18
Move a lot of files

### LINKED ISSUE #619: Split internal code for better maintainability
Currently all our `telegram.*` classes are all jumbled up in a single folder, and it's starting to get a bit unwieldy. I propose we split it into a few folders - much like the telegram documentation also does (Inline, Games, Payments).
We can still have everything exported in the `telegram` module (and we _really_ should), so it should be fully backwards compatible provided you've done something like `from telegram import User` or simply `import telegram`, - in fact the only time it wouldn't work is if they've hardcoded the module such as `from  telegram.user import User`, but I've never actually seen anyone do that, so I think we're fine.

bomjacob:
Moves a lot of files into seperate packages so it's easier to maintain.
I have a few questions:

- Should `files` be renamed to `media`?
- Should we give the `tests` directory the same treatment?
- Should all the subpackages have a proper `__init__.py` file with `__all__` despite not being meant for import by the user?
- Should the subpackages have an underscore prepended since they are internal?

Closes #619


Eldinnie:
I would love it if the imports would be grouped (all games imports together) and in alphabetical order too. It's not necessary, but it's looks a lot cleaner

PR #675, opened 2017-06-18
Allow filters to have a name.

bomjacob:
So their string representation is user friendly.

### Old
```
<telegram.ext.filters.MergedFilter consisting of <telegram.ext.filters.MergedFilter consisting of <telegram.ext.filters.MergedFilter consisting of <telegram.ext.filters.MergedFilter consisting of <telegram.ext.filters.Filters._Text object at 0x000002B28BDEFD68> or <telegram.ext.filters.Filters._Command object at 0x000002B28BDEFE48>> and <telegram.ext.filters.Filters._Reply object at 0x000002B28BDEFE10>> and <telegram.ext.filters.InvertedFilter inverting <telegram.ext.filters.Filters._Forwarded object at 0x000002B28BDA66D8>>> and <telegram.ext.filters.MergedFilter consisting of <telegram.ext.filters.Filters.entity object at 0x000002B28BDB3C50> or <telegram.ext.filters.Filters.entity object at 0x000002B28BDB3C88>>>
```
### New
```
<<<<Filters.text or Filters.command> and Filters.reply> and <inverted Filters.forwarded>> and <Filters.entity(url) or Filters.entity(text_link)>>
```

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -53,6 +53,9 @@ class BaseFilter(object):
     (so remember to initialize your filter classes).
     """
 
+    def __init__(self, *, name=None):
+        self.name = name
+
     def __call__(self, message):
         return self.filter(message)
 
@@ -64,6 +67,15 @@ class BaseFilter(object):
 
     def __invert__(self):
         return InvertedFilter(self)
+
+    def __str__(self):
+        # Do not rely on classes overwriting __init__ to set a name
+        # so we can keep backwards compatibility
+        if not hasattr(self, 'name') or self.name is None:
+            self.name = self.__class__.__name__
+        return self.name
+
+    __repr__ = __str__
 
     def filter(self, message):
         raise NotImplementedError
@@ -77,13 +89,14 @@ class InvertedFilter(BaseFilter):
     """
 
     def __init__(self, f):
+        super().__init__()
         self.f = f
 
     def filter(self, message):
         return not self.f(message)
 
     def __str__(self):
-        return "<telegram.ext.filters.InvertedFilter inverting {}>".format(self.f)
+        return "<inverted {}>".format(self.f)
 
     __repr__ = __str__
 
@@ -98,6 +111,7 @@ class MergedFilter(BaseFilter):
     """
 
     def __init__(self, base_filter, and_filter=None, or_filter=None):
+        super().__init__()
         self.base_filter = base_filter
         self.and_filter = and_filter
         self.or_filter = or_filter
@@ -109,9 +123,8 @@ class MergedFilter(BaseFilter):
             return self.base_filter(message) or self.or_filter(message)
 
     def __str__(self):
-        return ("<telegram.ext.filters.MergedFilter consisting of"
-                " {} {} {}>").format(self.base_filter, "and" if self.and_filter else "or",
-                                     self.and_filter or self.or_filter)
+        return "<{} {} {}>".format(self.base_filter, "and" if self.and_filter else "or",
+                                   self.and_filter or self.or_filter)
 
     __repr__ = __str__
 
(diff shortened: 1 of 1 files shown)


telegram/ext/filters.py
```diff
@@ -65,6 +68,15 @@ def __or__(self, other):
...
         return InvertedFilter(self)
 
+    def __str__(self):
+        # Do not rely on classes overwriting __init__ to set a name
+        # so we can keep backwards compatibility
+        if not hasattr(self, 'name') or self.name is None:
+            self.name = self.__class__.__name__
+        return self.name
+
+    __repr__ = __str__
```
tsnoam: it's sufficient to implement `__repr__`.  It will implicitly get `__str__` working as you want

telegram/ext/filters.py
```diff
@@ -77,13 +89,14 @@ class InvertedFilter(BaseFilter):
...
         self.f = f
 
     def filter(self, message):
         return not self.f(message)
 
     def __str__(self):
-        return "<telegram.ext.filters.InvertedFilter inverting {}>".format(self.f)
+        return "<inverted {}>".format(self.f)
 
     __repr__ = __str__
```
tsnoam: same comment about `__repr__` vs `__str__`

telegram/ext/filters.py
```diff
@@ -65,6 +68,15 @@ def __or__(self, other):
     def __invert__(self):
         return InvertedFilter(self)
 
+    def __str__(self):
+        # Do not rely on classes overwriting __init__ to set a name
+        # so we can keep backwards compatibility
```
tsnoam: what about defining `name` as class attribute which defaults to `'<user_filter>'` (or something of the sort)?
then we can save the ugly code below...
Following that idea, you can then add `name` attribute to each of the inheriting classes which will reduce the need to call `super().__init__()` and modify all the lines instantiating the filters.

telegram/ext/filters.py
```diff
@@ -53,6 +53,9 @@ class BaseFilter(object):
     (so remember to initialize your filter classes).
     """
 
+    def __init__(self, name=None):
+        self.name = name
+
```
tsnoam: This can be avoided, see my comment in `__str__` below

telegram/ext/filters.py
```diff
@@ -109,9 +123,8 @@ def filter(self, message):
             return self.base_filter(message) or self.or_filter(message)
 
     def __str__(self):
-        return ("<telegram.ext.filters.MergedFilter consisting of"
-                " {} {} {}>").format(self.base_filter, "and" if self.and_filter else "or",
-                                     self.and_filter or self.or_filter)
+        return "<{} {} {}>".format(self.base_filter, "and" if self.and_filter else "or",
+                                   self.and_filter or self.or_filter)
 
     __repr__ = __str__
```
tsnoam: same comment about `__repr__` vs `__str__`

telegram/ext/filters.py
tsnoam: on a second thought: what's wrong with using `self.__class__.__name__`?

bomjacob: Currently it's using`self.__class__.__name__` only if no other name is found. The reason we don't use that is that `telegram.ext.Filters.status_update.migrate.__class__.__name__` is `_Migrate` and not `migrate` as we want. So we have to overwrite it.
Or did I misunderstand your question?
I would prefer if it did use `self.__class__.__name__` when a better name is not found though. So if we had name as a class attribute and defaulted it to None (where it would then do `self.__class__.__name__`) would that work?

telegram/ext/filters.py
bomjacob: Alright thanks :)

telegram/ext/filters.py
tsnoam: so back to my first suggest. `name = <user_filter>` by default

bomjacob: But I'd rather have it be `self.__class__.__name__` by default though...?

PR #677, opened 2017-06-19
New filters for handling messages from specific chat/user id

evgfilim1:


bomjacob:
I'd personally prefer if these filter were made a bit smarter by allowing them to use a username or a User object.
Either by doing something akin to:
`def __init__(self,  user_id=None, username=None, user=None):`
or by doing some type checking using isinstance, so you don't need several kwargs (this might be cleaner since they would be mutually exclusive?) (or maybe not, since type checking isn't really pythonic?)

evgfilim1:
@bomjacob I think that user object is not very necessary because filters are usually used in handlers. I don't think that anyone would create `User` instance for it when they can specify user id or username

evgfilim1:
@bomjacob Done

bomjacob:
Okay yeah, that's probably true, looks good then :)
As far as error checking goes, we could probably have gotten away with simply:
`if not (chat_id or username):`
But I guess this works too :)

telegram/ext/filters.py
```diff
@@ -324,6 +324,65 @@ def filter(self, message):
 
     group = _Group()
 
+    class user(BaseFilter):
```
tsnoam: class names are CamelCase according to pep8. Please fix.

telegram/ext/filters.py
```diff
@@ -324,6 +324,65 @@ def filter(self, message):
...
+        """Filters messages to allow only those which are from specified user ID
+
+        Args:
+            user_id(Optional[int]): which user ID to allow through. Required if username is not
+                specified
+            username(Optional[str]): which username to allow through. Required if user_id is not
+                specified
+
```
tsnoam: 1. Documentation should be explicit on the fact that one and only one of `user_id` & `username` must be specified.
2. please add a period at the end of your sentences.
3. We use the extension of pep8 allowing 99 chars per line, so you can have the lines in matching size.

telegram/ext/filters.py
```diff
@@ -324,6 +324,65 @@ def filter(self, message):
...
+        Raises:
+            ValueError
+        """
+
+        def __init__(self, user_id=None, username=None):
+            if (user_id is None and username is None) or (bool(user_id) and bool(username)):
```
tsnoam: Too complicated. You can do instead:
```
if not (bool(user_id) ^ bool(username)):
```

telegram/ext/filters.py
```diff
@@ -324,6 +324,65 @@ def filter(self, message):
...
+                raise ValueError('You must specify either user_id or username')
```
tsnoam: I prefer:
```
raise ValueError('One and only one of user_id or username must be used')
```

telegram/ext/filters.py
```diff
@@ -324,6 +324,65 @@ def filter(self, message):
...
+
+        def filter(self, message):
+            if self.user_id is not None:
+                return bool(message.from_user and message.from_user.id == self.user_id)
+            else:
+                # self.username is not None
+                return bool(message.from_user and message.from_user.username and
+                            message.from_user.username == self.username)
+
+    class chat(BaseFilter):
```
tsnoam: Same comment about CamelCase

telegram/ext/filters.py
```diff
@@ -324,6 +324,65 @@ def filter(self, message):
...
+        """Filters messages to allow only those which are from specified chat ID
+
+        Args:
+            chat_id(Optional[int]): which chat ID to allow through. Required if username is not
+                specified
+            username(Optional[str]): which username to allow through. Required if chat_id is not
+                specified
```
tsnoam: Same comments about documentation.

telegram/ext/filters.py
```diff
@@ -324,6 +324,65 @@ def filter(self, message):
...
+
+        Raises:
+            ValueError
+        """
+
+        def __init__(self, chat_id=None, username=None):
+            if (chat_id is None and username is None) or (bool(chat_id) and bool(username)):
+                raise ValueError('You must specify either chat_id or username')
```
tsnoam: same comments about string to give to ValueError and how to make the `if` more simple.

tests/test_filters.py
```diff
@@ -213,6 +213,45 @@ def test_group_fileter(self):
         self.message.chat.type = "supergroup"
         self.assertTrue(Filters.group(self.message))
 
+    def test_filters_chat(self):
+        with self.assertRaisesRegexp(ValueError, 'either chat_id or username'):
+            Filters.chat(chat_id=0, username='chat')
+            Filters.chat()
```
tsnoam: This line will never be tested. I needs to have its own `with` statement.

bomjacob:
Could you make it so it accepts either what it does now, or a list of said types? Something akin to https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/filters.py#L353-L356 could probably work (and then use in instead of ==)

telegram/ext/filters.py
evgfilim1: I know about it, but in the same file we have classes, which names are not CamelCase (for example, `language` class below)

bomjacob: Yeah, we ignore pep8 in this  instance @tsnoam

telegram/ext/filters.py
evgfilim1: Thanks, I didn't know about this

evgfilim1:
@tsnoam @bomjacob Done

telegram/ext/filters.py
tsnoam: Sorry for the delay,  just saw your replies here.

Uhm. I'm not sure.
In contrary to other fillers we provide, this filter still needs to be instantiated and cannot be used otherwise.

tsnoam: after discussing about it with @bomjacob I remove my objection here.
I don't think it is "pretty" but it is reasonable and makes a clear API.

tsnoam:
@evgfilim1 thank you for your contribution.

PR #679, opened 2017-06-20
Allow specifying pass_* for handlers in callback annotations

bomjacob:
I often find it annoying to have to declare a ton of pass_* statements in handlers. This makes it so:
``` python
def x(bot, update, args):
    pass

dp.add_handler(CommandHandler('test', x, pass_args=True))
```
and
``` python
def x(bot, update, args: Args):
    pass

dp.add_handler(CommandHandler('test', x))
```
are equal.

_______________

Following annotations work

- Args
- Groups (regex)
- Groupdict (regex)
- JobQueue
- UpdateQueue
- ChatData
- UserData

I've changed code such that these types are actually passed to the callback methods, so typecheckers that check annotations are happy.

____________

One more this this allows you to you is have something like:
``` python
def x(b, u, g: Groups, gd: Groupdict):
    pass

dp.add_handler(RegexHandler('(test)(?P<number>\d+)', x))
```
Where the group and groupdict arguments can have other names (something that's always worked wih bot and update).


bomjacob:
Closing due to lack of interest (PR's shouldn't be long standing)

PR #681, opened 2017-06-20
Do so we can have multiple test bots?

bomjacob:
Note: I only made audio work with it so far... not sure this is a good way of doing it by any meaning of the word...


Eldinnie:
I modified you get_bot a bit and maybe can be used to be based this on further? See #683

PR #685, opened 2017-06-22
Cache pip for travis and appveyor

bomjacob:
Updating wheels package since caching was introduced in v7 and I'm not sure what version is default on travis.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,11 +15,13 @@ environment:
 install:
   # We need wheel installed to build wheels
   - "git submodule update --init --recursive"
-  - "%PYTHON%\\python.exe -m pip install wheel"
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -15,12 +15,14 @@ environment:
 install:
   # We need wheel installed to build wheels
   - "git submodule update --init --recursive"
-  - "%PYTHON%\\python.exe -m pip install wheel"
+  - "%PYTHON%\\python.exe -m pip install -U wheel"
```
Eldinnie: Why did you do this? On Appveyor wheel is not installed by default.

bomjacob: At all? Oh :/
I just wanted to be sure we had a recent enough (can't remember the exact version) so that pip knows how to use it for caching.

PR #686, opened 2017-06-22
Add folding to travis script section

bomjacob:
Maybe we should move to an actual script soon? :P


tsnoam:
@bomjacob what about now?

PR #694, opened 2017-06-29
log error trace if there is no error handler registered

jh0ker:


telegram/ext/dispatcher.py
```diff
@@ -365,5 +365,13 @@ def dispatch_error(self, update, error):
...
+        if self.error_handlers:
+            for callback in self.error_handlers:
+                callback(self.bot, update, error)
+
+        else:
+            try:
+                raise error
+
+            except:
+                self.logger.exception('No error handlers are registered, logging exception...')
```
evgfilim1: Maybe it'll be better to use `self.logger.exception('...', exc_info=error)` instead of this try-except clause?

bomjacob:
Could you maybe merge from master so tests are updated (so we can get the build to not fail)

tsnoam:
merged from master

PR #699, opened 2017-06-30
Fix #695 by checking if a user exists

### LINKED ISSUE #637: Message.from is not mandatory due to channels
Currently it's a required attribute in the Message constructor. We should fix that.

CC: @tsnoam

### LINKED ISSUE #695: Bot tries to assign user id to anonymous channel post
### Steps to reproduce
1. Add a bot to a channel that doesn't have post signing.
2. Post something to that channel.
3. Bot tries to assign a user id to the update, bot gets an AttributeError because there is no user.

### Expected behaviour
The bot checks if the user is in the update before trying to get the id.
### Actual behaviour
An AttributeError is thrown and the bot fails to process the update.
### Configuration
**Operating System:**
macOS Sierra

```
python-telegram-bot 6.1.0
urllib3 1.20
certifi 2017.04.17
future 0.16.0
Python 3.6.0 (default, Dec 24 2016, 08:01:42)  [GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)]
```

### Logs
```
  File "/Users/jelle/.venv/lib/python3.6/site-packages/telegram/ext/dispatcher.py", line 264, in process_update
    if handler.check_update(update):
  File "/Users/jelle/.venv/lib/python3.6/site-packages/telegram/ext/conversationhandler.py", line 181, in check_update
    key = self._get_key(update)
  File "/Users/jelle/.venv/lib/python3.6/site-packages/telegram/ext/conversationhandler.py", line 164, in _get_key
    key.append(user.id)
AttributeError: 'NoneType' object has no attribute 'id'
```

pingiun:
Check if a user exists for the update because it can't be used
otherwise in a conversationhandler.

Added a test to detect #695 

Fixes #695 

This PR doesn't fix #637 even though it was requested in 695 because making from_user optional would make it a keyword argument, which would break compatibility.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -32,6 +32,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Jacob Bom <https://github.com/bomjacob>`_
 - `JASON0916 <https://github.com/JASON0916>`_
 - `jeffffc <https://github.com/jeffffc>`_
+- `Jelle Besseling <https://github.com/pingiun>`_
 - `jh0ker <https://github.com/jh0ker>`_
 - `John Yong <https://github.com/whipermr5>`_
 - `jossalgon <https://github.com/jossalgon>`_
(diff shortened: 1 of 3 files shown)


Eldinnie:
Seems good to me. Let's merge

tsnoam:
@pingiun Thankyou for your contribution.

PR #703, opened 2017-07-01
Restructure again

bomjacob:
It's not in any way perfect, but I think it's a step in the right direction at least. This allows cleaner docs and increases maintainability.

Quite possibly backwards incompatible


Eldinnie:
I feel this branch is a bit outdated regarding changes now already in master.

bomjacob:
Yeah you're probably right, I'd still like to have the changes though, but they're not very pressing

PR #705, opened 2017-07-02
MessageQueue: fixed args dispatching in decorator

thodnev:
The devil is in detail, it didn't work because I forgot to add self to args tuple during dispatching.

Finally it's alive and breathing. See the screenshot (reply time of messages). I also add a simple example, which demonstrates MessageQueue usage.

Again, for now its only 90% of work done and MQ is already usable. But the next step is its proper native incorporation into Bot class without all these dirty decorator tricks.
Also seems that we need to dedicate at least one wiki page on how MQ works and how to use it.

The next week I plan to dedicate myself working on it.
times

[test_example.py](https://github.com/python-telegram-bot/python-telegram-bot/files/1118019/test_example.py.txt)

### DIFF
--- a/telegram/ext/messagequeue.py
+++ b/telegram/ext/messagequeue.py
@@ -323,7 +323,7 @@ def queuedmessage(method):
         queued = kwargs.pop('queued', self._is_messages_queued_default)
         isgroup = kwargs.pop('isgroup', False)
         if queued:
-            prom = promise.Promise(method, args, kwargs)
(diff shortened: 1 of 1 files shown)


jh0ker:
> I also add a simple example, which demonstrates MessageQueue usage.

Is it not included in this PR?

thodnev:
@jh0ker Example is in link under the image

PR #727, opened 2017-07-17
bot.py: global timeout argument for all methods (closes #676)

### LINKED ISSUE #676: [Feature] Add global timeout argument for all network methods
I think it will be useful to add one timeout argument for all network methods

For example, I have slow internet connection and I know that timeout for me should be 20 sec. I put `timeout=20` keyword argument in all send_* calls. But I have big code, so it will take me some minutes to do it. But when I have global timeout argument, I can change one line of code and everything will work as I wanted

python-telegram-bot 6.1.0, Python 3.6.1

rahiel:


bomjacob:
Not sure we should be using the longest one if supplied in the method... Imo it makes more sense to have the narrow one (directly on the method) completely overwrite the broad one, even if it's a shorter timeout.

rahiel:
@bomjacob I agree that it makes more sense, but while the majority of methods have a default timeout of `None`, some have a default of `20`. So for those the global timeout would never take effect if we prefer the narrower timeout. I also figure people only use the timeout to make it longer, so that's why I chose for the longer timeout.

bomjacob:
I agree that most people probably just make timeouts longer, but I think think that we should allow the other.
Maybe we can have all the method signatures default to timeout=None and then in _get_timeout have a keyword argument that's like method standard. So you if you don't specify anything it will use max(global timeout, message standard) but if you call it with say timeout=5 then that will always overwrite it?

tsnoam:
Per comments on #676 this PR is not necessary.
We should better document how to set timeouts though (possibly another PR).

PR #735, opened 2017-07-22
Sleep in webhook operations

Eldinnie:
Tests often fail on the setting/deleting of webhooks. I have a gut feeling it's because we immediately remove a webhook after setting it. I added a sleep of 2 seconds.


bomjacob:
That does no seem to be the case according to travis unfortunately.

Eldinnie:
nope, useless. Going to dig deeper :D

PR #737, opened 2017-07-23
TelegramObject.de_json became classmethod closes #734

### LINKED ISSUE #734: TelegramObject.de_json should be @classmethod
All `de_json` methods should be `@classmethod`, not `@staticmethod`. Because almost everywhere inside of them you can see something like this:
```python
    @staticmethod
    def de_json(data, bot):
        ...
        return Update(**data)
```
Instead of:
```python
    @classmethod
    def de_json(cls, data, bot):
        ...
        return cls(**data)
```

1. Did you have any valuable reasons to do so?
2. If not, I can change this and open PR. Do you want me to do this?

ihoru:


Eldinnie:
As you can see in the tests, they fail for python 2.7 (ignore the webhook timeout. And can you explain why this would be an improvement over static methods?

bomjacob:
They are an improvement since they make subclassing of our objects possible without necessarily needing to overwrite the de_json method

ihoru:
Errors fixed.

Take a look at my explanation here, please: #734

Eldinnie:
I was just wondering why it would make a difference. But seeing that tests keep up. I think it's fine.

tsnoam:
@ihoru 

PR #738, opened 2017-07-23
Flow control ability in Dispatcher

ihoru:
There are situations I have very often when I need to control a flow of handlers execution.

### Example for `Continue`:
I need to process a photo in different scopes (contexts): when a user got some state and in all other situations.
Using changes in this PR, I will create two handlers (in the same group):
1. first, will process more particular case (user has a state)
2. second, for general situations (everything else)

First will check if a user has not a state it will raise `Continue`.

### Example for `Stop`:
Some handlers have to be able to stop processing of update without any doubts if there are any other groups or not.

Look at `test_dispatcher_flow.py` if you need extra explanation.


Eldinnie:
This seems very similar to #667 Please take a look at the error's in travis' log.

ihoru:
@Eldinnie, you are right, #667 have similar logic as here but only one from two points.

Errors fixed. (Other problems are not mine :) )

ihoru:
Conflict with master fixed.

tsnoam:
notice the edit on the review comments.

ihoru:
TODO:
- [ ] line length
- [ ] revert touched imports
- [ ] rename Continue -> DispatcherHandlerContinue, Stop -> DispatcherHandlerStop
- [ ] global export for DispatcherHandlerContinue and DispatcherHandlerStop from ext
- [ ] tests over run_async
- [ ] if this does not work with `run_async` fix it
- [ ] Documentation for the feature should be added to `Dispatcher.add_handler()`

telegram/ext/dispatcher.py
```diff
@@ -255,24 +277,29 @@ def process_update(self, update):
         Processes a single update.
 
         Args:
-            update (:obj:`str` | :class:`telegram.Update`): The update to process.
+            update (:obj:`str` | :class:`telegram.Update` | :class:`telegram.TelegramError`):
+            The update to process.
```
tsnoam: Extra 4 spaces before `The update to process` is needed.

tests/test_dispatcher_flow.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python
```
tsnoam: can you please include these tests as part of test_updater.py?

tsnoam:
@ihoru 
as discussed:
1. I left two minor comments above ^^.
2. It doesn't make sense to support `run_async` here, so this is no longer a requirement. What we do need is to modify `Dispatcher.run_async()` so it will make sure that a flow control exception wasn't raised and if it was issue a warning that it was ignored/not-supported with run_async.
3. Documentation should use the new syntax, for example:
```
:class:`telegram.ext.DispatcherHandlerStop`
```

tsnoam:
@ihoru 
btw, I forgot to ask but if you're not on AUTHORS.rst and want to be included you can send a PR for that. (or add it to your other PR)

Issue #731, opened 2017-07-21
Convenient send methods


bomjacob:
Currently our send_* methods can accept either a string or an InputFile (filelike object really). This means that if I wanna for example resend a photo that got send to the bot I can do:
``` python
photo = update.message.photo
bot.send_photo(chat_id, photo.file_id)
```
I'm suggesting we add so it's possible to just do:
``` python
photo = update.message.photo
bot.send_photo(chat_id, photo)
```
So it's possible to send a photo that the bot has received very simply.

If we wanted to take this one step further (which I believe could be pretty cool, we can add a generic send method so something like this would work:
``` python
photo = update.message.photo
bot.send(chat_id, photo)
```
And it would automagically figure out the type of the 2nd argument and send it as such.

PR #742, opened 2017-07-24
send_* now accepts tg-objects

Eldinnie:
This adds the functionality suggested by @bomjacob to pass an object to the relevant send_* method.
For example send_video not accepts either a string, file or telegram.Video object (in which case it will use it;s file_id).

I also added this for send_location, _venue and _contact which means I added an extra parameter to those methods (in the end) and raise a `TypeError` when the required arguments nor the object parameter is given.

added tests for these new feature and adjusted the test_official to ignore the added parameters.

Lastly I noticed the TG-docs have updated the `MaskPosition` by renaming `zoom`  to  `scale`. I also updated this.


Eldinnie:
fixes #731

PR #751, opened 2017-07-26
fix-750

### LINKED ISSUE #750: RegexHandler error with stickers, voices, images...
RegexHandler does not check if update.effective_message.text exists.

### Steps to reproduce
1. Add a RegexHandler
2. Run the bot
3. Send a sticker

### Expected behaviour
The handler should not capture the sticker

### Actual behaviour
The handler capture the sticker and gives an error

### Configuration
Does not matter

``python-telegram-bot 7.0.0``

### Logs
2017-07-26 14:02:47,301 - telegram.ext.dispatcher - ERROR - An uncaught error was raised while processing the update
  File "/usr/local/lib/python3.5/dist-packages/telegram/ext/dispatcher.py", line 269, in process_update
    if handler.check_update(update):
  File "/usr/local/lib/python3.5/dist-packages/telegram/ext/regexhandler.py", line 150, in check_update
    match = re.match(self.pattern, update.effective_message.text)
  File "/usr/lib/python3.5/re.py", line 163, in match
    return _compile(pattern, flags).match(string)
TypeError: expected string or bytes-like object

Eldinnie:
fixes #750

### DIFF
--- a/telegram/ext/regexhandler.py
+++ b/telegram/ext/regexhandler.py
@@ -146,7 +146,7 @@ class RegexHandler(Handler):
         if any([(self.message_updates and update.message),
                 (self.edited_updates and update.edited_message),
                 (self.channel_post_updates and update.channel_post)]) and (
-                isinstance(update, Update)):
+                isinstance(update, Update)) and update.effective_message.text:
             match = re.match(self.pattern, update.effective_message.text)
             return bool(match)
         return False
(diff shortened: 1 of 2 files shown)


telegram/ext/regexhandler.py
```diff
@@ -146,7 +146,7 @@ def check_update(self, update):
         if any([(self.message_updates and update.message),
                 (self.edited_updates and update.edited_message),
                 (self.channel_post_updates and update.channel_post)]) and (
-                isinstance(update, Update)):
+                isinstance(update, Update)) and update.effective_message.text:
```
tsnoam: shouldn't we check that `update.effective_message` is not None?

PR #755, opened 2017-07-29
Decrease number of requests in tests

ihoru:
Use setUpClass instead of __init__ in BaseTest.

In my case speed change was: 324s->240s (25%)

### DIFF
--- a/tests/test_audio.py
+++ b/tests/test_audio.py
@@ -25,7 +25,6 @@ from flaky import flaky
 
 import telegram
 from tests.base import BaseTest, timeout
-from tests.bots import get_bot
 
 
 class AudioTest(BaseTest, unittest.TestCase):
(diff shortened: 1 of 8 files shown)


Eldinnie:
That would make for a descent increase indeed. Maybe you can take a look at `test_video` etc. It uses a new class (`bots.py`) to get a bot. This is to later implement using multiple bot's in one easy place.
I would appreciate it if you could

1. rewrite `BaseTest` to use `get_bot()`
2. remove `get_bot()` from `test_video`, `test_audio`, etc (all the media tests)

ihoru:
@Eldinnie , good notice! Done.

tests/test_audio.py
```diff
@@ -25,25 +25,22 @@
...
 from tests.base import BaseTest, timeout
-from tests.bots import get_bot
 
 
 class AudioTest(BaseTest, unittest.TestCase):
     """This object represents Tests for Telegram Audio."""
 
     @classmethod
     def setUpClass(cls):
+        super().setUpClass()
```
Eldinnie: To support python 2 (which we do) you need to use `super(TestAudio, self).setUpClass(cls)` (see Travis' logs)
Same for all others

ihoru: Oh, yeah... I did this mistake previously... Didn't learn that lesson :) Now I did ;)
Fixed.

Eldinnie:
Looks good to me

Murod1997:
привет

PR #758, opened 2017-07-30
Added myself to Authors list

ihoru:
Due to this #738 

P.S. I didn't do anything with `urllib3/util/connection.py`, just merged it from upstream... I don't understand, why is it so. If you can recommend anything how to solve this problem, I would be very glad!

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -16,8 +16,8 @@ Contributors
 The following wonderful people contributed directly or indirectly to this project:
 
 - `Alateas <https://github.com/alateas>`_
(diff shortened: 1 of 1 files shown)


tsnoam:
@ihoru 
there was a change in the git submodules.
you need to do:
```git submodule update --init --recursive``
but you first have to revert your changes.

ihoru:
@tsnoam thank you for your help!
We need to add a note about this to contribution file...

tsnoam:
@ihoru It's in the README but improvements are always welcome.

ihoru:
@tsnoam done #763

PR #759, opened 2017-07-30
Making requests in webhook answer

### LINKED ISSUE #713: Making requests when getting updates
I (like PTB user) wanna...
...be able to perform a request to the Bot API while sending an answer to the webhook [(read more)](https://core.telegram.org/bots/api#making-requests-when-getting-updates).

It helps to avoid telegram flood errors with popular bots, because Telegram has NO limit to webhook answers!

ihoru:
This is just a draft yet as a proof of concept that fixes #713

Let's discuss if I move in right direction.

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -66,7 +66,16 @@ def message(func):
     @functools.wraps(func)
     def decorator(self, *args, **kwargs):
         url, data = func(self, *args, **kwargs)
-        return self._message_wrapper(url, data, *args, **kwargs)
+        update_response = kwargs.pop('update_response', False)
+        data = self._message_wrapper(data, *args, **kwargs)
+        if update_response:
+            assert isinstance(update_response, Update)
+            if not update_response.get_webhook_response():
+                method = url.split('/')[-1]
+                data['method'] = method
+                update_response.set_webhook_response(data)
+                return True
+        return self._message_request(url, data, *args, **kwargs)
 
     return decorator
 
@@ -158,7 +167,7 @@ class Bot(TelegramObject):
 
         return '@{0}'.format(self.username)
 
-    def _message_wrapper(self, url, data, *args, **kwargs):
+    def _message_wrapper(self, data, *args, **kwargs):
         if kwargs.get('reply_to_message_id'):
             data['reply_to_message_id'] = kwargs.get('reply_to_message_id')
 
@@ -172,6 +181,9 @@ class Bot(TelegramObject):
             else:
                 data['reply_markup'] = reply_markup
 
+        return data
+
+    def _message_request(self, url, data, *args, **kwargs):
         result = self._request.post(url, data, timeout=kwargs.get('timeout'))
 
         if result is True:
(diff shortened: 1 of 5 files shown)


jh0ker:
I was just about to look into this myself when I saw your PR. I like your approach, but I have a few ideas that I'd like to share.

This is how I think it would look like in your current approach:
```python
bot.send_message(chat_id, 'foobar', update_response=update)
```
Not sure I like the naming here, but the pattern seems pretty nice. I also like how it is implemented in `bot.py`.

However, I feel like changing the threading model so drastically for one particular use-case is counter-intuitive. [There has to be a better way!](http://i.imgur.com/dswYieU.gif)

So this is my proposition: Instead of using an extra `NoQueueWebhookHandler`, we modify the existing handler so that it will wait for a response using a `threading.Event` as an attribute in `Update`. This has two advantages:
1. It fits better with the threading architecture and will allow for more multithreading
2. We can release the webhook early (or immediately, if we don't want to use this feature in a particular handler)

**Proposed changes:**

`Updater.start_webhook` has an additional kwarg `use_webhook_responses`. The value is forwarded to the `WebhookHandler`. 

When a new update is received, it sets the attribute `update.response_event = threading.Event()` before putting it into the `update_queue`. If `use_webhook_responses` is `False` (default), behaviour is just like it is now. If it is `True`, it will do `update.response_event.wait()`.

In your modified `Bot.message` decorator, after `update_response.set_webhook_response(data)`, also do `update_response.response_event.set()` or perhaps some convenience method like `update_response.release_webhook()` that can also be used by the user. It could be integrated with `set_webhook_response` as well.

We can now close the webhook connection before the handler is fully executed, instead of waiting until the end. Either by sending a response or releasing it manually. At the end of `Dispatcher.process_update`, it would release the webhook as a fallback. We can also have multiple webhook connections at once, without running into problems. Ultimately, I would like to replace the `BaseHTTPServer` with some WSGI solution anyways.

Let me know what you think!

telegram/bot.py
```diff
@@ -66,7 +66,16 @@ def message(func):
     @functools.wraps(func)
     def decorator(self, *args, **kwargs):
         url, data = func(self, *args, **kwargs)
-        return self._message_wrapper(url, data, *args, **kwargs)
+        update_response = kwargs.pop('update_response', False)
+        data = self._message_wrapper(data, *args, **kwargs)
+        if update_response:
+            assert isinstance(update_response, Update)
+            if not update_response.get_webhook_response():
+                method = url.split('/')[-1]
```
jh0ker: maybe use `os.path.basename` here?

telegram/bot.py
```diff
@@ -66,7 +66,16 @@ def message(func):
...
+                data['method'] = method
+                update_response.set_webhook_response(data)
+                return True
```
jh0ker: Returning `True` here is misleading IMO, better return `None`

jsmnbom:
I'm gonna go ahead and close due to inactivity and also affected code has changed so much that most of it no longer applies. Feel free to submit a new PR against master (or even better the newest V* branch - currently V12) :D

PR #761, opened 2017-07-30
add RESTRICTED as constant to chatmember

Eldinnie:


bomjacob:
Please ammend the docstring too
Currently says 
> The member's status in the chat. Can be 'creator', 'administrator', 'member', 'left' or 'kicked'.
It should include restricted too

Eldinnie:
good catch

PR #774, opened 2017-08-05
bot.py fixes discoverd while testing

Eldinnie:
offset for get_user_profile_photo can be zero.
edit_message_caption should raise ValueError in compliance with other methods. Also changed to snake_case
edit_message_reply_markup was even weirder

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -1156,7 +1156,7 @@ class Bot(TelegramObject):
 
         data = {'user_id': user_id}
 
-        if offset:
+        if offset is not None:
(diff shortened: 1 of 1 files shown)


bomjacob:
Not sure what you mean by "edit_message_reply_markup was even weirder", to me it looks to have been as weird as "edit_message_caption" ?

Eldinnie:
Exactly, it mentions editMessageCaption in the ValueError (instead of edit_message_reply_markup)

bomjacob:
Oh I see what you mean now ^^

PR #776, opened 2017-08-07
Remove Botan

### LINKED ISSUE #760: Remove botan from our library
according to [this](https://github.com/botanio/sdk#py) botan has it's own implementation for python. No need to reinvent the wheel. I suggest we remove it from ptb in the next major (8.0) version.

Eldinnie:
closes #760


tsnoam:
@Eldinnie We'll merge it closer to v8 release.

PR #777, opened 2017-08-07
restore text_*

### LINKED ISSUE #773: message.text_* returns link as hyperlink.
after latest updates the library using "message.text_html" method (i didn't check with markdown) returns links as text_link.

so a message like "`test www.google.com`" is returned by message.text_html as "`test <a href="www.google.com">www.google.com</a>`" despite it was not an text_link.

before of the latest udpates it worked fine and it was able to reflect how the sent message was formatted. so if the message didn't contain html tags it didn't return html tags and viceversa.

actually it's not possible to distinguish if the user sent a link or text_link using that attribute and also if a bot wants to echoes messages respecting the formatting style a possible limit is when an user sends 101 link and the bot will try to echoes with 101 text_link (while the maximum formatting things in a message is 100).

**to reproduce:**
print a link sent to the bot using "update.message.text_html" example: "`t.me/telegram`"
**expcted behaviour:**
you get "`t.me/telegram`"
**actual behaviour:**
you get `<a href="t.me/telegram">t.me/telegram</a>`

Eldinnie:
This restores the text_* properties for message to the way they were.
Implementing different methods for different behavior can be done either in this PR or in another.

fixes #773

### DIFF
--- a/telegram/message.py
+++ b/telegram/message.py
@@ -721,8 +721,6 @@ class Message(TelegramObject):
 
             if entity.type == MessageEntity.TEXT_LINK:
                 insert = '<a href="{}">{}</a>'.format(entity.url, text)
-            elif entity.type == MessageEntity.URL:
-                insert = '<a href="{0}">{0}</a>'.format(text)
             elif entity.type == MessageEntity.BOLD:
                 insert = '<b>' + text + '</b>'
             elif entity.type == MessageEntity.ITALIC:
@@ -772,8 +770,6 @@ class Message(TelegramObject):
 
             if entity.type == MessageEntity.TEXT_LINK:
                 insert = '[{}]({})'.format(text, entity.url)
-            elif entity.type == MessageEntity.URL:
-                insert = '[{0}]({0})'.format(text)
             elif entity.type == MessageEntity.BOLD:
                 insert = '*' + text + '*'
             elif entity.type == MessageEntity.ITALIC:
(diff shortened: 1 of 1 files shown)


Eldinnie:
I modified as requested.
properties now available are:
text_html <- does not encode MessageEntity.URL
text_html_urled < encodes MessageEntity.URL
text_markdown <- does not encode MessageEntity.URL
text_markdown_urled < encodes MessageEntity.URL

bomjacob:
Hmm maybe add a docstring link from the urled one to the not urled one and vise versa?

Eldinnie:
@bomjacob they should be listed right on top of each other in docstrings. I think it's a bit overkill

bomjacob:
That may be true, but I still think it makes more sense from a usability standpoint. Especially if you consider that people don't necessarily search through the docs themselves, they might be linked by roolsbot or be using the internal help() command.

91DarioDev:
@bomjacob 
about this in your comment:
`text_html <- does not encode MessageEntity.URL`

does not encode if the user didn't use html tags, but does it returns the html tags if the user sent the message using them?

Eldinnie:
@91DarioDev in that case it would be a MessageEntity.TEXT_LINK which is always encoded

tsnoam:
LGTM. @Eldinnie has proper unitest wip with the new unitests so no need for that here.

giovanniacanforah:
### 
### 
### 
### 
### 
## 
# 
# 
# 
## 
## 
#

Issue #778, opened 2017-08-07
Remove de_json from object that never use them


Eldinnie:
We have a lot of objects in the library that will only ever be sent, never received. But they still do have a de_json method. I propose to remove all de_json methods from objects that are never sent by tg. Below you'll find a list of the objects involved

- [x] Bot
- [x] ForceReply
- [x] InlineKeyboardButton
- [x] InlineKeyboardMarkup
- [x] InlineQueryResult*
- [x] InlineQueryResultCached*
- [x] InputMessageContent
- [x] Input*MessageContent
- [x] KeyboardButton
- [x] LabeledPrice
- [x] ReplyKeyboardMarkup
- [x] ReplyKeyboardRemove
- [x] ShippingOption

bomjacob:
The only case where this would be bad is if someone is serializing telegram objects as json and then deserializing them afterwards, but I don't believe that should be a usecase we should support since
1) We don't use it in the library internally anywhere
2) Pickle exists

Eldinnie:
@bomjacob it's even worse, de_json doesn't accept json, it works on dicts which is for another issue maybe

PR #757, opened 2017-07-29
Working on pytest

Eldinnie:
no merge or anything, want to see what travis does.

PR #788, opened 2017-08-11
Switch to pytest

bomjacob:
### **This is a joint PR between me and @Eldinnie.**

This completes the move to the [pytest](https://pytest.readthedocs.io/en/latest/) testing framework. Tests are now run using (after installing `requirements-dev.txt`:
``` bash
# -v means verbose
$ pytest -v 
```

With this our coverage once again reaches > 90% and with #778, it should increase even further *(this is due to removing de_json tests from the objects mentioned in that issue, since it seemed stupid to port tests that were just gonna get removed)*

## Other changes
*This PR also includes some other changes related to tests without actually being tests*

- Remove travis.py helper script *(not needed since folding is now a plugin and pre-commit-hooks and build are tested in test_meta.py)*
- Move converage config (`.coveragerc`) to `setup.cfg` for two reasons:
    1) Nicer to have tool configuration conbined in one place
    2) Less files clogging up our root directory (less scrolling for people wanting to read our readme)
- Exclude a couple of lines from coverage *(this was more of a test to see if the `pragma: no cover` comment worked, but those lines should indeed be excluded from coverage)*
- A few of the attributes of `Game `and `Message `now default to `list()` instead of `None `since they are indeed lists no matter what
- Update contributing guide, requirements and travis/appveyor configurations to new tests format
- Small change to `effective_attachment` "Because a few of the possible attachments will evaluate to an empty list if not present (photo for instance). This way it covers those too." - @Eldinnie.

*Opening a new PR instead of piggy bagging on #757, since it's gonna be squashed and @Eldinnie didn't care for the Github credit (on the projects collaborators view and on his profile), the other alternative was a manually squash a lot of stuff which wouldn't just be quite ugly but also time consuming*


tests/test_contact.py
```diff
@@ -5,108 +5,92 @@
...
+        b = Contact(self.phone_number, self.first_name)
+        c = Contact(self.phone_number, '')
+        d = Contact('', self.first_name)
+        e = Voice('', 0)
 
+        assert a == b
+        assert hash(a) == hash(b)
+        assert a is not b
 
-''' Commented out, because it would cause "Too Many Requests (429)" errors.
```
bomjacob: It's marked as xfail if it raises RetryError or if on pypy (where it's strangely more unstable)

pytest_migrate.py
```diff
@@ -0,0 +1,109 @@
...
+        class_vars = class_vars.group(1)
+        class_vars = class_vars.replace('    ', '')
+        class_vars = class_vars.replace('self.', '')
+        class_vars = '\n'.join(['    ' + x for x in class_vars.split('\n')])
+        new_class = re.sub(CLASS_VARS, class_vars, new_class)
+
+    new_class = re.sub(r'self.json_dict', r'json_dict', new_class)
+
+    new_text += new_class
+    new_file = Path('pytests/' + sys.argv[1]).open('w', encoding='UTF-8').write(new_text)
```
bomjacob: Whoops, no we do indeed not

.travis.yml
```diff
@@ -18,18 +18,20 @@ branches:
...
+  - pip install -U -r requirements.txt
+  - pip install -U -r requirements-dev.txt
   - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then pip install ujson; fi
 
 script:
-  - python travis.py
+  - pytest -v --cov=telegram
 
 after_success:
   coveralls
```
bomjacob: We are, but that's not in this PR

.coveragerc
```diff
@@ -1,8 +0,0 @@
-[run]
-source = telegram
-omit = telegram/vendor/*
-
-[report]
-omit =
-    tests/
-    telegram/vendor/*
```
bomjacob: Not quite, it's moved to setup.cfg so it's together with the pytest config

pytest_migrate.py
tsnoam: please remove then?

tests/test_contact.py
tsnoam: 👍

PR #790, opened 2017-08-12
Turn on branch coverage

bomjacob:
Just testing for now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,5 +34,5 @@ script:
   - pytest -v --cov=telegram
 
 after_success:
-  coverage combine
-  coveralls
+  - coverage combine
+  - coveralls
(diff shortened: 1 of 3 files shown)


Eldinnie:
@bomjacob regarding this, It seems to be working now?

bomjacob:
Sometimes yeah... but imo sometimes isn't good enough

Eldinnie:
I let travis build 5 times and it completed all successful, I would think that;s good enough?

bomjacob:
I mean we can try?
The fact of the matter is that it's failed before though and I have no idea if these changes actually changed anything

Eldinnie:
@bomjacob Having to rerun a build once in a while has happened before with other tests. It runs pretty stable now, I think this could be merged

PR #792, opened 2017-08-12
Remove DispatcherHandlerContinue + more unitests for dispatcher

tsnoam:
The idea was nice, but it really complicated things for us and for the
user.
If a user wants to run more than one handler on an update, he can put
the handlers in different groups or he can have a single handler.
If a user wants to have multiple handlers in the same group which only
one of them should run on the update, he should use check_update().

Since we haven't released this code yet, there's no problem with
backward compatability.

### DIFF
--- a/telegram/ext/dispatcher.py
+++ b/telegram/ext/dispatcher.py
@@ -301,7 +301,7 @@ class Dispatcher(object):
         A handler must be an instance of a subclass of :class:`telegram.ext.Handler`. All handlers
         are organized in groups with a numeric value. The default group is 0. All groups will be
         evaluated for handling an update, but only 0 or 1 handler per group will be used. If
-        :class:`telegram.DispatcherHandlerStop` is raised from one of the handler, no further
+        :class:`telegram.DispatcherHandlerStop` is raised from one of the handlers, no further
         handlers (regardless of the group) will be called.
 
         The priority/order of handlers is determined as follows:
--- a/telegram/ext/dispatcher.py
+++ b/telegram/ext/dispatcher.py
@@ -55,25 +55,8 @@ def run_async(func):
     return async_func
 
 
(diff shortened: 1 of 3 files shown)


telegram/ext/dispatcher.py
```diff
@@ -284,63 +268,54 @@ def process_update(self, update):
...
+                    self.logger.exception('An uncaught error was raised while handling the error')
+
+            # Errors should not stop the thread.
+            except Exception:
+                self.logger.exception('An uncaught error was raised while processing the update')
+
     def add_handler(self, handler, group=DEFAULT_GROUP):
-        """
-        Register a handler.
+        """Register a handler.
```
tsnoam: as discussed, i followed pep8

telegram/ext/dispatcher.py
```diff
@@ -284,63 +268,54 @@ def process_update(self, update):
...
 
         The priority/order of handlers is determined as follows:
 
           * Priority of the group (lower group number == higher priority)
-          * The first handler in a group which should handle an update will be
-            used. Other handlers from the group will not be used. The order in
-            which handlers were added to the group defines the priority.
-          * If :class:`telegram.DispatcherHandlerContinue` was raised, then next handler in the
-            same group will be called.
-          * If :class:`telegram.DispatcherHandlerStop` was raised, then zero handlers (even
```
tsnoam: it's kind of duplicate of the paragraph above, don't you think?

Eldinnie: maybe yes, but now it looks weird as a bullet list for order with just one bullet

tsnoam: two bullets. see below

tsnoam: two bullets. the code below this patch has a second bullet...

PR #806, opened 2017-08-23
Bot API 3.3

jeffffc:
Bot API 3.3
https://core.telegram.org/bots/api#august-23-2017

- added `forward_signature` and `author-signature` to `Message`
- added `is_bot` to `User`
- added `pinned_message` to `Chat`

- added `mention()` to `telegram.utils.helpers` for easy generation of inline text mentions with (id/User/Message) input

edit: i have no idea why it cannot import `Message`...

### DIFF
--- a/telegram/chat.py
+++ b/telegram/chat.py
@@ -19,7 +19,7 @@
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains an object that represents a Telegram Chat."""
 
-from telegram import TelegramObject, ChatPhoto
+from telegram import TelegramObject, ChatPhoto, Message
 
 
 class Chat(TelegramObject):
@@ -37,6 +37,8 @@ class Chat(TelegramObject):
         photo (:class:`telegram.ChatPhoto`): Optional. Chat photo.
         description (:obj:`str`): Optional. Description, for supergroups and channel chats.
         invite_link (:obj:`str`): Optional. Chat invite link, for supergroups and channel chats.
+        pinned_message (:class:`telegram.Message`): Optional. Pinned message, for supergroups.
+            Returned only in getChat.
 
     Args:
         id (:obj:`int`): Unique identifier for this chat. This number may be greater than 32 bits
@@ -57,6 +59,8 @@ class Chat(TelegramObject):
             Returned only in get_chat.
         invite_link (:obj:`str`, optional): Chat invite link, for supergroups and channel chats.
             Returned only in get_chat.
+        pinned_message (:class:`telegram.Message`, optional): Pinned message, for supergroups.
+            Returned only in getChat.
         bot (:class:`telegram.Bot`, optional): The Bot to use for instance methods.
         **kwargs (:obj:`dict`): Arbitrary keyword arguments.
 
@@ -83,6 +87,7 @@ class Chat(TelegramObject):
                  photo=None,
                  description=None,
                  invite_link=None,
+                 pinned_message=None,
                  **kwargs):
         # Required
         self.id = int(id)
@@ -96,6 +101,7 @@ class Chat(TelegramObject):
         self.photo = photo
         self.description = description
         self.invite_link = invite_link
+        self.pinned_message = pinned_message
 
         self.bot = bot
         self._id_attrs = (self.id,)
@@ -106,6 +112,7 @@ class Chat(TelegramObject):
             return None
 
         data['photo'] = ChatPhoto.de_json(data.get('photo'), bot)
+        data['pinned_message'] = Message.de_json(data.get('pinned_message'), bot)
 
         return cls(bot=bot, **data)
 
--- a/telegram/chat.py
+++ b/telegram/chat.py
@@ -19,7 +19,7 @@
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains an object that represents a Telegram Chat."""
 
-from telegram import TelegramObject, ChatPhoto, Message
+from telegram import TelegramObject, ChatPhoto
 
 
 class Chat(TelegramObject):
@@ -38,7 +38,7 @@ class Chat(TelegramObject):
         description (:obj:`str`): Optional. Description, for supergroups and channel chats.
         invite_link (:obj:`str`): Optional. Chat invite link, for supergroups and channel chats.
         pinned_message (:class:`telegram.Message`): Optional. Pinned message, for supergroups.
-            Returned only in getChat.
+            Returned only in get_chat.
 
     Args:
         id (:obj:`int`): Unique identifier for this chat. This number may be greater than 32 bits
@@ -60,7 +60,7 @@ class Chat(TelegramObject):
         invite_link (:obj:`str`, optional): Chat invite link, for supergroups and channel chats.
             Returned only in get_chat.
         pinned_message (:class:`telegram.Message`, optional): Pinned message, for supergroups.
-            Returned only in getChat.
+            Returned only in get_chat.
         bot (:class:`telegram.Bot`, optional): The Bot to use for instance methods.
         **kwargs (:obj:`dict`): Arbitrary keyword arguments.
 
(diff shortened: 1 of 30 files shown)


bomjacob:
1) You caused a circular import error please fix in `__init__.py`
2) I'm not a fan of mention tbh. It should be either follow our convention of "_html" and "_markdown" or work like send_message and accept a parsemode. 
3) I also feel like it's weird that it accepts a Message object, what's the basis for that decision?
4) Also why is the name overwriteable?

telegram/chat.py
```diff
@@ -57,6 +59,8 @@ class Chat(TelegramObject):
             Returned only in get_chat.
         invite_link (:obj:`str`, optional): Chat invite link, for supergroups and channel chats.
             Returned only in get_chat.
+        pinned_message (:class:`telegram.Message`, optional): Pinned message, for supergroups.
+            Returned only in getChat.
```
Eldinnie: Please use get_chat

telegram/chat.py
```diff
@@ -37,6 +37,8 @@ class Chat(TelegramObject):
         photo (:class:`telegram.ChatPhoto`): Optional. Chat photo.
         description (:obj:`str`): Optional. Description, for supergroups and channel chats.
         invite_link (:obj:`str`): Optional. Chat invite link, for supergroups and channel chats.
+        pinned_message (:class:`telegram.Message`): Optional. Pinned message, for supergroups.
+            Returned only in getChat.
```
Eldinnie: please use get_chat

telegram/utils/helpers.py
```diff
@@ -72,3 +73,45 @@ def from_timestamp(unixtime):
...
+    """
+    Args:
+        user (:obj:`int` | :class:`telegram.User` | :class:`telegram.Message`)
+            The user's id, or User object, or a Message from the user which you want to mention.
+        html (:obj:`bool`) Default output as markdown, pass True to use HTML.
+        name (:obj:`str`) The name the mention is showing. Required when id is provided in user.
+            Optional if User or Message is provided and will overwrite the name in user.
+
+    Returns:
+        str:
```
Eldinnie: Please define return as :obj :`str`: The inline mention for the user as html or markdown.

telegram/utils/helpers.py
```diff
@@ -72,3 +73,45 @@ def from_timestamp(unixtime):
...
+    """
+    if not user:
+        return None
+
+    if isinstance(user, Message):
```
Eldinnie: I would skip this, also see the total review comment.

telegram/utils/helpers.py
```diff
@@ -72,3 +73,45 @@ def from_timestamp(unixtime):
...
+
+    if isinstance(user, int) and name:
+        if html:
+            return '<a href="tg://user?id={}">{}</a>'.format(user, escape_html(name))
+        else:
+            return "[{}](tg://user?id={}".format(escape_markdown(name), user)
+
+    elif isinstance(user, User):
+        if not name:
+            if user.last_name:
```
Eldinnie: The user object already has a `name`-property, wouldn;t it make more sense to use that?
https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/user.py#L67

tests/test_message.py
```diff
@@ -74,14 +74,16 @@ def message(bot):
                     {'invoice': Invoice('my invoice', 'invoice', 'start', 'EUR', 243)},
                     {'successful_payment': SuccessfulPayment('EUR', 243, 'payload',
                                                              'charge_id', 'provider_id',
-                                                             order_info={})}
+                                                             order_info={})},
+                    {'forward_signature': 'some_forward_sign'},
+                    {'author_signature': 'some_autohr_sign'}
```
Eldinnie: Spell mistake (although no big deal in tests, but still.

Eldinnie:
Hey @jeffffc . Thanks for your work!

I made a few adjustment requests in a code comment, Als o please see my suggestion regarding the mention helper. What do you think about that approach?
Also @bomjacob I like the way you can choose the name for the mention, say you want to rank players and name them `first`, `second` etc. in your message.

Lastly can you fix the import trouble? So we can see what Travis thinks about it.

Pieter

jeffffc:
Let me answer and clarify before pushing updates.
@bomjacob 
For 2, I will divide it into 2 functions then. For 3, yea i just thought that from my instinct, maybe that is not necessary, i will remove it. For 4, it was because for the text mention you do not necessarily use the real name of the user, for example you can make it as `[I want something random](tg://user?id=123456789)`, but if you do not pass the `name` then it uses the user's name by default. I hope i made myself clear on this issue.
@Eldinnie 
Sure It is sensible to make it into user object, i didnt think of that! For the `User.name` object yes I did notice the existence of it but as I wrote above, the text_mention can be showing manual-input strings but not a must to be its name. Moreover, for users who have usernames, `User.name` is default outputting their username, i don't when people using text_mentions on these users would want to show their username directly?

Thank you both for your response and I will surely fix the circular import issue.

Eldinnie:
@jeffffc you make a good point. I was merely saying that the fallback could be the `name` property, but I agree you should be able to set the name. So maybe properties for the `mention_html/markdown` isn't useful.

How about two helper methods within `telegram.User`

```python
def mention_html(name=None) -> str (<a href ...>)
    
def mention_markdown(name=None) -> str ([name](tg://...))
```

jeffffc:
@Eldinnie sure.
Just one more thing atm: the new `pinned_message` is a message object within a user object, while the current message object already imported user object. How should i do to prevent the circular import? i tried to look it up but cant seems to find any good solution..

Eldinnie:
@jeffffc ouch, ehm, not at pc right now... I can look into it later.

jeffffc:
The errors in travis tests should not be mine now.
So the changes now are:

- `User` object
--> `User.mention_markdown(name=None)` and `User.mention_html(name=None)`
--> If name not passed, will use `User.name` by default

- `utils.helpers`
--> `mention_markdown(user_id, name)` and `mention_html(user_id, name)`
--> assuming users will pass custom names

Eldinnie:
@jeffffc Looks good to me, also thanks for fixing all the tests. There's one flake8 error:
`telegram/user.py:28:1: E303 too many blank lines (3)`
Can you fix that last one? Then imo it's good to merge @tsnoam @jh0ker

jeffffc:
you're welcome!
sure, i missed it my bad, just pushed a new commit.

jh0ker:
Looks good, thank you for your work :)

Eldinnie:
The tests failing are failing due to sticker trouble. #807 should fix those.

telegram/user.py
```diff
@@ -105,3 +111,29 @@ def de_list(cls, data, bot):
...
+    def mention_html(self, name=None):
+        """
+        Args:
+            name (:obj:`str`): If provided, will overwrite the user's name.
+
+        Returns:
+            :obj:`str`: The inline mention for the user as HTML.
+        """
+        if not name:
+            return util_mention_html(self.name, self.id)
```
evgfilim1: This line must be `return util_mention_html(self.id, self.name)`
Also fix line 139

evgfilim1: You've mixed up arguments. First is `user_id`, second is `name`

telegram/user.py
```diff
@@ -105,3 +111,29 @@ def de_list(cls, data, bot):
...
+        else:
+            return util_mention_html(name, self.id)
```
evgfilim1: See comment above

telegram/user.py
jeffffc: you are absolutely right. I mixed up with the old versions. thank you for spotting it out. pushing new commit now.

Eldinnie: oops, good spot.

Eldinnie:
Changed the base to 7.1-branch

evgfilim1:
🎉

jeffffc:
just wondering why travis keep saying build is in progress even all tests passed

Eldinnie:
@jeffffc I stopped travis to fix an error localy. Will push in a sec..

jeffffc:
i see, thanks

alexpacini:
I suggest to edit the readme to reflect that API v3.3 is supported

Eldinnie:
@alexpacini Good point, although I hope we can skip that and just jump to 3.4 in a few days.

PR #807, opened 2017-08-23
Fix sticker tests

Eldinnie:
Added flaky, timeout and xfails to the stickerset bot method tests.


tsnoam:
@Eldinnie Did we contact telegram support about this? Especially on the delete

Eldinnie:
Only tests failing are the test_official (which should be solved when #806 ) is merged

bomjacob:
Looks good to me :)
My only question is if we should really be raising an exception when the set is too large? Won't this fail the `test_bot_methods_3` too? (and stickers will never be deleted)

Eldinnie:
@bomjacob It will fail all sticker set tests. But it's in indicator we need to take action (I can add a comment in the code)

PR #809, opened 2017-08-25
Add `chat.name` property

evgfilim1:


Eldinnie:
Hey,
Looks nice. I have two questions about this.
1. If a chat has a title and a username. I think it would be preferred to return the title, not the username. Also atm you have no tests for the eventuality were title is returned as the name.
2. I'm not really sure, but is it possible for a group/supergroup to have neither a title nor a username? If so, should there be a fallback?

evgfilim1:
@Eldinnie 
Thanks.
1) Maybe, you are right. @jh0ker @tsnoam what do you think?
2) No, it's not possible for group to have no title. ([reference](https://core.telegram.org/bots/api#chat))

jh0ker:
We have discussed this in the dev group a few times, and we have reached the consensus that there is no way to implement this property in a way that it makes sense. Unlike the `User.name` property, there is not really a single obvious way to represent a chat, because there are simply too many different things it can be.
That is why we have decided to reject this PR.

Issue #591, opened 2017-05-06
Run tests on Windows


rahiel:
- [x] Enable/setup [AppVeyor Continuous Integration](https://www.appveyor.com/).
- [x] Add AppVeyor badge to Readme

There's a problem if multiple tests are run concurrently, because they'll fail if they try to use the same webhook/getupdates thing at the same time. So concurrent number of builds should be 1 (or low), and a delay should be used so the Appveyor tests run after Travis is done.

Eldinnie:
Usefull info:
https://packaging.python.org/appveyor/

PR #823, opened 2017-09-10
Enable appveyor

Eldinnie:
I made a new bot in my account and hid the relevant information in appveyors environment vars.
I modified appveyor.yaml to use the new tests and codecov.

If all is well this should mean I've enabled appveyor

### DIFF
--- a/tests/test_bot.py
+++ b/tests/test_bot.py
@@ -16,6 +16,7 @@
 #
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
+import os
 import time
 from datetime import datetime, timedelta
 from platform import python_implementation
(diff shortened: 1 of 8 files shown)


tests/test_meta.py
```diff
@@ -30,7 +30,7 @@ def call_pre_commit_hook(hook_id):
 
 @pytest.mark.nocoverage
 @pytest.mark.parametrize('hook_id', argvalues=('yapf', 'flake8', 'pylint'))
-@pytest.mark.skipif(not os.getenv('TRAVIS'), reason='Not running in travis.')
+@pytest.mark.skipif(not (os.getenv('TRAVIS') or os.getenv('APPVEYOR')), reason='Not running in travis.')
```
bomjacob: Could you edit the reason too?

tests/test_bot.py
```diff
@@ -364,6 +368,7 @@ def test_get_chat_member(self, bot, channel_id):
         assert chat_member.status == 'administrator'
         assert chat_member.user.username == 'EchteEldin'
 
+    @pytest.mark.skipif(os.getenv('APPVEYOR'), reason='Skip game tests on Appveyor')
```
bomjacob: Maybe we should write why we're skipping and not just that we are in the reason?
(I know we haven't really done this before but it makes more sense to do it like that I think)
(applies elsewhere)

bomjacob:
Mostly looks good! Good call on just doing it instead of waiting for us to set up more bots.
I left a few notes in my review :)

Eldinnie:
So, As far as I'm concerned this is the first step towards adding Appveyor to our CI cycle. For now I'm skipping tests for telegram-communication regarding stickers and games because they're not added on the bot (yet).
I'm also skipping tests for jobqueue and messagequeue because they rely on timing to a measure more precise then Appveyor can provide.

Coverage increased because now the branches that split for windows and posix systems now both are executed.

I addressed @bomjacob 's review comments.

Eldinnie:
closes #591

PR #826, opened 2017-09-11
Added utils.helpers.effective_message_type

JosXa:
I've added a TODO tag in the helpers file where cyclic imports would happen if I didn't put the import statements inside of the method. This should be fixed before merging.


Eldinnie:
I rebased on master. Two things I would like to see on this.
Tests for all occurencies (no type, and an update passed instead of a message)
And please revert the param line in `test_message`

Eldinnie:
@JosXa what is the status for this?

JosXa:
@Eldinnie I don't have much time atm. Hope to get to all my open issues end of the week

Eldinnie:
@JosXa 👍

JosXa:
Made changes as requested. 

Question remains: Can/should we do something about the cyclic import (see `helpers.py`, "TODO")?

JosXa:
Is codeclimate's "Function effective_message_type has a Cognitive Complexity of 6 (exceeds 5 allowed). Consider refactoring" too bad? Occurs to me that the method should __not__ be split up..

JosXa:
Integration tests fail because of flood timeouts etc. PR lgtm.

tsnoam:
@Eldinnie @JosXa 
lgtm
i added two small commits.
i think we can merge after CI

Eldinnie:
Coverage is not 100% because this case is never tested:
https://github.com/python-telegram-bot/python-telegram-bot/pull/826/files#diff-f070280279d9dc7a6c79eb51d4ab9156R140
Not a very big problem to me. 
Good to merge

PR #830, opened 2017-09-15
Docstring Fix for ext/filters.py

jeffffc:


Eldinnie:
Thanks for cuntributing again!

PR #832, opened 2017-09-15
edited_updates also for channel_posts

Eldinnie:
including tests

### DIFF
--- a/telegram/ext/messagehandler.py
+++ b/telegram/ext/messagehandler.py
@@ -20,8 +20,8 @@
 """This module contains the MessageHandler class."""
 import warnings
 
-from .handler import Handler
 from telegram import Update
+from .handler import Handler
 
 
 class MessageHandler(Handler):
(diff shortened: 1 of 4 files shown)


bomjacob:
To me it would be unintuitive to suddenly receive `edited_channel_posts` in my `MessageHandler(..., allow_edited=False)` without having `channel_post_updates=True` too, but that might just be me?

Eldinnie:
@bomjacob hmm, that's also True. What do you suggest? Although allow_edited is being deprecated. Do you feel the same for edited_updates?

bomjacob:
Yeah sorry that's what I meant. But looking at it now, that might turn out to me unintuitive behaviour since currently `edited_updates` doesn't require `message_updates` to be true for a `edited_message` to be allowed through. So I don't really know... This is just yet another reason for considering refractoring the filter system to handle these things.

Eldinnie:
@bomjacob that is definitely True. But for now I see no way to fix it (without breaking backwards compatibility) other then introducing edited_channel_post_updates as a marker.

bomjacob:
Hmm, that would be okay if `edited_updates` was called `edit_message_updates` but it's not... So I think for now that the change you've made already in the PR might be the best we can do while maintaining backwards compat and not confusing users too much.

PR #834, opened 2017-09-16
remove unnecessary to_dict()

Eldinnie:
We have some objects that have exactly the same to_dict() method, only specifying that `from_user` should be `from` in the `data`-dict. I refractored this logic to `TelegramObject` and removed those to_dicts() from the code.

### DIFF
--- a/telegram/callbackquery.py
+++ b/telegram/callbackquery.py
@@ -173,8 +173,7 @@ class CallbackQuery(TelegramObject):
 
         or::
 
-            bot.edit_message_reply_markup(
-            inline_message_id=update.callback_query.inline_message_id,
+            bot.edit_message_reply_markup(inline_message_id=update.callback_query.inline_message_id,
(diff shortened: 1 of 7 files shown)


telegram/callbackquery.py
```diff
@@ -180,7 +173,8 @@ def edit_message_reply_markup(self, *args, **kwargs):
 
         or::
 
-            bot.edit_message_reply_markup(inline_message_id=update.callback_query.inline_message_id,
+            bot.edit_message_reply_markup(
+            inline_message_id=update.callback_query.inline_message_id,
```
bomjacob: Please fix the indentation here :)

Eldinnie:
@bomjacob I fixed the indentation as you requested.
The reason I think `from/from_user` is an exception, is because it's the only one we rename in the library. And because it's the only one occurring more then once where it's the only reason to_dict exists.

PR #843, opened 2017-09-25
Add Missing Job Class to docs

jeffffc:
`Job` class was missing from the docs 
preview here: http://ptb.readthedocs.io/en/docs-fix/


jeffffc:
updated at  #845, closing

PR #846, opened 2017-09-25
Stabilize Coverage

Eldinnie:
Coverage reports can flutuate because in one build for the project error's are fired while in another build they're not. This makes for weird minimal increase/decreases.

I tried to stabelize this by adding a test for all exceptions we made and ignoring some other lines for coverage.

### DIFF
--- a/telegram/ext/updater.py
+++ b/telegram/ext/updater.py
@@ -260,7 +260,7 @@ class Updater(object):
                 return self.update_queue
 
     def _start_polling(self, poll_interval, timeout, read_latency, bootstrap_retries, clean,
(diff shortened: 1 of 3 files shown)


bomjacob:
Is it possible to run it a few times to test or will we just have to wait and see?

Eldinnie:
@bomjacob What it does is: run tests for all exceptions and remove coverage evaluation for code that is reliant on an error or other exception that do not always occur. Problem now is that project coverage fluctuates with every build for master, depending on if this coffee runs or not. I hope to eliminate that.

PR #854, opened 2017-10-02
Add extract_urls-helper

JosXa:
Added a new helper that extracts all URLs contained in a message and returned from Telegram


91DarioDev:
What about to include also captions in the helper using regex?

JosXa:
@91DarioDev Good point.

JosXa:
@91DarioDev 
- [x] Extracting URLs from Captions

Eldinnie:
Fails on all except py3.6 due to `urllib.parse`

Eldinnie:
@JosXa Can you please make sure the PR passes pre-commit and tests?

Eldinnie:
Note from @jh0ker Very nice, although perhaps it would be good to mention in the docstring that you can pass either location or lat/lon to edit_message_live_location

JosXa:
Unrelated tests fail.

JosXa:
`test_helpers` succeeds in all test cases. However, unrelated tests fails. So lgtm, but @reviewer please have another look at the tests.

Eldinnie:
@JosXa Can you remind me why we would need/want this? And anyway I think it's more logical to sort in order of appearance rather then alphabetical? Also to parse them from captions please use the parse_caption_entites

telegram/utils/helpers.py
```diff
@@ -113,3 +120,45 @@ def mention_markdown(user_id, name):
...
         return '[{}](tg://user?id={})'.format(escape_markdown(name), user_id)
+
+
+def _extract_urls_from_text(text):
+    """
+    Returns a list of urls from a text string.
+    URLs without a leading `http://` or `www.` won't be found.
+    """
+    out = []
+    for word in text.split(' '):
```
tsnoam: Is there a specific reason to split on explicit spaces instead of any white space?
IMHO, calling `text.split():` would make more sense. Do you agree?

telegram/utils/helpers.py
```diff
@@ -113,3 +120,45 @@ def mention_markdown(user_id, name):
     """
     if isinstance(user_id, int):
         return '[{}](tg://user?id={})'.format(escape_markdown(name), user_id)
+
+
+def _extract_urls_from_text(text):
```
tsnoam: Please make this function a generator (i.e. use `yield` instead of returning a list)

telegram/utils/helpers.py
tsnoam: Please consider doing something like this (shorter and more efficient):
```python
for word in (x for x in text.split() if urlparse(x).scheme):
    yield word
```

telegram/utils/helpers.py
```diff
@@ -113,3 +120,45 @@ def mention_markdown(user_id, name):
...
+
+    Returns:
+        :obj:`list`: A list of URLs contained in the message
+    """
+    from telegram import MessageEntity
+
+    types = [MessageEntity.URL, MessageEntity.TEXT_LINK]
+    results = message.parse_entities(types=types)
+    results.update(message.parse_caption_entities(types=types))
+    all_urls = [v if k.type == MessageEntity.URL else k.url for k, v in results.items()]
```
tsnoam: Please use a generator expression instead of a list comprehension.
i.e.
`all_urls = (v if k.type ...)`

telegram/utils/helpers.py
```diff
@@ -113,3 +120,45 @@ def mention_markdown(user_id, name):
...
+
+    # Strip trailing slash from URL so we can compare them for equality
+    stripped_urls = [x[:-1] if x[-1] == '/' else x for x in all_urls]
```
tsnoam: Use a generator expression here as well.

tsnoam: I think that `x.rstrip('/')` would make more sense in this context and be more readable:
1. It will make sure to strip all trailing slashes (regardless if there's 0, 1, or more).
2. It is much more readable than `x[:-1] if x[-1] == '/' else x`.

What do you think?

telegram/utils/helpers.py
```diff
@@ -113,3 +120,45 @@ def mention_markdown(user_id, name):
...
+        thing = urlparse(word.strip())
+        if thing.scheme:
+            out.append(word)
+    return out
+
+
+def extract_urls(message):
+    """
+    Extracts all Hyperlinks that are contained in a message. This includes
+    message entities and the media caption. Distinct links are returned in order of appearance.
```
tsnoam: Please clarify that message links will be always be specified before caption links.

tests/test_helpers.py
```diff
@@ -26,3 +28,47 @@ def test_escape_markdown(self):
...
+    def test_extract_urls_caption(self):
+        caption = "Taken from https://stackoverflow.com/questions/520031/whats" \
+                  "-the-cleanest-way-to-extract-urls-from-a-string-using-python"
+        test_message = Message(message_id=1,
+                               from_user=None,
+                               date=None,
+                               chat=None,
+                               caption=caption)
+        result = helpers.extract_urls(test_message)
+
```
tsnoam: Please add
```python
assert len(results)) == 1
```

examples/echobot2.py
```diff
@@ -17,9 +17,10 @@
 bot.
 """
 
-from telegram.ext import Updater, CommandHandler, MessageHandler, Filters
 import logging
 
+from telegram.ext import Updater, CommandHandler, MessageHandler, Filters
+
```
tsnoam: Please undo this.
It's good, but I prefer to focus on the main subject of the PR. So you can send another PR of style cleanups.
(I'd be more receptive to this if it was a file that you edit anyhow)

telegram/utils/helpers.py
tsnoam: actually, this is a dead code. maybe just remove this function?

telegram/utils/helpers.py
JosXa: Is this because it saves memory by not building up the whole list but only one element at a time?

telegram/utils/helpers.py
JosXa: Yep

tsnoam:
I see that I missed the new updates on this PR. It deserves a new review. 
I will attempt to make time for it in the near future.

JosXa:
Thank you

telegram/utils/helpers.py
```diff
@@ -17,12 +17,16 @@
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains helper functions."""
-from html import escape
```
Eldinnie: from html import escape should be fine. Why did you change it?

telegram/utils/helpers.py
```diff
@@ -94,7 +98,7 @@ def mention_html(user_id, name):
         :obj:`str`: The inline mention for the user as html.
     """
     if isinstance(user_id, int):
-        return '<a href="tg://user?id={}">{}</a>'.format(user_id, escape(name))
+        return '<a href="tg://user?id={}">{}</a>'.format(user_id, escape_html(name))
```
Eldinnie: and this?

tests/test_helpers.py
```diff
@@ -31,6 +31,48 @@ def test_escape_markdown(self):
...
+        test_text = 'Github can be found at http://github.com. Google is here.'
+        test_message = Message(message_id=1,
+                               from_user=None,
+                               date=None,
+                               chat=None,
+                               text=test_text,
+                               entities=[MessageEntity(**e) for e in test_entities])
+        result = helpers.extract_urls(test_message)
+
+        assert len(result) == 2
```
Eldinnie: I don;t get this. If there's three entities of type `url` and `text_link` in the message. Shouldn't `len(result) == 3`?

tests/test_helpers.py
```diff
@@ -31,6 +31,48 @@ def test_escape_markdown(self):
...
+        assert (test_entities[0]['url'][:-1] == result[0])
+        assert (test_entities[2]['url'] == result[1])
+
```
Eldinnie: In other words. Where did the url go?

tests/test_helpers.py
```diff
@@ -31,6 +31,48 @@ def test_escape_markdown(self):
...
+    def test_extract_urls_caption(self):
+        test_entities = [{
```
Eldinnie: Please test caption also with text_link and url entities

tests/test_helpers.py
nmlorg: extract_urls() strips trailing slashes and then dedupes, so the entity link `http://github.com/` and the inline link `http://github.com` are treated as the same.

(For what it's worth, in this case that's correct, but in cases where trailing slashes are anywhere else in the URL it's either unnecessary (`https://github.com/python-telegram-bot` versus `https://github.com/python-telegram-bot/`), meddlesome (any index page that uses relative links, and therefore must issue a redirect to restore the trailing slash), or actually broken (`http://www.example.com/?b64data=aGk/`).)

telegram/utils/helpers.py
Eldinnie: that actually makes no sense, because a message has either text or a caption, never both

Bibo-Joshi:
Brought this PR up to date and made the requested changes.

Since by @nmlorg s comment stripping the trailing slash is not always a good idea, I chaned the behavior: Exact duplicates will be removed. If afterwards two links differ only in a trailing slash, the slashed one will be kept:

`http://github.com/, http://google.com, http://github.com` gives `[http://github.com/, http://google.com]`.

Bibo-Joshi:
Well, that made it worse. My finding was, that `OrderedDict({k: None for k in urls})` was the problem: The order doesn't seem to be consistent (If you run [this example](https://rextester.com/MDEQG86718) a few times you will see, that the order is not always the same).
But apparently changing to

```
url = OrderedDict()
for k in urls:
        url[k] = None
```

doesn't solve it (although it seems to run fine [here](https://rextester.com/RTXD72330)).

I'll try and go for [this](https://stackoverflow.com/a/480227/10606962) solution that may be a bit slower but hopefully is more relyable

Bibo-Joshi:
Turns out the issue was hidden in the call of `message.parse_entities`. This builds a dict which is *not* sorted in python <=3.5. So now after parsing, the entities are just sorted.
One thing I want to mention: The method now can only handle *single* trailing slashes, i.e. `[github.com/, google.com, google.com/ github.com//]` will produce `[github.com/, google.com/ github.com//]`. However I double trailing slashes are very rare (if even existent?), so that should be ok IMO …

jh0ker:
@Bibo-Joshi I&#x27;m not quite sure why you would want to filter this at all. The difference between having a trailing slash or not in a URL is not necessarily insignificant. 

Unrelated, the comment in line 185 says &quot;dublicates&quot; instead of &quot;duplicates&quot;.

Bibo-Joshi:
@jh0ker I guess that 9-9-19 me interpreted @nmlorg s comment such that in doubt it&#x27;s better to have a trailing slash than not. But neither 9-9-19 me nor now-me have much expertise on that area (or any area at all 😅) …

JosXa:
@Spechide brought up an interesting point. He says that he's using something like this in his bot:

image

Maybe we could also provide some abstractions over this 🤔

telegram/utils/helpers.py
```diff
@@ -251,6 +250,41 @@ def effective_message_type(entity):
...
+        :obj:`list`: A list of URLs contained in the message
+    """
+    from telegram import MessageEntity
+
+    types = [MessageEntity.URL, MessageEntity.TEXT_LINK]
+    results = message.parse_entities(types=types)
+    results.update(message.parse_caption_entities(types=types))
+
+    # Sort results by order of appearence, i.e. the MessageEntity offset
+    sorted_results = sorted(results.items(), key=lambda e: e[0].offset)
```
tsnoam: First remove duplicates, then sort.
It will be more efficient and we wouldn't need OrderedDict

Bibo-Joshi:
Test fail is unrelated

tsnoam:
After long considerations and real intent to see this PR merged, we, the maintainers, decided that this PR is too specific to be included in a general purpose library.
Therefor we've decided to close the PR without merging.

Bibo-Joshi:
For future reference: this will be made available through the new [`ptbcontrib`](https://github.com/python-telegram-bot/ptbcontrib/) library

PR #856, opened 2017-10-02
README: add AppVeyor badge (closes #591)

### LINKED ISSUE #591: Run tests on Windows

rahiel:
More badges!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,6 +31,10 @@ We have made you a wrapper you can't refuse
    :target: https://travis-ci.org/python-telegram-bot/python-telegram-bot
    :alt: Travis CI Status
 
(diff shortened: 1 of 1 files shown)


Eldinnie:
Thanks!

Eldinnie:
Merging this, because why not? Badges are great.

Issue #864, opened 2017-10-11
October 11 update


bomjacob:
Moar updates!

- [x] Live locations
  - [x] add live_period to send_location 
  - [x] add edit_message_live_location
  - [x] add stop_message_live_location 
  - [x] add live_period to InlineQueryResultLocation and InputLocationMessageContent
  - [x] Tests
- [x] Group sticker management
  - [x] add get_chat_sticker_set
  - [x] add delete_chat_sticker_set
  - [x] add Chat.sticker_set_name
  - [x] add Chat.can_set_sticker_set
- [x] add caption_entities in Message
[ref](https://core.telegram.org/bots/api-changelog#october-11-2017)

PR #865, opened 2017-10-11
support 3.4 API

Eldinnie:
WIP
Live location stuff added See #864

### DIFF
--- a/telegram/chat.py
+++ b/telegram/chat.py
@@ -38,6 +38,9 @@ class Chat(TelegramObject):
         invite_link (:obj:`str`): Optional. Chat invite link, for supergroups and channel chats.
         pinned_message (:class:`telegram.Message`): Optional. Pinned message, for supergroups.
             Returned only in get_chat.
+        sticker_set_name (:obj:`str`): Optional. For supergroups, name of Group sticker set.
+        can_set_sticker_set	(:obj:`bool`): Optional. ``True``, if the bot can change group the
+            sticker set.
 
     Args:
         id (:obj:`int`): Unique identifier for this chat. This number may be greater than 32 bits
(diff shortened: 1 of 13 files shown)


Eldinnie:
Ok I think I got everything in. Something weird going on with test_official. I don't really get the error. I'm hoping @bomjacob can find time to figure it out.

Also we need to let the bot set a stickerset on our group every test (going to be problematic I'm affraid) to properly test the `set_chat_sticker_set` and `delete_chat_sticker_set` methods.

Review wanted.

Eldinnie:
I found the problem with test_official.
It seems TG has their table messed up on:
https://core.telegram.org/bots/api#inputlocationmessagecontent
and
https://core.telegram.org/bots/api#inlinequeryresultlocation

Eldinnie:
Note from @jh0ker Very nice, although perhaps it would be good to mention in the docstring that you can pass either location or lat/lon to edit_message_live_location

Eldinnie:
Need to up tests aswell. Hopefully have time tomorrow

Eldinnie:
@tsnoam added the requested check (and tests). If we can live with a small decrease in coverage due to not testing the groupstickerset stuff I think it's ready to merge/release

Eldinnie:
@tsnoam 
`if not (latitude is not None or longitude is not None) ^ bool(location):`
Ugly, but works.
If location and either lat or lon is not None it will raise.

tsnoam:
@Eldinnie As discussed, please add () around the entire logical expression which start just after if not.

Eldinnie:
@tsnoam Done

telegram/chat.py
```diff
@@ -38,6 +38,9 @@ class Chat(TelegramObject):
         invite_link (:obj:`str`): Optional. Chat invite link, for supergroups and channel chats.
         pinned_message (:class:`telegram.Message`): Optional. Pinned message, for supergroups.
             Returned only in get_chat.
+        sticker_set_name (:obj:`str`): Optional. For supergroups, name of Group sticker set.
+        can_set_sticker_set	(:obj:`bool`): Optional. ``True``, if the bot can change group the
```
bomjacob: I think a tab character snuck in here

telegram/chat.py
```diff
@@ -61,6 +64,10 @@ class Chat(TelegramObject):
         pinned_message (:class:`telegram.Message`, optional): Pinned message, for supergroups.
             Returned only in get_chat.
         bot (:class:`telegram.Bot`, optional): The Bot to use for instance methods.
+        sticker_set_name (:obj:`str`, optional): For supergroups, name of Group sticker set.
+            Returned only in get_chat.
+        can_set_sticker_set	(:obj:`bool`, optional): ``True``, if the bot can change group the
```
bomjacob: Same here

telegram/chat.py
Eldinnie: it did will fix

Issue #875, opened 2017-10-16
Add Filters.gif


JosXa:
Since the question comes up a lot in our Telegram group, I'm adding it as an issue:
The `Filters` should include a `gif`-Filter checking if the update contains a document which has the mime type of gifs.

tsnoam:
@JosXa Sounds useful. I'd say it should be somewhat more flexible - like instatinating fillers for known types and allowing users to instatinate their own custom fillers with a mimetype.

JosXa:
@tsnoam I agree, there's a lot more that could be added. I'd still say that since gifs are an integral part of Telegram and baked into every client, they should get their own thing. Hmm is there a list of mime types that can be sent over Telegram and are ameliorated with rich views?

JosXa:
Implemented with #1046: Filters.document.gif

PR #880, opened 2017-10-18
Adding Filters for mime-types and file_size

code1mountain:
Adding new subfilters to Filters.documents that allow to filter for different mime_type categories or file types. 

Adding new Category Filter for these file categories:
application/
audio/
image/
video/
text/

Adding Filters for these file types:
apk, docx, exe, gif, jpg, mp3, pdf, png, py, svg, txt, targz, wav, xml, zip

Adding Filters.FileSize(min=0, max=sys.maxsize) to filter all types of media by it's file_size.

The Category Filter and the FileType Filter are visible, because it should be possible to create own Filters for other mime_types.

This solves Issue #875

Contact for questions: t.me/bergfreak

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -197,22 +197,22 @@ class Filters(object):
 
         class Category(BaseFilter):
             """This Filter filters documents by their category in the mime-type attribute
-            
+
             Note:
-                This Filter only filters by the mime_type of the document, 
+                This Filter only filters by the mime_type of the document,
                     it doesn't check the validity of the document.
-                The user can manipulate the mime-type of a message and 
+                The user can manipulate the mime-type of a message and
                     send media with wrong types that don't fit to this handler.
-                    
+
             Examples:
-                Filters.documents.Category("audio/") returnes `True` for all types 
+                Filters.documents.Category("audio/") returnes `True` for all types
                 of audio sent as file, for example "audio/mpeg" or "audio/x-wav"
             """
             name = "Filters.document.Category()"
 
             def __init__(self, category):
                 """Initialize the category you want to filter
-                
+
                 Args:
                     category (str, optional): Category of the media you want to filter"""
                 self.category = category
@@ -231,9 +231,9 @@ class Filters(object):
             """This Filter filters documents by their mime-type attribute
 
             Note:
-                This Filter only filters by the mime_type of the document, 
+                This Filter only filters by the mime_type of the document,
                     it doesn't check the validity of document.
-                The user can manipulate the mime-type of a message and 
+                The user can manipulate the mime-type of a message and
                     send media with wrong types that don't fit to this handler.
 
             Examples:
@@ -282,9 +282,9 @@ class Filters(object):
             """Initialize the limits of the file_size in the `__init__` method.
             
             Args:
-                min (int, optional): Minimum `file_size` of the message media in Byte. 
+                min (int, optional): Minimum `file_size` of the message media in Byte.
                     Default is zero.
-                max (int, optional): Maximum `file_size` of the message media in Byte. 
+                max (int, optional): Maximum `file_size` of the message media in Byte.
                     Default is infinity.
             """
             self.min = min


telegram/constants.py
```diff
@@ -52,3 +52,42 @@
...
+    
+       They are used in `Telegram.ext.filters.Filters.document.`"""
+    application = "application/"
+    audio = "audio/"
+    image = "image/"
+    video = "video/"
+    text = "text/"
+
+
+class Document:
```
bomjacob: Is there a good reason you're not using the [stdlib mimetype](https://docs.python.org/3.2/library/mimetypes.html)?

telegram/ext/filters.py
```diff
@@ -192,12 +195,118 @@ def filter(self, message):
...
+
+                Args:
+                    category (str, optional): Category of the media you want to filter"""
+                self.category = category
+
+            def filter(self, message):
+                if message.document:
+                    return bool(message.document.mime_type.startswith(self.category))
+
+        application = Category(constants.MediaCategory.application)
```
bomjacob: Could we perhaps expose these as `filters.document.category.application` instead of just `filters.document.application`?

telegram/ext/filters.py
```diff
@@ -192,12 +195,118 @@ def filter(self, message):
...
+
+                Args:
+                    filetype (str, optional): mime_type of the media you want to filter"""
+                self.filetype = filetype
+
+            def filter(self, message):
+                if message.document:
+                    return bool(message.document.mime_type == self.filetype)
+
+        apk = FileType(constants.Document.apk)
```
bomjacob: Could we perhaps expose these as `filters.document.filetype.apk` instead of just `filters.document.apk`?

telegram/ext/filters.py
```diff
@@ -192,12 +195,118 @@ def filter(self, message):
...
 
     document = _Document()
     """:obj:`Filter`: Messages that contain :class:`telegram.Document`."""
 
+    class FileSize(BaseFilter):
+        """This Filter filters all messages with a `file_size` attribute.
+        """
+        name = "Filters.FileSize()"
+
+        def __init__(self, min=0, max=sys.maxsize):
```
bomjacob: I'm not a huge fan of using `sys.maxsize`. Could you instead have it default to `None` and if it is `None` just don't compare upper limit?

telegram/ext/filters.py
```diff
@@ -192,12 +195,118 @@ def filter(self, message):
...
+                min (int, optional): Minimum `file_size` of the message media in Byte.
+                    Default is zero.
+                max (int, optional): Maximum `file_size` of the message media in Byte.
+                    Default is infinity.
+            """
+            self.min = min
+            self.max = max
+
+        def filter(self, message):
+            if bool(message.audio):
```
bomjacob: There's no need for the `bool()` in these if statements

telegram/ext/filters.py
```diff
@@ -192,12 +195,118 @@ def filter(self, message):
     class _Document(BaseFilter):
         name = 'Filters.document'
 
+        class Category(BaseFilter):
```
bomjacob: Please use the convention of lowercasing filters.

telegram/ext/filters.py
```diff
@@ -192,12 +195,118 @@ def filter(self, message):
...
+                if message.document:
+                    return bool(message.document.mime_type.startswith(self.category))
+
+        application = Category(constants.MediaCategory.application)
+        audio = Category(constants.MediaCategory.audio)
+        image = Category(constants.MediaCategory.image)
+        video = Category(constants.MediaCategory.video)
+        text = Category(constants.MediaCategory.text)
+
+        class FileType(BaseFilter):
```
bomjacob: Lowercase here too

telegram/ext/filters.py
```diff
@@ -192,12 +195,118 @@ def filter(self, message):
...
+                This Filter only filters by the mime_type of the document,
+                    it doesn't check the validity of the document.
+                The user can manipulate the mime-type of a message and
+                    send media with wrong types that don't fit to this handler.
+
+            Examples:
+                Filters.documents.Category("audio/") returnes `True` for all types
+                of audio sent as file, for example "audio/mpeg" or "audio/x-wav"
+            """
+            name = "Filters.document.Category()"
```
bomjacob: Please do something like
`self.name = 'Filters.document.category({})'.format(self.category)` in `__init__` instead

telegram/ext/filters.py
```diff
@@ -192,12 +195,118 @@ def filter(self, message):
...
+            Note:
+                This Filter only filters by the mime_type of the document,
+                    it doesn't check the validity of document.
+                The user can manipulate the mime-type of a message and
+                    send media with wrong types that don't fit to this handler.
+
+            Examples:
+                Filters.documents.FileType("audio/mpeg") filters all audio in mp3 format.
+            """
+            name = "Filters.document.Filetype()"
```
bomjacob: Same comment about name

telegram/ext/filters.py
```diff
@@ -192,12 +195,118 @@ def filter(self, message):
...
+        xml = FileType(constants.Document.xml)
+        zip = FileType(constants.Document.zip)
+
         def filter(self, message):
             return bool(message.document)
 
     document = _Document()
     """:obj:`Filter`: Messages that contain :class:`telegram.Document`."""
 
+    class FileSize(BaseFilter):
```
bomjacob: Lowercasing

telegram/ext/filters.py
```diff
@@ -192,12 +195,118 @@ def filter(self, message):
...
+        """This Filter filters all messages with a `file_size` attribute.
+        """
+        name = "Filters.FileSize()"
```
bomjacob: Same comment about name

telegram/ext/filters.py
```diff
@@ -192,12 +195,118 @@ def filter(self, message):
...
+                filesize = message.photo.file_size
+            elif bool(message.sticker):
+                filesize = message.sticker.file_size
+            elif bool(message.video):
+                filesize = message.video.file_size
+            elif bool(message.voice):
+                filesize = message.voice.file_size
+            else:
+                return False
+            return bool(self.min <= filesize <= self.max)
```
bomjacob: Also no need for that `bool()`

telegram/constants.py
```diff
@@ -52,3 +52,42 @@
 MAX_MESSAGES_PER_MINUTE_PER_GROUP = 20
 MAX_MESSAGE_ENTITIES = 100
 MAX_INLINE_QUERY_RESULTS = 50
+
+
+class MediaCategory:
```
bomjacob: Just put these constants in the filters.py, since that's the only place they are used and they aren't meant to be exported to the general user.

telegram/ext/filters.py
tsnoam: @bomjacob 
is this nesting really necessary? I think that `filters.document.application` is much more easier to use.

telegram/ext/filters.py
code1mountain: I thought I use the PEP 8 Style here, because it's a class to call. It's different to the other filters yet

telegram/constants.py
code1mountain: No, just didn't found them. I will fix this.

telegram/ext/filters.py
code1mountain: Yes, that makes sense. I will fix it.

telegram/ext/filters.py
code1mountain: It's a class you call directly, and classes should be CamelCase according to PEP8

telegram/ext/filters.py
code1mountain: I will fix it

telegram/ext/filters.py
code1mountain: It's much more nesting. Not sure we really need this.

telegram/ext/filters.py
code1mountain: Again PEP8

telegram/ext/filters.py
code1mountain: Ok, I will leave it

telegram/ext/filters.py
code1mountain: I will leave it too. Fixed PR will come tomorrow. Thank you for your feedback.

telegram/constants.py
code1mountain: Not necessary if I use built-in mimetypes

telegram/constants.py
code1mountain: The stdlib mimetype doesn't work for many mimetypes in telegram. ".exe" documents for example have  the mimetype "application/x-ms-dos-executable" if you send it with an official telegram app, but ".exe" is matched to "application/octet-stream" in the stdlib mimetype.

Eldinnie:
@code1mountain please respond

bomjacob:
@code1mountain we're gonna go ahead and close this fairly soon if we receive no reply. It's been pending reply for almost 2 months now. Just give us a heads up if you're still working on it ^^

telegram/ext/filters.py
bomjacob: Yeah, but we ignore PEP8 for filters, since the other filters are simple attributes.

telegram/ext/filters.py
bomjacob: Yeah, no nesting is okay :)

tsnoam:
closing as of no response from the author of the PR

PR #883, opened 2017-10-19
New inline keyboard example with multiple callback query handler.

sahilsharma811:
This is an example of a bot with an inline keyboard that can handle user's response with a dedicated callback query handler. The bot implements a Conversation Handlers to handle the response of the user, which is sent by the inline keyboard.


JosXa:
Good one. People are often confused about combining Inline keyboards with ConvHandler.

PR #884, opened 2017-10-19
New inline keyboard example

sahilsharma811:
This is an example of a bot with an inline keyboard that can handle user's response with a dedicated callback query handler. The bot implements a Conversation Handlers to handle the response of the user, which is sent by the inline keyboard.

### DIFF
--- a/examples/inlinekeyboard2.py
+++ b/examples/inlinekeyboard2.py
@@ -64,6 +64,7 @@ def second(bot, update):
         message_id=query.message.message_id,
         text="Second CallbackQueryHandler"
     )
+    return ConversationHandler.END
 
 def error(bot, update, error):
     logger.warn('Update "%s" caused error "%s"', (update, error))    
--- a/examples/inlinekeyboard2.py
+++ b/examples/inlinekeyboard2.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#Simple inline keyboard bot with multiple callback Query Handler.
+"""
+This Bot uses the Updater class to handle the bot.
+First, a few callback functions are defined as callback query handler. Then, those functions are passed to
+the Dispatcher and registered at their respective places.
+Then, the bot is started and runs until we press Ctrl-C on the command line.
+Usage:
+Example of a bot that uses inline keyboard that has multiple callbacks Query Handler.
+Send /start to initiate the conversation.
+Press Ctrl-C on the command line to stop the bot.
+"""
+from telegram import InlineKeyboardButton, InlineKeyboardMarkup
+from telegram.ext import Updater, CommandHandler, CallbackQueryHandler, ConversationHandler
+import logging
+
+# Enable logging
+logging.basicConfig(format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+                    level=logging.INFO)
+
+logger = logging.getLogger(__name__)
+
+FIRST, SECOND = range(2)
+
+def start(bot, update):
+    user = update.message.from_user
+    logger.info("User %s started the conversation." % user.first_name)
+    keyboard = [
+        [InlineKeyboardButton(u"Next", callback_data=str(FIRST))]
+    ]
+    reply_markup = InlineKeyboardMarkup(keyboard)
+    update.message.reply_text(
+        u"Start handler, Press next",
+        reply_markup=reply_markup
+    )
+    return FIRST
+
+def first(bot, update):
+    query = update.callback_query
+    keyboard = [
+        [InlineKeyboardButton(u"Next", callback_data=str(SECOND))]
+    ]
+    reply_markup = InlineKeyboardMarkup(keyboard)
+    bot.edit_message_text(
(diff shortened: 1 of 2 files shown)


examples/inlinekeyboard2.py
```diff
@@ -0,0 +1,106 @@
...
+        chat_id=query.message.chat_id,
+        message_id=query.message.message_id,
+        text=u"Second CallbackQueryHandler"
+    )
+    return
+
+def error(bot, update, error):
+    logger.warn('Update "%s" caused error "%s"' % (update, error))    
+
+def  main():
```
bomjacob: Extra space

examples/inlinekeyboard2.py
```diff
@@ -0,0 +1,106 @@
...
+        
+    # Create the Updater and pass it your bot's token. Below is an example token
+    updater = Updater("476482598:AAGR26TlclQP1ZdsAbBuGy9AeCkybP8k-kA")
```
bomjacob: Please remove the token ;)

examples/inlinekeyboard2.py
```diff
@@ -0,0 +1,106 @@
...
+    return SECOND
+
+def second(bot, update):
+    query = update.callback_query
+    bot.edit_message_text(
+        chat_id=query.message.chat_id,
+        message_id=query.message.message_id,
+        text=u"Second CallbackQueryHandler"
+    )
+    return
```
bomjacob: Unnecessary return statement

examples/inlinekeyboard2.py
```diff
@@ -0,0 +1,106 @@
...
+
+def start(bot, update):
+    user = update.message.from_user
+    logger.info("User %s started the conversation." % user.first_name)
+    keyboard = [
+        [InlineKeyboardButton(u"Next", callback_data=str(FIRST))]
+    ]
+    reply_markup = InlineKeyboardMarkup(keyboard)
+    update.message.reply_text(
+        u"Start handler, Press next",
```
bomjacob: Do we mark strings in the other examples as unicode?

examples/inlinekeyboard2.py
```diff
@@ -0,0 +1,106 @@
...
+logging.basicConfig(format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+                    level=logging.INFO)
+
+logger = logging.getLogger(__name__)
+
+FIRST, SECOND = range(2)
+
+def start(bot, update):
+    user = update.message.from_user
+    logger.info("User %s started the conversation." % user.first_name)
```
bomjacob: Should look like `logger.info("User %s started the conversation.", user.first_name)` for performence reasons (really just to teach users to do it properly)

examples/inlinekeyboard2.py
sahilsharma811: No! We are not marking strings in other example as unicode, but I think it's a good practice to mark strings as unicode.

examples/inlinekeyboard2.py
sahilsharma811: Removed

examples/inlinekeyboard2.py
sahilsharma811: Removed

examples/inlinekeyboard2.py
sahilsharma811: Removed

examples/inlinekeyboard2.py
sahilsharma811: Done!

examples/inlinekeyboard2.py
sahilsharma811: Should I keep it or make this example symmetrical with other examples?

sahilsharma811:
@bomjacob. I've made all the changes as requested.

vr2xiq:
is it possible to return different routes?
let's say 2 buttons inlinekeyboard
press 1 return third route
press 2 return forth route

sahilsharma811:
It might be possible, but I'm not sure yet.

Eldinnie:
It looks okay to me. But I have some questions/additions I would like to see.
First of all, the conversation is never ended. I would prefer if you return `ConversationHandler.END` at the end of `second`.

Secondly I do agree with @vr2xiq that it would be nice if the example would show a keyboard with a choice. So for example on start you have a choice of two buttons (`1` and `2`)
When selected `1` you will have a new choice of `3` and `4`, and when selected `4` you will have a second choice of `4` and `2` (just an example)
This will have a more elaborate view on `CallbackQqueryHandler`s in a `ConversationHandler` and will have to use the `pattern` argument to discern between them.

You think it's possible for you to adjust the example that way?

sahilsharma811:
@Eldinnie I have added the ConversationHandler.END at the end of second.
For the keyboard which provide two or more choices, I will work on that ASAP. Till then we can merge this example so other developers can use the example to build awesome bots. Thanks

tsnoam:
@bomjacob @Eldinnie 
I see there are commits following your review. Can you please check if we can merge it?

PR #887, opened 2017-10-21
Fix race condition in dispatcher start/stop

### LINKED ISSUE #881: Race condition in Dispatcher.start/stop()
### Steps to reproduce
Run slighly modified example from https://python-telegram-bot.org/
```python
from telegram.ext import Updater, CommandHandler

def start(bot, update):
    update.message.reply_text('Hello World!')

def hello(bot, update):
    update.message.reply_text(
        'Hello {}'.format(update.message.from_user.first_name))

updater = Updater('YOUR TOKEN HERE')

updater.dispatcher.add_handler(CommandHandler('start', start))
updater.dispatcher.add_handler(CommandHandler('hello', hello))

updater.start_polling()
try:
    # this call to non-existing function may cause dispatcher.stop() to be called
    # before dispatcher.start() had time to set self.running=True,
    # thus stop() does nothing and dispatcher continues to run.
    CALL_TO_NON_EXISTING_FUNC()
finally:
    updater.stop()
```

### Expected behaviour
Application should stop

### Actual behaviour
Application continues to run and does not reacting to CTRL+C

### Configuration
**Operating System:**
Linux

``$ python -m telegram``
python-telegram-bot 8.1.1
certifi 2017.07.27.1
future 0.15.2
Python 2.7.13 (default, Jan 19 2017, 14:48:08)  [GCC 6.3.0 20170118]

### Logs
Not necessary

tsnoam:
Fixes #881

### DIFF
--- a/telegram/ext/dispatcher.py
+++ b/telegram/ext/dispatcher.py
@@ -183,10 +183,14 @@ class Dispatcher(object):
         self.__async_queue.put(promise)
         return promise
 
-    def start(self):
+    def start(self, ready=None):
         """Thread target of thread 'dispatcher'.
 
         Runs in background and processes the update queue.
+
+        Args:
+            ready (:obj:`threading.Event`, optional): If specified, the event will be set once the
+                dispatcher is ready.
 
         """
         if self.running:
@@ -201,6 +205,9 @@ class Dispatcher(object):
         self._init_async_threads(uuid4(), self.workers)
         self.running = True
         self.logger.debug('Dispatcher started')
+
+        if ready is not None:
+            ready.set()
 
         while 1:
             try:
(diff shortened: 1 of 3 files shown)


telegram/ext/dispatcher.py
```diff
@@ -183,11 +183,15 @@ def run_async(self, func, *args, **kwargs):
...
         """Thread target of thread 'dispatcher'.
 
         Runs in background and processes the update queue.
 
+        Args:
+            ready (:obj:`threading.Event`, optional): If specified, the event will be set once the
+                dispatcher is ready.
+
         """
         if self.running:
```
jh0ker: Shouldn't we do `ready.set()` here as well?

tsnoam: that makes sense. i'll fix.

tsnoam:
@bomjacob I don't think that we can properly test it.
We can pass an Event object to `dispatcher.start()` but what will it tell us? Only that we've called `set()` on the Event. It won't guarantee that it had been done properly.
So basically, I don't think that testing this is worth the effort.

PR #889, opened 2017-10-21
Fixed version to PR #880

code1mountain:
Changes to PR #880 :

- Added tests

- Renamed Category Filter to category

- Renamed FileSize Filter to file_size

- Renamed FileType Filter to file_type

- Removed mime_types from constants

- Standard values at file_size Filter are None now and are only compared if a value is given -> no sys.maxsize

- Usage of stdlib mimetypes, but sometimes types are missing or different from the mime_types sent by the official telegram app. Hardcoded the strings in this cases.

- Changed the name attribute of category, file_size and file_type, so their attributes are shown

Still solves Issue #875

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -379,7 +379,7 @@ class Filters(object):
         """Subset for messages containing a status update.
 
         Examples:
-            Use these filters like: ``Filters.status_update.new_chat_member`` etc. Or use just
+            Use these filters like: ``Filters.status_update.new_chat_members`` etc. Or use just
             ``Filters.status_update`` for all status update messages.
 
         """
@@ -449,7 +449,7 @@ class Filters(object):
 
         migrate = _Migrate()
         """:obj:`Filter`: Messages that contain :attr:`telegram.Message.migrate_from_chat_id` or
-            :attr: `telegram.Message.migrate_from_chat_id`."""
+            :attr: `telegram.Message.migrate_to_chat_id`."""
 
         class _PinnedMessage(BaseFilter):
             name = 'Filters.status_update.pinned_message'


telegram/ext/filters.py
```diff
@@ -192,12 +192,126 @@ def filter(self, message):
...
+            self.min = min
+            self.max = max
+            self.name = "Filters.file_size(min={0}, max={1})".format(self.min, self.max)
+
+        def filter(self, message):
+            if message.audio:
+                filesize = message.audio.file_size
+            elif message.document:
+                filesize = message.document.file_size
+            elif message.photo:
```
jh0ker: `message.photo` is an array of different photo sizes. Not sure if the filter even makes here

telegram/ext/filters.py
```diff
@@ -192,12 +192,126 @@ def filter(self, message):
...
+                filesize = message.photo.file_size
```
jh0ker: and this line will fail

jh0ker:
Ideas for `message.photo` ?

telegram/ext/filters.py
code1mountain: Oh, thanks. Didn't thought about that. What about just using the photo in the highest resolution?

telegram/ext/filters.py
```diff
@@ -192,12 +192,126 @@ def filter(self, message):
     class _Document(BaseFilter):
         name = 'Filters.document'
 
+        class category(BaseFilter):
+            """This Filter filters documents by their category in the mime-type attribute
+
+            Note:
+                This Filter only filters by the mime_type of the document,
```
jh0ker: You didn't put an `Args:` section, and you should specifically mention that `mime_type.startswith` is used

jh0ker: Ah I see you put the `Args:` in the `__init__`, sorry

telegram/ext/filters.py
```diff
@@ -192,12 +192,126 @@ def filter(self, message):
...
+                    it doesn't check the validity of the document.
+                The user can manipulate the mime-type of a message and
+                    send media with wrong types that don't fit to this handler.
+
+            Examples:
+                Filters.documents.category("audio/") returnes `True` for all types
+                of audio sent as file, for example "audio/mpeg" or "audio/x-wav"
+            """
+
+            def __init__(self, category):
```
jh0ker: perhaps it would be clearer if we renamed the `category` parameter to `mime_type_prefix`

telegram/ext/filters.py
```diff
@@ -192,12 +192,126 @@ def filter(self, message):
...
+                This Filter only filters by the mime_type of the document,
+                    it doesn't check the validity of document.
+                The user can manipulate the mime-type of a message and
+                    send media with wrong types that don't fit to this handler.
+
+            Examples:
+                Filters.documents.file_type("audio/mpeg") filters all audio in mp3 format.
+            """
+
+            def __init__(self, filetype):
```
jh0ker: and `filetype` to `mime_type`

telegram/ext/filters.py
```diff
@@ -192,12 +192,126 @@ def filter(self, message):
...
+                if message.document:
+                    return message.document.mime_type.startswith(self.category)
+
+        application = category("application/")
+        audio = category("audio/")
+        image = category("image/")
+        video = category("video/")
+        text = category("text/")
+
+        class file_type(BaseFilter):
```
jh0ker: and perhaps even name the filter `mime_type`, to avoid confusion

telegram/ext/filters.py
code1mountain: Yes, makes sence

telegram/ext/filters.py
code1mountain: Ok, I will change

jh0ker:
So after some discussion with @code1mountain, we came to the conclusion it would make the most sense to allow any message where at least one of the different `PhotoSizes` is within the `min` and `max` limits.

Then in a second PR, we add helper methods to `Message` to get a specific `PhotoSize` within acceptable limits. It could look something like `Message.get_photo_size(min_file_size=None, max_file_size=None, min_resolution=None, max_resolution=None, prefer_biggest=True)` but nothing is set in stone yet.

Opinions?

Eldinnie:
What's the status on this?

jh0ker:
closing in favor of #1046

PR #895, opened 2017-10-31
Added conversation timeout in ConversationHandler

### LINKED ISSUE #648: how to set conversation no-operation timeout
### Steps to reproduce
1. use telegram.ext.conversationhandler

2. set run_async_timeout = 90

### Expected behaviour
Tell us what should happen
after 90s, the conversation should be stopped
### Actual behaviour
Tell us what happens instead
after 90s, i still in the conversation and can use command set in conversation states list
### Configuration
**Operating System:**
arch linux arm, PRi 3

python 3.6.1 python-telegram-bot 6.0.3

### Logs

i want to know how to set no-operation timeout but after reading documents, i don't find or i just find `run_async_timeout`

evgfilim1:
Closes #648 

#### Short usage:
```python
updater = Updater(TOKEN)
conversation = ConversationHandler(
    [CommandHandler('start', start)], 
    {0: MessageHandler(Filters.text, continue_)},
    conversation_timeout=60
)
updater.dispatcher.add_handler(conversation)
```


Eldinnie:
Looks good to me.

Eldinnie:
Anyone else wants a look on this? @tsnoam or @jh0ker ? 
Or I'll add `pending_merge`

telegram/ext/conversationhandler.py
```diff
@@ -294,6 +303,14 @@ def handle_update(self, update, dispatcher):
         """
         new_state = self.current_handler.handle_update(update, dispatcher)
 
+        if self.timeout_job is not None:
+            self.timeout_job.schedule_removal()
```
jh0ker: Do I understand correctly that there is only a single timeout job which gets removed whenever any conversation is triggered? 

Wouldn't you want to use a different job for each conversation?

telegram/ext/conversationhandler.py
```diff
@@ -124,7 +130,8 @@ def __init__(self,
                  timed_out_behavior=None,
                  per_chat=True,
                  per_user=True,
-                 per_message=False):
+                 per_message=False,
+                 conversation_timeout=0):
```
jh0ker: I prefer `None` as the default value for no timeout

tests/test_conversationhandler.py
```diff
@@ -277,3 +277,27 @@ def test_channel_message_without_chat(self, bot):
         message = Message(0, None, None, Chat(0, Chat.CHANNEL, 'Misses Test'), bot=bot)
         update = Update(0, message=message)
         assert not handler.check_update(update)
+
+    def test_conversation_timeout(self, dp, bot, user1):
```
jh0ker: Related to the comment above, perhaps adding a test for 2 users would be prudent.

telegram/ext/conversationhandler.py
evgfilim1: I'll change it, but there is no difference, I think.

telegram/ext/conversationhandler.py
evgfilim1: This timeout job is triggered in any conversation state, if you mean this. For different ConversationHandler instances the job is different too.

evgfilim1:
@jh0ker Done

Eldinnie:
Hi, because you're latest commit has [ci skip] the latest version is not run on CI's and codecov. Could rebase on current master and commit so it will run a full CI?

Thanks

evgfilim1:
@Eldinnie I know, I marked it because there are no useful changes (I removed a comment that is not needed at all)

evgfilim1:
Maybe we remove "pending reply" label?

tsnoam:
@evgfilim1 
the last commit was marked "ci skip" and the commit before it failed unitests.
would you please be so kind to rebase and remove the "ci skip" from the commit message?

evgfilim1:
@tsnoam done, waiting for CI to check.

evgfilim1:
`test_official` fail status should be solved when #986 is merged.

tsnoam:
@evgfilim1 Thank you for your contribution

PR #919, opened 2017-11-21
Fix sendPhoto method description

erjanmx:
Change wrong description referencing "video" argument in sendPhoto method


Eldinnie:
Thanks for pointing it out and you're contribution

Issue #943, opened 2017-12-20
Add full name for user as a convenience method


graynk:
I didn't find any issues about this, but it seems to me as a useful feature. Basically, I want to address my users by their full name, always. Right now I can only use User.name which return full name _only_ if no actual @username is provided. And so I have to use this snippet every time:
```
def get_name(user):
    name = user.first_name
    if user.last_name:
        name += ' ' + user.last_name
    return name
```
Why not add it as User.full_name?

Eldinnie:
Hi,

I guess we never saw a need for it. You are welcome to make a PR for it.
See our [contributing guidelines](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/.github/CONTRIBUTING.rst) to get started.

graynk:
Hi,

Just [made](https://github.com/python-telegram-bot/python-telegram-bot/pull/949) a PR

PR #949, opened 2017-12-24
Convenience full_name method for User

graynk:
Returns full name (first + last) if last name is present or just the first name if not present.

### DIFF
--- a/telegram/user.py
+++ b/telegram/user.py
@@ -83,6 +83,17 @@ class User(TelegramObject):
             return '%s %s' % (self.first_name, self.last_name)
         return self.first_name
 
+    @property
+    def full_name(self):
+        """
+        :obj:`str`: The users :attr:`first_name` and if present :attr:`last_name`.
+
+        """
+
+        if self.last_name:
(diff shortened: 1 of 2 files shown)


robert-cody:
In many locales format should be "surname name" (which is BTW more useful for sorting), so it's better to make locale as incoming optional parameter and return full_name depending on locale.

graynk:
I don't think this information is in python [locale](https://docs.python.org/3/library/locale.html) module though? 
It seems to have been in CLDR at [some](https://unicode.org/cldr/trac/ticket/177) point, but not anymore, if I understand it correctly. In any case, I view this as just a simpler version of User.name method that works exactly the same way, only ensuring that returned names are consistent in style (I feel that mentioning some users via "@username" and others via "name" is unintuitive and doesn't look all that well in group chats).

robert-cody:
@graynk you're right, for some strange reason, there is no such info in Python's locale module.

Eldinnie:
Looks good to me. Closes #943

tsnoam:
thank you for your contribution

PR #951, opened 2017-12-25
Updater: Log info message when receiving signal

### LINKED ISSUE #946: Log an INFO-level message when SIGINT received
Hi ;)

I'm recently experimenting with your nice bot. However, I stumbled upon this during my "learning-process":

I've created a bot, it worked normally. I just did `updater.start_polling` like in the examples, but without using `updater.idle()` (I thought first `start_polling` is blocking).  Exiting with CTRL+C worked immediately, but has thrown a `KeyboardInterrupt`.
After a while I found the `idle()` function, but I was actually a bit confused of CTRL+C's behaviour, because I've configured logging with a minimum level of `INFO`. So I've seen nothing when I pressed the keys. The bot actually needs a while to shutdown, and because nothing was logged I assumed the SIGINT signal wasn't received (/ assuming due to internal threading stuff not properly processed), so I pressed always CTRL+C two times causing a SIGTERM which worked immediately.
SIGINT notifications are only visible when enabling `DEBUG`-level logging. I would suggest to add an `INFO` message, that tells something like `Received SIGINT, shutting down...`, so nobody is worried ;)

Makman2:
Closes #946

### DIFF
--- a/telegram/utils/helpers.py
+++ b/telegram/utils/helpers.py
@@ -19,12 +19,25 @@
 """This module contains helper functions."""
 
 import re
+import signal
 from datetime import datetime
 
 try:
     from html import escape as escape_html  # noqa: F401
 except ImportError:
     from cgi import escape as escape_html  # noqa: F401
+
+
+# From https://stackoverflow.com/questions/2549939/get-signal-names-from-numbers-in-python
+_signames = {v: k
+             for k, v in reversed(sorted(vars(signal).items()))
+             if k.startswith('SIG') and not k.startswith('SIG_')}
+
+
+def get_signal_name(signum):
+    """Returns the signal name of the given signal number."""
+    return _signames[signum]
+
 
 # Not using future.backports.datetime here as datetime value might be an input from the user,
 # making every isinstace() call more delicate. So we just use our own compat layer.
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -245,18 +245,16 @@ class TestUpdater(object):
         os.kill(os.getpid(), signal.SIGTERM)
 
     @signalskip
-    def test_idle(self, updater):
+    def test_idle(self, updater, caplog):
         updater.start_polling(0.01)
         Thread(target=self.signal_sender).start()
-        with LogCapture(level=logging.INFO) as capture:
+
+        with caplog.at_level(logging.INFO):
             updater.idle()
 
-        capture.check(
-            ('telegram.ext.updater',
-             'INFO',
-             'Received signal {} (SIGTERM), stopping...'.format(
-                 signal.SIGTERM))
-        )
+        assert caplog.record_tuples == [(
+            'telegram.ext.updater', 'INFO', 'Received signal {} (SIGTERM), '
+            'stopping...'.format(signal.SIGTERM))]
 
         # If we get this far, idle() ran through
         sleep(.5)
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -20,6 +20,7 @@ import logging
 import os
 import signal
 import sys
+from functools import partial
 from queue import Queue
 from random import randrange
 from threading import Thread
@@ -35,7 +36,6 @@ except ImportError:
 
 import pytest
 from future.builtins import bytes
-from testfixtures import LogCapture
 
 from telegram import TelegramError, Message, User, Chat, Update, Bot
 from telegram.error import Unauthorized, InvalidToken
@@ -240,21 +240,24 @@ class TestUpdater(object):
 
         return urlopen(req)
 
-    def signal_sender(self):
+    def signal_sender(self, updater):
         sleep(0.2)
+        while not updater.running:
+            sleep(0.2)
+
         os.kill(os.getpid(), signal.SIGTERM)
 
     @signalskip
     def test_idle(self, updater, caplog):
         updater.start_polling(0.01)
-        Thread(target=self.signal_sender).start()
+        Thread(target=partial(self.signal_sender, updater=updater)).start()
 
-        with caplog.at_level(logging.INFO):
+        with caplog.atLevel(logging.INFO):
             updater.idle()
 
-        assert caplog.record_tuples == [(
-            'telegram.ext.updater', 'INFO', 'Received signal {} (SIGTERM), '
-            'stopping...'.format(signal.SIGTERM))]
+        rec = caplog.records()[-1]
+        assert rec.msg.startswith('Received signal {}'.format(signal.SIGTERM))
+        assert rec.levelname == 'INFO'
 
         # If we get this far, idle() ran through
         sleep(.5)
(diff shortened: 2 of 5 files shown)


Makman2:
Wondering why coverage decreased, haven't introduced any new branches...

Eldinnie:
@Makman2 Sometimes coverage fluctuates because errors do or don't happen. The important part is that the diff is 100% covered. I'll look more detailed after Christmas. Thanks for contributing.

robert-cody:
`def signal_name_by_code(code):
    return next((n for n in dir(signal) if n.startswith("SIG") and "_" not in n and getattr(signal, n) == code))`

Looks better.

Makman2:
Okay that shifts lookup effort to each call instead of once preparing a dict inside the module. But we shouldn't raise an implicit `StopIteration` in case there's no such signal-code^^ Might have side-effects when used in for-loops (in case that happens). At least we should do this imo:
```python
try:
    return next(...)
except StopIteration:
    raise KeyError('Unknown signal number {}'.format(...))
```
May I call it `signal_name_from_number` though? `by` sounds less descriptive for me + python docs don't mention the term "code" for those signal numbers^^

robert-cody:
@Makman2 you're right about StopIteration. You can name your method as you wish, i'm not this repo developer, just proposed less code solution.

bomjacob:
We in the dev team are hesitant about adding new dependencies (even just for tests). Especially since in this case it doesn't seem to be required as log can already be captured using pure pytest. See [Loggging - pytest documentation](https://docs.pytest.org/en/latest/logging.html#logging) for details.

telegram/utils/helpers.py
```diff
@@ -19,13 +19,26 @@
...
 
 try:
     from html import escape as escape_html  # noqa: F401
 except ImportError:
     from cgi import escape as escape_html  # noqa: F401
 
+
+# From https://stackoverflow.com/questions/2549939/get-signal-names-from-numbers-in-python
+_signames = {v: k
+             for k, v in reversed(sorted(vars(signal).items()))
```
tsnoam: nitpicking here. the `reversed` & `sorted` are useless here as you're creating a dict. please remove them.

Makman2:
> We in the dev team are hesitant about adding new dependencies (even just for tests). Especially since in this case it doesn't seem to be required as log can already be captured using pure pytest. See Loggging - pytest documentation for details.
Could you maybe update the code so it uses that style instead of the extra dependency?

Sure no problem :+1:

telegram/utils/helpers.py
Makman2: I think the `sorted` was there for signals with different names but same number, so there's a deterministic outcome. However I haven't understood the `reversed` call too^^

Makman2: And kk `reversed` then picks signal names that are sorted at the beginning. On my system I get these difference when I don't use `reversed`:
```
SIGIOT vs SIGABRT
SIGCLD vs SIGCHLD
SIGPOLL vs SIGIO
```

Though not sure if it's that important here^^

Makman2:
About
> We in the dev team are hesitant about adding new dependencies (even just for tests). Especially since in this case it doesn't seem to be required as log can already be captured using pure pytest. See Loggging - pytest documentation for details.
Could you maybe update the code so it uses that style instead of the extra dependency?

I have tried using this feature, however:
- It requires the installation of `pytest-capturelog`, so a package is anyway needed for this
- Equivalent testing code with that package fails, pytest gets terminated too early:
  ```
  tests/test_updater.py FTerminated
  ```

So I believe using `testfixtures` is currently the easiest solution.

tsnoam:
@Makman2 
We prefer using pytest ecosystem if possible (so if it comes down to adding a new dependency the prefernce will be for it).
Can you please commit your code for testing with pytest to your branch and allow edits for maintainers as described here:
https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

we'll try to work on it together before using `testfixtures`.

Makman2:
Done :+1:

telegram/utils/helpers.py
tsnoam: Check manpage of `signal(7)` all the above are synonyms. If one name is chosen over another it's merely a matter of luck, as dictionary order is not guaranteed.
On my system I had different results on different executions.

Another option is to leave the signal name alone and just print the number.

tsnoam:
The problem with pytest in the test you've sent is that it catches the signal itself / prevents Updater from catching it.
I assume that testfixtures forks and runs the test in a different process, thus ignoring the issue.
I'll see what can be done here.

tsnoam:
Note to self: need to make sure that this PR works properly on windows as the unitest in question is skipped on appveyor.

tsnoam:
@Makman2  - I believe that I fixed the relevant test. It was a matter of race condition (the signal arrived too soon).

@Eldinnie - Can you please (manually) verify on windows that the new code doesn't break anything?

tests/test_updater.py
```diff
@@ -240,21 +240,24 @@ def _send_webhook_msg(self,
...
 
-        with caplog.at_level(logging.INFO):
+        with caplog.atLevel(logging.INFO):
             updater.idle()
 
-        assert caplog.record_tuples == [(
-            'telegram.ext.updater', 'INFO', 'Received signal {} (SIGTERM), '
-            'stopping...'.format(signal.SIGTERM))]
+        rec = caplog.records()[-1]
+        assert rec.msg.startswith('Received signal {}'.format(signal.SIGTERM))
```
Makman2: just curious, why not asserting for the complete message? :)

telegram/utils/helpers.py
Makman2: > If one name is chosen over another it's merely a matter of luck, as dictionary order is not guaranteed.

Looks like it's true, I'm just wondering: the order we iterate over should always produce a guaranteed outcome. `vars(signal).items()` has unique mapping from `name -> number`. Using `sorted` we get always the same list with the same order, and then we are filling the dict with those values, while switching `name` and `number`. If there are numbers that are equal, later elements in the list should overwrite the old ones. Although using dict-comprehension, citing the `dict`-docs:
>  If a key occurs more than once, the last value for that key becomes the corresponding value in the new dictionary.

Makman2: > Another option is to leave the signal name alone and just print the number.

I wouldn't do that, signal numbers differ on OSes for the "same" signals.

Eldinnie:
```> python .\examples\echobot2.py
2018-01-08 15:10:03,051 - telegram.ext.updater - INFO - Received signal 2 (SIGINT), stopping...
>
```
Works fine on windows.

telegram/utils/helpers.py
tsnoam: ok. not worth the argument, lets keep it as is.

tsnoam:
Build fails due to unrelated problems fixed in a different PR.
Will merge now. 

@Makman2 Thanks for your contribution.

Makman2:
Thanks for helping me out and bearing with me ;D

tsnoam:
@Makman2 All the thanks are for you

PR #952, opened 2017-12-28
fix broken link 

fangherk:
While looking at the examples, I found a broken link for paymentbot.py, so I modified the link to the right URL.


fangherk:
Fixed in #942.

PR #959, opened 2018-01-01
Check for update.effective_chat explicitly in ConversationHandler.check_update

### LINKED ISSUE #927: Payment is not working if written AFTER a conversation handler
### Steps to reproduce
1. Just make a test payment as shown in the official example https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/paymentbot.py

2. Add the payment handlers **AFTER a conversation handler**, like this:
```
conv_handler = ConversationHandler(
    ...
)

# Add command handler to start the payment invoice
dp.add_handler(CommandHandler("shipping", start_with_shipping_callback))
dp.add_handler(CommandHandler("noshipping", start_without_shipping_callback))

# Optional handler if your product requires shipping
dp.add_handler(ShippingQueryHandler(shipping_callback))

# Pre-checkout handler to final check
dp.add_handler(PreCheckoutQueryHandler(precheckout_callback))

# Success! Notify your user!
dp.add_handler(MessageHandler(Filters.successful_payment, successful_payment_callback))
```

3. Run it and try to make a payment (I used /noshipping).

### Expected behaviour
It should let you make the full payment successfully.

### Actual behaviour
It lets you type all payment information, but in the last step, when clicking the last "Pay" button, it crash (see [Logs](#logs)) and doesn't call precheckout_callback function throw PreCheckoutQueryHandler.

### Additional information

- It works perfectly if you add it BEFORE the ConversationHandler.
- It seems to be recognized as if payment handlers were located inside the ConversationHandler.
- I don't think it to affect this issue, but I'm using Stripe payment in testing mode.

### Configuration
I'm using the last version up to date. (e8fac19de31b3d246d0253ac0267f7253a22d526)

### Logs
```
2017-12-02 17:31:52,827 - telegram.ext.dispatcher - ERROR - An uncaught error was raised while processing the update
  File "C:\Users\Carlos\Documents\Programacion\TelegramBots\Pickatext\telegram\ext\dispatcher.py", line 278, in process_update
  File "C:\Users\Carlos\Documents\Programacion\TelegramBots\Pickatext\telegram\ext\dispatcher.py", line 278, in <genexpr>
  File "C:\Users\Carlos\Documents\Programacion\TelegramBots\Pickatext\telegram\ext\conversationhandler.py", line 215, in check_update
    key = self._get_key(update)
  File "C:\Users\Carlos\Documents\Programacion\TelegramBots\Pickatext\telegram\ext\conversationhandler.py", line 185, in _get_key
    key.append(chat.id)
AttributeError: 'NoneType' object has no attribute 'id'
```

nmlorg:
Fixes #927

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -208,7 +208,7 @@ class ConversationHandler(Handler):
         # Ignore messages in channels
         if (not isinstance(update, Update) or
                 update.channel_post or
-                self.per_chat and (update.inline_query or update.chosen_inline_result) or
+                self.per_chat and not update.effective_chat or
                 self.per_message and not update.callback_query or
                 update.callback_query and self.per_chat and not update.callback_query.message):
             return False
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -206,10 +206,11 @@ class ConversationHandler(Handler):
 
         """
         # Ignore messages in channels
-        if (not isinstance(update, Update) or update.channel_post or self.per_chat
-                and (update.inline_query or update.chosen_inline_result) or self.per_message
-                and not update.callback_query or update.callback_query and self.per_chat
-                and not update.callback_query.message):
+        if (not isinstance(update, Update) or
+                update.channel_post or
+                self.per_chat and (update.inline_query or update.chosen_inline_result) or
+                self.per_message and not update.callback_query or
+                update.callback_query and self.per_chat and not update.callback_query.message):
             return False
 
         key = self._get_key(update)
--- a/tests/test_conversationhandler.py
+++ b/tests/test_conversationhandler.py
@@ -20,7 +20,8 @@ from time import sleep
 
 import pytest
 
-from telegram import Update, Message, User, Chat, CallbackQuery
+from telegram import (CallbackQuery, Chat, ChosenInlineResult, InlineQuery, Message,
+                      PreCheckoutQuery, ShippingQuery, Update, User)
 from telegram.ext import (ConversationHandler, CommandHandler, CallbackQueryHandler)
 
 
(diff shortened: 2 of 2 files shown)


telegram/ext/conversationhandler.py
```diff
@@ -206,10 +206,11 @@ def check_update(self, update):
...
         # Ignore messages in channels
-        if (not isinstance(update, Update) or update.channel_post or self.per_chat
-                and (update.inline_query or update.chosen_inline_result) or self.per_message
-                and not update.callback_query or update.callback_query and self.per_chat
-                and not update.callback_query.message):
+        if (not isinstance(update, Update) or
+                update.channel_post or
+                self.per_chat and not update.effective_chat or
+                self.per_message and not update.callback_query or
+                update.callback_query and self.per_chat and not update.callback_query.message):
```
tsnoam: Honestly I'm confused form the logic here (it doesn't necessarily make it wrong though) so I have some questions:
1. Is anything "buggy" with the current logic? If yes, please explain what.
1. What is the new logic you're trying to introduce?
1. Was `and (update.inline_query or update.chosen_inline_result)` removed intentionally or is it by mistake? If intentionally please explain why.
1. When mixing `and`, `and not` & `or` conditionals it makes it hard to read (even if its correct). Can you be so kind and add parentheses to make the logic easier to read?

tsnoam: btw, this piece of code was barely readable before (at least to me) so this change made it much more difficult to undestand...

nmlorg: The original code was:
```python
        if (not isinstance(update, Update) or update.channel_post or self.per_chat
                and (update.inline_query or update.chosen_inline_result) or self.per_message
                and not update.callback_query or update.callback_query and self.per_chat
                and not update.callback_query.message):
```
and I first changed it to group related conditions by line:
```python
        if (not isinstance(update, Update) or
                update.channel_post or
                self.per_chat and (update.inline_query or update.chosen_inline_result) or
                self.per_message and not update.callback_query or
                update.callback_query and self.per_chat and not update.callback_query.message):
```
i.e. each line contains conditions that are joined by `and`, and the lines are separated by `or`. Due to order of operations rules, this means each line is now the same as if it had been enclosed in parentheses:
```python
       if ((not isinstance(update, Update)) or
                (update.channel_post) or
                (self.per_chat and (update.inline_query or update.chosen_inline_result)) or
                (self.per_message and not update.callback_query) or
                (update.callback_query and self.per_chat and not update.callback_query.message)):
```
The actual logic change (the second commit) was to remove the `(update.inline_query or update.chosen_inline_result)` check and replace it with a check for `not update.effective_chat`. It seemed pretty clear the intent was to exclude updates that don't have a `chat` assigned, which includes inline_query and chosen_inline_result (as explicitly tested in the original logic) as well as callback_query from inline messages, shipping_query, and pre_checkout_query (as documented in [Update.effective_chat](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/update.py#L138)). So I could have changed the logic to exclude those additional types explicitly, but ultimately the update gets passed into _get_key which blindly dereferences update.effective_chat, so either that would need to be modified to be `if self.per_chat and chat is not None:` (like the per_user check) or go ahead and just check for effective_chat before calling _get_key, which is what I did.

The one remaining issue is that last line, which excludes callback_query updates that don't have a message attached when per_chat is set, which doesn't seem to make sense to me. The only time update.callback_query is dereferenced is again in _get_key when per_message is set and update.callback_query.inline_message_id is not set, neither of which seem related to what's being tested. It's possible the original logic's author meant to include that logic tacked onto the previous per_message test:
```python
        if (not isinstance(update, Update) or
                update.channel_post or
                self.per_chat and not update.effective_chat or
                self.per_message and (not update.callback_query or
                                      update.callback_query and self.per_chat and not update.callback_query.message)):
```
i.e. if per_message and it's not a callback_query or it is a callback_query but callback_query.message is not set (though that's not strictly necessary, and still not sure what the per_chat check in the middle is for). Personally, if I was confident that _get_key's logic was perfect, I would just get rid of that check and have _get_key return `None` or something itself if the update isn't relevant, and do:
```python
        if not isinstance(update, Update):
            return False
        key = self._get_key(update)
        if not key:
            return False
```
but the smallest necessary change is to explicitly check for update.effective_chat before dereferencing it, and either exclude all updates without one (like I did) or include them even if per_chat is set ( `if self.per_chat and chat is not None:`).

(Do you still want me to add the extra explicit parens?)

tsnoam:
@nmlorg 

PR #975, opened 2018-01-16
Update commandhandler.py

gavrilovks:
Add processing of commands with arguments separated by underscore
examples:
/command_arg1_arg2
/download_875547
/item_78896
It will be useful if bot sends few commands like this in text message, and user can click on it

### DIFF
--- a/telegram/ext/commandhandler.py
+++ b/telegram/ext/commandhandler.py
@@ -41,6 +41,8 @@ class CommandHandler(Handler):
             edited messages.
         pass_args (:obj:`bool`): Optional. Determines whether the handler should be passed
             ``args``.
+        pass_uargs (:obj:`bool`): Optional. Determines whether the handler should be passed
+            ``args``. 
         pass_update_queue (:obj:`bool`): Optional. Determines whether ``update_queue`` will be
             passed to the callback function.
         pass_job_queue (:obj:`bool`): Optional. Determines whether ``job_queue`` will be passed to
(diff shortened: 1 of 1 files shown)


nmlorg:
Neat trick :)  I was actually thinking about proposing some kind of pre-processing stage that could do things like de-base64 the args to /start, attach bot-specific data about the sender, etc.  (My main concern is about Message.entities, once I start mucking around with payload text that becomes meaningless.) For right now, I'm doing pre-processing as a dummy initial Handler, and to get what you need I think you could use something like:
```python
class PreHandler(ext.Handler):
    def __init__(self):
        super(PreHandler, self).__init__(None)

    @staticmethod
    def check_update(update):
        if update.effective_message:
            update.effective_message.text = update.effective_message.text.replace('_', ' ')

 ...

    dispatcher.add_handler(PreHandler())
    dispatcher.add_handler(ext.CommandHandler('download', handle_download, pass_args=True))
    dispatcher.add_handler(ext.CommandHandler('item', handle_item, pass_args=True))
```

gavrilovks:
Thank you for the snippet! It will work for me

Eldinnie:
Hi,

Thanks for your PR. I think we will not implement this because it's a very specific usecase. The snippet provided will only work with two specific adjustments.
The PreHandler will have to implement a `check_update()` method  (https://github.com/python-telegram-bot/python-telegram-bot/blob/820f4e1d59ac2e05c237015843c855e2a2527c6b/telegram/ext/handler.py#L73) or it will raise a `NotImplementedError`.
And the prehandler should be added in a different group, or subsequent handlers will not be called.

I'm indending to close this PR. Any objections?

nmlorg:
You might need to add a dummy `handle_update` to make pylint happy but, since `check_update` essentially fails (falling off the end without returning `True`), Dispatcher will never try to call PreHandler.handle_update and will continue checking all the other handlers in the same group :)

tsnoam:
Hi @gavrilovks,

Thanks for your contribution.
As @Eldinnie specified, this patch is too specific for a single use case. Another elegant approach would be to modify `CommandHandler.__init__()` to accept a new parameter `separator` which will default to `None`.
Then you can simply split the text using `.split(self._seperator, ...)`.

In any case, I'll be closing this PR now. You are always welcome to consult with us on how to implement a more generic solution and/or open a new PR.

PR #934, opened 2017-12-15
Selecting jobs based on job's name

paradox70:
This new function `select_job(name=)` returns a tuple of jobs with the given name that are currently in the ``JobQueue``

PR #977, opened 2018-01-19
jobqueue: Thread safety fixes

### LINKED ISSUE #968: JobQueue.jobs() is not thread safe
Direct usage of `PriorityQueue.queue` is not thread safe.
We should either remove support for `JobQueue.jobs()` (undesireable) or fix the issue.

Blocks #934

tsnoam:
- Fix JobQueue.jobs to obtain a lock on the internal queue object prior
   to iterating over it.

 - Rename JobQueue.queue to JobQueue._queue. This shouldn't be
   accessible by the user directly, but rather only with sanitized
   thread safe methods.

 - JobQueue.interval_seconds - access self.interval only once to avoid
   race conditions.

Fixes #968

### DIFF
--- a/telegram/ext/jobqueue.py
+++ b/telegram/ext/jobqueue.py
@@ -38,7 +38,7 @@ class JobQueue(object):
     """This class allows you to periodically perform tasks with the bot.
 
     Attributes:
-        queue (:obj:`PriorityQueue`): The queue that holds the Jobs.
+        _queue (:obj:`PriorityQueue`): The queue that holds the Jobs.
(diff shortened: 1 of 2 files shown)


Eldinnie:
Looks good. But why not directly integrate #934 in this?

tsnoam:
@Eldinnie Yes it's possible.

tsnoam:
@Eldinnie Actually #934 is not complete. Still missing unitests.
I'll merge tihs PR and we'll wait for the author of #934 to add unitests.

PR #979, opened 2018-01-20
Update requirements: PySocks to fix #525

### LINKED ISSUE #525: How to set a proxy for this module?
I live in China, it's too hard to connent to Telegram's API Server because of some reason that as we all know. Can you have some ways to set a proxy for this module? Thanks.

kmahyyg:


kmahyyg:
We love it to help people build an immediately-open-to-use package.

tsnoam:
@kmahyyg 
Thank you for you contribution.

However, we've already  listed the PySocks as an extra dependency for ptb.
The reason that it's only an extra is that most people do not need it and we do not intend to force them to install unneeded dependencies.
Users who need socks support can always do:
`pip install python-telegram-bot[socks]`

Issue #835, opened 2017-09-16
Expand Filters and remove unwanted parameters from Handlers


Eldinnie:
It's been mentioned before, but I wanted a placemarker to keep it. Also see the discussion in #832 
There's a few handlers right now that have the arguments (`message_updates`, `channel_post_updates`, `edited_updates` and the deprecated `allow_edited`) These being `RegexHandler` and `MessageHandler`.

We have a good system in place to decide what update to deliver to a handler with the `Filters` module, and have these extra (legacy) arguments in these handlers is not needed and inconsistent. Also regard Issue #831 .
I propose to:

- Remove the arguments mentioned above from `RegexHandler` and `MessageHandler`
- Add the following filters to the `Filters` module:
  - message_update
  - edited_message_update
  - message_updates (both above)
  - channel_post_update
  - edited_channel_post_update
  - channel_updates (both above)
  - somethingtobenamed (for all above)

The end result will be that either of these update types can be separately filtered for `RegexHandler`, `MessageHandler` and `CommandHandler`.

However we need to decide on a default behavior and how to implement it.I think it could be easily implemented to let all these handlers default to `update.message` if filters list is empty:
```python
if not filters:
  self.filters = Filters.message
```
But whenever a user only specifies a `Filters.group` default behavior will be gone and will fire errors when a messagehandler tries to fire for `update.callback_query` for example.
Love to hear your thoughts.

jh0ker:
I'm wondering if we actually need to continue the `RegexHandler` or if we can integrate it with `MessageHandler`. Or perhaps we can inherit from `MessageHandler`. What do you think?

JosXa:
@jh0ker I've made a couple experiments that corroborate your idea. What I tried was to have a `Filters.regex(pattern)` and it made the handlers a lot cleaner since I could use other filters in conjunction. 
Now the only problem with that is the question of how to pass the `groups` and `groupdict` when there are *multiple* regex filters. One idea would be to require **named** `pass_groups` flags that are passed to the callback in a dict. Another possibility might be to build our own **abstraction** around what was matched by the several regex filters, in that we merge their individual groupdicts together to a bigger one that has all the captured groups. I'm still quite unsure.

On the same note, do we actually still need the `CommandHandler`? *Food for thought.*

[...]

PR #995, opened 2018-02-01
Implemented tools for Deep Linking

JosXa:
- Implemented new helper-method `create_deep_linked_url`
- Added `Filters.regex` and documented that it is not yet compliant with #835 (i.e. not passing groups or groupdict)
- Tests for all above
- Added example for deep linking
- Modified examples so that they all get their token from a `settings.py`. That way, you can set your token once and access all of the example bots. Added documentation to this in the appropriate places, especially that `Updater(TOKEN)`, where TOKEN stems from the `settings.py`, requires you to replace the constant with a **string** if you users want to use the example standalone. This might be controversial, so please respond if you think it is a good addition.


telegram/ext/filters.py
```diff
@@ -171,6 +173,33 @@ def filter(self, message):
...
+        """
+
+        def __init__(self, pattern):
+            if isinstance(pattern, string_types):
+                pattern = re.compile(pattern)
+            self.pattern = pattern
+            self.name = 'Filters.regex({})'.format(self.pattern)
+
+        def filter(self, message):
+            return bool(re.search(self.pattern, message.text))
```
tsnoam: Is there a reason not to do:
 ```python
return bool(self.pattern.search(message.text))
```

File: telegram/ext/filters.py [hunk 2]
```diff
@@ -171,6 +173,33 @@ def filter(self, message):
...
+
+
```
tsnoam: that's a redundant empty line.

telegram/ext/filters.py
```diff
@@ -171,6 +173,33 @@ def filter(self, message):
     command = _Command()
     """:obj:`Filter`: Messages starting with ``/``."""
 
+    class regex(BaseFilter):
```
tsnoam: Code for regex filter is good (regardless of some nitpicking).
However, it has nothing to do with the deep linking PR.
Please open a separate PR for it.

tests/test_helpers.py
```diff
@@ -26,3 +27,30 @@ def test_escape_markdown(self):
...
+        expected = "https://t.me/{}?start={}".format(username, payload)
+        actual = helpers.create_deep_linked_url(username, payload)
+        assert expected == actual
+
+        payload = ""
+        expected = "https://t.me/{}".format(username)
+        assert expected == helpers.create_deep_linked_url(username, payload)
+        assert expected == helpers.create_deep_linked_url(username)
+
+        try:
```
tsnoam: Please use `pytest.raises()` instead of the entire try...except... block

tests/test_helpers.py
```diff
@@ -26,3 +27,30 @@ def test_escape_markdown(self):
...
+            # Invalid characters
+            helpers.create_deep_linked_url(username, 'text with spaces')
+            pytest.fail()
+        except Exception as e:
+            assert isinstance(e, ValueError)
+
+        try:
```
tsnoam: Please use `pytest.raises()` instead of the entire try...except... block

telegram/ext/filters.py
code1mountain: self.pattern can either be a regex compiled pattern or a string. You're solutions doesn't work for strings.

JosXa: This should be okay

telegram/ext/filters.py
JosXa: See comment

tsnoam:
@JosXa 
Please send Regex filter as a different PR and then we can proceed.

JosXa:
#1028

jh0ker:
@JosXa #1028 is merged

JosXa:
@jh0ker Neat!

PR #1013, opened 2018-02-19
Add caption_(html|markdown)(_urled)?

### LINKED ISSUE #1010: [FeatureRequest] Bot API 3.6: caption_html and caption_markdown
As Bot API 3.6 now supports formatting and URL in message captions, it would be useful to have `caption_html`, `caption_html_urled`, `caption_markdown` and `caption_markdown_urled` in `Message` objects that has comparable features to their [`text_` counterparts](https://python-telegram-bot.readthedocs.io/en/stable/telegram.message.html#telegram.Message.text_html).

Eldinnie:
Closes #1010


telegram/message.py
```diff
@@ -938,11 +938,38 @@ def text_html_urled(self):
...
+        Use this if you want to retrieve the message caption with the caption entities formatted as
+        HTML. This also formats :attr:`telegram.MessageEntity.URL` as a hyperlink.
+
+        Returns:
+            :obj:`str`: Message caption with caption entities formatted as HTML.
+
+        """
+        return self._parse_html(self.caption, self.parse_caption_entities(), urled=True)
+
+    def _parse_markdown(self, text, entities, urled=False):
```
tsnoam: Can you change this to be `@staticmethod`?

telegram/message.py
```diff
@@ -873,17 +874,16 @@ def parse_caption_entities(self, types=None):
             for entity in self.caption_entities if entity.type in types
         }
 
-    def _text_html(self, urled=False):
-        entities = self.parse_entities()
-        message_text = self.text
+    def _parse_html(self, text, entities, urled=False):
```
tsnoam: Can you change this to be `@staticmethod`?

telegram/message.py
```diff
@@ -873,17 +874,16 @@ def parse_caption_entities(self, types=None):
...
+        message_text = text
```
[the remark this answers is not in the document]
Eldinnie: This is where I was being an idiot with double named parameters before. My fix is bad too :D. Will change

telegram/message.py
Eldinnie: Why?

telegram/message.py
Eldinnie: why again?

telegram/message.py
tsnoam: as discussed, because this method needs nothing from the instance which makes it a good enough reason to be static.

PR #1017, opened 2018-02-22
Persistence

Eldinnie:
Adding a construct for adding persistence to PTB.
- Added Basepersistence as an interface for persistence classes
- Added PicklePersistence for persistence using pickle
- Added DictPersistence for persistence using dict/JSON
- Integrating the persistence into the dispatcher, conversationhandler and updater classes
- Added helper methods for (de)serializing conversations and user/chat_data


bomjacob:
@Eldinnie Is this ready for review?

JosXa:
Hi @SehgalDivij, this is an implementation that will generally solve the persistence for ptb. An abstract base class was provided that will allow users to implement their own serialization interfaces and pass them on to the library, effectively giving you the ability to serialize almost every collection in the library.

telegram/ext/basepersistence.py
```diff
@@ -0,0 +1,123 @@
...
+    * If :attr:`store_user_data` is ``True`` you must overwrite :meth:`get_user_data` and
+      :meth:`update_user_data`.
+    * If you want to store conversation data with :class:`telegram.ext.ConversationHandler`, you
+      must overwrite :meth:`get_conversations` and :meth:`update_conversations`.
+    * :meth:`flush` will be called when the bot is shutdown, and must always be overwritten.
+
+    Attributes:
+        store_user_data (:obj:`bool`): Optional, Whether user_data should be saved by this
+            persistence class.
+        store_chat_data (:obj:`bool`): Optional. Whether user_data should be saved by this
```
bomjacob: chat_data*

telegram/ext/basepersistence.py
```diff
@@ -0,0 +1,123 @@
...
+            persistence class.
+
+    Args:
+        store_user_data (:obj:`bool`, optional): Whether user_data should be saved by this
+            persistence class. Default is ``True``.
+        store_chat_data (:obj:`bool`, optional): Whether user_data should be saved by this
```
bomjacob: chat_data*

telegram/ext/basepersistence.py
```diff
@@ -0,0 +1,123 @@
...
+                defaultdict to store.
+        """
+        raise NotImplementedError
+
+    def flush(self):
+        """Will be called by :class:`telegram.ext.Updater` upon receiving a stop signal. Gives the
+        persistence a chance to finish up saving or close a database connection gracefully. If this
+        is not of any importance just pass will be sufficient.
+        """
+        raise NotImplementedError
```
bomjacob: If this is not of importance, should it really raise NYE?

telegram/ext/conversationhandler.py
```diff
@@ -76,6 +76,10 @@ class ConversationHandler(Handler):
         per_user (:obj:`bool`): Optional. If the conversationkey should contain the User's ID.
         per_message (:obj:`bool`): Optional. If the conversationkey should contain the Message's
             ID.
+        name (:obj:`str`): Optional. The name for this conversationhandler. Required for
```
bomjacob: Could we default to an internal counter or something?

telegram/ext/conversationhandler.py
```diff
@@ -76,6 +76,10 @@ class ConversationHandler(Handler):
...
+            persistence
+        persistent (:obj:`bool`): Optional. If the conversations dict for this handler should be
```
bomjacob: If name is required, could we not just have a single `persistent_name` that turns persistence on if a string?

telegram/ext/dispatcher.py
```diff
@@ -70,6 +72,10 @@ class Dispatcher(object):
             instance to pass onto handler callbacks.
         workers (:obj:`int`): Number of maximum concurrent worker threads for the ``@run_async``
             decorator.
+        user_data (:obj:`dict`): A dictionary handlers can use to store data for the user.
```
bomjacob: Shouldn't this be defaultdict?

telegram/ext/dispatcher.py
```diff
@@ -86,16 +94,34 @@ class Dispatcher(object):
...
-    def __init__(self, bot, update_queue, workers=4, exception_event=None, job_queue=None):
+    def __init__(self, bot, update_queue, workers=4, exception_event=None, job_queue=None,
+                 persistence=None):
         self.bot = bot
         self.update_queue = update_queue
-        self.job_queue = job_queue
         self.workers = workers
+        if persistence:
+            if not isinstance(persistence, BasePersistence):
+                self.logger.warning("persistence should be based on Telegram.ext.BasePersistence")
```
bomjacob: Telegram should be lowercase I think

telegram/ext/dispatcher.py
```diff
@@ -86,16 +94,34 @@ class Dispatcher(object):
...
+            if self.persistence.store_user_data:
+                self.user_data = self.persistence.get_user_data()
+                if not isinstance(self.user_data, defaultdict):
+                    raise ValueError("user_data must be of type defaultdict")
+            else:
+                self.user_data = defaultdict(dict)
+            if self.persistence.store_chat_data:
+                self.chat_data = self.persistence.get_chat_data()
+                if not isinstance(self.chat_data, defaultdict):
+                    raise ValueError("`chat_data must be of type defaultdict")
```
bomjacob: misplaced `

telegram/ext/dispatcher.py
```diff
@@ -277,6 +303,17 @@ def process_update(self, update):
             try:
                 for handler in (x for x in self.handlers[group] if x.check_update(update)):
                     handler.handle_update(update, self)
+                    if self.persistence:
+                        if self.persistence.store_chat_data:
+                            try:
+                                self.persistence.update_chat_data(self.chat_data)
+                            except Exception:
```
bomjacob: Is there a need to have separate exception handlers for chat/user data?

telegram/ext/picklepersistence.py
```diff
@@ -0,0 +1,235 @@
...
+# You should have received a copy of the GNU Lesser Public License
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+"""This module contains the BasePersistence class."""
+import pickle
+from collections import defaultdict
+
+from telegram.ext import BasePersistence
+
+
+class PicklePersistence(BasePersistence):
```
bomjacob: Is there ever a file where it makes sense to use single_file=False?

telegram/ext/picklepersistence.py
```diff
@@ -0,0 +1,235 @@
...
+        filename (:obj:`str`): The filename for storing the pickle files. When :attr:`single_file`
+            is false this will be used as a prefix.
+        store_user_data (:obj:`bool`): Optional. Whether user_data should be saved by this
+            persistence class.
+        store_chat_data (:obj:`bool`): Optional. Whether user_data should be saved by this
+            persistence class.
+        single_file (:obj:`bool`): Optional. When ``False`` will store 3 sperate files of
+            `filename_user_data`, `filename_chat_data` and `filename_conversations`. Default is
+            ``True``.
+        on_flush (:obj:`bool`): Optional. When ``True`` will only save to file when :meth:`flush`
```
bomjacob: It would be nice if we could have an in-between too. So instead of "always whenever data change" and "almost never, only when exiting", we have something like "every 10 min" or "every 10 updates".
Do you think something like that would be possible?

telegram/ext/picklepersistence.py
```diff
@@ -0,0 +1,235 @@
...
+
+class PicklePersistence(BasePersistence):
+    """Using python's builtin pickle for making you bot persistent.
+
+    Attributes:
+        filename (:obj:`str`): The filename for storing the pickle files. When :attr:`single_file`
+            is false this will be used as a prefix.
+        store_user_data (:obj:`bool`): Optional. Whether user_data should be saved by this
+            persistence class.
+        store_chat_data (:obj:`bool`): Optional. Whether user_data should be saved by this
```
bomjacob: Should it not have a store_conversations too?

telegram/ext/conversationhandler.py
Eldinnie: That would rely on the conv_handlers keeping in the same order when the code is changing. Better to force naming IMO

telegram/ext/conversationhandler.py
Eldinnie: Interesting. But this way we also have the name seperate if not persistent. Might be usefull some day?

telegram/ext/dispatcher.py
Eldinnie: I always learned to have as few statements in a `try` block as possible.

telegram/ext/picklepersistence.py
Eldinnie: dunno, but it might

telegram/ext/picklepersistence.py
Eldinnie: It is, when you create a job that calls `flush()` on the persistence.

telegram/ext/picklepersistence.py
Eldinnie: I had a hard time with this, But basically the conversationhandler determines if the conversations should be stored.

telegram/ext/dispatcher.py
Eldinnie: yest

telegram/ext/updater.py
```diff
@@ -485,6 +492,8 @@ def _join_threads(self):
 
     def signal_handler(self, signum, frame):
         self.is_idle = False
+        if self.persistence:
```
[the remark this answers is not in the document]
Eldinnie: why?

telegram/ext/dispatcher.py
```diff
@@ -86,16 +94,34 @@ class Dispatcher(object):
...
-        self.job_queue = job_queue
         self.workers = workers
+        if persistence:
+            if not isinstance(persistence, BasePersistence):
+                self.logger.warning("persistence should be based on Telegram.ext.BasePersistence")
+            self.persistence = persistence
+        else:
+            self.persistence = None
+
+        if self.persistence:
```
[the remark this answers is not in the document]
Eldinnie: there was no good one

Eldinnie:
Ok for future reference (And for the future `CHANGES.RST`)
Beside the obvious persistence classes and construct in this PR I added 3 helper methods to `telegram.utils.helpers` namely:
- `enocde_conversations_to_json(conversations)`
  to json enocde a conversations dict which uses tuples as keys (which give a ValueError when trying to encode as JSON. This helper JSON-serializes the tuples to before serializing the dict.
- `decode_conversations_from_json(json_string)`
  To reverse the operation from the encoding. It takes a string as encoded by the previous function and returns a conversations dict as the conversationhandler expects it
- decode_user_chat_data_from_json(data)
  When JSON-serializing user/chat_data to json the keys (which are ints) are converted to strings. 
  This function tries to cast dicts keys to ints.

telegram/ext/updater.py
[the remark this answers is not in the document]
jsmnbom: Cause otherwise we impede the ability to exit immediately? Like if the user presses CTRL-C (or whatever) twice, it shouldn't save persistence twice, it just quit ASAP imo

telegram/ext/basepersistence.py
```diff
@@ -0,0 +1,123 @@
...
+
+    All relevant methods must be overwritten. This means:
+
+    * If :attr:`store_chat_data` is ``True`` you must overwrite :meth:`get_chat_data` and
+      :meth:`update_chat_data`.
+    * If :attr:`store_user_data` is ``True`` you must overwrite :meth:`get_user_data` and
+      :meth:`update_user_data`.
+    * If you want to store conversation data with :class:`telegram.ext.ConversationHandler`, you
+      must overwrite :meth:`get_conversations` and :meth:`update_conversations`.
+    * :meth:`flush` will be called when the bot is shutdown, and must always be overwritten.
```
jsmnbom: This docstring should probably be changed, since flush mustn't always be overwritten as it is implemented atm (which I think is a good thing)

telegram/ext/dictpersistence.py
```diff
@@ -0,0 +1,197 @@
...
+
+try:
+    import ujson as json
+except ImportError:
+    import json
+from collections import defaultdict
+from telegram.ext import BasePersistence
+
+
+class DictPersistence(BasePersistence):
```
jsmnbom: How is this class useful? Like it doesn't actually save anywhere does it?

telegram/ext/dispatcher.py
```diff
@@ -86,16 +94,30 @@ class Dispatcher(object):
...
-        self.job_queue = job_queue
         self.workers = workers
-
         self.user_data = defaultdict(dict)
-        """:obj:`dict`: A dictionary handlers can use to store data for the user."""
         self.chat_data = defaultdict(dict)
-        """:obj:`dict`: A dictionary handlers can use to store data for the chat."""
+        if persistence:
+            if not isinstance(persistence, BasePersistence):
+                self.logger.warning("persistence should be based on telegram.ext.BasePersistence")
```
jsmnbom: Imo this should raise an exception

telegram/ext/picklepersistence.py
jsmnbom: I'm not really a fan of having features that we aren't sure anyone can use... It's just more maintenance work for us :/

telegram/ext/picklepersistence.py
```diff
@@ -0,0 +1,235 @@
...
+            pickle.dump(data, f)
+
+    def get_user_data(self):
+        """Returns the user_data from the pickle file if it exsists or an empty defaultdict.
+
+        Returns:
+            :obj:`defaultdict`: The restored user data.
+        """
+        if self.user_data:
+            pass
```
jsmnbom: I'm not sure I quite understand these passes.. Could you explain it for me?

telegram/ext/picklepersistence.py
jsmnbom: Also you can't choose the filetype (suffix) when using single_file=False.?

telegram/utils/helpers.py
```diff
@@ -139,3 +145,65 @@ def effective_message_type(entity):
...
+    tmp = json.loads(json_string)
+    conversations = {}
+    for handler, states in tmp.items():
+        conversations[handler] = {}
+        for key, state in states.items():
+            conversations[handler][tuple(json.loads(key))] = state
+    return conversations
+
+
+def decode_user_chat_data_from_json(data):
```
jsmnbom: What's the inverse of this func?

tests/test_persistence.py
```diff
@@ -0,0 +1,751 @@
...
+        del (pickle_persistence)
+        pickle_persistence_2 = PicklePersistence(filename='pickletest',
+                                                 store_user_data=True,
+                                                 store_chat_data=True,
+                                                 singe_file=False,
+                                                 on_flush=False)
+        u = Updater(bot=bot, persistence=pickle_persistence_2)
+        dp = u.dispatcher
+        dp.add_handler(h2)
+        dp.process_update(update)
```
jsmnbom: Should this check to make sure that first and second even got called?

telegram/ext/updater.py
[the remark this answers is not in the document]
Eldinnie: Good point

telegram/ext/dictpersistence.py
Eldinnie: What class do you mean?

telegram/ext/dispatcher.py
Eldinnie: I thought about that. But if someone really wants to create their own version without basing it on Basepersistence they should be allowed

telegram/ext/dictpersistence.py
jsmnbom: DictPersistence
I don't really see the use of it, since it doesn't save anywhere, right? And therefore isn't persistent?

telegram/ext/picklepersistence.py
Eldinnie: What do you mean with: `aren't sure anyone can use... I` 

It will check if it has user_data loaded in memory in the `PicklePersistence`. If it does (pass) it will just return a copy of it. If not it will try to load it from a pickle.

telegram/ext/dispatcher.py
jsmnbom: I equate it to how we handle handlers in dispatche https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/dispatcher.py#L329 
Also BaseFilter iirc

telegram/ext/dictpersistence.py
Eldinnie: It gives you easy acces to the dicts/json from the persistence class. So you can use a job to store them. It accepts json to restore them. @tsnoam requested thuis to be present

telegram/ext/picklepersistence.py
jsmnbom: Riiight, pass.. not return. Got it. Brain derped a bit :)

jsmnbom: What do you mean with
`What do you mean with: aren't sure anyone can use... I`
Now I'm confused...

telegram/ext/dispatcher.py
Eldinnie: ok seems legit

telegram/ext/dictpersistence.py
jsmnbom: Alright, so it's for more manual handling of persistent data. Got it. :)

telegram/ext/picklepersistence.py
Eldinnie: No, but I don;t really see a problem here. I think it's versatile enough this way.

telegram/ext/picklepersistence.py
Eldinnie: I was looking in the wrong view and comments got messed up. NVM me. found it :D

telegram/utils/helpers.py
Eldinnie: `json.dumps` (which will convert `{1: 'test', 2: 3}` into: `'{"1": "test", "2": 3}'`

jsmnbom: Ah, yes, of course :D

Eldinnie:
only pypy failing on Travis.
Coverage can be chacked [here](https://codecov.io/gh/python-telegram-bot/python-telegram-bot/pull/1017/changes). Diff coverage is 94.04%
Please check if you're happy now @jsmnbom and others welcome for review (@tsnoam @jh0ker )

owneroxxor:
Already using PTB 11.1.0 but my script gives error when trying to import BasePersistence fromtelegram.ext.
Is persistence feature already launched?

Eldinnie:
@owneroxxor it's merged into master, but not yet released. In the next released version it will be available.

alisherafat01:
@Eldinnie when do you release this feature? we need this!

davidh-:
> @Eldinnie when do you release this feature? we need this!

Also would like a release! I guess for the meantime we can install from master.

PR #1020, opened 2018-02-24
CommandHandler - ignore strings in entities and "/" followed by whitespace

PaulSonOfLars:
Telegram entities show that `/ test` is not a valid command. However,
PTB accepts this as a valid command due to how the message splitting is
done.

This commit makes sure that any commands are indeed in the format of
`/test`, with no space between `/` and `test`.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -53,8 +53,8 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Oleg Shlyazhko <https://github.com/ollmer>`_
 - `Oleg Sushchenko <https://github.com/feuillemorte>`_
 - `overquota <https://github.com/overquota>`_
-- `Paul Larsen <https://github.com/PaulSonOfLars>`_
 - `Patrick Hofmann <https://github.com/PH89>`_
+- `Paul Larsen <https://github.com/PaulSonOfLars>`_
 - `Pieter Schutz <https://github.com/eldinnie>`_
 - `Rahiel Kasim <https://github.com/rahiel>`_
 - `Sascha <https://github.com/saschalalala>`_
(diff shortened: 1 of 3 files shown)


tsnoam:
@PaulSonOfLars 
Thanks for your contribution. It is a great observation that leading spaces should be properly handled, as well as ignoring commands if they are part of a markup entity.

I pushed a really small fix for the PR and making sure that unitest pass.
I'm not sure why the unitests on Windows failed before, but if they'll pass this time I'll merge, otherwise I'll look into the failure more thoroughly.

Eldinnie:
Small question, why do you check on `message.text_markdown` instead of `message.text`?

tsnoam:
@Eldinnie That's a smart change. Suppose the message sent to the bot is __/start__ - this is not a command...

Eldinnie:
@tsnoam Ah that is indeed true. However /吴语 isn't a command either, and we do accept those as commands. Changing the behavior breaks with what people axpect and this changes more then only the proposed change.

tsnoam:
@Eldinnie We'll still support "/吴语" (up to potential problems with py2)
however, a single "/" with a space afterwards is definitely not a command, i really don't think we should support that.

tsnoam:
Fixed the PR title to be a bit more accurate.

Eldinnie:
Ok so I found out what the trouble was. (And I;m pretty surprised it didn't give any trouble on other platforms).
The test bot we use has the username `ptb_test_bot` which was only escaped properly on AppVeyor to `ptb\_text\_bot` Why this only happened on appveyor is a mystery for me. But it seems using `text_html` instead of `text_markdown` fixes the problem for this PR.

If CI finishes properly now it;s good to merge

PaulSonOfLars:
Awesome, glad I could help! And thanks for the debugging, I really had no clue what was going on there... Let me know if there's anything else that should be changed.

PR #1021, opened 2018-02-25
Added prefix argument to CommandHandler

JosXa:
This will be useful for making CommandHandlers that do not adhere to the `/`-prefix convention.
Hashtag commands or ones without any prefix at all become possible through this PR.

### DIFF
--- a/telegram/ext/commandhandler.py
+++ b/telegram/ext/commandhandler.py
@@ -21,8 +21,8 @@ import warnings
 
 from future.utils import string_types
 
-from .handler import Handler
 from telegram import Update
+from .handler import Handler
 
 
 class CommandHandler(Handler):
@@ -39,6 +39,8 @@ class CommandHandler(Handler):
             Filters.
         allow_edited (:obj:`bool`): Optional. Determines Whether the handler should also accept
             edited messages.
+        prefix (:obj:`bool`): Optional. Denotes the leading character of commands with this
+            handler.
         pass_args (:obj:`bool`): Optional. Determines whether the handler should be passed
             ``args``.
         pass_update_queue (:obj:`bool`): Optional. Determines whether ``update_queue`` will be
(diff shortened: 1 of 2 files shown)


telegram/ext/commandhandler.py
```diff
@@ -110,6 +115,15 @@ def __init__(self,
...
         self.allow_edited = allow_edited
+        if prefix is None or not isinstance(prefix, string_types) or len(prefix) > 1:
+            raise ValueError("The prefix argument to CommandHandler must be a single character or "
+                             "an empty string.")
+        if prefix == ' ':
+            # A leading space character is stripped away in Telegram messages, so we can consider
+            # it as an empty string prefix.
+            self.prefix = ''
+        else:
+            self.prefix = prefix
```
tsnoam: * I don't think prefix should be limited to a single char. Same as you accept an empty string you can accept any string longer than 1 char.

* Instead of calculating (over and over) the `empty_leader` during `check_update`, you can calculate `self.leader_len` here (at the c'tor) and simplify everything else later on.

* Calling `len(prefix)` here will also make the check for `None` redundant. It will be the one to raise `TypeError` if necessary.

wagnerluis1982:
The idea behind this PR is interesting. Since there is two months without activity, I assume it's abandoned, so can I assume the work?

tsnoam:
@wagnerluis1982 Hi,
Thanks for the suggestion. The reason this PR was on hold is because we discussed a different design:
1. CommandHandler should be implemented strictly for commands as they are defined and supported by bot father.
2. PrefixHandler (as it would be more appropriate to call the suggested feature here) is nothing more than a convenience on top of RegexHandler.
3. We have plans to refactor the handler mechanism which will make this PR somewhat of wasted work.

If you want to contribute to the project, feel free to contact me and we can discuss the details. http://t.me/tsnoam

tsnoam:
@wagnerluis1982 see issue #1026 about the redesign of handlers.

wagnerluis1982:
@tsnoam thanks for the feedback, I understood the point.

bomjacob:
Once again thanks for getting us started on this idea.
This PR has been superseded by #1114 - closing.

PR #1042, opened 2018-03-12
Fix in telegram.Message

Eldinnie:
The `_parse_(html|markdown)` methods now properly return `None` on an empty `Message.text` or an empty `Message.caption`

### DIFF
--- a/tests/test_message.py
+++ b/tests/test_message.py
@@ -176,7 +176,7 @@ class TestMessage(object):
     def test_text_html_empty(self, message):
         message.text = None
         message.caption = "test"
-        assert not message.text_html
+        assert message.text_html is None
 
     def test_text_html_urled(self):
         test_html_string = ('Test for &lt;<b>bold</b>, <i>ita_lic</i>, <code>code</code>, '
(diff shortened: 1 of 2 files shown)


tests/test_message.py
```diff
@@ -173,6 +173,11 @@ def test_text_html_simple(self):
         text_html = self.test_message.text_html
         assert text_html == test_html_string
 
+    def test_text_html_empty(self, message):
+        message.text = None
+        message.caption = "test"
+        assert not message.text_html
```
jh0ker: Perhaps change to `assert message.text_html is None` ?

jh0ker: same for the other tests of course

Eldinnie: I could, but why?

jh0ker: Because it would properly test the feature. We don't want to allow empty strings, right?

Eldinnie: I don't see why not, actually I was think of doing:
```
if not message_text:
  return message_text
```

PR #1049, opened 2018-03-16
Implemented Tools for deep linking

JosXa:
Allows
`url = helpers.create_deep_linked_url(bot.get_me().username, "Payload")`


jh0ker:
Looks good, but the example seems very convoluted. Do we really need 3 levels of deep linking to demonstrate how a single function works?

JosXa:
@jh0ker It's more about explaining what's possible with deep links and to spark the imagination a bit

jh0ker:
Fair enough, but in that case perhaps you could make one of the levels use an inline button with a URL instead of just another text link

JosXa:
Good point

telegram/utils/helpers.py
```diff
@@ -139,3 +139,43 @@ def effective_message_type(entity):
...
+
+def create_deep_linked_url(bot_username, payload=None):
+    """
+    Creates a deep-linked URL for this ``bot_username`` with the specified ``payload``.
+    See  https://core.telegram.org/bots#deep-linking  to learn more.
+
+    The ``payload`` may consist of the following characters: ``A-Z, a-z, 0-9, _, -``
+
+    Note: Works well in conjunction with
+    ``CommandHandler("start", callback, Filters.regex('payload')``.
```
bomjacob: Missing a paramthesis here?
Also shouldn't the filter be passed as a keyword argument?

telegram/utils/helpers.py
```diff
@@ -139,3 +139,43 @@ def effective_message_type(entity):
...
+
+    Examples:
+        Example ``create_deep_linked_url(bot.get_me().username, "some-params")``
```
bomjacob: "Examples: Example"
seems a bit redundant
(remove 2nd example text)

examples/deeplinking.py
```diff
@@ -0,0 +1,111 @@
...
+    """Reached through the CHECK_THIS_OUT payload"""
+    url = helpers.create_deep_linked_url(bot.get_me().username, SO_COOL)
+    text = "Awesome, you just accessed hidden functionality!\n\nContinue here: " + url
+    update.message.reply_text(text)
+
+
+def deep_linked_level_2(bot, update):
+    """Reached through the SO_COOL payload"""
+    url = helpers.create_deep_linked_url(bot.get_me().username, USING_ENTITIES)
+    text = "You can also mask the deep-linked URLs as links: " \
```
bomjacob: Does this run on py2?

tsnoam:
@JosXa 
What is the status of this PR? ( @bomjacob requested certain changes)

JosXa:
Currently not having the time to fix it up. Anyone else interested in getting the kudos for this contribution? @bomjacob

Bibo-Joshi:
Brought this PR up to date and implemented the requested changes.
Also made it possible create `startgroup` deep linking urls using @JosXa s new `helpers` method

Issue #1048, opened 2018-03-16
Consider making chat_id a positional argument in chat.send_message and similar to avoid error messages when not using kwargs


xates:
When doing `chat.send_message("Hello")`  the following error occurs: 
> TypeError: send_message() got multiple values for keyword argument 'chat_id'

This can be solved by providing each argument through its keyword (like that: `chat.send_message(text="Hello")`), however this is not very convenient.

By replacing keyword argument `chat_id=self.id` with positional argument `self.id` (it is the first argument in of the `bot.send_message` method) in the `Chat` class, you can avoid to be forced to always use kwargs when calling this method.

The same thing happens for the other shortcut methods (`send_photo`, `send_audio`, ...) and for the corresponding methods of the `User` class.
https://github.com/python-telegram-bot/python-telegram-bot/blob/1530ed20e51960296b0edb9bfa483047b59187e0/telegram/chat.py#L227

jh0ker:
@xates Yeah I agree with that. Would you like to contribute a PR to change this behavior?

xates:
Yeah sure!

jh0ker:
@xates Thanks a lot! :)

Please read the contributing guide before you begin: https://github.com/python-telegram-bot/python-telegram-bot/blob/master/.github/CONTRIBUTING.rst

PR #1050, opened 2018-03-16
Make chat_id a positional argument inside shortcut methods of Chat and User classes #1048

xates:


jh0ker:
@xates Looks like you also have to update the tests

xates:
Ok, I have tried to figure out how they work but I am not entirely sure... would something like that work?

```python

    def test_instance_method_send_message(self, monkeypatch, chat):
        def test(*args, **kwargs):
            return args[0] == chat.id and args[1] == 'test'

        monkeypatch.setattr('telegram.Bot.send_message', test)
        assert chat.send_message('test')

    def test_instance_method_send_audio(self, monkeypatch, chat):
        def test(*args, **kwargs):
            return args[0] == chat.id and args[1] == 'test_audio'

        monkeypatch.setattr('telegram.Bot.send_audio', test)
        assert chat.send_audio('test_audio')
```

Moreover, I have noticed that both Chat and the User class lack the tests for `send_picture`. Is that intentional or should I add them?

jh0ker:
I believe that will work, but please confirm it locally before committing (as per contributing guide). 

And yes, that looks like an oversight, adding tests for `send_photo` would be appreciated :+1:

xates:
I cannot run the tests locally because I get an ImportError: No module named ptb_urllib3.urllib3

Eldinnie:
@xates Looks like you didn't follow this step from the contributing guide:
```
Clone your forked repository of python-telegram-bot to your computer:

$ git clone https://github.com/<your username>/python-telegram-bot --recursive
```
To fix it run:
```
$ git submodule update --init --recursive

jh0ker:
@xates Sounds like you didn't clone with `--recursive` to also clone git submodules (as per contribution guide)

xates:
You're right, sorry, I missed that `--recursive` that was at the end

jh0ker:
If you get errors in `test_get_chat` you can ignore them

xates:
Oh, I see! That was driving me crazy...

jh0ker:
I'm sorry about that, someone unwittingly renamed the group that is used in that test so the test suddenly failed everywhere.

jh0ker:
Looks good to me :+1:

jh0ker:

xates:
Yeah, I guess that SSL problem is not my fault :smile: 

Well, thank you, for running such a wonderful project :wink:

jh0ker:
Eh, tests sometimes pass and fail depending on the alignment of the planets :new_moon:

PR #1067, opened 2018-04-07
Added video note filter

OrBin:
Filters.video_note was missing from the message filters file (`telegram/ext/filters.py`), so I added it.


jh0ker:
Sweet, thanks a lot!

PR #1068, opened 2018-04-08
Add a caption_entity filter for filtering caption entities

PaulSonOfLars:
Currently, the Filters.entity filter only checks message.entities.
This PR adds Filters.caption_entities to check message.caption_entities.

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -532,7 +532,7 @@ class Filters(object):
             self.name = 'Filters.entity({})'.format(self.entity_type)
 
         def filter(self, message):
-            return any([entity.type == self.entity_type for entity in message.entities])
+            return any(entity.type == self.entity_type for entity in message.entities)
 
     class caption_entity(BaseFilter):
         """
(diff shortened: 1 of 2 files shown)


keessonnema:
This would be really helpful. Code looks good aswell. +1

telegram/ext/filters.py
```diff
@@ -534,6 +534,27 @@ def __init__(self, entity_type):
...
+                in :class:`telegram.MessageEntity`.
+
+        """
+
+        def __init__(self, entity_type):
+            self.entity_type = entity_type
+            self.name = 'Filters.caption_entity({})'.format(self.entity_type)
+
+        def filter(self, message):
+            return any([entity.type == self.entity_type for entity in message.caption_entities])
```
jh0ker: nitpick, but why not just use a generator expression instead of a list comprehension?

PaulSonOfLars: totally agree, i copy pasted the previous filter code is why

jh0ker:
Thanks for your contribution :slightly_smiling_face:

PR #1072, opened 2018-04-13
Catch json_data decode errors

udf:
Fixes the case where non-utf8 data would cause an exception and halt the dispatcher. See https://t.me/pythontelegrambotgroup/129353 for more info.

Note that the telegram docs at https://core.telegram.org/bots/api#callbackquery warn about this specific case ("Be aware that a bad client can send arbitrary data in this field.").

### DIFF
--- a/telegram/utils/request.py
+++ b/telegram/utils/request.py
@@ -148,7 +148,7 @@ class Request(object):
         try:
             decoded_s = json_data.decode('utf-8')
             data = json.loads(decoded_s)
-        except ValueError:
+        except (ValueError, UnicodeDecodeError):
             raise TelegramError('Invalid server response')
 
         if not data.get('ok'):  # pragma: no cover
--- a/telegram/utils/request.py
+++ b/telegram/utils/request.py
@@ -145,8 +145,8 @@ class Request(object):
             dict: A JSON parsed as Python dict with results - on error this dict will be empty.
 
         """
-        decoded_s = json_data.decode('utf-8')
         try:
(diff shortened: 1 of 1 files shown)


jh0ker:
Thanks a lot for your contribution @udf 

The bug and fix look good, but it would be great to have a test to ensure we won't have regressions on this in the future.

I see you are a first-time contributor, would you like to add your name to the [`contributor list`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/AUTHORS.rst)?

Lonami:
How does one write tests for a single line change like this…? There should have been tests *before* the change, testing random bytes.

jh0ker:
@udf A good test case would be to show that the correct exception is raised for a callback query that contains bytes that can't be decoded using the utf8 encoding

Lonami:
I feel like we can open a separate issue to track that, this should be merged ASAP as it brings down any bots using the library. Unless you restart them automatically they will be down until you find out.

udf:
@jh0ker I don't have the time right now to figure out how to write a test. Perhaps someone else can do it, so we can get this merged quicker?

udf:
Closing this because apparently critical bug fixes aren't important enough.

jh0ker:
I'm actually in the process of writing unit tests right now. If you don't want your commits merged, I'll commit the fix myself. Otherwise, feel free to reopen this PR.

Lonami:
Thanks, we're just glad it will be fixed one way or another.

Issue #1073, opened 2018-04-13
CommandHandler should not check filters if not listening to the command


wagnerluis1982:
I have a filter to restrict some users from call some commands. I call `AdminFilter` and the relevant code can be seen below (actually, I check the admin users from a database).

```python
class AdminFilter(BaseFilter):
    def __init__(self, *admin_users):
        self.admin_users = admin_users

    def filter(self, message):
        if message.from_user.username in self.admin_users:
            return True
        else:
            message.reply_text("You're not an admin!", quote=True)
            return False
```

With this filter applied, when an user that is not an admin calls the forbidden command, then receives a reply warning. And here is the issue.

### Steps to reproduce
The following code illustrate the problem. In this scenario I have an administrative command `/jungle` and a normal command `/rain`.

If I am the admin, I call any command and see no issue, but If am not admin, I receive the warning for both commands.

```python
def jungle(bot, update):
    bot.send_message(update.message.chat_id, 'Welcome to the Jungle!')

def rain(bot, update):
    bot.send_message(update.message.chat_id, 'November Rain...')

dispatcher.add_handler(CommandHandler('jungle', jungle, AdminFilter('wagnermacedo')))
dispatcher.add_handler(CommandHandler('rain', rain))
```

### Expected behaviour
Suppose I'm not the bot administrator, then in a conversation with the bot, I expect the following:

```
me: /jungle
bot: > reply /jungle
       You're not an admin!
```
```
me: /rain
bot: November Rain...
```

### Actual behaviour
What happens instead is that bot reply that I'm not an admin for both commands.

For `/jungle`, it's okay, it was expected
```
me: /jungle
bot: > reply /jungle
       You're not an admin!
```

But `/rain` hasn't the admin filter, so it should not reply the warning.
```
me: /rain
bot: > reply /rain
       You're not an admin!
bot: November Rain...
```

:point_up_2: Surprisingly the command still works, even replying I'm not admin...

### Configuration
**Version of Python, python-telegram-bot & dependencies:**

I tested with the last version from master (59659ea).

```
$ python -m telegram
python-telegram-bot 10.0.1
certifi 2018.01.18
future 0.16.0
Python 3.4.8 (default, Apr 13 2018, 16:18:01)  [GCC 5.4.0 20160609]
```

### Fix
I have already fixed the issue in my machine, I'm going soon to create a pull request.

PR #1074, opened 2018-04-13
CommandHandler faster check

wagnerluis1982:
Make the filters applied to a command to execute only if the command name matches. This solves the problem described in #1073

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -65,6 +65,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `thodnev <https://github.com/thodnev>`_
 - `Valentijn <https://github.com/Faalentijn>`_
 - `voider1 <https://github.com/voider1>`_
+- `Wagner Macedo <https://github.com/wagnerluis1982>`_
 - `wjt <https://github.com/wjt>`_
 
 Please add yourself here alphabetically when you submit your first pull request.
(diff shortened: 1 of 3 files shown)


wagnerluis1982:
I forgot to say, this fixes #1073

jh0ker:
Thanks a lot for your contribution @wagnerluis1982 

The fix looks good, but it would be great to have a test to ensure we won't have regressions on this in the future. 

I see you are a first-time contributor, would you like to add your name to the [`contributor list`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/AUTHORS.rst)?

jh0ker:
@wagnerluis1982 A good test case would be to show that a filter on a command handler is not executed when the command in a message does not match the command defined in the handler.

jh0ker:
I added a unit test :+1:

wagnerluis1982:
@jh0ker Sorry the delay, I was off in the weekend.

> The fix looks good, but it would be great to have a test to ensure we won't have regressions on this in the future.

I was so eager to fix the issue that I didn't remember tests :roll_eyes:. I see you already did, so I'm too late now...

> I see you are a first-time contributor, would you like to add your name to the contributor list?

It's always good :smile:. Since I actively develop one bot, I intend to contribute more to the project.

tsnoam:
@wagnerluis1982 thanks for your contribution

PR #1079, opened 2018-04-21
InputFile: Fix proper naming of file when reading from subprocess.PIPE

### LINKED ISSUE #1055: Allow subprocess.Popen.stdout as file input
### Steps to reproduce
1.

```python
@run_async
def pipe(bot, update):
  proc = subprocess.Popen(
    "cat image.jpg".split(),
    stdout=subprocess.PIPE
  )
  update.message.reply_voice(proc.stdout)
  while proc.returncode is None:
    proc.poll()
```

### Expected behaviour

Allow `subprocess.Popen.stdout` be treated as valid io stream, just like `open("image.jpg", "rb")` is.

### Actual behaviour

```
...
  File "/usr/lib/python3.6/site-packages/telegram/utils/request.py", line 262, in post
    data = InputFile(data)
  File "/usr/lib/python3.6/site-packages/telegram/files/inputfile.py", line 76, in __init__
    self.filename = os.path.basename(self.input_file.name)
  File "/usr/lib/python3.6/posixpath.py", line 144, in basename
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not int
```

### Configuration
**Operating System:**

`Linux xiaomi 4.15.9-1-ARCH #1 SMP PREEMPT Sun Mar 11 17:54:33 UTC 2018 x86_64 GNU/Linux`

```
Python 3.6.4 (default, Jan  5 2018, 02:35:40) 
[GCC 7.2.1 20171224] on linux
```

### PROPOSED FIX

https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/files/inputfile.py#L76 should stringify `self.input_file.name` to look as:

```python
self.filename = os.path.basename(str(self.input_file.name))
```

tsnoam:
fixes #1055

### DIFF
--- a/telegram/files/inputfile.py
+++ b/telegram/files/inputfile.py
@@ -70,7 +70,9 @@ class InputFile(object):
             self.input_file_content = self.input_file.read()
             if 'filename' in data:
                 self.filename = self.data.pop('filename')
-            elif hasattr(self.input_file, 'name'):
+            elif (hasattr(self.input_file, 'name') and
+                  not isinstance(self.input_file.name, int) and  # py3
+                  self.input_file.name != '<fdopen>'):  # py2
                 # on py2.7, pylint fails to understand this properly
                 # pylint: disable=E1101
                 self.filename = os.path.basename(self.input_file.name)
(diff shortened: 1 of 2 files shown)


bomjacob:
This looks good, except on windows you'd want the the png path to use `\` instead of `/` in the test.

PR #1088, opened 2018-04-25
Fix the default value of "timeout" argument to match documentation

xamgore:
It is said "20 seconds" is the default value, but in fact, it is None.


tsnoam:
@xamgore 
Thankyou for your contribution.

PR #1089, opened 2018-04-25
Improve documentation of 'telegram.ext.filters' module

jh0ker:
The `Filters._Document` documentation was completely missing from the sphinx build because it's a private member. Not sure if there's a better way of doing this with sphinx, but I'm happy with how it looks now.

I fixed a few minor documentation messes as well.


Eldinnie:
@jh0ker I'll check tomorrow, remind me if I forget :D

Eldinnie:
As discussed in chat. I think the solution for the `document` filter docs should basically be the same as for the `status_update`.

jsmnbom:
Superseeded by #1348.

PR #1096, opened 2018-04-30
Updating sphinx to py3

Eldinnie:


docs/source/conf.py
```diff
@@ -291,17 +291,9 @@
 
 # -- script stuff --------------------------------------------------------
 
-import inspect
-
 
 def autodoc_skip_member(app, what, name, obj, skip, options):
```
bomjacob: We can probably remove the whole "script stuff" and below now :)

tsnoam:
@Eldinnie @bomjacob 
I think this is good merge, no?

docs/source/conf.py
Eldinnie: It's a part of the template so leaving it in.

Eldinnie:
Yes

Issue #1026, opened 2018-03-01
Handler callback revamp


tsnoam:
This issue is here to expose to the community a proposed design for the Handlers callback prototype.
We have recently closed/rejected PR #859 in this subject as we preferred to go in a different direction. Here we intend to discuss with the community what will be the chosen design.

**Current situation:**
Minimal callback prototype is `def callback(bot, update)`.
If users require more data/facilities they can initialize the Handler with extra `pass_` (bool) arguments  which mandates the callback to accept more arguments (like `groups`, `args` and others).

Problems:
 * The API is very cumbersome and confusing for new users.
 * Adding/pushing new context data into the callbacks requires code changes from the users.
Lets say that we want the RegexHandler to provide the match object to the callback. With the current design it will be impossible to add that information without changing the callback prototype and (most likely) impose a new `pass_` parameter during Handler initialization.

**Proposed API**
IMPORTANT: This change is non backward compatible. We might implement a backward compatibility layer for a certain period like we have done with other deprecated mechanisms in the past.
 * Deprecate `pass_` arguments when initializing Handlers.
 * Handler callbacks prototype will change to: `def callback(update, ctx)`.
 * The above `ctx` will be a context of all information/facilities that users may need. The context will be a Mixin of various context objects, depending on the Handler.
Examples of such information/facilities:
    * `bot`
    * `update_queue`
    * `job_queue` (something to think about: instead of updater_queue & job_queue, pass the `updater`)
    * `user_data`
    * `chat_data`
    * `match` (instead of `groups`, `groups_dict` and similar for).

JosXa:
@tsnoam I'd love for the ctx argument to be optional somehow. This was one of the nice results of #859, enabling short callbacks like
https://github.com/python-telegram-bot/python-telegram-bot/blob/9c6efb8b6ba2b0d9c0da5ca59c48714e805d5cdc/examples/autowiring.py#L39

tsnoam:
@JosXa 
I don't see a reason why to make the context argument optional. Users should have all the power they need. They can ignore (not use) the argument if they want, but they should still be able to easily use it.

tsnoam:
To clarify what I'm thinking:
Passing a context argument to the callback is the standard approach. This is what every programmer is looking for when he tries to understand what he can do from within the callback.
So in order to make things simple, the callback prototype should be simple enough so every one can immediately see what he has at hands.

Eldinnie:
[...]

PR #1100, opened 2018-05-03
Context

### LINKED ISSUE #1080: Integrating chat_data and user_data with respective Objects
Instead of passing them via `pass_chat_data` and `pass_user_data` explicitly, the user and chat data should be attached to the actual wrapper objects:
`update.effective_user.data`,
`update.effective_chat.data`

This should happen in the same fashion as many of the ptb objects get a `bot` injected.

As user_data and chat_data are only pass_ -able when they actually exist, I think this could be a good way of integrating everything a bit better by removing all the pass_* arguments.

This may happen either in preparation for #1026, or **as a replacement for it** (as the number of `pass_*` statements would sink). Also, it'd be consistent with [Pyrogram](https://github.com/pyrogram/pyrogram).

bomjacob:
This adds optional support for using context objects instead of all the pass_* in the handlers, as we've talked about a lot in the past.
Motivation is #1080 being seemingly really difficult to implement nicely, and it seems redundant if this got implemented instead (therefore closes #1080).
Not sure if this is exactly how we talked about it, but it seems to implement all that's needed (and as a bonus is 100% backwards compatible!!).

Currently employs a bit of magic to figure out whether to use the old pass_ or the new context. Namely it checks the `callback` signature to check how many parameters it expects. This is expected to work very well on python >= 3.5 even with decorators. On python < 3.5 it will fail completely if decorators are used, but otherwise work fine.
The user can disable the new functionality by setting `use_context=False` in a given handler. This is also how it can be manually turned on, in the case of python < 3.5 and decorated callback functions.


tsnoam:
The design discussion is at #1026

bomjacob:
This is a bit different from that discussion. Mainly that everything is in the context object, but that it acts like a update object, so it's still very convenient for the user. (it also make it much easier to do backwards compat, since it's a different amount of parameters)
The only thing that stood out to be as something we maybe should change is whether to keep groups/groupdict or just have the Match object in there instead.

telegram/ext/handler.py
```diff
@@ -17,6 +17,13 @@
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains the base class for handlers as used by the Dispatcher."""
+import inspect
+import sys
+
+from telegram.ext import Context
+
+has_follow_wrapped = sys.version_info >= (3, 5)
+has_full_argspec = sys.version_info[0] == 3
```
tsnoam: Regardless of the issue of do we want bkwd compat here, please use `CAPITAL_LETTERS` for globals.
Also, since these are internal implementation details, they should be prefixed with `_`.

telegram/ext/handler.py
```diff
@@ -63,12 +81,28 @@ def __init__(self,
...
+                 use_context=None):
         self.callback = callback
         self.pass_update_queue = pass_update_queue
         self.pass_job_queue = pass_job_queue
         self.pass_user_data = pass_user_data
         self.pass_chat_data = pass_chat_data
+        self.use_context = use_context
+
+        if use_context is None:
+            if has_follow_wrapped:
```
tsnoam: Since you've favoured using globals here (in contrary to code locality) please add a comment signifying which python version each of the `if... elif... else...` blocks supports.

telegram/ext/handler.py
```diff
@@ -63,12 +81,28 @@ def __init__(self,
...
+                parameters = [p for p in sig.parameters.values() if p.kind ==
+                              inspect.Parameter.POSITIONAL_OR_KEYWORD or
+                              p.kind == inspect.Parameter.POSITIONAL_ONLY]
+                self.use_context = len(parameters) == 1
+            elif has_full_argspec:
+                argspec = inspect.getfullargspec(callback)
+                self.use_context = len(argspec.args) == (2 if inspect.ismethod(callback) else 1)
+            else:
+                argspec = inspect.getargspec(callback)
+                self.use_context = len(argspec.args) == (2 if inspect.ismethod(callback) else 1)
```
tsnoam: Please add here something like
```python
if not self.use_context:
    warning('deprecated ... nice text ....')
```

telegram/ext/handler.py
```diff
@@ -17,6 +17,13 @@
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains the base class for handlers as used by the Dispatcher."""
+import inspect
+import sys
+
+from telegram.ext import Context
```
tsnoam: Do this instead:
```python
from .context import Context
```
(though it might be irrelevant if you'd move the `Context` definition into this file.

telegram/ext/choseninlineresulthandler.py
```diff
@@ -67,13 +77,15 @@ def __init__(self,
...
+                 pass_chat_data=False,
+                 use_context=None):
         super(ChosenInlineResultHandler, self).__init__(
             callback,
             pass_update_queue=pass_update_queue,
             pass_job_queue=pass_job_queue,
             pass_user_data=pass_user_data,
-            pass_chat_data=pass_chat_data)
+            pass_chat_data=pass_chat_data,
+            use_context=use_context)
```
tsnoam: Any reason not to delete `__init__()` completely? It does nothing except for calling `super()` with the same default arguments.

telegram/ext/choseninlineresulthandler.py
```diff
@@ -86,20 +98,3 @@ def check_update(self, update):
...
-
-        """
-        optional_args = self.collect_optional_args(dispatcher, update)
-
-        return self.callback(dispatcher.bot, update, **optional_args)
-
-    # old non-PEP8 Handler methods
-    m = "telegram.ChosenInlineResultHandler."
-    checkUpdate = deprecate(check_update, m + "checkUpdate", m + "check_update")
-    handleUpdate = deprecate(handle_update, m + "handleUpdate", m + "handle_update")
```
tsnoam: Please update `CHANGES.rst` and specify that these methods were removed.
It would be unpleasant if we'd forget doing so.

telegram/ext/commandhandler.py
```diff
@@ -129,8 +142,8 @@ def check_update(self, update):
             :obj:`bool`
 
         """
-        if (isinstance(update, Update)
-                and (update.message or update.edited_message and self.allow_edited)):
+        if (isinstance(update, Update) and
+                (update.message or update.edited_message and self.allow_edited)):
```
tsnoam: Since it's deprecated - remove this code.
(and don't forget to update CHANGES.rst)

telegram/ext/precheckoutqueryhandler.py
```diff
@@ -66,13 +77,15 @@ def __init__(self,
...
+                 pass_chat_data=False,
+                 use_context=None):
         super(PreCheckoutQueryHandler, self).__init__(
             callback,
             pass_update_queue=pass_update_queue,
             pass_job_queue=pass_job_queue,
             pass_user_data=pass_user_data,
-            pass_chat_data=pass_chat_data)
+            pass_chat_data=pass_chat_data,
+            use_context=use_context)
```
tsnoam: Any reason not to delete `__init__()` completely? It does nothing except for calling super() with the same default arguments.

telegram/ext/shippingqueryhandler.py
```diff
@@ -66,13 +77,15 @@ def __init__(self,
...
+                 pass_chat_data=False,
+                 use_context=None):
         super(ShippingQueryHandler, self).__init__(
             callback,
             pass_update_queue=pass_update_queue,
             pass_job_queue=pass_job_queue,
             pass_user_data=pass_user_data,
-            pass_chat_data=pass_chat_data)
+            pass_chat_data=pass_chat_data,
+            use_context=use_context)
```
tsnoam: Any reason not to delete `__init__()` completely? It does nothing except for calling super() with the same default arguments.

telegram/ext/commandhandler.py
bomjacob: That seems like something for another PR.

telegram/ext/choseninlineresulthandler.py
bomjacob: They should've been removed ages ago... the other ones were... though I cannot figure out when...

tsnoam: I think that we delayed all deprecations regarding Handlers until this refactoring effort.

telegram/ext/commandhandler.py
tsnoam: Everything deprecated with Handlers - this is the time...

telegram/ext/stringcommandhandler.py
```diff
@@ -86,18 +97,13 @@ def check_update(self, update):
         return (isinstance(update, string_types) and update.startswith('/')
                 and update[1:].split(' ')[0] == self.command)
```
tsnoam: This code is not aligned with the code in `collect_additional_context()`.  I suggest to do:
```python
if isinstance(update, string_types) and update.startswith('/'):
    args = update[1:].split(' ')
    if args[0] == self.command:
        return args
return None
```

telegram/ext/stringcommandhandler.py
```diff
@@ -86,18 +97,13 @@ def check_update(self, update):
...
+        optional_args = super(StringCommandHandler, self).collect_optional_args(dispatcher,
+                                                                                update,
+                                                                                check_result)
         if self.pass_args:
             optional_args['args'] = update.split()[1:]
+        return optional_args
 
-        return self.callback(dispatcher.bot, update, **optional_args)
+    def collect_additional_context(self, context, update, dispatcher, check_result):
+        context.args = update.split()[1:]
```
tsnoam: see comment above regarding matching behaviour.

telegram/ext/stringregexhandler.py
```diff
@@ -90,28 +103,28 @@ def check_update(self, update):
...
+        if self.pattern:
+            if self.pass_groups:
+                optional_args['groups'] = check_result.groups()
+            if self.pass_groupdict:
+                optional_args['groupdict'] = check_result.groupdict()
+        return optional_args
+
+    def collect_additional_context(self, context, update, dispatcher, check_result):
+        if self.pattern:
+            context.groups = check_result.groups()
```
tsnoam: same comment about keeping only the `match` object.

telegram/ext/commandhandler.py
```diff
@@ -142,32 +156,22 @@ def check_update(self, update):
...
+        optional_args = super(CommandHandler, self).collect_optional_args(dispatcher, update)
         if self.pass_args:
+            message = update.message or update.edited_message
             optional_args['args'] = message.text.split()[1:]
+        return optional_args
 
-        return self.callback(dispatcher.bot, update, **optional_args)
+    def collect_additional_context(self, context, update, dispatcher, check_result):
+        message = update.message or update.edited_message
+        context.args = message.text.split()[1:]
```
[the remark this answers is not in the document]
Eldinnie: check_result is either True or False. Not the arg list

telegram/ext/dispatcher.py
```diff
@@ -275,8 +275,10 @@ def process_update(self, update):
 
         for group in self.groups:
             try:
-                for handler in (x for x in self.handlers[group] if x.check_update(update)):
-                    handler.handle_update(update, self)
+                for handler in self.handlers[group]:
+                    check = handler.check_update(update)
+                    if check is not None and check is not False:
```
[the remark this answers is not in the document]
Eldinnie: I agree on this

telegram/ext/dispatcher.py
```diff
@@ -275,8 +275,10 @@ def process_update(self, update):
...
+                        handler.handle_update(update, self, check)
                     break
```
Eldinnie: I think this `break` needs one more indent. It should only break when an update is handled in a group. Not after only checking the first.

telegram/ext/messagehandler.py
```diff
@@ -98,7 +109,8 @@ def __init__(self,
                  pass_chat_data=False,
                  message_updates=True,
                  channel_post_updates=True,
```
Eldinnie: See my note about filters.

telegram/ext/regexhandler.py
```diff
@@ -50,12 +50,16 @@ class RegexHandler(Handler):
             the callback function.
         pass_chat_data (:obj:`bool`): Optional. Determines whether ``chat_data`` will be passed to
             the callback function.
+        use_context (:obj:`bool`): Optional. Determines whether all `pass_` arguments will be
```
Eldinnie: I think we can deprecate regexhandler in total. See my note about filters.

telegram/ext/commandhandler.py
```diff
@@ -142,32 +156,22 @@ def check_update(self, update):
 
                     if not (command[0].lower() in self.command
                             and command[1].lower() == message.bot.username.lower()):
-                        return False
+                        return None
 
                     if self.filters is None:
-                        res = True
+                        return True
```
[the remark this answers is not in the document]
tsnoam: So it won't need to be calculated again (see below)

telegram/ext/commandhandler.py
```diff
@@ -142,32 +156,22 @@ def check_update(self, update):
...
                     elif isinstance(self.filters, list):
-                        res = any(func(message) for func in self.filters)
+                        return any(func(message) for func in self.filters)
```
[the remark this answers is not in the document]
tsnoam: What do you suggest instead?

telegram/ext/commandhandler.py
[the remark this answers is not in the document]
tsnoam: I want it to be the arg list. See above.

telegram/ext/handler.py
```diff
@@ -63,12 +78,25 @@ def __init__(self,
                  pass_update_queue=False,
                  pass_job_queue=False,
                  pass_user_data=False,
-                 pass_chat_data=False):
+                 pass_chat_data=False,
+                 use_context=None):
```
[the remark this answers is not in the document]
tsnoam: So how long do you propose to have this default to old handler semantics (with deprecation warning obviously) and how long do you think we should keep backward compatibility code once we change the default to the new handler semantics?

telegram/ext/handler.py
```diff
@@ -63,12 +78,25 @@ def __init__(self,
...
         self.callback = callback
         self.pass_update_queue = pass_update_queue
         self.pass_job_queue = pass_job_queue
         self.pass_user_data = pass_user_data
         self.pass_chat_data = pass_chat_data
+        self.use_context = use_context
+
+        if use_context is None:
+            warnings.warn('You are currently not using context based handlers. This is being '
```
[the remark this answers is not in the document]
tsnoam: :+1:

telegram/ext/commandhandler.py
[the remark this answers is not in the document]
Eldinnie: It knows what command is, what it needs later on is the args. So if anything it should be `return message.text.split()[1:] but that might be empty...

telegram/ext/commandhandler.py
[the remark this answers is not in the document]
Eldinnie: remove it. We had a deprecation warning for a while (see constructor)

telegram/ext/handler.py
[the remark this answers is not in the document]
Eldinnie: I would say 2 months for both?

telegram/ext/handler.py
```diff
@@ -123,3 +175,55 @@ def collect_optional_args(self, dispatcher, update=None):
...
+    """
+
+    def __init__(self, update, dispatcher):
+        self.update = update
+        self.bot = dispatcher.bot
+
+        self.chat_data = None
+        self.user_data = None
+
+        if update is not None and isinstance(update, Update):
```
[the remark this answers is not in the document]
Eldinnie: yes in case of `StringRegexHandler`, `StringCommandHandler` and `TypeHandler`

telegram/ext/choseninlineresulthandler.py
bomjacob: But... all the other deprecate statements are gone already....

telegram/ext/conversationhandler.py
```diff
@@ -147,8 +147,6 @@ def __init__(self,
 
         self.timeout_jobs = dict()
         self.conversations = dict()
-        self.current_conversation = None
-        self.current_handler = None
```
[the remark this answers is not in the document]
bomjacob: Oh... it would seem I have... whoops?

telegram/ext/handler.py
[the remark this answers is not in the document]
bomjacob: So you don't wanna give people a way to turn the warning off then?

telegram/ext/choseninlineresulthandler.py
tsnoam: What does it matter? The code is there? We remove it **now**? We need to make sure its in the release notes. Even if we didn't intend to have the code there.

telegram/ext/commandhandler.py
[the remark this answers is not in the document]
tsnoam: I see your point.
If we're going for CommandHandler handling only valid commands, this becomes irrelevant?

telegram/ext/conversationhandler.py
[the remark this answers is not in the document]
tsnoam: I say it's a win. It's more of a thumbs up :-)

telegram/ext/handler.py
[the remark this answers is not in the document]
tsnoam: Maybe: Next major version (so if this is going to be v11, then on v12 we change the default to True, and on v13 or v14 we'll remove the backward compatibility code)

telegram/ext/handler.py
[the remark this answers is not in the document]
tsnoam: exactly

telegram/ext/regexhandler.py
tsnoam: I suggest the RegexHandler be a convenience wrapper over MessageHandler.

tsnoam:
Replying to @Eldinnie review:
1,1. I agree for CommandHandler to be only for MessagEntity.COMMAND.
Can be in the next PR, but must be before release.

1.2. MessageHandler for everything else. PrefixHandler should be a convenience wrapper on top of it.
Can be in the next PR, but must be before release.

1.3. What is wrong with list of filters? I do not recall that conversation.

2.1.  *_updates should still exist in MessageHandler.__init__() prototype. The implementation detail of using filters is a very good idea.
Can be in the next PR, doesn't have to be before release.

2.2. I would keep RegexHandler, but change it to be a convinient wrapper on top of MessageHandler.

telegram/ext/handler.py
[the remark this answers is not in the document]
Eldinnie: agreed

Eldinnie:
1.3 Long time ago the filters argument was supposed to be a list of callables and the handler woud handle the update if any of those would return True In october 2016 the bitwise filters have been added #411 and we've supported both (list of filters and bitwise filters) since then. there's a warning list is deprecated since then: https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/messagehandler.py#L123
I say we remove it now

tsnoam:
1.3. Ok. Removing the filters list is a good idea. But I think that we should have in the migration guide a code snippet demonstrating how to convert a list of filters to bitwise-or filters

JosXa:
Could we please gather up the individual user handlers that do not fit the new callback signature and show **a single** deprecation warning  message to the user in the following fashion:

```
"TelegramDeprecationWarning: You are currently not using context based handlers. This is being deprecated, and will soon not be possible. Please change your callback function signatures from (bot, update, others...) to (update, context), and create your handlers using use_context=True instead of setting any pass_ parameters. You can set use_context=False to turn this warning off. Please see https://git.io/vpVe8 for more info.

Your following callables are affected by this:
- MessageHandler(whatever_the_user_set_up, pass_chat_data=True)
- MessageHandler(whatever_the_user_set_up, pass_chat_data=True)
- MessageHandler(whatever_the_user_set_up, pass_chat_data=True)
- MessageHandler(whatever_the_user_set_up, pass_chat_data=True)
- MessageHandler(whatever_the_user_set_up, pass_chat_data=True)
- MessageHandler(whatever_the_user_set_up, pass_chat_data=True)
- MessageHandler(whatever_the_user_set_up, pass_chat_data=True)
- MessageHandler(whatever_the_user_set_up, pass_chat_data=True)
- ...
"
```

Might be possible to do this with a little timeout that checks a list of wrongdoings after like 2 seconds or so after the initialization step. Reasoning behind this is that I'm currently being flooded with ~140 warning messages in one of my bigger bots, and we don't want to make a bad impression towards our users with this change but the opposite :)

telegram/ext/choseninlineresulthandler.py
bomjacob: But the original removal isn't even in the release notes... but alright, it's there now.

telegram/ext/handler.py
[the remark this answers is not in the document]
bomjacob: Sounds good to me

tsnoam:
@JosXa Lets first see what the actual strings will look like.
As stated on the developers group, I want them to look something like:
<object_name>  is deprecated. See http://...

telegram/ext/handler.py
```diff
@@ -138,26 +136,30 @@ def handle_update(self, update, dispatcher, check_result):
...
-            check_result: The result from check_update
+            dispatcher (:class:`telegram.ext.Dispatcher`): The calling dispatcher.
+            check_result: The result (return value) from :attr:`check_update`.
+
         """
         pass
 
     def collect_optional_args(self, dispatcher, update=None, check_result=None):
-        """Prepares the optional arguments. If the handler has additional optional args,
+        """
```
Eldinnie: no whitespace here

tsnoam:
I pushed fixes for documentation and prettier warning messages.

We still need to complete the wiki article at https://github.com/python-telegram-bot/python-telegram-bot/wiki/Migrating-to-context-based-handlers

JosXa:
We should think about adding `__slots__` to the `HandlerContext` in order to save on memory.
More info: https://stackoverflow.com/questions/472000/usage-of-slots?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

bomjacob:
Okay so I know this is coming quite late in the process. But I'm gonna suggest that we move the `use_context` onto the `Dispatcher` itself.
It was okay (though slightly annoying for an upgrading user) to have it on all the Handlers. But now we have it on the `add_error_handler` too, which makes for confusing semantics like
``` python
dp.add_handler(CommandHandler('test', cb, use_context=True))
dp.add_error_handler(error_cb, use_context=True))
```
In one case it's the handler that has the `use_context`, in the other it's the dispatcher method.

This gets complicated even further with `JobQueue` where we'd have confusing stuff like
``` python
job_queue.run_once(cb, 1, context=something, use_context=True)
```
_(note that those two `context`s refer to two completely seperate things.)_
_This also turned out to be really annoying to actually implement_

Therefore I think it would be a lot cleaner to just have
``` python
updater = Updater('TOKEN', use_context=True)
```
_(which is then passed to the `Dispatcher` & `JobQueue`)_
And then all callbacks spawned from that, will use the new API.
This makes it much cleaner during our transistion period IMO, since a user doesn't have to, come version 12, remove a bunch of `use_context=True` from all their Handlers, add_error_handler, and JobQueue calls.

Eldinnie:
@bomjacob I like it.

examples/paymentbot.py
```diff
@@ -70,17 +71,17 @@ def start_without_shipping_callback(bot, update):
...
 
-def shipping_callback(bot, update):
+def shipping_callback(update, context):
     query = update.shipping_query
     # check the payload, is this from your bot?
     if query.invoice_payload != 'Custom-Payload':
         # answer False pre_checkout_query
-        bot.answer_shipping_query(shipping_query_id=query.id, ok=False,
-                                  error_message="Something went wrong...")
+        query.answer(shipping_query_id=query.id, ok=False,
```
nmlorg: Remove `shipping_query_id=query.id, `.

examples/paymentbot.py
```diff
@@ -89,31 +90,33 @@ def shipping_callback(bot, update):
         # an array of LabeledPrice objects
         price_list = [LabeledPrice('B1', 150), LabeledPrice('B2', 200)]
         options.append(ShippingOption('2', 'Shipping Option B', price_list))
-        bot.answer_shipping_query(shipping_query_id=query.id, ok=True,
-                                  shipping_options=options)
+        query.answer(shipping_query_id=query.id, ok=True,
```
nmlorg: Remove `shipping_query_id=query.id, `.

examples/paymentbot.py
```diff
@@ -89,31 +90,33 @@ def shipping_callback(bot, update):
...
 # after (optional) shipping, it's the pre-checkout
-def precheckout_callback(bot, update):
+def precheckout_callback(update, context):
     query = update.pre_checkout_query
     # check the payload, is this from your bot?
     if query.invoice_payload != 'Custom-Payload':
         # answer False pre_checkout_query
-        bot.answer_pre_checkout_query(pre_checkout_query_id=query.id, ok=False,
-                                      error_message="Something went wrong...")
+        query.answer(pre_checkout_query_id=query.id, ok=False,
```
nmlorg: Remove `pre_checkout_query_id=query.id, `.

examples/paymentbot.py
```diff
@@ -89,31 +90,33 @@ def shipping_callback(bot, update):
...
+                     error_message="Something went wrong...")
     else:
-        bot.answer_pre_checkout_query(pre_checkout_query_id=query.id, ok=True)
+        query.answer(pre_checkout_query_id=query.id, ok=True)
```
nmlorg: Remove `pre_checkout_query_id=query.id, `.

telegram/ext/messagehandler.py
```diff
@@ -142,28 +147,10 @@ def check_update(self, update):
...
-
-        Args:
-            update (:class:`telegram.Update`): Incoming telegram update.
-            dispatcher (:class:`telegram.ext.Dispatcher`): Dispatcher that originated the Update.
-
-        """
-        optional_args = self.collect_optional_args(dispatcher, update)
-
-        return self.callback(dispatcher.bot, update, **optional_args)
+                if self.filters is None:
```
nmlorg: self.filters will never be None if it evaluated to True on line 149.

telegram/ext/jobqueue.py
```diff
@@ -37,16 +41,23 @@ class JobQueue(object):
...
+            warnings.warn("Passing bot to jobqueue is deprecated. Please use set_dispatcher "
+                          "instead!", TelegramDeprecationWarning, stacklevel=2)
+
+            class MockDispatcher(object):
+                def __init__(self):
+                    self.bot = bot
+                    self.use_context = False
+
+            self._dispatcher = MockDispatcher()
+        self._dispatcher = None
```
nmlorg: This clobbers lines 50-59, maybe reorder or:
```python
  if bot:
   ...
      self._dispatcher = MockDispatcher()
! else:
!     self._dispatcher = None
```

bomjacob:
Thanks for the review @nmlorg :D You're totally right in all the comments :)

PR #1104, opened 2018-05-05
Channel filter

robert-cody:
Added channel filter for convenience.


robert-cody:
Thanks for your reply, i'll add required tests soon.

jsmnbom:
@robert-cody I'm gonna close this issue due to inactivity.

As far as I can see, similar functionality is also implemented #1221 (albeit slightly different).

PR #1106, opened 2018-05-08
Add 3.7-dev to travis

Eldinnie:
It might be good to start testing on 3.7 too, as the expected release draws nearer


Eldinnie:
See https://github.com/yaml/pyyaml/issues/126

Eldinnie:
pyyaml is only used by pre-commit. So it shouldn't affect installing/running ptb on 3.7 (and it's testing good :D)

PR #1114, opened 2018-05-21
rework CommandHandler

Eldinnie:
* Only register valid botcommands, else raise ValueError

* Make CommandHandler strict
Only register valid botcommands, else raise ValueError

* Add PrefixHandler


telegram/ext/commandhandler.py
```diff
@@ -119,6 +125,10 @@ def __init__(self,
             self.command = [command.lower()]
         else:
             self.command = [x.lower() for x in command]
+        for comm in self.command:
+            if not re.match(r'^[\da-z_]{1,32}$', comm):
```
bomjacob: I'm not sure that 32 chars is a hard limit, dispite the docs... Do we care?

Both my phone and my desktop app, highlights commands that are longer than 32 chars at least.

telegram/ext/commandhandler.py
```diff
@@ -135,21 +145,21 @@ def check_update(self, update):
...
-            message = update.message or update.edited_message
+            message = update.effective_message
 
-            if message.text and message.text.startswith('/') and len(message.text) > 1:
-                first_word = message.text_html.split(None, 1)[0]
-                if len(first_word) > 1 and first_word.startswith('/'):
-                    command = first_word[1:].split('@')
-                    command.append(
-                        message.bot.username)  # in case the command was sent without a username
+            if message.entities and message.entities[0].type == MessageEntity.BOT_COMMAND and \
```
bomjacob: Really not a fan of `\` in code.. could we crap it in parenthesis instead?

telegram/ext/commandhandler.py
```diff
@@ -135,21 +145,21 @@ def check_update(self, update):
...
+                    message.entities[0].offset == 0:
+                command = message.text[1:message.entities[0].length]
```
bomjacob: I know this technically works because command can only be ascii, but maybe using the parse entity method would be a good idea here

telegram/ext/commandhandler.py
```diff
@@ -135,21 +145,21 @@ def check_update(self, update):
...
+                args = message.text.split()[1:]
```
bomjacob: Wouldn't it be more correct to do `message.text[len(command):].split()` ?

telegram/ext/commandhandler.py
```diff
@@ -159,3 +169,147 @@ def collect_optional_args(self, dispatcher, update=None, check_result=None):
 
     def collect_additional_context(self, context, update, dispatcher, check_result):
         context.args = check_result
+
+
+class PrefixHandler(CommandHandler):
+    """Handler class to handle custom prefix commands
+
+    This is a intermediate handler between :class:`MessageHandler` and :class:`CommandHandler`.
+    It supports configurable commands with the same options as commandhandler. It will respond to
```
bomjacob: CommandHandler*

telegram/ext/commandhandler.py
```diff
@@ -159,3 +169,147 @@ def collect_optional_args(self, dispatcher, update=None, check_result=None):
...
+    Attributes:
+        prefix (:obj:`str` | List[:obj:`str`]): The prefix(es) that will precede :attr:`command`.
+        command (:obj:`str` | List[:obj:`str`]): The command or list of commands this handler
+            should listen for.
+        callback (:obj:`callable`): The callback function for this handler.
+        filters (:class:`telegram.ext.BaseFilter`): Optional. Only allow updates with these
+            Filters.
+        allow_edited (:obj:`bool`): Determines Whether the handler should also accept
+            edited messages.
+        pass_args (:obj:`bool`): Determines whether the handler should be passed
```
bomjacob: Is there really a need to have the pass_ stuff on this new handler?

telegram/ext/commandhandler.py
```diff
@@ -159,3 +169,147 @@ def collect_optional_args(self, dispatcher, update=None, check_result=None):
 
     def collect_additional_context(self, context, update, dispatcher, check_result):
         context.args = check_result
+
+
+class PrefixHandler(CommandHandler):
```
bomjacob: Should we add a docstring note about arguments being availible on the CallbackContext object? (applies to CommandHandler too I suppose)

telegram/ext/commandhandler.py
```diff
@@ -159,3 +169,147 @@ def collect_optional_args(self, dispatcher, update=None, check_result=None):
...
+            '#test'.
+
+        Miltiple prefixes and commands:
+
+            PrefixHandler(['!', '#'], ['test', 'help`], callback) will respond to '!test',
+            '#test', '!help' and '#help'.
+
+    Attributes:
+        prefix (:obj:`str` | List[:obj:`str`]): The prefix(es) that will precede :attr:`command`.
+        command (:obj:`str` | List[:obj:`str`]): The command or list of commands this handler
```
bomjacob: Missing a full stop (.)

telegram/ext/commandhandler.py
```diff
@@ -159,3 +169,147 @@ def collect_optional_args(self, dispatcher, update=None, check_result=None):
...
+        pass_job_queue (:obj:`bool`): Determines whether ``job_queue`` will be passed to
+            the callback function.
+        pass_user_data (:obj:`bool`): Determines whether ``user_data`` will be passed to
+            the callback function.
+        pass_chat_data (:obj:`bool`): Determines whether ``chat_data`` will be passed to
+            the callback function.
+
+    Note:
+        :attr:`pass_user_data` and :attr:`pass_chat_data` determine whether a ``dict`` you
+        can use to keep any data in will be sent to the :attr:`callback` function.. Related to
```
bomjacob: .. -> .

telegram/ext/inlinequeryhandler.py
```diff
@@ -103,7 +103,8 @@ def __init__(self,
                  pass_groups=False,
                  pass_groupdict=False,
                  pass_user_data=False,
-                 pass_chat_data=False):
+                 pass_chat_data=False,
+                 use_context=False):
```
bomjacob: This isn't right ...

telegram/ext/commandhandler.py
Eldinnie: image
image
There's a discrepancy here. @botfather won't let you add too long commands for suggestion, but they are recognised as entities. 

I vote to stay with the docs and enforce 32 chars

telegram/ext/commandhandler.py
Eldinnie: I think it's easiest and safest this way.
If we used `parse_entities` we would have to search the dict for either message.text.split()[0] or check every entity for it's offset. Dicts are not ordered on <py3.5.
Since as you pointed out the entity is ascii and we know for sure it should be the first one in `message.entites` I think this is the best way.

telegram/ext/commandhandler.py
Eldinnie: This is the way we always calculated args, And the result should be the same I guess, except for an extra `len()` call in your proposal.

telegram/ext/commandhandler.py
Eldinnie: hmm, In my mind It would be best to keep all handlers compatible to the same model for v11.0 You disagree?

telegram/ext/commandhandler.py
Eldinnie: ? On the next line at the end of the sentence...

telegram/ext/commandhandler.py
Eldinnie: I'll fix this in every handler then :D

telegram/ext/inlinequeryhandler.py
Eldinnie: leftover from my base. Fixed

telegram/ext/commandhandler.py
bomjacob: Depending on the internals of .split() and telegrams entity calculations, but yeah, you're probably right ^^

telegram/ext/commandhandler.py
bomjacob: Kinda, but mostly from a "we shouldn't be writing code that's already deprecated" standpoint, but since you've already written it. I think it's fine :D

telegram/ext/commandhandler.py
bomjacob: Oh derp, sorry :P

telegram/ext/commandhandler.py
```diff
@@ -29,7 +29,9 @@ class CommandHandler(Handler):
     """Handler class to handle Telegram commands.
 
     Commands are Telegram messages that start with ``/``, optionally followed by an ``@`` and the
-    bot's name and/or some additional text.
+    bot's name and/or some additional text. It will add a ``list`` to the :class:`CallbackContext`
```
bomjacob: "It" doesn't really make sense here (it works fine in the prefixhandler)

PR #1117, opened 2018-05-22
Update Filters, CommandHandler and MessageHandler

Eldinnie:
* update_filter attribute on filters
* add update_type filter
* Messagehandler rework
  - remove allow_edited (deprecated for a while)
  - set deprecated defaults to None
  - Raise deprecation warning when they're used
  - add sensible defaults for filters.
  - rework tests
* Commandhandler rework
  - deprecate allow_edited
  - Raise deprecation warning when used
  - default to allowing edited

### DIFF
--- a/telegram/ext/commandhandler.py
+++ b/telegram/ext/commandhandler.py
@@ -38,7 +38,7 @@ class CommandHandler(Handler):
     which is the text following the command split on single or consecutive whitespace characters.
 
     By default the handler listens to messages as well as edited messages. To change this behavior
-    use ``~Filters.update_type.edited_message``.
+    use ``~Filters.updates.edited_message`` in the filter argument.
 
     Attributes:
         command (:obj:`str` | List[:obj:`str`]): The command or list of commands this handler
@@ -88,7 +88,7 @@ class CommandHandler(Handler):
         allow_edited (:obj:`bool`, optional): Determines whether the handler should also accept
             edited messages. Default is ``False``.
             DEPRECATED: Edited is allowed by default. To change this behavior use
-            ``~Filters.update_type.edited_message``.
+            ``~Filters.updates.edited_message``.
         pass_args (:obj:`bool`, optional): Determines whether the handler should be passed the
             arguments passed to the command as a keyword argument called ``args``. It will contain
             a list of strings, which is the text following the command split on single or
@@ -141,16 +141,16 @@ class CommandHandler(Handler):
                 raise ValueError('Command is not a valid bot command')
 
         if filters:
-            self.filters = Filters.update_type.messages & filters
+            self.filters = Filters.updates.messages & filters
         else:
-            self.filters = Filters.update_type.messages
+            self.filters = Filters.updates.messages
 
         if allow_edited is not None:
             warnings.warn('allow_edited is deprecated. See https://git.io/vp113 for more info',
                           TelegramDeprecationWarning,
                           stacklevel=2)
             if not allow_edited:
-                self.filters &= ~Filters.update_type.edited_message
+                self.filters &= ~Filters.updates.edited_message
         self.pass_args = pass_args
 
     def check_update(self, update):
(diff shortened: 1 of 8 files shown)


telegram/ext/commandhandler.py
```diff
@@ -131,8 +140,17 @@ def __init__(self,
             if not re.match(r'^[\da-z_]{1,32}$', comm):
                 raise ValueError('Command is not a valid bot command')
 
-        self.filters = filters
-        self.allow_edited = allow_edited
+        if filters:
```
bomjacob: Should we test that self.filters is a subclass of BaseFilter? Otherwise it will give an obscure crash when we try to & it.
Also, wouldn't this actually break simple function filters? Or have we not supported those for a while?

telegram/ext/commandhandler.py
```diff
@@ -274,15 +290,14 @@ def __init__(self,
...
-                 allow_edited=False,
                  pass_args=False,
                  pass_update_queue=False,
                  pass_job_queue=False,
                  pass_user_data=False,
                  pass_chat_data=False):
 
         super(PrefixHandler, self).__init__(
-            'nocommand', callback, filters=filters, allow_edited=allow_edited, pass_args=pass_args,
+            'nocommand', callback, filters=filters, allow_edited=None, pass_args=pass_args,
```
bomjacob: Could we change 'nocommand' to None, to make it more clear that it's simply not used?
This of course requires changing a few things in the CommandHandler

telegram/ext/filters.py
```diff
@@ -701,3 +712,77 @@ def __init__(self, lang):
...
+
+            def filter(self, update):
+                return update.channel_post is not None or update.edited_channel_post is not None
+
+        channel_posts = _ChannelPosts()
+
+        def filter(self, update):
+            return self.messages(update) or self.channel_posts(update)
+
+    update_type = _UpdateType()
```
bomjacob: Should we name it `updates` instead? It's simply shorter and in my opinion as descriptive

telegram/ext/commandhandler.py
```diff
@@ -131,8 +140,17 @@ def __init__(self,
             if not re.match(r'^[\da-z_]{1,32}$', comm):
                 raise ValueError('Command is not a valid bot command')
 
-        self.filters = filters
-        self.allow_edited = allow_edited
+        if filters:
+            self.filters = Filters.update_type.messages & filters
```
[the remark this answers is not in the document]
Eldinnie: self.filters = Non-existent at this point.

telegram/ext/commandhandler.py
Eldinnie: we never checked before, we could introduce the check here but just let them get an error if they make a mistake imo.. function filters are removed in this version anyway

telegram/ext/commandhandler.py
Eldinnie: That would mean changing CommandHandlers logic to accomodate this. This seemed pretty clean to me.

telegram/ext/filters.py
Eldinnie: yes aggreed

Eldinnie:
@bomjacob I changed some things accoring to your review, left some comments on others.
The wiki is already up to date with these changes

telegram/ext/commandhandler.py
```diff
@@ -80,6 +87,8 @@ class CommandHandler(Handler):
             operators (& for and, | for or, ~ for not).
         allow_edited (:obj:`bool`, optional): Determines whether the handler should also accept
             edited messages. Default is ``False``.
+            DEPRECATED: Edited is allowed by default. To change this behavior use
```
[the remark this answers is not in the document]
Eldinnie: Hmm, maybe it;s personal, but I hate bots that don't respond to me fixing my type in a message and it doesn't respond like it should. IMO handling edited message is a sane default.

I don;t really understand what you mean with a `reply_or_edit` method.

telegram/ext/filters.py
```diff
@@ -159,6 +168,7 @@ class _Text(BaseFilter):
         name = 'Filters.text'
 
         def filter(self, message):
+            print('text_filter_filter {}'.format(repr(self)))
```
tsnoam: why do you `print` here?
also nitpicking: `'text_filter_filter {:r}'.format(self)` is more idiomatic

PR #1119, opened 2018-05-25
examples: updated timerbot to work with the latest version of the library

Mindavi:
I've updated the timerbot example to work with the latest version of the library.
I was working on a bot the other day and found out the example did not work correctly any more.
This pull request should fix that problem.
Perhaps it's also a good idea to have some automated testing for the examples to ensure they won't get out of date?


Eldinnie:
Hi @Mindavi ,

Thanks for you contribution. You couldn't really know this, but what you actually did was reversing an example. We already merged #1100 in preperation for the V11 release, including the examples. Unfortunately this means the examples won't work with v10.1. 
You can find working examples for the latest version here:
https://github.com/python-telegram-bot/python-telegram-bot/tree/v10.1.0/examples

Closing this PR because it's actually a rollback. But thanks for noticing and wanting to help.

Pieter

Issue #1125, opened 2018-05-29
test_subprocess_pipe fails with ProcessLookupError


socram8888:
### Steps to reproduce
1. Install a dev setup as specified in CONTRIBUTING.rst

2. Run `make test`

### Expected behaviour
Tests should pass

### Actual behaviour
test_subprocess_pipe fails with ProcessLookupError. The process finishes before the `os.kill` call is made, and thus the exception is being raised.

### Configuration
**Operating System:** Ubuntu 16.04 on Windows 10

**Version of Python, python-telegram-bot & dependencies:** Python 3.5.2, bot on commit 42daf96d20a26609fe7e6f1922be39e7d49e24cb (latest available)

### Logs
````================================================================================================================= FAILURES ==================================================================================================================
____________________________________________________________________________________________________ TestInputFile.test_subprocess_pipe _____________________________________________________________________________________________________

self = <tests.test_inputfile.TestInputFile object at 0x7ff6fce846a0>

    def test_subprocess_pipe(self):
        if sys.platform == 'win32':
            cmd = ['type', self.png]
        else:
            cmd = ['cat', self.png]

        proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, shell=(sys.platform == 'win32'))
        in_file = InputFile({'photo': proc.stdout})

        assert in_file.input_file_content == open(self.png, 'rb').read()
        assert in_file.mimetype == 'image/png'
        assert in_file.filename == 'image.png'

>       proc.kill()

tests/test_inputfile.py:43:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/lib/python3.5/subprocess.py:1787: in kill
    self.send_signal(signal.SIGKILL)
[...]

PR #1126, opened 2018-05-29
Fix ProcessLookupError if process finishes before we kill it

socram8888:
Fix for #1125


tsnoam:
@socram8888 

socram8888:
You're welcome.

I'll look into the other PR as soon as I get some more free time.

Issue #1124, opened 2018-05-29
Add retrying to requests


socram8888:
Right now, the amount of retries per request isn't configurable, and uses urllib3's [default of 3 attempts](http://urllib3.readthedocs.io/en/latest/user-guide.html#retrying-requests). Worse even, the [backoff](https://en.wikipedia.org/wiki/Exponential_backoff) is [disabled by default](http://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.retry.Retry), which means that on network error instead of waiting it will consume all three attempts immediately.

JosXa:
That's some nice insight on why it seems to fail so oftem

socram8888:
There's something that's caught my eye. This library seems to be using a [custom urllib3 version](https://github.com/python-telegram-bot/urllib3/) which just changes [the handling of the connection timeouts](https://github.com/python-telegram-bot/urllib3/commits/master).

I think the reason is that the timeout was handled as a single float, and the connection open time was not measured, but the latest version of urllib3 upstream [does have a way to specify a max timeout for both connection and read](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.timeout)

Removing this vendor-packed urllib3 would be outside the scope of this request for sure, but maybe @tsnoam could give some insight if it's posible? It'll be nice for both reducing the size of the library, and for getting actually updates.

tsnoam:
Hi @socram8888 

Thank you for your time reporting this issue. Your insights are important to us.

I see a few questions on this issue. I'll attempt to answer them fully:

1. The reason for vendoring our forked version of urllib3 is described in #533, and I'm quoting:
> The problem is that urllib3 doesn't honour the `read_timeout` when using a newly created connection. 

It's a tricky problem, almost impossible to recreate in lab which caused us pain pushing the fix to upstream (I don't blame the maintainers either, I would be suspicious too with a change in such a delicate place). We might have needed to ping the maintainers more. But the fact is that there's a bug which we can solve for ptb users and that's what we did.
Since urllib3 is a delicate component, which we found difficult to tune for ptb needs. We are not thrilled to upgrade to a newer release without a good reason.
We follow the (official) releases of urllib3 and so far we haven't seen a specific fix which may be relevant for ptb. If you've spotted one, please let us know (well, I've lied a little - we once cherry-picked a fix for proper socks5h/socks5 support).

2. About retries & backoff:
[...]

Issue #354, opened 2016-07-19
Updater: automatic retries for bot.send* calls


tsnoam:
The main idea is to allow the user to call the `bot.send*()` methods and get QOS of automatic retries. Specific requirements are:
- Default behaviour will be: no retries (semantic backward-compatible).
- User will be able to control amount of retries up to infinite.
- The retries will be performed in async threads in order to prevent blocking the Updater thread and/or hogging threads from the `Dispatcher` thread pool.
- Allow the user to set a callback action after the message had been sent successfully. That is for cases where the user need to send a number of messages in a very specific order.

Note: we currently have a retry capability (free of charge) in the `urllib3` component. This mechanism will need to be disabled.

Eldinnie:
is this ever going to happen or dealt with @tsnoam ?

Eldinnie:
@Eldinnie @tsnoam

Bibo-Joshi:
Is this different from #1124 ? If no, can we close this?

Bibo-Joshi:
A first attempt for this was made in #1127. That PR is no longer pursued, however.

Bibo-Joshi:
As of v20, users can now implement a custom request backend via the `BaseRequest` interface. This allows to implement a customized retry-behavior and since v20 is based on `asyncio`, doing this in a concurrency-friendly way should be rather easy and there is probably no need to set a "done-callback".

I'll hence close this long-standing issue.

PR #1127, opened 2018-05-29
Added retries and exponential backoff to Request

socram8888:
Adds #1124.

I'm not sure if it will be a good idea to add an option to use Retry object as-is, so backoff can be configured, even if it's inside the suposedly private "vendor" package.

### DIFF
--- a/telegram/utils/request.py
+++ b/telegram/utils/request.py
@@ -34,6 +34,7 @@ try:
     import telegram.vendor.ptb_urllib3.urllib3 as urllib3
     import telegram.vendor.ptb_urllib3.urllib3.contrib.appengine as appengine
     from telegram.vendor.ptb_urllib3.urllib3.connection import HTTPConnection
+    from telegram.vendor.ptb_urllib3.urllib3.util.retry import Retry
     from telegram.vendor.ptb_urllib3.urllib3.util.timeout import Timeout
 except ImportError:  # pragma: no cover
     warnings.warn("python-telegram-bot wasn't properly installed. Please refer to README.rst on "
@@ -64,7 +65,7 @@ class Request(object):
             consecutive read operations for a response from the server. None will set an infinite
             timeout. This value is usually overridden by the various ``telegram.Bot`` methods.
             (default: 5.)
-
+        retries (int): Number of retries for each request. Use 0 to disable retrying. (default: 3)
     """
 
     def __init__(self,
@@ -72,7 +73,8 @@ class Request(object):
                  proxy_url=None,
                  urllib3_proxy_kwargs=None,
                  connect_timeout=5.,
-                 read_timeout=5.):
+                 read_timeout=5.,
+                 retries=3):
         if urllib3_proxy_kwargs is None:
             urllib3_proxy_kwargs = dict()
 
@@ -80,6 +82,8 @@ class Request(object):
 
         sockopts = HTTPConnection.default_socket_options + [
             (socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)]
+
+        retries = self._parse_retries(retries)
 
         # TODO: Support other platforms like mac and windows.
         if 'linux' in sys.platform:
@@ -94,6 +98,7 @@ class Request(object):
             cert_reqs='CERT_REQUIRED',
             ca_certs=certifi.where(),
             socket_options=sockopts,
+            retries=retries,
             timeout=urllib3.Timeout(
                 connect=self._connect_timeout, read=read_timeout, total=None))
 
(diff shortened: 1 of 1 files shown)


telegram/utils/request.py
```diff
@@ -64,15 +65,16 @@ class Request(object):
...
     """
 
     def __init__(self,
                  con_pool_size=1,
                  proxy_url=None,
                  urllib3_proxy_kwargs=None,
                  connect_timeout=5.,
-                 read_timeout=5.):
+                 read_timeout=5.,
+                 retries=3):
```
tsnoam: I wonder if we should have one or all of the arguments: `retries`, `backoff_factor` and/or `urllib3_req_kwargs`.
Personally I lean towards `retries` and `urllib3_req_kwargs` for the following advantages:
1. Simple things are visible and easily accessible.
1. It hints that everything else this are advanced options, but still controllable/accessible by the user.

So basically this will allow users understanding the nuances of networking and/or urllib3 use it. Others, will stay away.

However, I'd like to hear the other maintainers take on this.

telegram/utils/request.py
```diff
@@ -224,14 +229,38 @@ def _request_wrapper(self, *args, **kwargs):
...
+
+        Args:
+            retries (:obj:`int`): Retries parameter.
+
+        Returns:
+          A :obj:`telegram.vendor.ptb_urllib3.urllib3.util.retry.Retry` object, or False if
+          retrying is disabled.
+
+        """
+        # TODO: allow configuring the backoff. Ideally we would just let the programmer use the
```
tsnoam: See my comment above about `urllib3_req_args`

telegram/utils/request.py
```diff
@@ -224,14 +229,38 @@ def _request_wrapper(self, *args, **kwargs):
...
+        # Retry class for fine-tuning, but that's inside the "private" urllib3 package.
+
+        if isinstance(retries, int) and retries > 0:
+            return Retry(total=retries, backoff_factor=0.1)
```
tsnoam: That looks like a reasonable default. :+1:

telegram/utils/request.py
```diff
@@ -224,14 +229,38 @@ def _request_wrapper(self, *args, **kwargs):
...
+    def get(self, url, timeout=None, retries=None):
         """Request an URL.
 
         Args:
             url (:obj:`str`): The web location we want to retrieve.
             timeout (:obj:`int` | :obj:`float`): If this value is specified, use it as the read
                 timeout from the server (instead of the one specified during creation of the
                 connection pool).
+            retries (:obj:`int`): If this value is specified, use it as the number of retries for
+                the request (instead of the one specified during creation of the connection pool).
```
tsnoam: Maybe add `urllib3_req_kwargs` here as well? (here and the other functions below)

telegram/utils/request.py
```diff
@@ -279,35 +316,41 @@ def post(self, url, data, timeout=None):
...
         """Retrieve the contents of a file by its URL.
 
         Args:
             url (:obj:`str`): The web location we want to retrieve.
             timeout (:obj:`int` | :obj:`float`): If this value is specified, use it as the read
                 timeout from the server (instead of the one specified during creation of the
                 connection pool).
-
+            retries (:obj:`int`): If this value is specified, use it as the number of retries for
+                the request (instead of the one specified during creation of the connection pool).
```
tsnoam: nitpick: please add back the empty line before the ending `"""`

telegram/utils/request.py
```diff
@@ -279,35 +316,41 @@ def post(self, url, data, timeout=None):
...
 
         Args:
             url (str): The web location we want to retrieve.
             timeout (:obj:`int` | :obj:`float`): If this value is specified, use it as the read
                 timeout from the server (instead of the one specified during creation of the
                 connection pool).
+            retries (:obj:`int`): If this value is specified, use it as the number of retries for
+                the request (instead of the one specified during creation of the connection pool).
 
           filename:
```
tsnoam: ehm. if we touch the docstring here it looks like we can fix it a bit.
would you mind moving `filename` after `url` and fix the alignment for all args to be like `filename`'s?

socram8888:
From the issue:

Regarding the retries, I was thinking about just creating a custom Retry object (with only retry and backoff parameters), or reusing the Retry object [importing it like urllib3 does for ConnectionError](https://github.com/python-telegram-bot/urllib3/blob/67591cdfd3f29f168baa05cc1fcc1fc92c7e9deb/urllib3/connection.py#L25) so it can be imported like `import telegram.utils.request.Retry`.

I'm not sure if allowing passing parameters directly to the urllib3 connetion pool would be a good idea - wouldn't exposing such implementation details prevent in the future from updating to any other HTTP library if needed (such as, say, requests) without breaking changes?

telegram/utils/request.py
bomjacob: retries and urllib3_retry_kwargs sounds good to me for those reasons :)

telegram/utils/request.py
bomjacob: Don't we need to specify a `status_forcelist`? [docs](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry)

tsnoam: 1. What http error codes should be retried in your opinion?
2. I'm not sure I want to retry on http errors. IMHO, It should be a much higher level logic decision, outside the control of urllib3.

tsnoam:
@socram8888 
it's a fine balance between wrapping a library for every of its functionality vs. letting the user interact with it when he's advanced enough.
I lean towards the latter. we can put in the documentation that we don't promise backward compatibility for this in the future. Also, we're already allowing such interaction with the proxy parameters, so I don't see the difference.

Bibo-Joshi:
Hi. It's been a long time :D @socram8888 would you still like to finish this pr? If yes, we'll need to make sure that it covers everything of #354

socram8888:
Hello @Bibo-Joshi. I am no longer using this project so I have no further interest in further working on the PR. Greetings.

Bibo-Joshi:
Thanks for the quick reply! Then I'll go ahead and close this.
Still, it would be great if you kept your branch alive so that a possible future PR can build upon this one :)

PR #1131, opened 2018-06-06
Have added another example of using library

iskhomutov:


iskhomutov:
I'm not a native english speaker, so there may be some mistakes in comments and docstrings.

PR #1144, opened 2018-06-25
Add msg_in filter

icecom-dg:
Filter that allows exact matches with messages in a list.

Can be used to make sure that user replies with one of the custom keyboard buttons.

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -716,7 +716,7 @@ class Filters(object):
 
         def __init__(self, list_):
             self.list_ = list_
-            self.name = 'Filters.in_list({})'.format(repr(self.list_))
+            self.name = 'Filters.in_list({!r})'.format(self.list_)
 
         def filter(self, message):
             return message.text in self.list_
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -711,13 +711,15 @@ class Filters(object):
         Args:
             list_ (List[:obj:`str`]): Which messages to allow through. Only exact matches
                 are allowed.
+            caption (:obj:`bool`): Optional. Should the caption be used instead of text?
+                Default is ``False``.
 
         """
 
         def __init__(self, list_, caption=False):
             self.list_ = list_
             self.caption = caption
-            self.name = 'Filters.msg_in({!r})'.format(self.list_)
+            self.name = 'Filters.msg_in({!r}, caption={!r})'.format(self.list_, self.caption)
 
         def filter(self, message):
             if self.caption:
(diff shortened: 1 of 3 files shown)


jsmnbom:
Thank you a lot for taking the time to write this code, and submitting this PR! :D
I am, however, unsure of the benefit of adding this filter, for these 2 reasons:
- Doesn't provide which keyword triggered the filter to the handler function, which means you need to check it in your handler function afterwards
- Usually one would just use a RegexHandler for this sort of thing. This has the added benefit of actually passing to the handler function which keyword matched.

Any of the devs feel differently?

icecom-dg:
The purpose of this filter is different. It doesn't check if a keyword exists in message, but if the message is an exact match with one of the strings in the list. 

For example if you have a keyboard markup generated from a list:
```
buttons = ['Start', 'Settings', 'Back']
markup = ReplyKeyboardMarkup([[x] for x in buttons])
bot.send_message(chat_id, text, reply_markup=markup)
```
You can use `MessageHandler(Filters.in_list(buttons), callback)` in one of the states of `ConversationHandler` to filter out any replies that don't match one of the buttons and direct them to fallback handler. And you can get the string that triggered that handler from `update.message.text`, as it is an exact match to the one in the list.

jsmnbom:
Alright, you're right. I agree that we need much a filter. I think we should simply name it `in` instead of `in_list` since it should work on everything that the python keyword `in` does.

JosXa:
@jsmnbom It doesn't specify what attribute of an update you're actually comparing though. Shouldn't it be text_in(), caption_in(), callback_query_in() perhaps?

telegram/ext/filters.py
```diff
@@ -701,3 +701,22 @@ def __init__(self, lang):
...
+
+        Args:
+            list_ (List[:obj:`str`]): Which messages to allow through. Only exact matches
+                are allowed.
+
+        """
+
+        def __init__(self, list_):
+            self.list_ = list_
+            self.name = 'Filters.in_list({})'.format(repr(self.list_))
```
tsnoam: Nitpick:
`'{!r}'.format(self._list)` is more elegant than using repr

icecom-dg:
Yes, `in` is a reserved keyword. Maybe `msg_in`?

jsmnbom:
It might be a good idea to go with @JosXa 's suggestion then: `text_in` ?

JosXa:
@icecom-dg I'd like msg_in with parameters to control whether to check message text or caption

icecom-dg:
Review requested. Probably not the best docstring.

Issue #1145, opened 2018-06-26
message decorator is never getting default values from wrapped functions


jonbesga:
I just realized of this after working with the [send_audio](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/bot.py#L396) function and see that the default timeout value, 20, is not getting used. The [message](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/bot.py#L68) decorator doesn't know anything about the parameters of the wrapped function and its default values. So the timeout argument and other default values are None unless is explicitly passed.

This is a problem, for example, with the [send_audio](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/bot.py#L396) because big files take more than 5 seconds, and without the `timeout` argument passed to the `self._request.post` function it raise a NetworkError exception.

I submitted #1146 with the fixes

jsmnbom:
Per discussion in #1146, the best fix for this is probably gonna be to remove the `@message` decorator in favour of calling method that has the same functionality (except that it properly passes the default args of course).

y2kbug-hk:
It seems I have found the same issue. Even though I have set a 300sec timeout, the request get timeout easily within a short time.

jsmnbom:
Fixed in #1156 .

PR #1146, opened 2018-06-26
Fixed message decorator to use default arguments of wrapped functions

jonbesga:


Eldinnie:
Hi,

I'm not sure I understand the problem in the issue you made, but as far as I understand you're thinking that if you pass: `bot.send_audio(chat_id, audio, timeout=20)`. The timeout will be ignored by the bot class because of the message decorator right?

This concern is unwarrented though.
If you look at this line in de `@message` decorator in the bot class:
https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/bot.py#L85
You see it take's the timeout from the kwargs.

jonbesga:
Hi,

The concern is that if you don't pass the argument timeout to `bot._send_audio`. There is no default timeout.

Eldinnie:
Well as you can see on this line. `send_audio` (I think you meant that) has a defauklt timeout of 20

https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/bot.py#L396

jonbesga:
Exactly. And that value is not passed to the decorator.

jsmnbom:
You would indeed be correct in that default timeout are not getting applied at all. And I gotta give you probs for taking the initiative to write a pr.
However, I'm not a fan of how this has been fixed - using the inspect module makes code much harder to read and comprehend (and correct me if I'm wrong, but its performance is pretty bad too, right?).
Therefore I think a better fix is gonna be to get rid of the message decorator, and instead turn it into a method and call it explicitly at the bottom of send_* etc.
@jonbesga Would you like to write this fix, or should we quickly fix it ourselves? :)

(@tsnoam is totally gonna have a field day with this xD)

jsmnbom:
We'll also need a test to make sure something like this doesn't happen again in the future. I adapted one quickly:
``` python
    def test_timeout_propagation_implicit(self, monkeypatch, bot, chat_id):
        class OkException(Exception):
            pass

        def post(*args, **kwargs):
            print(kwargs)
            if kwargs.get('timeout') == 20:
                raise OkException

        monkeypatch.setattr('telegram.utils.request.Request.post', post)

        with pytest.raises(OkException):
            bot.send_photo(chat_id, open('tests/data/telegram.jpg', 'rb'))
```

jsmnbom:
Alright I will close this PR then and add some info to #1145.

PR #1152, opened 2018-07-07
Add 2.7 EOL Shield

JosXa:
More of a practical joke than a real suggestion :D but if you like it we can keep it


jsmnbom:
Considering this was meant as a joke, and it could be potentially confusing to users (especially since it doesn't link anywhere useful), I'm gonna close this PR)

Issue #405, opened 2016-09-04
Cannot use a ConversationHandler as part of another ConversationHandler


liiight:
### Steps to reproduce
1. Define a ConversationHandler where one if its states corresponds to another ConversationHandler
2. Trigger the nested ConversationHandler
3. End the nested ConversationHandler using the `Conversation.END` state.
### Expected behaviour

There should be an option to manipulate the "external" conversation state, as possible with other Handlers.
### Actual behaviour

Since the last return value of the nested conversation is END, there is no option to alter the external conversation state, and it's "stuck" in that state.

Linked to #385

jh0ker:
I think this (with #385 in mind) requires a small but backwards-compability breaking adjustment to the Handler subsystem , so I'll put this up for milestone 6.0

IMO we should get #373 resolved, release 5.1 and then start working on this.

jh0ker:
Thoughts:

When you want to exit the current `ConversationHandler` and also pass on a new state for the parent handler, return a `list`: 

- `return [ConversationHandler.END, FOO]` 
- or to exit two stages: `return [ConversationHandler.END, ConversationHandler.END]` 

In [`ConversationHandler.handle_update`](https://github.com/python-telegram-bot/python-telegram-bot/blob/a91fe5f8f6058abdb12eb8ebaf47ad860313bd17/telegram/ext/conversationhandler.py#L208), we could then test if the return type is `list`, and if the first item in that list is `END`. If so, slice the first element off and return the remainder of the list, which will then be interpreted by the parent `ConversationHandler`. If only one item is left in the list, return that.

Another idea: `ConversationHandler.END` is defined as `-1`. We define another exit condition. If the return value is a `tuple` (named `t`) with `len(t) == 2` and the first element is a negative number, the `ConversationHandler` ends the conversation and does 

```python
if t[0] < -1:
    return (t[0] + 1, t[1])
else:
    return t[1]
```

That way, we don't have to build big lists that could be messed up.

Let me know what you think!

jh0ker:
Alright this issue is pretty old, and we have decided to close it because 

1. Any solution to allow for nested conversation handlers is incredibly convoluted
2. From a computer science perspective, nested state machines are the same as state machines

PR #1153, opened 2018-07-08
Simple support for nesting ConversationHandlers

syntx:
Allowed **ConversationHandler** to possibly return a value from its **handle_update** method

Useful for allowing **ConversationHandler** to be nested in another **ConversationHandler** (see #405).

This solution was inspired by [transitions](https://github.com/pytransitions/transitions) library's [implementation of Hierarchical State Machines (HSM)](https://github.com/pytransitions/transitions#reuse-of-previously-created-hsms) (see the 'remap' keyword).

With this change it's now possible to return arbitrary state names in the nested **ConversationHandler** as long as the state name is mapped to an external **ConversationHandler** state (e.g. you can have your callback functions return SUCCESS, FAILURE, CANCELED etc. as you wish. You can also map ConversationHandler.END on one or both ends).

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -63,6 +63,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Shelomentsev D <https://github.com/shelomentsevd>`_
 - `Simon Schürrle <https://github.com/SitiSchu>`_
 - `sooyhwang <https://github.com/sooyhwang>`_
+- `syntx <https://github.com/syntx>`_
 - `thodnev <https://github.com/thodnev>`_
 - `Trainer Jono <https://github.com/Tr-Jono>`_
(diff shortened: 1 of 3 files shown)


Eldinnie:
Hi @syntx 

I'm not really seeing how this structure would go down. It's also missing tests. Can you please add tests for expected and failing bahavior? They will be needed for the PR anyway and it might help me understand what this is supposed to do.

Pieter

jh0ker:
Just wanna say, nice solution a problem that came up a couple times already. I never thought it was worth it implementing something cause there didn't seem to be a neat solution, but i think this is good. Even allows you to go up the conversation handler tree arbitrarily high, which is even better. If we have a bit more documentation and an example for this, I'd approve this PR. Oh and there's a couple `flake8` errors in the tests.

PR #1154, opened 2018-07-08
Add Conflict error (HTTP error code 409)

### LINKED ISSUE #1143: Show conflicting bot id in error when bot token is used multiple times
Tell which of the bots causes conflict:
```Telegram NetworkError : Conflict: terminated by other getUpdates request; make sure that only one bot instance is running (409)```
This is essential when you run more than one bot in an app, how to know what bot token is used twice.

### Steps to reproduce
1. Run 2 (or more) bot with the same token in different python-telegram-bot app instances
2. Shortly you get the message: ```Telegram NetworkError : Conflict: terminated by other getUpdates request; make sure that only one bot instance is running  (409)```

### Expected behaviour
Add bot id to this message:
```Telegram NetworkError : Conflict: terminated by other getUpdates request; make sure that only one bot instance is running **(bot: 12345456)** (409)```

### Actual behaviour
```Telegram NetworkError : Conflict: terminated by other getUpdates request; make sure that only one bot instance is running (409)```

### PR DESCRIPTION
**Pull request by jsmnbom**
Fixes #1143.

Do we need tests for stuff like this?

### CODE REVIEW

**Comment by Eldinnie** [comment]
hmm, Not to be blunt but why should we patch a user_error? I think the case in #1143 is too specific to add this feature..

**Comment by jsmnbom** [comment]
The way I see it, is that telegram is returning a custom http error code, which we should handle on par with how we handle other 4xx error codes. Adding the conflicting id is just a quality of life type of improvement in my opinion. Doesn't really harm anything, but will probably make it easier to debug issues for some people.

File: telegram/error.py [hunk 1]
```diff
@@ -100,3 +99,12 @@ def __init__(self, retry_after):
         super(RetryAfter,
               self).__init__('Flood control exceeded. Retry in {} seconds'.format(retry_after))
         self.retry_after = float(retry_after)
+
+
+class Conflict(TelegramError):
+    def __init__(self, msg, url):
+        match = re.search(r'bot(\d+):.*/', url)
```
tsnoam: Can you please add a comment on the code itself  what is the input string and why it means a conflicting bot id? (in favour of future us who won't remember all the details...)

Eldinnie:
What's the status on this? We discussed it a few times. I seem to remember we decided not to add this except for the exception?

tsnoam:
> What's the status on this? We discussed it a few times. I seem to remember we decided not to add this except for the exception?

Yes, exactly. @jsmnbom  was on it. I thought it was already merged.

jsmnbom:
Yes, I'm on it, will update the PR tomorrow when I have time :)

PR #1156, opened 2018-07-09
Remove message decorator to fix default timeouts.

### LINKED ISSUE #1145: message decorator is never getting default values from wrapped functions

jsmnbom:
Closes #1145.


Eldinnie:
Seems like you missed `Bot.edit_reply_markup`

jsmnbom:
@Eldinnie I'm not sure what you mean... `Bot.edit_reply_markup` doesn't exist? and `Bot.edit_message_reply_markup` seems proper to me?

Eldinnie:
@jsmnbom It gives an error in the logs on ci, that's as far as I looked

jsmnbom:
Also fixes #1180 since that seemed to fit well here :)

jsmnbom:
I need to stop using github to resolve merge conflicts... keep missing stuff. Will fix later :)

Eldinnie:
Any fix on this incomming @jsmnbom ?

jsmnbom:
Will fix tomorrow :)

telegram/bot.py
```diff
@@ -132,6 +106,27 @@ def __init__(self, token, base_url=None, base_file_url=None, request=None, priva
                                                                   password=private_key_password,
                                                                   backend=default_backend())
 
+    def _message(self, url, data, **kwargs):
+        if kwargs.get('reply_to_message_id'):
+            data['reply_to_message_id'] = kwargs.get('reply_to_message_id')
```
Eldinnie: give assignment operators now! :D
But as long as we can't use those. double `get`'s are not needed. if `kwargs.get` returns a value we can use `kwargs['item']` within the iff clause.
I have no idea if it has any change in speed or not :D

telegram/bot.py
```diff
@@ -132,6 +106,27 @@ def __init__(self, token, base_url=None, base_file_url=None, request=None, priva
...
+
+        if kwargs.get('disable_notification'):
+            data['disable_notification'] = kwargs.get('disable_notification')
```
Eldinnie: same as before and same goes for all below :D

telegram/bot.py
```diff
@@ -132,6 +106,27 @@ def __init__(self, token, base_url=None, base_file_url=None, request=None, priva
...
+        if kwargs.get('reply_markup'):
+            reply_markup = kwargs.get('reply_markup')
+            if isinstance(reply_markup, ReplyMarkup):
+                data['reply_markup'] = reply_markup.to_json()
+            else:
+                data['reply_markup'] = reply_markup
+
+        result = self._request.post(url, data, timeout=kwargs.get('timeout'))
+
+        if result is True:
```
Eldinnie: Just to be sure. You checked it's always either `True` or a `Message`?

tests/test_bot.py
```diff
@@ -641,3 +641,16 @@ def post(*args, **kwargs):
 
         with pytest.raises(OkException):
             bot.send_photo(chat_id, open('tests/data/telegram.jpg', 'rb'), timeout=timeout)
+
+    def test_timeout_propagation_implicit(self, monkeypatch, bot, chat_id):
+        class OkException(Exception):
```
Eldinnie: Interesting way to test this (with the exception). I know it was already in, but why don;t we just assert in the monkeypatched method? That should work.

tests/test_callbackquery.py
```diff
@@ -88,18 +88,20 @@ def test(*args, **kwargs):
 
     def test_edit_message_text(self, monkeypatch, callback_query):
         def test(*args, **kwargs):
+            print(args, kwargs)
```
Eldinnie: debug line?

tests/test_callbackquery.py
```diff
@@ -88,18 +88,20 @@ def test(*args, **kwargs):
...
             try:
+                text = args[1] == 'test'
```
Eldinnie: if the line is in both `try` and `except` it should be outside of them.

telegram/bot.py
jh0ker: Wouldn't the most pythonic way be 

```suggestion
        if 'reply_to_message_id' in kwargs:
            data['reply_to_message_id'] = kwargs['reply_to_message_id']
```

telegram/bot.py
jsmnbom: I mean... that exact code worked before... so I would assume?

jsmnbom: Alright, just checked, and all the methods that we use the _message helper in, do indeed return either True or a Message.

jsmnbom:
Gave another lookover, merging with the intention of releasing a V12 beta to squash bugs.

PR #1158, opened 2018-07-11
Fix Filters.regex failing on non-text message

### LINKED ISSUE #1115: Filters.regex fails on non-text messages
### Steps to reproduce
1. given a handler like the following:
`def msg(bot, update):
    answer = random.choice(bot_misc.ANSWERS)
    update.message.reply_text(answer)
updater.dispatcher.add_handler(MessageHandler(Filters.regex('Hi'), msg))`

2. send the bot some kind of media (like a sticker, an image or a voice-message)

3. error (success)

### Expected behaviour
I usually expect a regex filter to imply Filters.text (that the message is text), because you can only regex over text...

### Actual behaviour
the regex filter fails with the following error when it tries to process the update:
`2018-05-21 18:15:43,279 - telegram.ext.dispatcher - ERROR - An uncaught error was raised while processing the update
  File "/home/thorbijoern/env/ptb/lib/python3.6/site-packages/telegram/ext/dispatcher.py", line 278, in process_update
  File "/home/thorbijoern/env/ptb/lib/python3.6/site-packages/telegram/ext/dispatcher.py", line 278, in <genexpr>
  File "/home/thorbijoern/env/ptb/lib/python3.6/site-packages/telegram/ext/messagehandler.py", line 152, in check_update
    res = self.filters(message)
  File "/home/thorbijoern/env/ptb/lib/python3.6/site-packages/telegram/ext/filters.py", line 65, in __call__
    return self.filter(message)
  File "/home/thorbijoern/env/ptb/lib/python3.6/site-packages/telegram/ext/filters.py", line 201, in filter
    return bool(self.pattern.search(message.text))
TypeError: expected string or bytes-like object`

### workaround
a workaround would be just using `(Filters.text & Filters.regex('foo'))`.

icecom-dg:
Fix #1115 and improve `test_filters_regex`

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -198,7 +198,9 @@ class Filters(object):
         # the matched groups and groupdict to the context object.
 
         def filter(self, message):
-            return bool(self.pattern.search(message.text))
+            if message.text:
+                return bool(self.pattern.search(message.text))
+            return False
 
     class _Reply(BaseFilter):
         name = 'Filters.reply'
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -68,6 +68,9 @@ class TestFilters(object):
         message.text = 'i love python'
         assert Filters.regex(r'.\b[lo]{2}ve python')(message)
 
+        message.text = None
+        assert not Filters.regex(r'fail')(message)
+
     def test_filters_reply(self, message):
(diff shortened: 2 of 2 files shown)


alisherafat01:
I need this. when ?

jsmnbom:
Merging. tests will fail until #1168 is merged since we are not compliant with the official api until then.

Issue #1165, opened 2018-07-26
Telegram BotAPI Update July 26


code1mountain:
The Telegram Bot API got a new update. Changelog is here: 
https://core.telegram.org/bots/api-changelog#july-26-2018

**New Features:**

1. Added support for Telegram Passport. See the official announcement on the [blog](https://telegram.org/blog) and the [manual](https://core.telegram.org/passport) for details.

2. Added support for editing the media content of messages: added the method editMessageMedia and new types inputMediaAnimation, [inputMediaAudio]/bots/api(#inputmediaaudio), and inputMediaDocument.

3. Added the field thumb to the Audio object to contain the thumbnail of the album cover to which the music file belongs.

4. Added support for attaching custom thumbnails to uploaded files. For animations, audios, videos and video notes, which are less than 10 MB in size, thumbnails are generated automatically.

5. tg:// URLs now can be used in inline keyboard url buttons and text_link message entities. (No changes needed?)

6. Added the method sendAnimation, which can be used instead of sendDocument to send animations, specifying their duration, width and height. (PR #1172 )

7. Added the field animation to the Message object. For backward compatibility, when this field is set, the document field will be also set. (PR #1172 )

8. Added two new MessageEntity types: cashtag and phone_number.

9. Added support for Foursquare venues: added the new field foursquare_type to the objects Venue, InlineQueryResultVenue and InputVenueMessageContent, and the parameter foursquare_type to the sendVenue method. (PR #1170 )

10. You can now create inline mentions of users, who have pressed your bot's callback buttons. (No changes needed?)

11. You can now use the Retry-After response header to configure the delay after which the Bot API will retry the request after an unsuccessful response from a webhook.

12. If a webhook returns the HTTP error 410 Gone for all requests for more than 23 hours successively, it can be automatically removed.

13. Added vCard support when sharing contacts: added the field vcard to the objects Contact, InlineQueryResultContact, InputContactMessageContent and the method sendContact. (PR #1169 )

This issue can be used as reference for the pull requests.

PR #1166, opened 2018-07-26
Add vCard field to Contact and related

code1mountain:
This adds the vCard field to Contact, InlineQueryResultContact, InputContactMessageContent and bot.send_contact as mentioned in Issue #1165 

Also added the corresponding tests

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -3,8 +3,8 @@ Credits
 
 ``python-telegram-bot`` was originally created by
 `Leandro Toledo <https://github.com/leandrotoledo>`_ and is now maintained by
-`Jannes Höke <https://github.com/jh0ker>`_ (`@jh0ker <https://t.me/jh0ker>`_ on Telegram) and
-`Noam Meltzer <https://github.com/tsnoam>`_.
+`Jannes Höke <https://github.com/jh0ker>`_ (`@jh0ker <https://t.me/jh0ker>`_ on Telegram),
+`Noam Meltzer <https://github.com/tsnoam>`_, `Pieter Schutz <https://github.com/eldinnie>`_ and `Jasmin Bom <https://github.com/jsmnbom>`_.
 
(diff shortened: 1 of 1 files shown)


telegram/bot.py
```diff
@@ -1093,6 +1093,7 @@ def send_contact(self,
                      phone_number=None,
                      first_name=None,
                      last_name=None,
+                     vcard=None,
```
Eldinnie: Please move the argument to the last in the method signature (just before **kwargs).

telegram/inline/inlinequeryresultcontact.py
```diff
@@ -62,6 +66,7 @@ def __init__(self,
                  phone_number,
                  first_name,
                  last_name=None,
+                 vcard=None,
```
Eldinnie: Same as above

telegram/bot.py
```diff
@@ -1141,11 +1144,14 @@ def send_contact(self,
             phone_number = contact.phone_number
             first_name = contact.first_name
             last_name = contact.last_name
+            vcard = contact.vcard
```
Eldinnie: Since vcard is optional I would prefer it if vcard was only added here if it was available in the original contact.

tests/test_contact.py
```diff
@@ -43,20 +44,23 @@ def test_de_json_required(self, bot):
...
         assert contact.user_id == self.user_id
+        assert contact.vcard == self.vcard
 
     def test_send_with_contact(self, monkeypatch, bot, chat_id, contact):
         def test(_, url, data, **kwargs):
             phone = data['phone_number'] == contact.phone_number
             first = data['first_name'] == contact.first_name
             last = data['last_name'] == contact.last_name
-            return phone and first and last
+            vcard = data['vcard'] == contact.vcard
```
Eldinnie: I think this should split int two seperate tests. One with vcard, one without. Also see comment above

Eldinnie:
Left a review. Please adress those issues.

PR #1174, opened 2018-07-29
Passport

code1mountain:
**Work in Progress**

Added Classes:
- PassportData
- PassportFile
- EncryptedPassportElement
- EncryptedCredentials
- PassportElementError*

Added function:
- set_passport_data_errors

Added `passport_data` field to message.

Added helper function to generate an url that can be used to request access to telegram passport.

File decryption is not implemented yet. **Help wanted**.

Added corresponding tests


jsmnbom:
Alright, I've added some basic passport decryption. Mostly just focussed on getting it working for now.
We probably wanna move the actual decryption currently at https://github.com/python-telegram-bot/python-telegram-bot/pull/1174/commits/a09394b218d3ae05dc1a1f74e782c701283fb82b#diff-eaefd62d25b87433b035868713d3437aR39 somewhere better.

We might also want to look into implementing all or some of the classes described at https://core.telegram.org/passport.

I also added an example that simply prints all the passport data and fetches the url of files. We definitely need to improve it though.

jsmnbom:
All in all really good work so far though! :D

jsmnbom:
Alright, I think we're only missing a couple of tests and a bunch of docs now as far as telegram passport support goes.

PR #1176, opened 2018-07-29
Make ConversationHandler and run_async work together properly

### LINKED ISSUE #1175: ConversationHandler and run_async racing condition
In conversation handler we may face racing condition in check_update method:
1) `res = new_state.result(timeout=self.run_async_timeout)` may return None since the result is not ready
2) ` if not error and new_state.done.is_set():` may return **True** since the result may be calculated by this time

After that we have a state (old_state, Promise_object) since `self.update_state(None, key)` will leave previous state and may go by the fallback scenario since state dictionary does not contain this state.

Code lines with bug: https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/conversationhandler.py#L229 - https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/conversationhandler.py#L255

**Actually problem may be even bigger because res may be calculated and may actually be None.**

This bug was originally noticed by Николай during a conversation in official telegram chat.

vasinkd:
fixes #1175


jsmnbom:
@jh0ker Do you have time to check this out sometime? You're the one with the best insight into ConvHandler.

vasinkd:
A description of the code:
1) We control whether a promise is already resolved with Event().wait() and only in case of the timeout, we use timed_out_behavior pattern.
2) If promise returns an error we leave the previous state and not trying to resolve this promise again when we receive a new update
3) If promise returns None state, we replace the state with old state and not trying to resolve this promise again when we receive a new update

I'm going to check whether all the unit tests associated with ConversationHandler pass when I have more free time.

tsnoam:
LGTM.
Lets see how unitests pass after merging latest master.

tsnoam:
Failed unitests has nothing with this change. Will merge now.

Eldinnie:
Can we rerun and see coverage please?

tsnoam:
@Eldinnie pypy will keep failing.
i can rerun for py-3.7 if that will make a difference

tsnoam:
We want code coverage to work and for that we need the unitests passing. 
We'll fix the unitests and then most likely merge.

Eldinnie:
Ok, I looked over the original Issue and the fix. Since we're talking about a race condition (I didn't get that before) testing before/after scenarios is impossible. Hence I;m going to merge now.

Thanks for you contribution!

PR #1191, opened 2018-08-19
Replaces http.server with Tornado

### LINKED ISSUE #1189: Anyone may block your bot
### Steps to reproduce
1. Start bot using webhook

2. On the other machine use the following code:
```
import socket

def client(ip, port, message):
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
        sock.connect((ip, port))
        sock.sendall(bytes(message, 'ascii'))
        response = str(sock.recv(1024), 'ascii')
        print("Received: {}".format(response))
client(BOT_IP, BOT_PORT, "hello")
```

3. Try to call some commands inside bot, press buttons

### Expected behaviour
Bot should keep receiving calls from Telegram

### Actual behaviour
Bot is stuck and won't answer until client from step 2 stops his "attack"

### Configuration
**Operating System:**
Ubuntu 16.04

python 3.6, python-telegram-bot 10.1.0
``$ python -m telegram``

### Logs
Nothing in logs but I managed to localize the problem: 
get_request in BaseServer from socketserver.py got blocked.
Here an inheritance path: socketserver.BaseServer -> socketserver.TCPServer -> http.server.HTTPServer -> telegram.utils.WebhookServer

vasinkd:
fixes #1189 
adds 2 dependencies: tornado and asyncio

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -42,6 +42,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Joscha Götzer <https://github.com/Rostgnom>`_
 - `jossalgon <https://github.com/jossalgon>`_
 - `JRoot3D <https://github.com/JRoot3D>`_
+- `Kirill Vasin <https://github.com/vasinkd>`_
 - `Kjwon15 <https://github.com/kjwon15>`_
 - `Li-aung Yip <https://github.com/LiaungYip>`_
 - `macrojames <https://github.com/macrojames>`_
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,4 @@
 future>=0.16.0
 certifi
+asyncio
+tornado
--- a/telegram/ext/updater.py
+++ b/telegram/ext/updater.py
@@ -21,6 +21,7 @@
 import logging
 import os
 import ssl
+import asyncio
 from threading import Thread, Lock, current_thread, Event
 from time import sleep
 import subprocess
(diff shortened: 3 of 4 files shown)


tsnoam:
Thanks. We'll check this out.

vasinkd:
Sure. Thank you for supporting this great library! BTW, I didn't spend time on fixing of tests since you may still decide to fix this issue with http.server, just made minimal working solution.

vasinkd:
Ok. I guess it is almost final version.

This code can be executed with Python 2.7 and Python 3.4+
It fixes #1189

Now it returns a little different HTTP status codes: 500 (used to be 403) if Content-Length header is not readable and 404 (used to be 200, why?) if someone uses not POST-request or incorrect URL. It may be easily fixed if necessary.

Also, I didn't get what to do with tests aimed at Telegram API 4.0 coverage. Should I comment it out?

tsnoam:
@vasinkd Thanks. I'm on vacation until next week so I will be able to properly review the code only then.

Regarding the API unitests, just ignore them for now.

vasinkd:
@tsnoam Ok. I'll leave the code as it is until you review it.
Enjoy your vacation!

tsnoam:
Sorry for the delay.

1. I like it. IMHO, we should get this solution into the library.
2. We'll need to work on py2.7 compatibility. Specifically I think we should use the Tornado eventloop (instead of asyncio's).
3. I merged latest master and pushed. Lets see which unitests pass and which needs to be fixed.

tsnoam:
@vasinkd 
Thank you very much for this contribution.

vasinkd:
I found out that answer 200 for GET request was there for a reason :)
Actually, get_webhook_info will always return an error with the code above.

vasinkd:
I added a fix to the branch "vasinkd:tornado_webhook".
Should I open a new PR?

tsnoam:
@vasinkd yes please

PR #1200, opened 2018-09-01
feat: @send_action(action: ChatAction) decorator

mitinarseny:
Implemented `@send_action` decorator to decorate callback functions of handlers.

Just simply:
```python
from telegram import (Bot, ChatAction, Update)
@send_action(ChatAction.TYPING)
def start_callback(bot: Bot, update: Update):
    update.message.reply_text('Before this message ChatAction.TYPING was sent.')
```

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -47,6 +47,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `macrojames <https://github.com/macrojames>`_
 - `Michael Elovskikh <https://github.com/wronglink>`_
 - `Mischa Krüger <https://github.com/Makman2>`_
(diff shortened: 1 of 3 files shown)


jsmnbom:
Hi @mitinarseny, thanks a lot for taking the time to send this PR.
Unfortunately we are not currently looking to implement this sort of functionality in the library.
The primary reason for this, is that it really only saves a couple of keypresses for the user, and in turn makes a lot of assumptions about what kinds of functions are decorated with it (only handler callbacks, but not job callbacks etc.)
There's also a few potential problems I see with it:

- What if the decorated function takes more than 5 sec to execute?
- We still support py2.7, so we can't use annotations.

Once again, we really appreciate the effort, and if you can come up with something that addresses these issues, you're very welcome to suggest it :D

PR #1206, opened 2018-09-04
add text mention for message parse

ehsanonline:
Now we can get the mentioned user in text's that are parsed with HTML or Markdown formatting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,9 +63,6 @@ target/
 # Sublime Text 2
 *.sublime*
 
-# Visual Studio Code
-.vscode
-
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -63,6 +63,9 @@ target/
 # Sublime Text 2
 *.sublime*
 
+# Visual Studio Code
```
jsmnbom: Could you remove this from the PR since it isn't really relavant?

Hint: you can add this rule in your local .gitignore

ehsanonline: Sure.

PR #1207, opened 2018-09-05
An attempt to fix the docs page

alexbagirov:
image


nmlorg:
You can move those words back to the previous line, just indent the second line another level.

jsmnbom:
@alexbagirov We'd love you to fix this, but as @nmlorg said, there is a better way to do it :)

PR #1213, opened 2018-09-08
Update data.py to be compatible with example

reablaz:
for now, if you process personal_info with example code, then you got an error if there is no set option to get native fist and last name.

setting default value will allow to process personal_info without native name/surname transation

### DIFF
--- a/telegram/passport/data.py
+++ b/telegram/passport/data.py
@@ -39,7 +39,7 @@ class PersonalDetails(TelegramObject):
     """
 
     def __init__(self, first_name, last_name, birth_date, gender, country_code,
-                 residence_country_code, first_name_native, last_name_native, middle_name=None,
+                 residence_country_code, first_name_native=None, last_name_native=None, middle_name=None,
                  middle_name_native=None, bot=None, **kwargs):
         # Required
         self.first_name = first_name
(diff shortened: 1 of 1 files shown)


jsmnbom:
Thanks a lot for the PR @reablaz. You are indeed correct that the native arguments should be optional.

It seems like you didn't quite set up the development environment as described in https://github.com/python-telegram-bot/python-telegram-bot/blob/master/.github/CONTRIBUTING.rst Most specifically the pre-commit-hook flake8, which is complaining that 
```
telegram/passport/data.py:42:100: E501 line too long (105 > 99 characters)
```

Could we perhaps get you to fix that line? :D

jsmnbom:
You still with us @reablaz ? :D

reablaz:
Hi there! I shortened this line, thank you for your pointing!

reablaz:
failed again.. can anyone teach me how to do things right? :D

jsmnbom:
Actually it seems the build failed for other reasons, so this looks good :D 

Would you like to add yourself to AUTHORS.rst, please do so :) Otherwise just reply and we'll get this merged :D

reablaz:
Hey there!

I think my contribution is too trivial to add myself to authors list. So we can skip this step. Let's merge :+1:

jsmnbom:
Alright, thanks once again :D

Issue #1210, opened 2018-09-07
Typo in comment in conversationbot2.py


freelancervvp:
### Steps to reproduce
1. Not really a bug... wrong comment line in conversationbot2.py :)

### Expected behaviour
Should be: # Add conversation handler with the states CHOOSING, TYPING_CHOICE and TYPING_REPLY

### Actual behaviour
Actual comment in conversationbot2.py (leftover from conversationbot.py I guess :))
# Add conversation handler with the states GENDER, PHOTO, LOCATION and BIO

### Configuration
**Operating System:**
Windows

**Version of Python, python-telegram-bot & dependencies:**
3.6

``$ python -m telegram``

### Logs
Insert logs here (if necessary)

marcelogdeandrade:
I will work on this just to start contributing to the project.

jsmnbom:
Great! :D Nice to have you :)

Just write on the telegram group if you need help setting up the development environment :D

marcelogdeandrade:
Hello, I fixed the requested comment and opened a PR but it seems that the CI test didn't pass, which is really strange because I just changed a simple comment on one file. Could you help me understand what went wrong on the test? Thanks!

PR #1216, opened 2018-09-10
fixes comment on examples/conversationbot2.py

marcelogdeandrade:
fixes issue #1210


jsmnbom:
This looks good to me. The tests are not caused by this PR but some pypy issues we're having :)

Thanks a lot for the PR @marcelogdeandrade, I hope you'll stay with us :D

fixes #1210 
(needs that specific format :)

Issue #1136, opened 2018-06-13
Add conversation timeout handler/  How to know last state?


luciaasen:
### 1.  conversation_ timeout handler
The Conversation Handler provides the possibility to add a timeout to conversations, so that conversations end after been inactive for a period of time. 

When that timeout takes place, I would like to be notified so that I can tell the user that its conversation has ended, but I can't find a way to do it! 

The timed_out_behaviour attribute handles run_async decorated handlers timeout, but there's no way of configuring behaviour after conversation_timeout

### 2. Converstion Handler: return to previous state
During any state of a conversation, I would like the user to be able to type /help command, give him some info and then return to the conversation's previous state.

Let's say my states are A, B
If I add a HELP state, I can go from A to HELP and from B to HELP. But, from the HELP handler, how do I know whether to return A or B?

I added the help command handler as a fallback which always returns conversation.END, but it is  annoying to end the conversation just because you wanted some help.
Also, I could check at the beginning of each state handler if the user has typed /help, but I hope there is a more elegant way to do it

jsmnbom:
### 1
Monkeypatching ConversationHandler._trigger_timeout, should theoretically work, but is of course not really recommended.
Pinging @python-telegram-bot/developers: should we provide a proper way to notify the user when a conversation handler timeout occurs?  
### 2
This is more of a usage question, we invite you to ask questions about the usage of this library in our [Telegram group](https://t.me/pythontelegrambotgroup). In case you are unable to join our group due to Telegram restrictions, you can use our [IRC channel](https://webchat.freenode.net/?channels=##python-telegram-bot) to participate in the group.

jsmnbom:
@luciaasen It's been a while without reply from you, is your issue resolved?

Eldinnie:
1. It would be possible to add a callable to the conversationhandler like `conversation_timeout_callback` with signature: `def timeout_callback(bot, state)` But to be honest I think it's a niche not worth pursueing

simonvorobjev:
I have the same issue, handler for conversation timeout still not implemented?

PR #1217, opened 2018-09-11
Cannot handle timeout by callback in ConversationHandler #1136

simonvorobjev:
https://github.com/python-telegram-bot/python-telegram-bot/issues/1136

Met same issue, if conversation ended by timeout, we cannot handle it and make some finalization like clearing data, stopping threads etc. Now we can add handler for ConversationHandler.END, for example:

```python
conv_handler = ConversationHandler(
        entry_points=[CommandHandler('find', begin)],
        states={
            PRODUCT_CHOOSE: [MessageHandler(Filters.text,
                                           product_reply,
                                           pass_user_data=True),
                            ],
            ConversationHandler.END: [MessageHandler(Filters.text,
                                         conversation_timeout,
                                         pass_user_data=True),
                            ]
        }
```

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -38,8 +38,8 @@ class ConversationHandler(Handler):
 
     The second collection, a ``dict`` named :attr:`states`, contains the different conversation
     steps and one or more associated handlers that should be used if the user sends a message when
-    the conversation with them is currently in that state. You will probably use mostly
-    :class:`telegram.ext.MessageHandler` and :class:`telegram.ext.RegexHandler` here.
+    the conversation with them is currently in that state. Here you can also define a state for
+    :attr:`TIMEOUT` to define the behavior when :attr:`conversation_timeout` is exceeded.
 
     The third collection, a ``list`` named :attr:`fallbacks`, is used if the user is currently in a
     conversation but the state has either no associated handler or the handler that is associated
@@ -79,8 +79,9 @@ class ConversationHandler(Handler):
             ID.
         conversation_timeout (:obj:`float`|:obj:`datetime.timedelta`): Optional. When this handler
             is inactive more than this timeout (in seconds), it will be automatically ended. If
-            this value is 0 (default), there will be no timeout.
-            Handled in state :attr:`TIMEOUT` or ``-2``
+            this value is 0 (default), there will be no timeout. when it's triggered. The last
+            received update will be handled by ALL the handler's who's `check_update` method
+            returns True that are in the state :attr:`ConversationHandler.TIMEOUT`.
 
     Args:
         entry_points (List[:class:`telegram.ext.Handler`]): A list of ``Handler`` objects that can
@@ -112,9 +113,11 @@ class ConversationHandler(Handler):
             Default is ``True``.
         per_message (:obj:`bool`, optional): If the conversationkey should contain the Message's
             ID. Default is ``False``.
-        conversation_timeout (:obj:`float`|:obj:`datetime.timedelta`, optional): When this handler
-            is inactive more than this timeout (in seconds), it will be automatically ended. If
-            this value is 0 or None (default), there will be no timeout.
+        conversation_timeout (:obj:`float` | :obj:`datetime.timedelta`, optional): When this
+            handler is inactive more than this timeout (in seconds), it will be automatically
+            ended. If this value is 0 or None (default), there will be no timeout. The last
+            received update will be handled by ALL the handler's who's `check_update` method
+            returns True that are in the state :attr:`ConversationHandler.TIMEOUT`.
 
     Raises:
         ValueError
@@ -337,8 +340,7 @@ class ConversationHandler(Handler):
         self.logger.debug('conversation timeout was triggered!')
         del self.timeout_jobs[job.context['current_conversation']]
         handlers = self.states.get(self.TIMEOUT, [])
-        for candidate in handlers:
-            handler = candidate
+        for handler in handlers:
             if handler.check_update(job.context['update']):
                 handler.handle_update(job.context['update'], job.context['dispatcher'])
         self.update_state(self.END, job.context['current_conversation'])
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -55,7 +55,8 @@ class ConversationHandler(Handler):
     To change the state of conversation, the callback function of a handler must return the new
     state after responding to the user. If it does not return anything (returning ``None`` by
     default), the state will not change. To end the conversation, the callback function must
-    return :attr:`END` or ``-1``.
+    return :attr:`END` or ``-1``. To handle the conversation timeout, use must add handler
+    for :attr:`TIMEOUT` or ``-2``.
 
     Attributes:
         entry_points (List[:class:`telegram.ext.Handler`]): A list of ``Handler`` objects that can
@@ -79,6 +80,7 @@ class ConversationHandler(Handler):
         conversation_timeout (:obj:`float`|:obj:`datetime.timedelta`): Optional. When this handler
             is inactive more than this timeout (in seconds), it will be automatically ended. If
             this value is 0 (default), there will be no timeout.
+            Handled in state :attr:`TIMEOUT` or ``-2``
 
     Args:
         entry_points (List[:class:`telegram.ext.Handler`]): A list of ``Handler`` objects that can
@@ -120,6 +122,8 @@ class ConversationHandler(Handler):
     """
     END = -1
     """:obj:`int`: Used as a constant to return when a conversation is ended."""
+    TIMEOUT = -2
+    """:obj:`int`: Used as a constant to handle state when a conversation is timed out."""
 
     def __init__(self,
                  entry_points,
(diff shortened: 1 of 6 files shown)


Eldinnie:
Hi @simonvorobjev 

Thanks for your contribution. We have discussed it a bit and have some additional requests, questions.
First of all, we do all agree this is a missing part of the conversationhandler that has to be added. So thanks for that.
1. Now, I like the approach of adding the timeout behavior as a state to the `states` dict. However to reduce the confusion we would prefer it if it would use a new constant `ConversationHandler.TIMEOUT` as the key instead of `ConversationHandler.END`. This has to be added to the `ConversationHandler` and could default to `"TIMEOUT"`.
2. I also like the approach of saving the last update and sending that to the timeout handler. However right now it does not check if the handler is appropriate for the update. So seeing your example, if the last update was a photo, the timeout handler and the update would have a conflict. It should use `check_update`
3. I would like you to add some debug messages telling the logging system the conversation is timed out.
4. You're missing tests. New behavior like this should be tested. Please add tests to test this behavior with all possibilities (and expected failures)
5. Please add yourself to our authors list :D 

Thanks,
Pieter

Eldinnie:
@Eldinnie oh and lastly, please add the new attributes and it;s behaviours to the docstrings

simonvorobjev:
> Hi @simonvorobjev
> 
> Thanks for your contribution. We have discussed it a bit and have some additional requests, questions.
> First of all, we do all agree this is a missing part of the conversationhandler that has to be added. So thanks for that.
> 
> 1. Now, I like the approach of adding the timeout behavior as a state to the `states` dict. However to reduce the confusion we would prefer it if it would use a new constant `ConversationHandler.TIMEOUT` as the key instead of `ConversationHandler.END`. This has to be added to the `ConversationHandler` and could default to `"TIMEOUT"`.
> 2. I also like the approach of saving the last update and sending that to the timeout handler. However right now it does not check if the handler is appropriate for the update. So seeing your example, if the last update was a photo, the timeout handler and the update would have a conflict. It should use `check_update`
> 3. I would like you to add some debug messages telling the logging system the conversation is timed out.
> 4. You're missing tests. New behavior like this should be tested. Please add tests to test this behavior with all possibilities (and expected failures)
> 5. Please add yourself to our authors list :D
> 
> Thanks,
> Pieter

1. Done
2. We cannot use check_update, because in that case we will be able only to use MessageHandler, because for Handler check_update is not defined (the most obvious option for TIMEOUT handler), and CommandHandler checking last command and it will not set the callback.
3. Done
4. Done
5. Done
6. Docs done

telegram/ext/conversationhandler.py
```diff
@@ -328,5 +334,10 @@ def update_state(self, new_state, key):
             self.conversations[key] = new_state
 
     def _trigger_timeout(self, bot, job):
-        del self.timeout_jobs[job.context]
-        self.update_state(self.END, job.context)
+        self.logger.debug('conversation timeout was triggered!')
+        del self.timeout_jobs[job.context['current_conversation']]
+        handlers = self.states.get(self.TIMEOUT)
```
Eldinnie: please use: `handlers = self.states.get(self.TIMEOUT, [])` so the next line becomes
`for candidate in handlers:`

telegram/ext/conversationhandler.py
```diff
@@ -328,5 +334,10 @@ def update_state(self, new_state, key):
...
+        for candidate in (handlers or []):
+            handler = candidate
+            handler.handle_update(job.context['update'], job.context['dispatcher'])
```
Eldinnie: We need to run a check_update here. If we don;t it could send an update to a handler not fit for it.

simonvorobjev:
@Eldinnie I've fixed the issues, and about docstrings I'm not sure, it will be better if you will "take a shot" in it :-)

tests/test_conversationhandler.py
```diff
@@ -63,8 +64,12 @@ def reset(self):
                 CommandHandler('gettingThirsty', self.start),
                 CommandHandler('drinkMore', self.drink)
             ],
+            ConversationHandler.TIMEOUT: [
+                MessageHandler(None, self.passout)
```
Eldinnie: Please use a list of handlers here and add a test to check if both cvan be reached

tests/test_conversationhandler.py
```diff
@@ -90,6 +95,10 @@ def drink(self, bot, update):
     def code(self, bot, update):
         return self._set_state(update, self.CODING)
 
+    def passout(self, bot, update):
+        self.is_timeout = True
+        return self._set_state(update, self.END)
```
Eldinnie: This should not be returned, because the timeout will set end automatically

tests/test_conversationhandler.py
```diff
@@ -341,6 +350,7 @@ def test_conversation_timeout(self, dp, bot, user1):
         assert handler.conversations.get((self.group.id, user1.id)) == self.BREWING
         sleep(0.5)
         dp.job_queue.tick()
+        assert self.is_timeout is True
```
Eldinnie: Please write new tests instead of using the old ones

tests/test_conversationhandler.py
```diff
@@ -375,6 +385,7 @@ def test_conversation_timeout_keeps_extending(self, dp, bot, user1):
         assert handler.conversations.get((self.group.id, user1.id)) == self.DRINKING
         sleep(.1)  # t=1.1
         dp.job_queue.tick()
+        assert self.is_timeout is True
```
Eldinnie: Please write new tests instead of using the old ones

tests/test_conversationhandler.py
```diff
@@ -397,5 +408,6 @@ def test_conversation_timeout_two_users(self, dp, bot, user1, user2):
         assert handler.conversations.get((self.group.id, user2.id)) == self.THIRSTY
         sleep(0.5)
         dp.job_queue.tick()
+        assert self.is_timeout is True
```
Eldinnie: Please write new tests instead of using the old ones

Eldinnie:
@simonvorobjev you;re still with us here?

simonvorobjev:
> @simonvorobjev you;re still with us here?

Yes, I just don't know how to properly make test for different handlers (for example I still don't know how to use CommandHandler so it will not fail on check_update).

Eldinnie:
Ok, I'll chip in on tests and docs.

jsmnbom:
Something is wrong with the tests it seems.. Flake8 is also freaking out.

Eldinnie:
@jsmnbom working on it, came from the v12 merge

Eldinnie:
I think I addressed all issues, test run local :D

Eldinnie:
it's getting green

Eldinnie:
@tsnoam @jsmnbom Would like a review on this please

Eldinnie:
@tsnoam added the context object. If tests run can you please review?

telegram/ext/conversationhandler.py
```diff
@@ -26,6 +26,13 @@
 from telegram.utils.promise import Promise
 
 
+class ConversationTimeoutContext(object):
```
tsnoam: This is an internal class, you probably want to call it `_ConversationTimeoutContext`

telegram/ext/conversationhandler.py
```diff
@@ -26,6 +26,13 @@
...
+    def __init__(self, conversation_key=None, update=None, dispatcher=None):
```
tsnoam: Do you really need the defaults as `None`? (not that it really matters)

telegram/ext/conversationhandler.py
```diff
@@ -331,8 +338,9 @@ def handle_update(self, update, dispatcher, check_result):
         if self.conversation_timeout and new_state != self.END:
             self.timeout_jobs[conversation_key] = dispatcher.job_queue.run_once(
                 self._trigger_timeout, self.conversation_timeout,
-                context={'current_conversation': conversation_key, 'update': update,
-                         'dispatcher': dispatcher})
+                context=ConversationTimeoutContext(conversation_key=conversation_key,
```
tsnoam: Why not just `_ConversationTimeoutContext(conversation_key, update, dispatcher)`?
(i'm just nitpicking after our long discussion what's an unneeded overhead)

telegram/ext/conversationhandler.py
```diff
@@ -358,10 +366,13 @@ def update_state(self, new_state, key):
 
     def _trigger_timeout(self, bot, job):
         self.logger.debug('conversation timeout was triggered!')
-        del self.timeout_jobs[job.context['current_conversation']]
+        conversation_key = job.context.conversation_key
+        update = job.context.update
+        dispatcher = job.context.dispatcher
```
tsnoam: nitpicking again: why the over verbosity of the above 3 lines? why not just call `job.context.<something>` when needed?

Eldinnie:
@tsnoam I was being overly verbose, mainly for myself. I took all your suggestions

Eldinnie:
Merging.
Thanks for the startup on this @simonvorobjev

A-Iskakov:
can anyone provide an example of using `ConversationHandler` with `conversation_timeout`?

PR #1226, opened 2018-09-27
Added parse_mode parameter in Updater and in Bot class

Ak4zh:
Added parse_mode parameter in Updater and in Bot class to allow to set a default parse mode for your bot.

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -108,12 +108,16 @@ class Bot(TelegramObject):
             :obj:`telegram.utils.request.Request`.
         private_key (:obj:`bytes`, optional): Private key for decryption of telegram passport data.
         private_key_password (:obj:`bytes`, optional): Password for above private key.
+        parse_mode (:obj:`str`, optional): Send Markdown or HTML, if you want Telegram apps to
+                show bold, italic, fixed-width text or inline URLs in the media caption. See the
+                constants in :class:`telegram.ParseMode` for the available modes.
 
     """
 
     def __init__(self, token, base_url=None, base_file_url=None, request=None, private_key=None,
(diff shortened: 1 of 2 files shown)


Eldinnie:
Hi @Ak4zh 

We discussed this feature several times in the developers chat. But never came to a good way to offer it.
The following are the requirements and thinkpoints we've come up with so far

- If a default parse_mode is set and none is given. Add it.
- If a default parse_mode is set and another parse_mode is given. Use the given one
- If a default parse_mode is set and `None` is given. Use no parse_mode
- Default parse_mode should be used for text messages and captions for every media type
- Default parse_mode should be used for answers to an inline_query <- This is the very tricky one.
- Tests for all the above.

Right now your PR is not up to these requests. Are you willing to put in some work to try a solution as I just described?

Ak4zh:
Regarding:
```
If a default parse_mode is set and None is given. Use no parse_mode
```

What seems a better approach.

-> Default parse mode is set, so it always uses the default one and use other parse_mode if something is passed specifically.

To use no parse_mode you can pass parse_mode=False and no parse mode will be used.

Will look something like:

```
if parse_mode:
    data[‘parse_mode’] = parse_mode
elif parse_mode is False:
    data[‘parse_mode’] = None
elif self.parse_mode:
    data[‘parse_mode’] = self.parse_mode
```


Or another approach can be, adding another parameter

```
use_default_parse_mode=True
```

So whenever you pass nothing in parse_mode it’s None and will use default parse mode (if anything is set)

If you want to use no parse mode, just pass 
```
use_default_parse_mode=False
```

Poolitzer:
will be implemented with #1490, closing

PR #1238, opened 2018-10-01
Refactor

djdany01:
- Replaced *else:* clause with *return False* #1236


telegram/ext/conversationhandler.py
```diff
@@ -264,10 +264,8 @@ def check_update(self, update):
                         # Save the current user and the selected handler for handle_update
                         self.current_conversation = key
                         self.current_handler = candidate
-
                         return True
-
-                else:
+                      
                     return False
```
Eldinnie: oops, the indent is of. We don;t want to return False if the first `if` fails.

Eldinnie:
Hey, I understand you want a t-shirt. But this fix is just creating a new problem

Issue #957, opened 2017-12-31
In test_jobqueue.py (maybe others), wait longer (but explicitly) for expected jobs to run


nmlorg:
Hi, during my last review the [CI reported an error](https://travis-ci.org/python-telegram-bot/python-telegram-bot/jobs/322945105) in tests/test_jobqueue.py TestJobQueue test_error after scheduling a job to run after 20 ms and, after waiting 30 ms, seeing it hadn't run. One possibility is that the environment the tests are run in can't reliably resume execution in under 30 ms, and the two threads just ran out of order. If that's the case, would you guys be interested in changing some/all of those tests from:

```python
      def test_run_once(self, job_queue):
          job_queue.run_once(self.job_run_once, 0.01)
-         sleep(0.02)
          assert self.result == 1
```

to something like:

```python
      def test_run_once(self, job_queue):
          job_queue.run_once(self.job_run_once, 0.01)
+         assert self.job_ran.wait(10)
          assert self.result == 1
```

with:

```python
  class TestJobQueue(object):
      result = 0
      job_time = 0
+     job_ran = threading.Event()

      @pytest.fixture(autouse=True)
      def reset(self):
          self.result = 0
          self.job_time = 0
+         self.job_ran.clear()

      def job_run_once(self, bot, job):
          self.result += 1
+         self.job_ran.set()
```

which would cause the tests to wait explicitly until the job executes, but give up (and signal failure) after 10 seconds instead of just 10 ms?

tsnoam:
@nmlorg 
That is a good idea. Would you like to send a PR for it?

Ambro17:
The approach of having a threading.Event() does not work for tests that call the `job_run_once` more than one time. Because as per `Event.wait` behaviour all waits are started as soon as an event is `.set()`.
In those cases, as soon as the callback is executed the first time, it continues to try to validate the assertion (which fails because the callback was executed just 1 time, not `n` as expected)
Example:
```python
    def test_run_repeating(self, job_queue):
        job_queue.run_repeating(self.job_run_once, 0.02)
        # sleep(0.05)
        assert self.job_ran.wait(timeout=2)
        # As soon as job_run_once sets the flag, wait continues but job_run_once was called only once.
        assert self.result == 2
```

jsmnbom:
@Ambro17 I haven't read too much into it but could a \[semiphore](https://docs.python.org/3/library/threading.html#threading.Semaphore) work?

[...]

PR #1266, opened 2018-10-18
Add wait longer for expected jobs to run in tests.

francamacdowell:
I saw that one guy created a good idea for the tests and disappear.
So I'm sending the PR idea.
Solving #957

### DIFF
--- a/tests/test_jobqueue.py
+++ b/tests/test_jobqueue.py
@@ -40,6 +40,7 @@ def job_queue(bot):
 class TestJobQueue(object):
     result = 0
     job_time = 0
+    job_ran = threading.Event()
 
     @pytest.fixture(autouse=True)
     def reset(self):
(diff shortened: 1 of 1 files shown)


jsmnbom:
Thanks a lot for taking the time to make this contribution! :D
Currently it doesn't quite fix everything that we'd like, you're very welcome to simply add more commits to your branch if you'd like to keep working on this.

We are go need to do this for every single one of the jobqueue tests, not just test_run_once. This means you'll probably also have to put the job_ran = threading.Event() inside the reset fixture so it is reset for every test that is run. Once this is done, try also removing the @skipif line that we have on the class, since that should no longer be needed :D

Eldinnie:
I would like encourage you to fix the rest of the tests. Verhy welcome!

Ambro17:
Hi, i'm interested in refactoring these tests but i don't want to overlap with other dev. Is it okay if i take it?

francamacdowell:
Yes, It is. You can take it

Bibo-Joshi:
Closing in favour of of #1357

PR #1270, opened 2018-10-23
Fix unresolvable promises

vasinkd:
**Situation:**
We use ConversationHandler with run_async decorator and an entry_point function returns `None`

**Result:**
We stuck within conversation state `(None, Promise)` since Promise returns `None` and `None` means we do not update conversation state.

**Solution:**
If Promise returns `None` AND  old state is `None` - exit Conversation Handler.

**Caveats:**
May be not suitable if a programmer uses `None` as a valid state in Conversation states dict. But why would anyone do that?


jh0ker:
Good catch, thanks for your contribution! 

This also seems to be the behavior for when a synchronous entry point handler returns `None`, although this is never explicitly documented. Perhaps it would be a good idea to do so. 

It would also be great if you could add a unit test that tests for this bug, so we won't have any regressions in the future. It should probably look similar to [this test](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/tests/test_conversationhandler.py#L277).

jsmnbom:
@vasinkd is test and docs something you would like to do an attempt at?

vasinkd:
Totally forgot about this. @jsmnbom , thanks for a reminder!

Ok, I'll add tests, no problemo.
What about documentation? Should I mention this behavior in ConversationHandler class description?

vasinkd:
Done. Better late than never :)

Eldinnie:
Test for this instance passed. Can we merge for V12 @jsmnbom ?

Eldinnie:
CI is running, should be able to be merged in v12

Eldinnie:
Merged V12 after making small adjustment there regarding conversationhandler. If CI agrees we can merge.

PR #1271, opened 2018-10-25
Allow persistence with non telegram.Update updates

jsmnbom:
Update types used with TypeHandler, doesn't necessarily have a `effective_chat` and `effective_user`. Therefore we need to make sure that it's an instance of telegram.Update before assuming those attributes exist.

I'm unsure how to test... any suggestions?

### DIFF
--- a/telegram/ext/dispatcher.py
+++ b/telegram/ext/dispatcher.py
@@ -30,7 +30,7 @@ from queue import Queue, Empty
 
 from future.builtins import range
 
-from telegram import TelegramError
+from telegram import TelegramError, Update
 from telegram.ext.handler import Handler
 from telegram.utils.promise import Promise
 from telegram.ext import BasePersistence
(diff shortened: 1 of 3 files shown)


jh0ker:
Looks good. As for testing, run some arbitrary update through a persistence-enabled dispatcher and `assert(True)` after? Or something to that effect.

JosXa:
@jsmnbom Shouldn't we rather go with duck typing and assume that when it looks like an Update and behaves like an Update, it probably is something like an update?
Long story short: if hasattr(obj, 'effective_chat') checks rather than isinstance

Eldinnie:
I think we should add this.

Eldinnie:
@JosXa can you explain why we would do that?

JosXa:
Well if users choose to create their own versions of the `Update` class but do not necessarily inherit from or our built-in one, then the check will fail even though the object has the same signature

PR #1274, opened 2018-10-27
Added echobot type 3

chaitanyabaranwal:
Created another version of an echobot which on typing '/echo', sends a message that echo mode is on. The user can then type the message to be echoed, and the bot echoes the message. Uses ConversationHandler to achieve the same.

### DIFF
--- a/examples/echobot3.py
+++ b/examples/echobot3.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""Simple Bot to reply to Telegram messages.
+This program is dedicated to the public domain under the CC0 license.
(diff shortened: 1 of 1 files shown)


Eldinnie:
Hi @chaitanyabaranwal,

Thanks for making this PR.
Although I'm not against having more examples. I think this one is not adding anything more to the examples then `conversationbot(2).py` does. I would like to see (for example) an example with a conversationhandler using callbackqueries.
Also take a look at the comments [codacity](https://app.codacy.com/app/python-telegram-bot/python-telegram-bot/pullRequest?prid=2446011) made on your PR.

Closing this, but thanks for wanting to contribute.

PR #1283, opened 2018-11-01
Make dispatcher use one context per update

Eldinnie:
It gives user the option to `overload` context with their own properties in a lower group handler if they like

### DIFF
--- a/telegram/ext/callbackcontext.py
+++ b/telegram/ext/callbackcontext.py
@@ -34,6 +34,8 @@ class CallbackContext(object):
         get passed the same `CallbackContext` object (of course with proper attributes like
         `.matches` differing). This allows you to add custom attributes in a lower handler group
         callback, and then subsequently access those attributes in a higher handler group callback.
+        Note that the attributes on `CallbackContext` might change in the future, so make sure to
+        use a fairly unique name for the attributes.
 
     Warning:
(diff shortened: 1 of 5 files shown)


Eldinnie:
1. Yes in the wiki I guess?
2. I think that's overthingking it personally...
3. ehm.... I guess it might not be on run_async handlers....

jsmnbom:
1. maybe in the ContextCallback class too? Like "You can add attributes to this class and they will propagate to handlers in higher dispatcher groups."
2. Alright, I do think we should add another note to somehow describe this behaviour
3. We need to test that somehow I think

jsmnbom:
Okay, so this feature simply doesn't work with `@run_async` at all... Is that okay?

Test code to show not working/working (uncomment/comment the `@run_async`
``` python
# @run_async
def a(u, c: CallbackContext):
    print(c)
    c.q = 1

# @run_async
def b(u, c: CallbackContext):
    print(c)

u = Updater(token, use_context=True)
d = u.dispatcher
d.add_handler(MessageHandler(Filters.all, b))
d.add_handler(MessageHandler(Filters.all, a), -1)

u.update_queue.put(Update(0, message=Message(0, None, None, None)))

u.start_polling()
u.idle()
```

PR #1299, opened 2018-11-21
Fix description for JobQueue's run_daily and run_repeating methods.

ssandke:
I also fixed a couple of incorrect English contractions in the same file.

Note: I have no idea if this contribution qualifies for addition to the Authors list; I presume not.


ssandke:
Hm. Not sure why some python versions would fail tests and others not. The change doesn't touch code at all.

jsmnbom:
Thanks for a lot for this change! :D
And you're totally right that it shouldn't affect our tests at all, our tests are a bit flaky sometimes.

Issue #1297, opened 2018-11-20
PicklePersistence of user_data set within conversations


Bibo-Joshi:
### Steps to reproduce
1. In [persistentconversationbot.py](https://github.com/python-telegram-bot/python-telegram-bot/blob/378784f55e0c00c5b85659cdc7edb278557bdc0b/examples/persistentconversationbot.py#L150), change `persistent=True` to `persistent=False`
2. Start the bot, give the bot some information to store

3. Stop bot
4. Start bot again, send command `/show_data`. It will be empty

### Expected behaviour
`user_data` changed within a conversation should be persisted even if the conversation is not
(i.e., I would expect this)

### Actual behaviour
`user_data` changed within a conversation that is not persisted, will not be persisted

### Configuration
Raspbian GNU/Linux 9 (stretch)

**Version of Python, python-telegram-bot & dependencies:**

```
python-telegram-bot 11.1.0
certifi 2018.01.18
future 0.17.1
Python 3.5.3 (default, Sep 27 2018, 17:25:39)  [GCC 6.3.0 20170516]
```

### Research so far:

1.  When the conversation is persisted, the `user_data` will only be persisted, if `single_file = True`. This is because this will trigger `dump_singlefile()` in [picklepersistence.py](https://github.com/python-telegram-bot/python-telegram-bot/blob/378784f55e0c00c5b85659cdc7edb278557bdc0b/telegram/ext/picklepersistence.py#L97) on update in Conversation. When setting `single_file = False`, `user_data` will not be persisted
2. Things go south in [`update_user_data()`](https://github.com/python-telegram-bot/python-telegram-bot/blob/378784f55e0c00c5b85659cdc7edb278557bdc0b/telegram/ext/picklepersistence.py#L195). Printing the result of

    ```
    self.user_data.get(user_id) == data:
    ```
    always gives `True` even though `self.user_data.[user_id]` _has_ changed after user input (as one can test by printing it).

I therefore conclude, that `self.user_data[user_id]` must be changed somewhere along the way _before_ `update_user_data()` is called. Still, I can't pinpoint, where that happens. Will try and find it.

If you have a clue, please let me know :)

Bibo-Joshi:
Found the issue:
[...]

PR #1301, opened 2018-11-25
Fix and enhance PicklePersistence: deepcopy

Bibo-Joshi:
As reported in [#1297](https://github.com/python-telegram-bot/python-telegram-bot/issues/1297), the use of shallow copies in `PicklePersistence` leads to data not being persisted. Same holds for `DictPersistence` A fix would be the use of `copy.deepcopy`, which however may become time consuming for large data sets.

I therefore implemented an additional argument `on_update` for `Pickle/DictPersistence` that let's you choose wether to check for updates before dumping or not.
If `on_update = False`, checking is obmitted and one can skip time consuming deepcopies (i.e. just pass the actual data). If `on_update = True`, deepcopies are used.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -19,6 +19,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Anton Tagunov <https://github.com/anton-tagunov>`_
 - `Avanatiker <https://github.com/Avanatiker>`_
 - `Balduro <https://github.com/Balduro>`_
+- `Bibo-Joshi <https://github.com/Bibo-Joshi`_
 - `bimmlerd <https://github.com/bimmlerd>`_
 - `d-qoi <https://github.com/d-qoi>`_
 - `daimajia <https://github.com/daimajia>`_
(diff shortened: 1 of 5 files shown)


Bibo-Joshi:
This PR might be redundant (see [#1297](https://github.com/python-telegram-bot/python-telegram-bot/issues/1297#issuecomment-452033841) and [#1322](https://github.com/python-telegram-bot/python-telegram-bot/pull/1322#issue-242742755)).
Leaving this open until #1297 is resolved.

jsmnbom:
I would like to merge this for the V12 beta that I'm working on.
But I'm gonna hold of for now because of the following (please correct me if I'm wrong :))
- Tests are updated but don't actually include a regression test for #1297.
- I still don't 100% understand why we need to copy the data at all?
- And I'm not really sure if on_update is a descriptive enough name for the parameter if we do need it.
- It also feels the parameter should be on BasePersistence and not only subclasses of it?

jsmnbom:
Oh and feel free to reach out to me on telegram @jsmnbom if you wanna help convince and fix these things together so we can get it v12 beta ready :)

jsmnbom:
Superseeded by #1342. We choose to go a simpler route and always deepcopy the data. Thanks a lot for both finding the issue and taking the time to make the PR :D

Issue #1302, opened 2018-11-25
Persistence of Updates without effective_chat/effective_user


Bibo-Joshi:
### Steps to reproduce
1. In the example-bot [inlinebot](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/inlinebot.py), import `PicklePersistence`, and add persistence to the dispatcher.
2. Run bot
3.  Make `InlineQuery`

### Expected behaviour
Handle InlineQuery and skip persistence

### Actual behaviour
Error get's thrown in [dispatcher.py](https://github.com/python-telegram-bot/python-telegram-bot/blob/0cf0cccbc5abd538b6c6ea7487cce4988328da03/telegram/ext/dispatcher.py#L303).

### Configuration
**Operating System:**
Raspbian GNU/Linux 9 (stretch)

**Version of Python, python-telegram-bot & dependencies:**

```
python-telegram-bot 11.1.0
certifi 2018.01.18
future 0.17.1
Python 2.7.13 (default, Sep 26 2018, 18:42:22)  [GCC 6.3.0 20170516]
```

### Logs
```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/python_telegram_bot-11.1.0-py2.7.egg/telegram/ext/dispatcher.py", line 303, in process_update
    if self.persistence.store_chat_data and update.effective_chat.id:
AttributeError: 'NoneType' object has no attribute 'id'
```

### Proposed Fix
Change lines 303 to 
```
if self.persistence.store_chat_data and update.effective_chat:
# instead of: update.effective_chat.id
```
and line 309
```
if self.persistence.store_user_data and update.effective_user:
# instead of: update.effective_user.id
```
That does the trick. :)
Already tested it, just did't want to open a pull request for that little change ;)

jsmnbom:
This change looks good, please do go ahead and open up a PR, if you have time ^^

PR #1303, opened 2018-11-29
Fix check for effective chat/user in persistence

Bibo-Joshi:
Changes as proposed in [#1302](https://github.com/python-telegram-bot/python-telegram-bot/issues/1302)


jsmnbom:
Thanks a lot for submitting this PR. I will try to review it shortly.

The failed tests don't seem to have anything with the changes at first glance btw :)

jsmnbom:
Given that effective_chat and effective_user should always have an id, if they're not None. This change looks good to me. Thanks a lot for the PR :D

PR #1308, opened 2018-12-08
Fix bug: unable to save jobs with timezone aware dates

Ambro17:
There was a bug when trying to save a job with a timezone aware date. The error was the following:
`TypeError: can't subtract offset-naive and offset-aware datetimes`

This was caused by the line 65 of `telegram/ext/jobqueue.py` that not surprisingly, tried to substract the timezone aware date of the job with `datetime.now()` which is naive as follows.
`next_t = (next_t - datetime.datetime.now(next_t.tzinfo)).total_seconds()`

This PR fixes that and adds tests that pass on the new implementation and fail on the previous one.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -16,6 +16,7 @@ Contributors
 The following wonderful people contributed directly or indirectly to this project:
 
 - `Alateas <https://github.com/alateas>`_
+- `Ambro17 <https://github.com/Ambro17>`_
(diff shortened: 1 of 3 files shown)


jsmnbom:
After discussion in internal group we have decided to revert this as some users ran into some issues with it.
Please open a new issue if you really believe this is absolutely critical to have, and then we can potentially discuss how to implement it properly :)
Thanks for opening the PR anyways, we love it when people help :)

PR #1312, opened 2018-12-16
Adjust persistence on exit behaviour

Bibo-Joshi:
As `user_data` and `chat_data` might be changed between updates (e.g. when storing information from background processes that need to be visible to the user on request), they should be updated before calling `flush()` on shutdown of the bot. I therefore added the method `update_persistence()` to the `Dispatcher` class, that is called from the `Updater` before flushing on shutdown.
Also, since it will be called on shutdown, `flush()` should *always* flush. I therefore edited the method in `PicklePersistence` to be independent of `on_flush`.

### DIFF
--- a/telegram/ext/dispatcher.py
+++ b/telegram/ext/dispatcher.py
@@ -395,6 +395,15 @@ class Dispatcher(object):
                 del self.handlers[group]
                 self.groups.remove(group)
 
+    def update_persistence(self):
+        """Update ``user_data`` and ``chat_data`` in :attr:`persistence`.
+        """
+        if self.persistence:
+            for chat_id in self.chat_data:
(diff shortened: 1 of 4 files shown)


Bibo-Joshi:
If #1325 is merged, the new `update_persistence` method needs to be adjusted to also update `bot_data`

jsmnbom:
As per #1301 I'd love to have this in v12, but I'm not comfortable enough the persistence component yet, so I'll hold off till the author can review or I get more time.
Thanks a whole lot for all your PRs though, it's great to see people helping :)

Eldinnie:
Right, good addition!

Eldinnie:
I pushed two commits to fix a docstring and add tests. If CI approves it;s good to go

jsmnbom:
CI fail unrelated, will fix on master

PR #1313, opened 2018-12-16
InlineQueryHandler: Add missing return in check_update and add Filters support

Bibo-Joshi:
In `check_update()` there was a missing `return False` in case `isinstance(update, Update) and update.inline_query` returns `False`

Also, since I have a use case, I'd like to add Filters support to InlineQueryHandler. Since an InlineQuery has very little information that can be checked (only the `query` and the `from_user`) and the `pattern` can be used to filter the `query`, this may seem to be an overkill. One could argue that a simple `if - else`-statement checking the user ID does the trick. However, since the Filters module is already provided and one may already have set up a bunch of user-Filters to be used with the other handlers, why not make them useable with InlineQueryHandlers?


jsmnbom:
Thanks a lot for this PR @Bibo-Joshi. However, I think we'll have to put it on hold for a little while, until we figure out how we are gonna handle filters and patterns in the v12 release.

Bibo-Joshi:
Closing in favor of #1520 and #1620

PR #1314, opened 2018-12-17
Improve regex filter docstring and avoid compiling compiled regex

Ambro17:
This PR improves the docstring of Filters.regex. Changed format to comply with the rest of the file, and specify that pattern can be either a string or a pattern object. Also avoids recompiling regex on regex instantiation.
Add test case and define `Filter` attribute so sphinx can show the proper message instead of the placeholder text

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -173,10 +173,11 @@ class Filters(object):
     command = _Command()
     """:obj:`Filter`: Messages starting with ``/``."""
 
-    class regex(BaseFilter):
+    class _Regex(BaseFilter):
         """
-        Filters updates by searching for an occurence of ``pattern`` in the message text.
+        Filters updates by searching for an occurrence of ``pattern`` in the message text.
         The ``re.search`` function is used to determine whether an update should be filtered.
+
         Refer to the documentation of the ``re`` module for more information.
 
         Note: Does not allow passing groups or a groupdict like the ``RegexHandler`` yet,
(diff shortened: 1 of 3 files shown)


telegram/ext/filters.py
```diff
@@ -173,25 +173,32 @@ def filter(self, message):
     command = _Command()
     """:obj:`Filter`: Messages starting with ``/``."""
 
-    class regex(BaseFilter):
+    class _Regex(BaseFilter):
```
jsmnbom: We actually have several filters that don't follow this pattern further down in the file.
As far as I can see we've set done it such that:
- All filters that don't need an input, name the class `_Filter` and init it as `filter = _Filter()`
- All filters that do need input, break PEP8 style and name the class `filter` - this also improves docs if I recall correctly.

Ambro17: You are right, i only saw the first filters so i asumed it was a distraction. Effectively, filters that expect an argument don't have the underscore prefix.
I'll change it today if i have time

Ambro17: Done, i'm not very familiar with sphinx so i would like to know if there's a way to build the docs to check the comment with the filter description (L212) is rendered as expected

Bibo-Joshi: Hey @Ambro17,
you can find a installation guide to sphinx [here](http://www.sphinx-doc.org/en/master/usage/installation.html). Maybe you need to install [spinx.ext.napoleon](https://pypi.org/project/sphinxcontrib-napoleon/), [sphinx_rtd_theme ](https://pypi.org/project/sphinx_rtd_theme/) or other dependencies of [conf.py](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/docs/source/conf.py).
You can build the docs by running `make html` in the `docs` directory. The HTML pages will be saved to `docs/build/html`, with `index.html` being the start page. The output of the `make` will also tell you, if dependencies are missing.
Hope this helps :)

Ambro17: @Bibo-Joshi Thanks! I managed to get it to look as i like. Not sure if you agree, here's a screenshot of how it looks now
image

Ambro17: 1. Added the proper spacing after ```Note:``` so it's rendedered nicely.
2. I checked this option and it seems cool but it sets a green background for the code. As all the other filters have the ``` `` ``` notation, i would prefer to mantain the same style over all filters, but tell me what you think
3. I agree that filters that don't receive arguments don't have the filter right now, but i'm not sure that was on purpose. As it currently renders a default message which is not correct or at least is confusing.
image
The overriding of the filter method is done under the hood, users should not override the filter method, only build the filter with the desired regex.
Personally i think that adding the filter behaviour for that case makes more sense than rendering the ```BaseFilter``` default message.
image

jsmnbom:
Tested locally and seems good to me (CI is a bit broken right now)
Merging primarily to avoid conflicts with improving the regex filter for V12.
Thanks a lot for the PR :)

PR #1315, opened 2018-12-20
FIXED: ConversationHandler warnings were logged to root logger

Tom-Brouwer:
Several warnings for the conversationhandler were logged to the root logger (logging.logger), instead of the module level logger (self.logger)

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -173,8 +173,8 @@ class ConversationHandler(Handler):
             raise ValueError("'per_user', 'per_chat' and 'per_message' can't all be 'False'")
 
         if self.per_message and not self.per_chat:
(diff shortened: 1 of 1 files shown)


jsmnbom:
I wonder if we shouldn't really be using `warnings.warn` for these cases.
Per the python docs:

> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
> logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

Eldinnie:
I think `warnings.warn` is appropriate here. Also please merge, because these warnings have a live change.

jsmnbom:
This has been superseeded #1343. Thanks a lot for the PR @Tom-Brouwer, your commit also exists in that PR, we must wanted to make a couple of commits on top.

PR #1319, opened 2019-01-03
Add break to for loops to avoid multiple repetitive warnings on conve…

Ambro17:
As spoken earlier today with @Eldinnie, it seems fair to show the warnings explaining the edge cases for the conversation handlers. But if one has a complex conversation, those messages are replied too many times and don't give new information.

This PR makes those warnings appear only once and add tests for this behaviour.


Eldinnie:
Looks good to me.

Issue #1321, opened 2019-01-07
Unable to commit: Error during pylint check


Bibo-Joshi:
Since I have little knowledge of the automatic checking setup, I don't know, if this is even PTB related. However, since I was not able to resolve the issue in not-PTB-related ways, I at least wanted to ask.
I finally hat the time, to implement the global memory suggested in #1318 and would very much like to open a PR 😅 

### Steps to reproduce
1. Follow the Contribution guideline

2. Change something (for testing, I added a whitespace in a comment …)

3. try to commit the changes

### Expected behaviour
Commits

### Actual behaviour
During the `pylint` procedure an error gets thrown: `AttributeError: module 'astroid' has no attribute 'YES'`.

### Configuration
**Operating System:**

```
DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=19
```

**Version of Python, python-telegram-bot & dependencies:**

``$ python -m telegram``

```
python-telegram-bot 11.1.0
certifi 2018.11.29
future 0.17.1
Python 2.7.15rc1
```

### What I have tried

- Update `pylint` and `astroid`
- Reinstall the OS
- check if `astroid.YES` exists (it does!)

jsmnbom:
I will look into this later, for now just use the `--no-verify` arg in git commit to skip the pre-commit hooks :)

Bibo-Joshi:
Thanks for the quick reply!
I just noted, that `make test` also fails (A lot of `INTERNALERROR`s).
If it seems to you, that this is not a general problem, please feel free to close the issue. Sorry for beeing noobish :grimacing:

jsmnbom:
It sounds like there's something wrong with your setup somehow :/
`make test` works fine for me.
Could you humor me and just try using python 3?...

Bibo-Joshi:
Got `make test` fixed by upgrading `flaky`. BTW: I noticed, that AppVeyor and Travis CI throw the same error I got (see outputs in #1322 ) 
`pylint` seems to be rum from `python3`. This is the complete traceback:

```
yapf.....................................................................Passed
Flake8...................................................................Passed
pylint...................................................................Failed
hookid: pylint

No config file found, using default configuration
Traceback (most recent call last):
  File "/home/hinrich/.cache/pre-commit/repotmz9w1sv/py_env-python3.6/bin/pylint", line 11, in <module>
    sys.exit(run_pylint())
[...]

Issue #1318, opened 2018-12-28
Enhancement: global memory


Bibo-Joshi:
It would be lovely to have a global, persistable memory available. That is, add an attribute `global_data` or such to the `Dispatcher` class, that can be passed to handlers and is integrated in persistence.
This would make it easy to process stuff that can be changed by multiple different chats.

Putting this here, because I don't know when/if I will be able to open a PR. If someone is faster than me, please don't hesitate ;)

jsmnbom:
I agree that we need some way to have arbitrary data (not accociated with a chat/user) be persisted via persistence.

Bibo-Joshi:
Closed by #1325

PR #1322, opened 2019-01-07
bot_data as global memory

Bibo-Joshi:
This PR implements the global memory proposed in #1318 by providing a dictionary `bot_data` in addition to `user_data` and `chat_data`.

I've integrated `bot_data` in the dispatcher class and the handler classes. Since the `bot_data` is a global memory, I've also integrated it in those handlers, that are not meant for telegram.Updates (e.g. StringCommandHandler).

Integration in persistence is also provided. Note, that in `update_bot_data`, the data is updated by

```
self.bot_data = data.copy()
````
in contrast to the current implentation for `user/chat_data`

```
self.user_data[chat_id] = data
```
This is to avoid problems as described in #1297. However, to leave this PR self-contained, I didn't change it for `user_data` and `chat_data`.

In `DictPersistence`, I just used `json.loads()` to decode serialized data. For `user_data` and `chat_data`, a function from the helpers module is used to allow integer keys. imo, this is not really necessary and may lead to problems (e.g. when using `'3'` as a key, that will be decoded as `3`), so I left it out. If I just didn't get the point of it, I will gladly add this.

Also, I've started to adjust the test routines. I will add them, once I get `make test` to work again (see #1321 )

### DIFF
--- a/telegram/ext/basepersistence.py
+++ b/telegram/ext/basepersistence.py
@@ -25,6 +25,8 @@ class BasePersistence(object):
 
     All relevant methods must be overwritten. This means:
 
+    * If :attr:`store_bot_data` is ``True`` you must overwrite :meth:`get_bot_data` and
+      :meth:`update_bot_data`.
     * If :attr:`store_chat_data` is ``True`` you must overwrite :meth:`get_chat_data` and
       :meth:`update_chat_data`.
     * If :attr:`store_user_data` is ``True`` you must overwrite :meth:`get_user_data` and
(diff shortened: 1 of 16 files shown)


jsmnbom:
I really like what you've done here, but I'm sorry to report that you'll probably have to make a couple of changes. I haven't looked very much into it, but since you've based it on master instead of the V12 development branch, you don't use the new Context object which this feature definitely should.
You're very welcome to base it on V12 branch and rewrite the stuff that needs rewriting and then we'll take a look. It should be fairly simple - mostly switching the pass_xxx over to the Context system.
Feel free to contact me on telegram under this username if you have questions :)

Bibo-Joshi:
Replaced by #1325

PR #1324, opened 2019-01-11
Check for attributes on persisting custom update

Bibo-Joshi:
Custom updates may not have the attributes `effective_user` and `effective_chat`  and therefore persisting `user_data` and `chat_data` will fail. This is because in V12, `isinstance(update, telegram.Update` was removed.
This PR adds a `try - except` statement for this case.


jsmnbom:
It looks like after merging and updating the V12 branch that the `isinstance(update, telegram.Update)` is back. Is the approach in this pr in some way better than having that check?

Bibo-Joshi:
Mh … Well, I had assumed, that ``isinstance(update, telegram.Update)`` was gone on purpose to allow more customizable Updates. However, I don't really see why an customized Update should not be based on ``telegram.Update``. So, yeah, if ``isinstance(update, telegram.Update)`` is back for good, feel free to close this PR.

jsmnbom:
Honestly, thinking about it further. I think your solution is better actually. I've even on one of my bots had custom updates not subclassing Update (since they effectively had nothing to do with telegram).
And since we have a TypeHandler, I feel like this section of the code should support this usecase too.
Could you fix merge from v12 to fix the conflict, and then we can get it into the v12 beta?

Eldinnie:
`user_data` and `chat_data` are however pretty tied in with `telegram.Update`. I guess that;s why I assumed them only for `Update` classes. I could change it.

Poolitzer:
I mean, I dont really think its needed, but its just one try - except clause on the other side. We can go for it if you ask me.

Bibo-Joshi:
This was originally ment as a small bug fix. But if `ininstance(update, Update)` remains, I'm good with closing this.

PR #1327, opened 2019-01-21
Create Filters.channel

marcroga:
Filters messages sent in a channel chat.

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -592,6 +592,15 @@ class Filters(object):
 
     group = _Group()
     """:obj:`Filter`: Messages sent in a group chat."""
+    
+    class _Channel(BaseFilter):
+        name = 'Filters.channel'
+
+        def filter(self, message):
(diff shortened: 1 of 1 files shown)


jusim333:
good

jsmnbom:
As with #1104 I'm gonna close this. The reason is that it's implemented as part of #1221. Thanks for taking the time to submit a PR though, we always appreciate it :D

jsmnbom:
Actually scratch that, #1221 has the update.channel_post, but this filter would also detect forwarded channel messages so we might actually want both. Reopening and we'll check it out :)

Eldinnie:
I don;t think `message.chat.type == 'channel'` on forwarded channel messages....

jsmnbom:
@Eldinnie You're indeed right. If you need that info you'd need to check `forwarded_from_chat.type`. And I feel like having filters for those is too specific.
Closing since #1221 includes update.channel which does pretty much the same :)

Once again thanks for taking the time to submit a PR though, we always appreciate it :D

PR #1342, opened 2019-02-13
Fix #1297

### LINKED ISSUE #1297: PicklePersistence of user_data set within conversations

Eldinnie:
This makes a deepcopy of the user_data and chat_data dict as suggested by @Bibo-Joshi
Fixes #1297

### DIFF
--- a/telegram/ext/dictpersistence.py
+++ b/telegram/ext/dictpersistence.py
@@ -17,6 +17,8 @@
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains the DictPersistence class."""
+from copy import deepcopy
+
 from telegram.utils.helpers import decode_user_chat_data_from_json,\
     decode_conversations_from_json, enocde_conversations_to_json
 
@@ -129,7 +131,7 @@ class DictPersistence(BasePersistence):
             pass
         else:
             self._user_data = defaultdict(dict)
-        return self.user_data.copy()
+        return deepcopy(self.user_data)
 
     def get_chat_data(self):
         """Returns the chat_data created from the ``chat_data_json`` or an empty defaultdict.
@@ -141,7 +143,7 @@ class DictPersistence(BasePersistence):
             pass
         else:
             self._chat_data = defaultdict(dict)
-        return self.chat_data.copy()
+        return deepcopy(self.chat_data)
 
     def get_conversations(self, name):
         """Returns the conversations created from the ``conversations_json`` or an empty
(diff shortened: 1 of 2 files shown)


jsmnbom:
@Eldinnie So we are fine with the performance penalty "issue" as raised by @Bibo-Joshi in #1301 then?

jsmnbom:
I'm personally fine with it. If people were gonna store massive amounts of data, they would use a proper library/database for it instead I think.

Eldinnie:
I think we have no choice, and I think the performance penalty will be minimal. Also the fix proposed in #1301 (oops, missed that one before stating on this) means it;s a choice between `deepcopy()` and unneeded writes to disk (which are even more time consuming) so I guess just deepcopying is fine

jsmnbom:
CI fail unrelated, will fix on master.

PR #1343, opened 2019-02-13
Use warnings.warn for conversationhandler warnings.

jsmnbom:
Superseeds #1315 
(tried to push to Tom-Brouwer/python-telegram-bot, but it wouldn't let me.


jsmnbom:
CI fail unrelated, will fix on master.

Issue #1250, opened 2018-10-11
timeout behavior and kwargs in conversationhandler


Eldinnie:
In `ConversationHandler` we have quite a lot of kwargs regarding `@run_async` decorated handler callbacks and timeouts.
these are:

- `run_async_timeout` - The time conversationhandler will wait on a run_async task to complete.
- `timed_out_behavior` - Here you can specify a list of handlers that will handle the new update if the timeout is exceeded but nog finished on a `@run_async` decorated callback
- `conversation_timeout` - If this value is exceeded the conversation will be ended (with an internal job)

in v12 we add a mechanic to give the user an interaction with the conversation on 
- `conversation_timeout` by adding a `TIMEOUT` state which will be offered the last received update.

Now while debugging I noticed the following things that are a bit off.
If I specify a `run_async_timeout=5` it should mean that if the decorated handler takes longer then 5 seconds it will switch to `timed_out)behavior` but this is not the case.
The waiting for it to complete starts on an incomming new update, within `check_update`. This means a couple of bad things.

- If I set a timeout for 10 seconds. It could still be running for 3 days and 10 seconds if the next update is only received after 10 days
- If I specify a `run_async_timeout=20` and the handler is doing something infinite. The processing of the `update_queue` will be paused for 20 seconds while waiting.
- see this example: https://gist.github.com/Eldinnie/a94f313eb74d7d1ef8579a514f18f1a4 If you send /start -> /brew it will timeout (which will set the state to END) but after 20 seconds it will still send the message from brew

I also think `timed_out_behavior` is badly named. From the name I would expect whatever is in there to happen when it's timedout (especially with the `conversation_timeout` also an argument)

Question:
Why was `run_async_timeout` and `timed_out_behavior` added, and mostly do we still need it with the new mechanics?
Is it possible to stop a promise from running if timeout is reached?

Proposal.
Remove `run_async_timeout` and `timed_out_behavior` kwargs and functionality. I think it's superseeded with the newer timeout mechanics.

jsmnbom:
I'm not sure if new timeout behaviour superseeds run_async_timeout and timed_out_behavior.
conversation_timeout will usually be set to something like a couple of hours right, and is the timeout where conversationhandler forgets about a state.
[...]

PR #1344, opened 2019-02-13
Add WAITING state and behavior

Eldinnie:
* Remove `run_async_timeout` and `timed_out_behavior` arguments
* replace with `WAITING` constant and behavior from states
* never wait for promise to resolve (will hang up entire update queue
* see #1250 for discussion


falkolab:
Hi @Eldinnie please pay attention. Docs still has mention about `run_async_timeout`

PR #1345, opened 2019-02-14
Add `last` kwarg to `run_repeating`

### LINKED ISSUE #1333: `last` kwarg for `JobQueue.run_repeating` method
Just as `JobQueue.run_repeating` takes a `first` optional keyword argument, that specifies the first time a job should be run (as a `datetime.datetime` object, for example), I believe it would be convenient to add a `last` optional keyword argument that specified, if relevant, the last time a job should be run.

## Use cases

I can think of a few use cases, but I'm sure there are more. For example, suppose the bot is part of a game, during which it has to send regular updates (by using a `Job`). Suppose the game lasts a fixed amount of time, which set prior to its start; say, 1 hour. Then, by specifying the `last` keyword argument, the bot programmer could set the "regular updates job" to end when the game ends.

## Current alternative

A workaround I have been using is to specify a separate `run_once` job that will run the `schedule_removal` method on the job at the desired time. A better way is to check inside the job's callback, and remove the job from there using `job.schedule_removal()`.

In either case, I find these solutions to be, if not cumbersome, inelegant, compared to having a much simpler-to-use and elegant dedicated feature.

------

A proposed implementation can be found in #1345.

plammens:
Added a `last` keyword argument to the `run_repeating` method of job queues that takes in a time specifier (in the same formats supported by `first`). It specifies the last time the job should be run. Currently, checking if the job should be run or not involves a certain "delay buffer" (so that if `last` is an exact multiple of `interval`, the internal delays don't mess up the intended behaviour), but it's a temporary solution. Also, I've made some minor refactorings.

Tests run correctly from my side. I've added one test for the new feature, but maybe a few more would be in place. The docs haven't been updated yet.

Closes #1333.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -60,6 +60,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Oleg Sushchenko <https://github.com/feuillemorte>`_
 - `Or Bin <https://github.com/OrBin>`_
 - `overquota <https://github.com/overquota>`_
+- `Paolo Lammens <https://github.com/plammens>`_
 - `Patrick Hofmann <https://github.com/PH89>`_
 - `Paul Larsen <https://github.com/PaulSonOfLars>`_
 - `Pieter Schutz <https://github.com/eldinnie>`_
--- a/telegram/ext/jobqueue.py
+++ b/telegram/ext/jobqueue.py
@@ -130,7 +130,7 @@ class JobQueue(object):
         self._put(job, next_t=when)
         return job
 
-    def run_repeating(self, callback, interval, first=None, context=None, name=None):
+    def run_repeating(self, callback, interval, first=None, last=None, context=None, name=None):
         """Creates a new ``Job`` that runs at specified intervals and adds it to the queue.
 
         Args:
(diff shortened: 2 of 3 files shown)


Eldinnie:
Hi,

Thanks for your contribution. As you can see tests fail on py2. This is because you modified the calculation for `_next_t` in a way a `datetime` on py2 will never be accepted.
Can you please fix this?

Pieter

plammens:
Ah! Sorry I missed that. By the way, could time objects be stored as `datetime.datetime` instead of plain timestamps? That would simply it a bit.

plammens:
@Eldinnie Should be fixed now. Though Travis keeps failing on the unrelated `test_idle`.

plammens:
@Eldinnie Any updates on this? I can make changes if the PR is not up to scratch

Bibo-Joshi:
This was closed when the V12 branch was deleted. Would you mind reopening the PR to master, @plammens ? :)

plammens:
@Bibo-Joshi Oh, that makes sense. Reopened as #1497!

Bibo-Joshi:
@plammens Thanks!

PR #1346, opened 2019-02-17
test: Fix 'test_idle' test

plammens:
This test annoyingly caused some Travis CI checks to appear as failed, when the changes are unrelated to that test. The cause was the fact that the last message logged isn't the 'Received signal 15' message, but the 'Error while getting Updates: Conflict: terminated by setWebhook request'.

### DIFF
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -334,9 +334,16 @@ class TestUpdater(object):
         with caplog.at_level(logging.INFO):
             updater.idle()
 
-        rec = caplog.records[-1]
-        assert rec.msg.startswith('Received signal {}'.format(signal.SIGTERM))
-        assert rec.levelname == 'INFO'
+        rec1, rec2 = caplog.records[-1], caplog.records[-2]
+
+        def test1(rec):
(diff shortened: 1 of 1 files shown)


Eldinnie:
Ehm, I'm guessing you;re running multiple tests at the same time. I've never ran into this problem. Never on CI either.

plammens:
> Ehm, I'm guessing you;re running multiple tests at the same time. I've never ran into this problem. Never on CI either.

I've found this issue on CI (not locally). Check out the Travis CI builds of #1325 and #1345; they both failed on `test_idle` and didn't change (as far as I can see) any code that should affect that test.

Admittedly I don't know very well what the issue at hand is, I just tried to see if this would work (I can't test CI from my fork).

Eldinnie:
Hi,

It seems travis was having some trouble with it's environment variables which made multiple jobs try to run tests on the same bot. hence the error you experienced. It should be fixed now so there's no need for this PR.

Thanks anyway!

plammens:
@Eldinnie Hi; the issue seems to persist: https://travis-ci.org/python-telegram-bot/python-telegram-bot/jobs/500342043 (just FYI).

PR #1357, opened 2019-02-28
Remove sleep from jobqueue tests. Solves #957

### LINKED ISSUE #957: In test_jobqueue.py (maybe others), wait longer (but explicitly) for expected jobs to run

Ambro17:
I added freezegun depdendency to requirements-dev to avoid the need of sleeping to test the jobqueue. This will hopefully help to make tests more deterministics, as sleeps are a bit treasonous with precise timings.

My approach was creating a frozen job queue fixture that yields the job queue itself and the frozen time to be manually advanced.
I also added a friendly-named helper function to 'time travel' and execute jobs due in the future, by manuallly ticking the frozen time and explicitly run due jobs calling job_queue.tick()

Any suggestions are welcome

P.S: If you feel like adding a dependency for this is overkill please read #957. Both `threading.Event` and semaphores were evaluated and discarded because either they fail to cover some cases (threading.Event ca)or they make too complex-to-read code (Semaphores).

Closes #957


Eldinnie:
Hi,

Thanks for you contribution. As you can see, tests fail because pre-commit is not happy. Did you install your environment as suggested in our [contribution guide](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/.github/CONTRIBUTING.rst)?. Can you fix the pre-commit errors so travis is happy? I will make a more in depth review afterwards.

Pieter

jsmnbom:
Turns out the issue wasn't necessarily with your code, but rather a config issue. Please merge from master, so we can actually see if pylint is satisfied :)

Eldinnie:
Whoops, thanks for finding that @jsmnbom

Ambro17:
Hi, i merged master and made some code-style fixes to improve pylint score. I did follow contributing guidelines but pylint is skipped while commiting. I inspected `.pre-commit-config` and it seems it only cares about files under `telegram` directory. 
```yaml
    hooks:
    -   id: pylint
        files: ^telegram/.*\.py$
```  
Maybe we should modify files regex to include files under `tests`?

jsmnbom:
Now that's kinda curious that the CI would then get mad at you, we'll look into it at some point.

It looks like there's some problem on py2? https://travis-ci.org/python-telegram-bot/python-telegram-bot/jobs/500894791#L2188
Failures on other versions are unrelated afaik, but these seem related. Anything you can reproduce->fix ?

Ambro17:
Yeah, i'll see it when i have time. It's strange because it didn't happen locally on any of the times i ran it so it must be some weird config. I'll comment here any update

plammens:
Any updates on this? I'd be happy to help if I can. I feel like the need for deterministic tests is intensifying every time more tests for `JobQueue` and other time-related stuff are added, or the existing ones are extended.

Ambro17:
I'm not working on this anymore. it seems a problem of Travis rather than the code now so it should be easier to fix

Bibo-Joshi:
> I'm not working on this anymore. it seems a problem of Travis rather than the code now so it should be easier to fix

Hey. We switched to GitHub Actions by now and personally I haven't experienced any issues with `test_jobqueue`.

@Ambro17 I didn't really dive into the discussion of this pr and #957. Do you think we can close this or is there something we still need to change?

Ambro17:
This issue can be closed

PR #1388, opened 2019-04-14
Add polls and methods

jsmnbom:
Draft for now since
1) didn't try it
2) am very tired
3) missing tests


jsmnbom:
If CI passes (except test_official and such) this should be ready for review.

Issue #1414, opened 2019-06-04
Clarify documentation about error handler


Lonami:
[`add_error_handler`](https://python-telegram-bot.readthedocs.io/en/latest/telegram.ext.dispatcher.html#telegram.ext.Dispatcher.add_error_handler) reads:

> Registers an error handler in the Dispatcher.
> […]
> The callback function for this error handler. Will be called when an error is raised

This isn't very helpful, [and it leads to questions](https://t.me/pythontelegrambotgroup/230189), such as…
```python
def test(update, context):
    wuhu = {}
    wuhu.append("yay")
```

…apparently not triggering it. In addition it would be helpful to [list common use-cases](https://t.me/pythontelegrambotgroup/230203) for it.

Poolitzer:
I for one see its as a replacement for a try except catch around the whole code. So if I have a not excepted error anywhere in my bot, I can use this handler and for example let the bot PM me so I know I have to fix smth.

eegad:
I wonder if it can be used for normal python exceptions. If yes, can save a lot of try-except blocks

Poolitzer:
It looks like it potentially can, I am quite deep in the code right now. If you want to have a look at it yourself, these are the lines:
https://github.com/python-telegram-bot/python-telegram-bot/blob/9d005d5124eca1787c70a09b471c9450238abf83/telegram/ext/dispatcher.py#L316-L371

In the example lonami started with, these lines must have been called, because I found those in my logs :)

https://github.com/python-telegram-bot/python-telegram-bot/blob/9d005d5124eca1787c70a09b471c9450238abf83/telegram/ext/dispatcher.py#L368-L370 

I cant quite figured out why my error handler wasn't called, because the try block before this exception should have handled it. I will investigate further.

Poolitzer:
After more investigating, I found it.

https://github.com/python-telegram-bot/python-telegram-bot/blob/9d005d5124eca1787c70a09b471c9450238abf83/telegram/ext/dispatcher.py#L467-L474

So the error_handler is intended to catch telegram errors, nothing else. It's not enough for me, so I changed it locally. But at this point, I think its rather a question of the definition/idea behind it.

Poolitzer:
@Lonami we can close this, according to #1173 and the dev, it will take care of all Exceptions in the end.

Issue #1173, opened 2018-07-29
Error in handlers is not shown or handled by default


avishai-ish-shalom:
### Steps to reproduce
```
def bad_handler(bot, update):
    raise new Error('something bad happened')

updater.dispatcher.add_handler(CommandHandler('boom', bad_handler))
updater.dispatcher.add_error_handler(on_error)
```
Error will not be handled by `on_error` nor printed anywhere unless python logging is configured. This is confusing to users and not properly documented.

### Expected behaviour
Error handler should be either be called or the main documentation (readme, tutorial) should contain a link to the [wiki](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Exception-Handling#other-exceptions) or a section about this.

jsmnbom:
@python-telegram-bot/developers ? What do you think? Should arbitrary exceptions raised in handlers be passed to the error handlers?

Eldinnie:
I think it's counterintuïtive that they don't.
We have an errorhandler, let it handle errors. Whichever type (instead of only TelegramError)

tsnoam:
duplicate of #682

Eldinnie:
@tsnoam No it's not

tsnoam:
Sorry, my bad. Not a duplicate.
Thinking of it, I suggest the following behavior:
1. In case of unhandled exception (any kind) - if there's a error handler defined - call it.
2. If no error handler is defined: log error the exception.
3. TelegramError may receive prettier logs then unknown errors.

nikolaevigor:
Hi, +1 here.

Spent several hours on this. I have configured my own logger, so didn't see exceptions and tracebacks (100% my fault for not seeing it in wiki).
Still, in my opinion any kind of non-Telegram related exceptions should not be suppressed. Or may be it could be an explicit option to suppress it or not.
Version: 12.0.0b1

Poolitzer:
This was introduced with #1481 and will be part of the V12 release, thanks for the suggestion.

PR #1416, opened 2019-06-05
Clarify error_handler documentation (#1414)

Poolitzer:


Eldinnie:
Although I thank you for your contribution. We've decided that changing the dispatchers behavior to dispatch all errors instead of only TelegramErrors as discussed in #1173 
Would you be interested in posting some code to that end?

Poolitzer:
Nice, I actually have done this already at my local dispatcher. Let me make a new branch/PR/commit/whatever.

PR #1418, opened 2019-06-06
All api 4.2 and 4.3 changes

Eldinnie:
Tests and coverage check needed

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,20 +1,20 @@
 repos:
 -   repo: git://github.com/python-telegram-bot/mirrors-yapf
-    rev: master
+    sha: 5769e088ef6e0a0d1eb63bd6d0c1fe9f3606d6c8
     hooks:
     -   id: yapf
         files: ^(telegram|tests)/.*\.py$
         args:
         - --diff
 -   repo: git://github.com/pre-commit/pre-commit-hooks
-    rev: v2.0.0
+    sha: 0b70e285e369bcb24b57b74929490ea7be9c4b19
     hooks:
     -   id: flake8
         exclude: ^(setup.py|docs/source/conf.py)$
         args:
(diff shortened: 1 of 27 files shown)


AyraHikari:
Hello, i test that branch, and starting editing my bot script for future update.
But i got an error:
```
  File "/home/ayra/bot/my/__main__.py", line 535, in process_update
    handler.handle_update(update, self)
TypeError: handle_update() missing 1 required positional argument: 'check_result'
```
My script is
```
                handler.handle_update(update, self)
                break
```

I don't know what check_result is mean, when i'm insert check_result="test", bot is not responding when i send a command /start.

Also there is no guide for that in [this guide](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Transition-guide-to-Version-12.0#handler-changes).

Thank you

Eldinnie:
ehm...
Why do you run this? Choosing which handler is/and ashould be done by the dispatcher. It might be helpfull if you shared all your code and the intention of what you;re trying to do.

AyraHikari:
Sure, [here it is](https://github.com/AyraHikari/EmiliaHikari/blob/be067cbaed457571fc0fc5555f51506034d6016f/emilia/__main__.py#L535)

Eldinnie:
@AyraHikari you rewrote the dispatchers <code>process_update</code> method. If you want to do that with the new version, you will have to make sure it works the same as <code>Dispatcher.process_update</code>.

AyraHikari:
Oh i see, maybe i should rewrite all things since many of that deprecated.
Thank you for your response. Sorry for late reply

madl-ash:
why is this not merged?

Nukesor:
Would love to see this merged, especially since 4.4 is already out there!

Bibo-Joshi:
I noticed, that an `telegram.loginurl.rst` file is missing for the docs.

Issue #1435, opened 2019-07-05
problem when when sending 0.0, 0.0 as latitude and longitude 


kercos:
### Steps to reproduce
1.
```python
BOT.send_location(chat_id = TEST_CHAT_ID, latitude=0, longitude=0)
```

### Expected behaviour
Sending the location of the north pole

### Actual behaviour
ValueError: Either location or latitude and longitude must be passed asargument.

### Suggestion on how to solve
Do not use `all` function in https://github.com/python-telegram-bot/python-telegram-bot/blob/9d005d5124eca1787c70a09b471c9450238abf83/telegram/bot.py#L1033

Eldinnie:
hmm,
I think we never considered that `0` is a valid alt/long.
Needs fixing, thanks for reporting

PR #1437, opened 2019-07-05
bool(latitude) -> latitude is not None

Marchello00:


filipre:
This should fix #1435, right?

Marchello00:
> This should fix #1435, right?

Yes exactly

tsnoam:

PR #1451, opened 2019-07-20
Support for tz-aware datetime.datetime objects

torresjrjr:
## Original issue ##
Issue: python-telegram-bot#1409 
version: 12.0.0b1

## Changes ##
The `._put()` method can now accept utc_offset-aware/timezone-aware `datetime.datetime` objects. This helps bots to be hosted on a server with any timezone without complications.

To convert datetimes to UTC, a "UTC" subclass of the `datetime.tzinfo` class was added, and used in `._put()`.

## Other options considered ##
- Changing all the `.now()` comparisons in `._put()` to `.utcnow()`. This was not chosen, since it would cause a major compatibility issue. Bots would have to be rewritten to pass in UTC datetimes, instead of local (computer time) datetimes, into the jobqueue.run_once and other methods. This would likely be unwanted behaivour.

### DIFF
--- a/telegram/ext/jobqueue.py
+++ b/telegram/ext/jobqueue.py
@@ -36,6 +36,22 @@ class Days(object):
     EVERY_DAY = tuple(range(7))
 
 
+class UTC(datetime.tzinfo):
+    """UTC"""
+
+    ZERO = datetime.timedelta(0)
+    HOUR = datetime.timedelta(hours=1)
+
+    def utcoffset(self, dt):
+        return ZERO
+
+    def tzname(self, dt):
+        return "UTC"
+
+    def dst(self, dt):
+        return ZERO
+
+
 class JobQueue(object):
     """This class allows you to periodically perform tasks with the bot.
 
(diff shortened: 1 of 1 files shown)


Eldinnie:
Hi @torresjrjr 
Thanks for making this PR, and our excuse for not responding sooner.
I like the addition and the idea behind it, but I would like to see you add some tests reflecting the changes.

Could you add those please?

tsnoam:
Actually, I like to open this for discussion with the other maintainers.
IMHO, servers should run on UTC . Timezones add so many questions and problems. 
My vote is not to add this code to the library.

Eldinnie:
@tsnoam the point they make in the issue, is that sometimes you want to use <code>run_daily</code> for the user-aware timezone. This PR makes it possible to say:
Run daily on timezone +5 and it will convert to UTC to put in the job_queue.
It assumes the server on UTC and makes it easier to add jobs for all timezones without the hassle of calculating the timedelta manually or program it in every bot

tsnoam:
Well, that's a subset of what a timezone is.
A timezone can also be "Asia/Jerusalem". Now go and map between summer and winter clock. Or, if you're on the edge of switching between them, switch.
Not to mention countries which change their decision when to move between summer & winter clock every year.
I prefer not to "put a healthy head inside a sick bed".

torresjrjr:
Actually, I'm not so invested in this PR as much, since this was an attempt at a solution to a problem I later solved. I wanted to "help my bot to be hosted on a server with any timezone without complications" but I ended up simply adding an extra step in my code.

I created a `timedelta` object of the time from `now()` to the `datetime`, and submitted that to my job creations. It's a quick solution others might find useful :) and could be used in the final PR in the `.put()` method.

I should also stress that I'm a python noob/intermediate. I find PRs scary. So please, take my idea and do what you will. Thanks.

Eldinnie:
We were debating whether this should be a here, in helpers or in the programmers hands. I guess you helped us answering that. I'm closing this PR.

PR #1461, opened 2019-08-03
Fixed example conversationbot2.py

jjocram:
A friends (https://github.com/Butterneck) and I were working on a bot when we need to use a conversation and save datas in context. 

The example (conversationbot2.py) was broken so, searching and trying, we figured out how to solve the problem.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -51,6 +51,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Li-aung Yip <https://github.com/LiaungYip>`_
 - `Loo Zheng Yuan <https://github.com/loozhengyuan>`_
 - `macrojames <https://github.com/macrojames>`_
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
It seems to me that you overlooked the new context based syntax in version 12.0.0b1. Are you running the 11.1.0 version?

jjocram:
Yes, my fault, I used the version available with pip, I should have checked before 😔.

Anyway, I forced to use 12.0.0b1 version but the example still doesn't work, something about `RegexHandler is deprecated.` and the link to wiki on `Transition guide to Version 12.0`

Bibo-Joshi:
The link to the Transition guide is propably because you still use the `pass_user_data` syntax.
As for the depricated `RegexHandler`: This should be fixed by #1426, when mergend, as far as I can see.

jjocram:
Ok, I understand. I think we can close the pull request

PR #1465, opened 2019-08-07
Make MessageEntity objects comparable

Bibo-Joshi:
Right now comparison of `MessageEntity` object fails, e.g.
```
MessageEntity('bold', 2, 4) == MessageEntity('code', 5, 10)
```
yields `True`.

To make `MessageEntity` objects comparable, I used `TelegramObject`s attribute `_id_attr`. Also, I added a `test_equality` to `test_messageentity`.

### DIFF
--- a/telegram/messageentity.py
+++ b/telegram/messageentity.py
@@ -55,6 +55,8 @@ class MessageEntity(TelegramObject):
         self.url = url
         self.user = user
 
+        self._id_attrs = (self.type, self.offset, self.length)
+
     @classmethod
     def de_json(cls, data, bot):
         data = super(MessageEntity, cls).de_json(data, bot)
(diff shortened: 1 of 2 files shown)


telegram/messageentity.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, type, offset, length, url=None, user=None, **kwargs):
         self.url = url
         self.user = user
 
-        self._id_attrs = (self.type, self.offset, self.length)
+        self._id_attrs = (self.type, self.offset, self.length, self.url, self.user)
```
Eldinnie: Like I mentioned in chat, I prefer to remove the `self.url` and `self.user`.
type, offset and length are the unique identifiers for a MessageEntity.

Bibo-Joshi: @Eldinnie Reverted

PR #1474, opened 2019-08-18
Use context.error for v12 context-based error callbacks

wzyboy:
Some of the examples are using context-based callbacks, but the error callbacks are not updated. This PR fixes these.


tsnoam:
Thankyou for your contribution

PR #1476, opened 2019-08-22
Prepare debian packetization

mmaridev:


tsnoam:
@mmaridev Nice!
Can you please move the files under a new subdirectory named &quot;contrib/&quot;?

mmaridev:
I'm afraid but moving the debian directory would break the whole system as debhelper won't be able to find the package source directory (telegram).

If you want you may push a new target to the Makefile that copies the debian subdirectory out from contrib and runs the build but this wouldn't be a clean solution, IMHO.

tsnoam:
@mmaridev The thing is quite simple. Packaging for Debian (or any other distro/OS) is out of scope for this project.

We do not have the resources to constantly maintain  the various packaging.
However, since this is something very nice to have in the git repo, we&#x27;ll be happy to include it in a subdirectory which shows the commitment level (therefor: contrib). It is a very common practice in other opensource projects as well.

tsnoam:
@mmaridev Also, changing the Makefile is also not desired. 
But a bootstrap/build shell script inside the contrib/debian directory will probably be the best way to tackle this.

mmaridev:
Right, @tsnoam .

My principal goal right now is to see your library packaged and available in the repositories as far as I'm developing a software (that will be either distributed via APT) that is based on your API.

Let me get in touch with the Debian/Ubuntu upstream team to understand what they prefer. I already worked on packaging and I think they need the debian directory in the root of the project. If so I'll submit this version to their own vcs (e.g. salsa) and I'll be re-packaging under contrib/ for you (it will become a double work but never mind).

mmaridev:
Btw, have you ever thought to manage builds automatically via CI/CD? It could be a patch to your lack of resources. A one-shot setup and then go ahead with that (for debian it would be quite easy, it just needs to override the version number).

tsnoam:
@mmaridev We do have a CI, but then again, packaging for various distros is out of scope.

mmaridev:
Right, this should fit your requests @tsnoam

tsnoam:

mmaridev:
You're welcome; thank you for the great job done with this library!

PR #1481, opened 2019-08-25
Error handler all exceptions

Poolitzer:
and away we go

### DIFF
--- a/telegram/ext/dispatcher.py
+++ b/telegram/ext/dispatcher.py
@@ -26,6 +26,7 @@ from threading import Thread, Lock, Event, current_thread, BoundedSemaphore
 from time import sleep
 from uuid import uuid4
 from collections import defaultdict
+import traceback as tb
 
 from queue import Queue, Empty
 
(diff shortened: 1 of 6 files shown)


telegram/ext/dispatcher.py
```diff
@@ -337,37 +337,45 @@ def process_update(self, update):
...
-                self.logger.warning('A TelegramError was raised while processing the Update')
-
+            # Dispatch any error - errors should not stop the thread.
+            except Exception as e:
                 try:
-                    self.dispatch_error(update, te)
+                    self.dispatch_error(update, e)
                 except DispatcherHandlerStop:
-                    self.logger.debug('Error handler stopped further handlers')
+                    self.logger.exception('An uncaught error was raised while handling the error')
```
Bibo-Joshi: Why did you change this? `DispatcherHandlerStop` is just used to prevent handlers in higher groups to be executed and is no actual exception.

Eldinnie: correct

PR #1483, opened 2019-08-26
improving error_handler

Poolitzer:
third time a charme


Eldinnie:
Only thing failing is test_official, which is to be expected until we merge bot api 4.4
Merging

Issue #1407, opened 2019-05-18
No control over dispatcher parameters when creating an Updater instance


Ambro17:
### Steps to reproduce
1. Create an instance of the updater with updater = Updater('token')
2. Try to control the instantiation of the asociated dispatcher object
3. Fail becuase there's no way to control its instantiation. It always creates a dispatcher with the same non-configurable args

### Expected behaviour
One should be able to control object creation either by extra kwargs or by passing a complete dispatcher instance.
My use case is adding a decorator factory method on Dispatcher Class, in order to decorate handler functions and automatically add them to the dispatcher handlers list. I have working code, but i would like to avoid forking the lib to use my extension

### Actual behaviour
One has to edit lib code (dispatcher.py), or set atributes dinamically to extend dispatcher functionality

### Configuration
**Operating System:**
Linux - Kubuntu 18 - KDE

**Version of Python, python-telegram-bot & dependencies:**

``$ python -m telegram``
``python-telegram-bot 12.0.0b1 (v12.0.0b1-10-g2860318)``

Bibo-Joshi:
Closed by #1484

PR #1484, opened 2019-08-27
Add dispatcher argument to Updater

Bibo-Joshi:
This allows to pass a pre-initialized `Dispatcher` instance to `Updater`, closing #1407.

If the new `dispatcher` argument is used, attributes like `bot` and `job_queue` are fetched from the `dispatcher` rather than initializing them.
In the docstring tried to make clear that proper initialization of the `dispatcher` is up to the user, i.e. it is not checked if `dispatcher.job_queue is not None`.

Also added tests for creation with `dispatcher`. I hope those are enough …

### DIFF
--- a/telegram/ext/updater.py
+++ b/telegram/ext/updater.py
@@ -63,24 +63,29 @@ class Updater(object):
         token (:obj:`str`, optional): The bot's token given by the @BotFather.
         base_url (:obj:`str`, optional): Base_url for the bot.
         workers (:obj:`int`, optional): Amount of threads in the thread pool for functions
-            decorated with ``@run_async``.
-        bot (:class:`telegram.Bot`, optional): A pre-initialized bot instance. If a pre-initialized
-            bot is used, it is the user's responsibility to create it using a `Request`
-            instance with a large enough connection pool.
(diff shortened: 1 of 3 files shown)


Eldinnie:
Can you fix the open problems marked by CI?

Bibo-Joshi:
Fixed Flake8 and Codacy (by adding a private property to `Dispatcher`). AFAICS the failed tests are due to `test_official` (i.e. missing API 4.4)

Eldinnie:
pre-commit still gives a flake8 error

telegram/ext/updater.py
```diff
@@ -97,58 +102,84 @@ def __init__(self,
                  base_url=None,
                  workers=4,
                  bot=None,
+                 dispatcher=None,
```
Eldinnie: Please add new kwargs at the end of the parameter list.
I know it should be trivial, but you'd be surprised hom many people still like to use args instead of kwargs

Eldinnie:
I'm also confused to why codecov isn;t running on this.

Eldinnie:
We can ignore the codecov problem here. Ready to merge

PR #1490, opened 2019-08-28
Add default values

### LINKED ISSUE #1527: [FEATURE] Enter missing defaults at runtime
This was an idea we had in mind (and discussed) somewhere a while back. Users should be able to pass defaults to a convenient method, so all the requests to the bot API which have this paramater defaults to the one the user provided. Examples would be parse mode, disable notification, hide link preview. This shall be edited with a sufficient list of parameters and where exactly we want to implement this (updater, dispatcher, bot instance?) in the future, this is merely a reminder for now.

Bibo-Joshi:
Supersedes/Closes #1226 and closes #1527

To meet the requirements stated by @Eldinnie in #1226, I added a new class `DefaultValue` to the `helpers` module. It's purpose is to allow to differentiate check if a value is just the default or was set specifically:

With `def f(arg=None)` we don't know, if `arg is None` results from `f()` or from `f(None)`.
With `def f(arg=DefaultValue(None)`, we can handle those differently.

This makes it necessary to add some stuff to the `InlineQueryResult*` for `bot.answer_inline_query` to parse the new `default_parse_mode` correctly. But it does the job :) Maybe `DefaultValue` also comes in handy with a future use case …
How do the @python-telegram-bot/maintainers feel about this?

For the tests, I tried and added them for the `send_*` and `edit_message_*` methods. I'm still not too familiar with `pytest`, so please tell if those can be improved.
However, I couldn't find a clever way to test the `default_parse_mode` for `answer_inline_query` since I'd somehow have to get hold of the resulting message. Any idea is appreciated :)


telegram/inline/inlinequeryresult.py
```diff
@@ -41,3 +41,11 @@ def __init__(self, type, id, **kwargs):
...
 
         self._id_attrs = (self.id,)
+
+    @property
+    def _has_parse_mode(self):
+        return hasattr(self, 'parse_mode')
+
+    @property
+    def _has_input_message_content(self):
+        return hasattr(self, 'input_message_content')
```
Poolitzer: I dont exactly see what this has to do with the PR?

Poolitzer:
What I saw from my little dive into your changes is that this is a good solutions which tackels all the problems we had, everything is covered. If the other maintainer agree, we can built the other default ideas we had (timeout, disable web preview mainly) on top of this. Lets use #1527 to track this if we come so far. Good job though, I really like this :)

jh0ker:
Hey,

I was wondering if it was possible to move this check:

https://github.com/python-telegram-bot/python-telegram-bot/pull/1490/files#diff-3b086505b88889eca9352d47789b348dL101-L102

from all the different `Input` classes into this check:

https://github.com/python-telegram-bot/python-telegram-bot/pull/1490/files#diff-56fa4ffcaf1325dae7804deba7d447afR1467-R1473

so that we wouldn't have as much code duplication.

jh0ker:
Regarding the `Bot` class, i had the idea of using a factory function to make the defaults a bit nicer.

Assuming we want to be able to have defaults for other parameters, I suggest a `Defaults` class like this:

```python
class Defaults():
    def __init__(parse_mode=None, disable_notification=False):
        self.parse_mode = parse_mode
        self.disable_notification = disable_notification
```

Then, we can use it like this:

```python

def Bot(token, base_url=None, base_file_url=None, request=None, private_key=None,
        private_key_password=None, defaults=Defaults()):
    class Bot(TelegramObject):
        [...]

        @log
        def send_message(self,
                        chat_id,
                        text,
                        parse_mode=defaults.parse_mode,
                        disable_web_page_preview=None,
                        disable_notification=defaults.disable_notification,
                        reply_to_message_id=None,
                        reply_markup=None,
                        timeout=None,
                        **kwargs):

    [...]

    return Bot(token, base_url, base_file_url, request, private_key, private_key_password)
```

which means you can create a bot like `Bot(..., defaults=Defaults(parse_mode='html'))` and don't need any extra checks in all the api methods.

Now, i realize this seems a little janky with the factory and all, but i still wanted to put it out there as an idea.

telegram/inline/inlinequeryresult.py
Bibo-Joshi: Good question … Don't recall, why I added it …

Bibo-Joshi: Mh, actually it seems that the `InlineQueryResult…` classes don't always have an `input_message_content` attribute. This is due to the fact, that the base class `InlineQueryResult` doesn't have one and in the child classes the optionals are parsed in the manner

```
if input_message_content:
            self.input_message_content = input_message_content
```
*without* preceding `self.input_message_content = None`.
I.e. if the optional `input_message_content` is not passed, the class instance won't have such an attribute.

I'm not sure that this is acutally wanted behavior, but changing that would surely be the content of another PR …

Bibo-Joshi:
Your idea for a `Defaults` class looks interesting, @jh0ker, especially since it makes it easy to add more defaults.
It would still require the `DefaultValue` to allow to differentiate between say in your example `None` and `defaults.parse_mode`, e.g. the class definition should read

```
class Defaults():
    def __init__(parse_mode=DefaultValue(None), disable_notification=DefaultValue(False)):
...
```

My main concern is usability. To me it seems a bit harder to understand

```
def Bot(token, base_url=None, base_file_url=None, request=None, private_key=None,
        private_key_password=None, defaults=Defaults(parse_mode=MARKDOWN,disable_notification=True))
```

than 

```
def Bot(token, base_url=None, base_file_url=None, request=None, private_key=None,
        private_key_password=None, default_parse_mode=MARKDOWN, default_disable_notification=True)
```

Still, I guess I'd put maintainability before usability 😄 
How do the other @python-telegram-bot/developers fell about that?

jh0ker:
@jsmnbom Seems equally janky, if not more so, but it's really cool ^^ 
Probably still better than the factory cause we still have the one class so we don't break `isinstance` checks and similar.

Bibo-Joshi:
Your approach good, @jsmnbom, but it only seems to work for bot methods that get the `parse_mode` (or other defaultable) argument *directly*. Most notably, I don't see how your solution would handle `answer_inline_query`:

`default_parse_mode` is not passed to `answer_inline_query` as `parse_mode`, since it has no such parameter. But say we modify your solution to pass `default_parse_mode` to *all* methods as kwarg. Then we still have the problem, that we can't defferentiate between

```
InlineQueryResultArticle(
            ...,
            input_message_content=InputTextMessageContent(
                `some text',
                parse_mode=None))
```
and
```
InlineQueryResultArticle(
            ...,
            input_message_content=InputTextMessageContent(
                `some text'))
```
within `answer_inline_query`.

Or am I missing something?

jh0ker:
@Bibo-Joshi We could still save the <code>Defaults</code> object in the <code>Bot</code> for that, couldn&#x27;t we?

Bibo-Joshi:
Simply storing `default_parse_mode` as attribute for `Bot` (or as attribute for `Bot.defaults`, if you will) wouldn't suffice (If I'm understanding you correctly @jh0ker). We need to make sure that the default value for `parse_mode` is different from `None`, wherever `parse_mode` is a parameter. That's why I came up with the `DefaultValue` class …

jh0ker:
@Bibo-Joshi In my mind, there are two different problems here:

1. Setting the defaults for the methods on the `Bot` class
2. Setting the defaults for the `Input` classes passed to `AnswerInlineQuery`

Your PR already handles the second problem pretty well, IMO, with the help of the `DefaultValue` class you introduced, and the one additional check for this special value inside `Bot.answer_inline_query`.
But it would be nice to have a similarly neat solution for the `Bot` class as well, where the code isn't spread out/duplicated as much as it is right now.

Is there a third case or am i missing some other problem with this?

jsmnbom:
Yea in my opinion the best way to do it is using the snippet i posted, but then keeping a lot of your code for the InlineQuery stuff ^^

Bibo-Joshi:
Okidok, then we're all on the same page :) Just wanted to clarify, before starting to work on it ;)
I'll go ahead and implement the changes.

Bibo-Joshi:
mh … I implemented the changes @jsmnbom suggested for the default parameters, but now I'm having trouble adjusting the tests. Since I can't just alter `bot.defaults.parse_mode`, I do something like

```
setattr(bot, 'send_animation', functools.partial(bot.send_animation, **{'parse_mode': 'Markdown'}))
```

in the routines (and reset to 'None' afterwards). However that causes (at least i thinks that's the connection) other tests to fail. E.g. after `test_animation.py` fiddles with the bot, `test_instance_method_send_animation` in `test_user.py` will fail. As far as I can see this is because the `mokeypatch.setattr('telegram.Bot.send_animation', test)` doesn't work and `user.send_animation` still calls the actual method instead of the provided dummy.

Do you have any hint on how to handle this, @jh0ker @jsmnbom ?

telegram/bot.py
```diff
@@ -82,13 +83,49 @@ class Bot(TelegramObject):
...
+            }
+            # ... apply the defaults using a partial
+            if default_kwargs:
+                setattr(instance, method_name, functools.partial(method, **default_kwargs))
+
+        return instance
+
     def __init__(self, token, base_url=None, base_file_url=None, request=None, private_key=None,
-                 private_key_password=None):
+                 private_key_password=None, default_parse_mode=None):
```
jsmnbom: This probably won't work cause `del kwargs[kwarg]` (line 98) deletes this kwarg if i'm not mistaken.

A better way instead of `self.defaults = Defaults(parse_mode=default_parse_mode)` below, might be to simply do a `instance.defaults = defaults` in `__new__` above.

Bibo-Joshi: Actually it does. Running

```
from telegram import Bot
bot = Bot('TOKEN', default_parse_mode='Markdown')
print(bot.defaults.parse_mode)
```

prints `Markdown` as expected. If I understand correctly, `__new__` can not directly alter the arguments passed to `__init__`. Instead for this to work one would have to overwrite the `__call__` method or something (see [here](https://bytes.com/topic/python/answers/751865-modify-arguments-between-__new__-__init__)).

Why did you add the `del kwargs[kwarg]` in the first place, if I may ask? The described behaviour in mind I think we can savely delete it (all tests pass without it, too).

jsmnbom: Originally i had `instance = super(Bot, cls).__new__(cls, *args, **kwargs)` note the args and kwargs. so i had to strip out the kwargs that `object.__new__` didn't want. I would prefer to have it like that, as it should provide better integration with custom Bot classes as far as I understand (instead of just ignoring the args and kwargs we actually pass them up according to the MRO)

Bibo-Joshi: Okay, I see your point about passing the arguments. A solution for the problem that `object` takes no parameters (mentioned in the dev chat) could be to overwrite `TelegramObjects` (which inherits from `object`) `__new__` method to

```
def __new__(cls, *args, **kwargs):
    return super(TelegramObject, cls).__new__(cls)
```

i.e. stop the passing of args here. This passes all our tests but I'm not enough of a pythonian to tell if this might cause trouble later …

Still, I don't really see why the `default_` args should be deleted (and none of the other) before passing them along to `TelegramObject`. What am I missing?

jsmnbom: Bot inherits from `object` and should continue to do so. I was under the impression that it would be possible to override the MRO so that super would go up to a user defined class that does accept `__new__` but now that i think of it, that's not possible.
So as long as Bot inherits from object there's no need to pass the arguments.

Bibo-Joshi:
It seems test cases are not everything … While getting started on `default_disable_notification` I noticed, that the current version does not actually work. :/
 This is because `inspect.getargspec` can't see through the `@log` decorators of the bot methods and hence the defaults substitution can't possibly work (`inspect.getargspec` only sees `self` as argument of all the methods). The test cases can't pick up this behaviours since the new default values are passed manually in the tests.

**Possible solutions for the acutal problem:**
1. Go back to using `DefaultValue` everywhere with the downside of duplicated code
2. Find a workaround for the decorator problem. E.g. there is this [package ](https://pypi.org/project/decorator/) that promises to resolve the problem. Also [here](https://hynek.me/articles/decorators/) is a more elaborate article on the topic. This however would potentially mean adding dependencies in favor of a not *so* much convenience enhancement feature.

**Possible solution to make the tests more effective:**
Provide tests bots that are acutally *created* with the correcsponding default settings. On this I'd be glad for help, since I don't fully understand the pytest setup.

Bibo-Joshi:
On second thought, why are we checking the signature of the methods in the first place? They all accept kwargs anyway, so we could just pass the default values to all of them (and that does work, just tested it). Or would that be regarded unproper coding?

Still, it would be good to provide proper bots for testing. Also I still have no idea how to test `answer_inline_query` with `default_parse_mode` …

jsmnbom:
Per the article that you link in 2., it seems like we are encouraged to simply vendor the decorator library.  This would be my preferred solution since as far as i remember this isn't even the first time that our decorators being "wrong" has tripped me up.

As for the tests i suggestion would be to do something like this:
``` python
 @pytest.mark.parametrize('bot', [{'default_disable_notification': True}], indirect=True)
    def test_bot_defaults(self, bot):
        assert bot.defaults.disable_notification # or whatever we're testing
```
then change the bot fixture and make_bot to
``` python
def bot(request, bot_info):
    return make_bot(bot_info, **request.param if hasattr(request, 'param') else {})

def make_bot(bot_info, **kwargs):
    return Bot(bot_info['token'], private_key=PRIVATE_KEY, **kwargs)
```

I'm not sure what you mean by testing with proper bots? You want it so that we have a userbot sending requests to the bots?

Bibo-Joshi:
Latests commits do two things:

1. Add `decorator` library to dependencies to allow for a `@log` decorator that preserves the function signature and hence making @jsmnbom s solution acutally work. May also come in handy in other situations.
2. Add a test fixture `default_bot`. Using `@pytest.mark.parametrize('defaut_bot', [{'default_...': ...}], indirect=True)` this can be used to run tests for bots with predefined `default_...` settings. Those bots are cached for reuse. Right now that is to too important but we want to add more default values later …
3. Adjust some of the `monkeypatch.setattr('telegram.Bot.method', test)` directives in the tests to `monkeypatch.setattr(chat/user/... .bot, 'method', test)`. This is necessary, since `bot.__new__()` now replaces some of bots methods so that  `chat/user/... .bot.method` doesn't actually call `telegram.Bot.method` anymore. More adjustmets will be needed, when the other default values are added. This is a bit tiresome but not complicated.

AppVeyor fail is unrelated.

Bibo-Joshi:
Latest commits add the `default_ ` arguments for `disable_web_page_preview`, `timeout` and `disable_notification` and close #1527.

Regarding the failed tests:

* I have no idea, what codecov wants from me. Please tell, what can be improved.
* py2.7 windows install dependencies fails b/c of deprecation warning
* py3.5 windows install dep fails b/c of `ujson` (don't know, why)
* py3.7 windows fails for deleting old messages b/c of:

  ```
  AssertionError: Pattern 'Message to delete not found' not found in 'Message can't be deleted'
  ```
  Does `Massage can't be deleted` have a different meaning than `Message to delete not found` here or should I just add it as another expected output to the test?

telegram/inline/inlinequeryresult.py
Bibo-Joshi: Okay, above mentioned issue is solved by #1600. But this is still needed, since e.g. `InlineQueryResultGame` has *neither* a `parse_mode` nor ar `input_message_content` attribute.

Bibo-Joshi:
`telegram.Message` should get a default value for `quote`. This can't be handled by the bot class, but we could just define a `__new__` for Message as we did for `Bot`. on my todo list.

Bibo-Joshi:
Well, a `__new__` method won't help at all. But adding an attribute to `Message` and passing the `default_quote` argument to `Message.de_json` (through all the classes the call the method, being `WebhookServer`, `CallbackQuery`, `Chat`, `Bot` and `Update`) should work.

Bibo-Joshi:
Latest commits add `default_disable_web_page_preview` to `InputTextMessagContent` and `default_quote` for `Message.reply_text` and friends.

Regarding failed tests:
* Codecov complains but iisc, only lines not changed in this PR are marked red
* Codacy complains that `WebhookHandler._default_quote` is set in `initialize` instead of `__init__`, but this is also done for `WebhookHandler.{bot, update_queue}` …
* Pytest fails on `test_pin_and_unpin_message` even on multiple runs. Though I have added another test that pins and unpins (`test_chat_data_default_quote`), I don't see how this should be connected.
* test-official fails b/c API 4.5 is not yet merged

PR #1495, opened 2019-08-29
raise attribute errors when someone tries to assign values to chat/us…

Poolitzer:


telegram/ext/callbackcontext.py
```diff
@@ -73,13 +73,35 @@ def __init__(self, dispatcher):
...
         self.error = None
         self.job = None
 
+    @property
+    def chat_data(self):
+        return self._chat_data
+
+    @chat_data.setter
+    def chat_data(self, value):
+        if self._chat_data is not None:
```
Eldinnie: I think this should just always warn. There is no scenario in which setting chat_data is usefull right?

telegram/ext/callbackcontext.py
```diff
@@ -73,13 +73,35 @@ def __init__(self, dispatcher):
...
+                                 "https://git.io/fjxKe")
+        self._chat_data = value
+
+    @property
+    def user_data(self):
+        return self._user_data
+
+    @user_data.setter
+    def user_data(self, value):
+        if self._user_data is not None:
```
Eldinnie: Same as above

tests/test_callbackcontext.py
```diff
@@ -105,3 +105,13 @@ def test_match(self, cdp):
         callback_context.matches = ['test', 'blah']
 
         assert callback_context.match == 'test'
+
+    def test_data_assignment(self, cdp):
+        update = Update(0, message=Message(0, User(1, 'user', False), None, Chat(1, 'chat')))
+
+        callback_context = CallbackContext.from_update(update, cdp)
+
+        with pytest.raises(AttributeError):
```
Eldinnie: Please use one close per context manager. (so three `with pytest.raises`) here.
Now even if one of them raises and the rest don't the test passes.

telegram/ext/callbackcontext.py
Poolitzer: There is. It is called every time there is an chat and/or user. I dont understand it either, but this check is needed and it works :D

telegram/ext/callbackcontext.py
Poolitzer: same as above (two can play this game)

tests/test_callbackcontext.py
Poolitzer: oh really? Okay, sad. Fixed

telegram/ext/callbackcontext.py
Eldinnie: https://github.com/python-telegram-bot/python-telegram-bot/pull/1495/files#diff-a47242a7a011f5b57df445ace75583baR119
It's because of these lines. Makes these set `self._chat/user_data` and you can remove this check

Poolitzer: you just dont like my solution :(

yeah, yours is better, bah. *grummel*

Eldinnie:
If CI passes, it's ready to merge

PR #1496, opened 2019-08-30
answer callbackqueries when no handler handled them

### LINKED ISSUE #1364: Automatic answerCallbackQuery
This would be, in my opinion, desirable behaviour. If the user uses any other method (e.g. edits the message) **and** they haven't answered the query themselves before, it's done automatically with no parameters (such as text or show alert).

Once it is answered, it cannot be answered again (I believe Telegram ignores it?).

I can think of two caveats:
* User answers the query as the last thing they do (in my opinion it should be the first thing anyway). This may make the change breaking.
* User answers the query through the bot object (so it's a lot harder to track if a callback was answered). This shouldn't be a big deal, since the last call is ignored.

Poolitzer:
This closes #1364. I could turn the skip argument in one for the dispatcher, so user could disable this feature. I dont see any harm this could do though, so I didnt make it an option.

If the query is somehow answered, this call is just ignored by telegram. If it doesnt, this just enhances the user experience while not changing anything for the developer.

### DIFF
--- a/tests/test_dispatcher.py
+++ b/tests/test_dispatcher.py
@@ -237,21 +237,21 @@ class TestDispatcher(object):
         assert self.count == 3
 
     def test_callbackquery_answered_when_no_handler(self, dp, bot, monkeypatch):
-        update = Update(1, callback_query=CallbackQuery(1, User(1, "pool", False), 1, bot=bot))
-        wuhu = []
+        update = Update(1, callback_query=CallbackQuery(1, User(1, "Test", False), 1, bot=bot))
+        passed = []
 
         def test(*args, **kwargs):
-            wuhu.append(1)
+            passed.append(1)
 
         def test_2(update, context):
-            wuhu.append(2)
+            passed.append(2)
 
         monkeypatch.setattr('telegram.Bot.answerCallbackQuery', test)
 
         dp.process_update(update)
         dp.add_handler(CallbackQueryHandler(test_2))
         dp.process_update(update)
-        assert wuhu == [1, 2]
+        assert passed == [1, 2]
 
     def test_add_handler_errors(self, dp):
         handler = 'not a handler'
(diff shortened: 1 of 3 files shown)


Bibo-Joshi:
Stupid question: This just makes the litte spinning loading thingy on callback buttons disappear if the button press is not handled by any handler?

If so:
* The naming of `skip` is not very suggestively to me. How about `is_handled` / `update_handled` ?
* Even if we don't make this behaviour optional, I thinks it should be documented, probably both in the docstring of `CallbackQuery` and `Dispatcher`

Poolitzer:
You got it. Also, clients will show a this vit doesnt respond popup after a while.


I can rename the variable and improve the docs, good point

Poolitzer:
@Bibo-Joshi I added your suggestions, thanks for them. Feel way better now :)

tests/test_dispatcher.py
```diff
@@ -236,6 +236,23 @@ def test_groups(self, dp):
         sleep(.1)
         assert self.count == 3
 
+    def test_callbackquery_answered_when_no_handler(self, dp, bot, monkeypatch):
+        update = Update(1, callback_query=CallbackQuery(1, User(1, "pool", False), 1, bot=bot))
+        wuhu = []
```
Eldinnie: please use serious, descriptive, variable names. Even in tests

Poolitzer:
> I thought this was supposed to answer the callbackquery, if the user does not do that in their own code.

Uhm, but that is exactly what is happening here? If a callbackquery update goes through the group loop without ever being handled, this will jump in and answer it.

tests/test_dispatcher.py
Poolitzer: yeah ok

Eldinnie:
I meant. If the user has a handler for the update, but does not call `answer()` in that handler

Poolitzer:
Hm, nah. He could just edit the message/reply markup, thats a vaild "answer" as well. And how would you even check for that?

PR #1497, opened 2019-08-30
Add `last` keyword argument to `run_repeating`

### LINKED ISSUE #1333: `last` kwarg for `JobQueue.run_repeating` method

plammens:
See #1345 ([reopened on `master` instead of `V12`](https://github.com/python-telegram-bot/python-telegram-bot/pull/1345#issuecomment-526491062)):

> Added a `last` keyword argument to the `run_repeating` method of job queues that takes in a time specifier (in the same formats supported by `first`). It specifies the last time the job should be run. Currently, checking if the job should be run or not involves a certain "delay buffer" (so that if `last` is an exact multiple of `interval`, the internal delays don't mess up the intended behaviour), but it's a temporary solution. Also, I've made some minor refactorings.
>
> Tests run correctly from my side. I've added one test for the new feature, but maybe a few more would be in place. The docs haven't been updated yet.

Closes #1333.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -63,6 +63,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Oleg Sushchenko <https://github.com/feuillemorte>`_
 - `Or Bin <https://github.com/OrBin>`_
 - `overquota <https://github.com/overquota>`_
+- `Paolo Lammens <https://github.com/plammens>`_
 - `Patrick Hofmann <https://github.com/PH89>`_
 - `Paul Larsen <https://github.com/PaulSonOfLars>`_
 - `Pieter Schutz <https://github.com/eldinnie>`_
--- a/telegram/ext/jobqueue.py
+++ b/telegram/ext/jobqueue.py
@@ -89,7 +89,7 @@ class JobQueue(object):
                 Specification of the time for which the job should be scheduled. The precise
                 semantics of this parameter depend on its type (see
                 :func:`telegram.ext.JobQueue.run_repeating` for details).
-                Defaults to now + ``job.interval``.
+                Defaults to now + ``job.interval``.  # TODO: look into this
             previous_t (optional):
                 Time at which the job last ran (``None`` if it hasn't run yet).
 
@@ -144,7 +144,7 @@ class JobQueue(object):
         self._put(job, time_spec=when)
         return job
 
-    def run_repeating(self, callback, interval, first=None, context=None, name=None):
+    def run_repeating(self, callback, interval, first=None, last=None, context=None, name=None):
         """Creates a new ``Job`` that runs at specified intervals and adds it to the queue.
 
         Args:
@@ -171,6 +171,11 @@ class JobQueue(object):
                   tomorrow.
 
                 Defaults to ``interval``
+            last (:obj:`int` | :obj:`float` | :obj:`datetime.timedelta` |                        \
+                  :obj:`datetime.datetime` | :obj:`datetime.time`, optional):
+                Time after which the job should stop running.
+                This parameter will be interpreted depending on its type (same as for ``first``).
+                If ``None``, the job will run indefinitely.
             context (:obj:`object`, optional): Additional data needed for the callback function.
                 Can be accessed through ``job.context`` in the callback. Defaults to ``None``.
             name (:obj:`str`, optional): The name of the new job. Defaults to
@@ -189,6 +194,7 @@ class JobQueue(object):
         job = Job(callback,
                   interval=interval,
                   repeat=True,
+                  finish_time=last,
                   context=context,
                   name=name,
                   job_queue=self)
@@ -268,6 +274,9 @@ class JobQueue(object):
                 self._set_next_peek(t)
                 break
 
+            delay_buffer = 0.01  # tolerance for last
+            if job.finish_time is not None and now > job.finish_time + delay_buffer:
+                job.schedule_removal()  # job shouldn't run anymore
             if job.removed:
                 self.logger.debug('Removing job %s', job.name)
                 continue
@@ -288,7 +297,7 @@ class JobQueue(object):
             if job.repeat and not job.removed:
                 self._put(job, previous_t=t)
             else:
-                self.logger.debug('Dropping non-repeating or removed job %s', job.name)
+                self.logger.debug('Dropping non-repeating, removed, or finished job %s', job.name)
 
     def start(self):
         """Starts the job_queue thread."""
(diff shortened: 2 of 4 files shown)


plammens:
I agree that the `_to_timestamp` helper should be moved to the `telegram.utils.helpers` module. I have been working on that, but I came across some things I'd like to discuss before going forward. There are also some minor things which I will mention now that I've stopped.

1. There was already a `to_timestamp` helper in that module. It seems it is exclusively used to convert `datetime.datetime` objects into integer timestamps to be sent in HTTP requests to Telegram's API. <br><br>
	- To clear any ambiguity on the term `timestamp`, I renamed some identifiers in such a way that `timestamp` means "integer timestamp" (as I think that's what's usually meant when using that word) and explicitly used `float_timestamp` to refer to timestamps with sub-second precision. <br><br> Namely, I renamed the moved function to `to_float_timestamp`, and the former `_timestamp` helper (for the former `telegram.utils.helpers.to_timestamp`) to `_datetime_to_float_timestamp` (to explicitly clarify that it only converts `datetime.datetime` objects). It is now used in the `to_float_timestamp` helper. <br><br>
	- The original `to_timestamp` function's name has been left the same, but its implementation has been redirected through the new `to_float_timestamp` for better maintainability. Now it also is able to convert any of the other relative time values (e.g. `int`s or `float`s as "offset from a particular reference time"). <br><br> I don't know if you agree on that: as I said above, this function is only used in the repository to convert from a `datetime.datetime` to a timestamp, but not any of the other ways to specify time, as in the `first` argument for `telegram.ext.jobqueue.JobQueue.run_repeating`. So,  should restrict the usage of `to_timestamp` to `datetime.datetime` objects? Or is it ok to allow the other types too? <br><br><br>
2. There was inconsistent use of UTC vs local times. For instance, in the former `_timestamp` helper (now `_datetime_to_float_timestamp`), assumed that naïve `datetime.datetime` objects were in the local timezone, while the `from_timestamp` helper —which I would have thought was the corresponding inverse function— returned naïve objects in UTC. <br> 
This meant that, for instance, `telegram.Message` objects' `date` field was constructed as a naïve `datetime.datetime` (from the timestamp sent by Telegram's server) in *UTC*, but when it was stored in `JSON` format through the `to_json` method, the naïve `date` would be assumed to be in *local time*, thus generating a different timestamp from the one it was built from. <br>
    So I sought to unify all handling of naïve `datetime.datetime` by assuming they're always in UTC.  <br><br>
	* This, though, breaks the tests related to the parameter `first` of `JobQueue.run_repeating`, since they assume that naïve `datetime` objects are in local time. That makes me assume that users of PTB have been feeding local `datetime` objects to this parameter, so changing it to UTC would break it for them. But I definitely believe that something has to be done about this: either we unify all naïve `datetime`s to UTC or to local time. And since unifying to local would almost surely break backwards compatibility elsewhere, I think UTC is the better of the two options.<br><br>
	* Maybe even better: ditch naïve `datetime`s altogether and only work with `datetime`s that have an explicit `tzinfo`?<br><br>
    * Finally, I also made sure that `datetime`s that *are* timezone-aware are also treated validly.<br><br>
    * With Python 2's lack of the `datetime.timezone` class, and in particular, of `datetime.timezone.utc`, I had to "hard-code" a UTC singleton to use as a `tzinfo` instance to indicate UTC. I wonder whether it would be better to move that to a new module, e.g. `telegram.utils.datetime`, since it's not a "helper" strictly speaking. 
<br><br>
3. Besides the aforementioned timezone stuff, the original `_datetime_to_float_timestamp` had an issue of its own. The replacement method used to calculate the timestamp in Python 2 (in absence of `datetime.datetime.timestamp`) truncated the timestamp to an integer since that's the precision of `datetime.datetime.timetuple`/`time.mktime`. Of course, as it was before it didn't matter, but with the changes outlined in `1.` it does. So I replaced the alternative calculation with calculating a timedelta from the epoch and using `.total_seconds()` on it. At this point, I don't know if it's worth keeping the switch between Python 3.3+'s `datetime.datetime.timestamp` and the alternative calculation, since the former's implementation is basically the same as the latter.
<br><br>
4. When the job queue `tick`s, there is an inevitable delay between the exact time at which it was scheduled to peek and when `tick` is entered and the current time gets "frozen" (`now = time.time()`). Meaning that, for instance, if you set it to run every 5 seconds and stop running *after* 20 seconds from now, in theory it should run exactly five times: at `Δt = 0s`, `Δt = 5s`, `Δt = 10s`, `Δt = 15s` and `Δt = 20s`. But, in practice, when it gets to the fifth time, due to delays, `Δt` will be something like `20 + 6.28e-5`, meaning that if the code makes a strict comparison (`Δt <= 20.0s`), then the job won't be executed the fifth time. When I first made this PR, I thought that was undesired behaviour, so I added a "delay buffer" so that the check is more like `Δt <= 20.0s + ε`, where `ε` is something like `0.001s`. But now that I think of it, I'd rather have the strict check, partly because of the fact that if the user needed to control the exact times the job was executed, they wouldn't use an end date/time, burt rather something like counting up from within the callback. In this case I'd rename the kwarg to `end`, since it indicates that after that time the job shouldn't run anymore — but the job won't necessarily run at the specified time for the last time.
<br><br>
5. Is this project going to keep supporting Python 2 for the next version? All of this compatibility handling and restraining oneself to older feature sets is a bit of a pain :sweat_smile: 

I realized that rambling on about the changes I've made without showing them is a bit useless. Tomorrow I'll push what I've got (we'll roll back if necessary). In any case, what do you think about all this?

Bibo-Joshi:
Thank for your detailed report, @plammens  !
For the discussion of the time related stuff, I'd like to move that to #1505 (or a corresponding PR). I guess, we'll have to think about that, before moving on with this one …

About `4.` : The way you put it, I tend to agree with your proposal about a strict check.

jh0ker:
About 5.: We're looking to drop Python 2 support in version 13 (see #1538). Perhaps it makes sense to wait for that?

tsnoam:
@plammens 
would you like to rebas/merge this pr with the latest master so we can look at it with a fresh eye?

plammens:
@tsnoam I've rebased this using the stuff introduced in #1506; I've also squashed a little.

I think now the only thing to discuss is [4. (see comment above)](https://github.com/python-telegram-bot/python-telegram-bot/pull/1497#issuecomment-527256274). Do you agree that it should be a strict check? And thus `last` should be renamed to `end`?

Bibo-Joshi:
TL;DR: Closing in favor of #1981 

After internal discussion, we came to the conclusion that we don't want to continue implementing scheduling logic on our own and instead rely on a 3rd party library for that (see #1936, #1981).
One of the upsides is, that a `last` keyword is already included.
We appreciate the effort that was put in this PR and hope that you'll find the new `JobQueue` useful once it's finished :)

PR #1500, opened 2019-09-01
Add info on documentation to CONTRIBUTING.rst

plammens:
I didn't find this information elsewhere in the repository; I thought it might be useful for new contributors.

### DIFF
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -66,6 +66,26 @@ Here's how to make a one-off code change.
    - You can refer to relevant issues in the commit message by writing, e.g., "#105".
 
    - Your code should adhere to the `PEP 8 Style Guide`_, with the exception that we have a maximum line length of 99.
+   
+   - Document your code. This project uses `sphinx`_ to generate static HTML docs. To build them, first make sure you have the required dependencies:
+
+     .. code-block:: bash
+
+        $ pip install -r docs/requirements-docs.txt
+
+     then run the following from the PTB root directory:
+   
+     .. code-block:: bash
+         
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.rst
```diff
@@ -66,6 +66,12 @@ Here's how to make a one-off code change.
    - You can refer to relevant issues in the commit message by writing, e.g., "#105".
 
    - Your code should adhere to the `PEP 8 Style Guide`_, with the exception that we have a maximum line length of 99.
+   
+   - Document your code. This project uses `sphinx`_ to generate static HTML docs. To build them, use the ``Makefile`` in the ``docs`` directory:
```
Bibo-Joshi: Instead of `use the ``Makefile``... `, maybe just say `run … from within your PTB directory`? Else it reads like one should `cd docs` before running the snippet

plammens: Yes, that was confusing. I've amended the commit!

Poolitzer:
this does not cover windows.

On there, you need to run sphinx-build docs\source temp_html or so :)

plammens:
> this does not cover windows.
> 
> On there, you need to run sphinx-build docs\source temp_html or so :)

Right, forgot about that. I'm using Windows, and I have installed `make` through `cygwin`, but probably its best to add the alternative in any case!

plammens:
> uhm, there exists requirements-docs.txt in the docs folder, which people need to install with pip -r. Sorry that I only think of that now ;P

Ha! I also missed that. I've added it now.

Eldinnie:
Hi,
Thanks again for contributing!

PR #1503, opened 2019-09-02
Remove note about V12b in examples

Bibo-Joshi:


Eldinnie:
Yup, that needs to be out. Merging

PR #1506, opened 2019-09-04
Fix UTC/local inconsistencies for naive datetimes

### LINKED ISSUE #1505: Inconsistent use of UTC/local naïve datetimes
PR #1485 made `telegram.utils.helpers.from_timestamp` return a naïve `datetime.datetime` object in UTC, but this created an inconsistency with the rest of the repository in the handling of naïve `datetime.datetime`s, where it is assumed that these objects represent *local* time. In fact, there is a direct inconsistency between `from_timestamp` and `to_timestamp` helpers, since the latter still assumes that the `datetime` object is in local time.

See [#1497 (comment)](https://github.com/python-telegram-bot/python-telegram-bot/pull/1497#issuecomment-527256274) (point `3.`) for extended discussion.

I've already fixed the issue in #1497 (and added tests), but should I make a separate PR to isolate the change?

### Steps to reproduce

0. Go to a place with nonzero UTC offset. I live in CEST (GMT +2).

1. Construct a `telegram.Message` from a JSON data dictionary. This is a message I sent to a test bot of mine (stripped of the optional fields), as received from an HTTP request to Telegram:

    ```python
    import telegram as tg
    
    message_json = {
        'message_id': 462,
        'from': {
            'id': 41214001,
            'is_bot': False,
            'first_name': 'Paolo',
        },
        'chat': {
            'id': 41214001,
            'type': 'private'
        },
        'date': 1567596574,
    }
    
    original_timestamp = message_json['date']

    bot = tg.Bot('000:not_a_token')
    message = tg.Message.de_json(message_json, bot=bot)
    ```

2.  Re-form the `Message`'s JSON `dict` and extract the timestamp:
    ```python
    restored_timestamp = message.to_dict()['date']
    ```

3. Check that the timestamps are the same:
    ```python
    assert restored_timestamp == original_timestamp, \
        f"Timestamps differ by {restored_timestamp - original_timestamp} seconds"
    ```

### Expected behaviour
The timestamps should be the same.

### Actual behaviour
The timestamps differ by the local UTC offset (reversed):

```text
AssertionError: Timestamps differ by -7200 seconds
```

When the `Message` is constructed, the `datetime` object it gets represents UTC time, but when it gets re-converted to a timestamp, that same object is assumed to be in local time, which in my case is `7200` seconds ahead of UTC, so the resulting timestamp is `7200` seconds behind the original.

### Configuration
<details>

**Operating System:**  Windows

``$ python -m telegram``

```
python-telegram-bot 12.0.0 (v12.0.0-0-gc84e21d)
certifi 2019.03.09
future 0.17.1
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)]
```

</details>

plammens:
Closes #1505.


telegram/ext/jobqueue.py
```diff
@@ -379,7 +376,7 @@ def __init__(self,
         self.context = context
         self.name = name or callback.__name__
 
-        self._repeat = repeat
+        self._repeat = None
```
jh0ker: What prompted this change?

telegram/ext/jobqueue.py
```diff
@@ -264,7 +260,8 @@ def tick(self):
 
             if job.enabled:
                 try:
-                    current_week_day = datetime.datetime.now().weekday()
+                    # TODO: set timezone for checking weekday
```
jh0ker: I would note that making the timezone configurable per `Job` is important. Being able to set a default in the `JobQueue` might be nice, but i don't think it's essential.
Do you think adding this is still in the scope of this PR?

jh0ker:
Some more questions that popped up during the discussion in the dev group. Are timezone-aware datetimes in the job queue supported with this? Were they supported before? There seem to be no tests for this yet.

telegram/ext/jobqueue.py
plammens: To be honest it was only for aesthetic reasons. Since all the other properties were also being set this way—i.e. initialising the internal attribute to `None` and then triggering the `<property>.setter` method by assigning to the "public" attribute (e.g. `self._interval = None; self.interval = interval`)—, I changed it to make it consistent. But now that I think of it, it's more than just aesthetic, because this way we ensure that any checks done by the `<property>.setter` method get executed at initialisation too.

telegram/ext/jobqueue.py
plammens: > I would note that making the timezone configurable per `Job` is important.
> Do you think adding this is still in the scope of this PR?

Yep, I've just pushed the commit.

telegram/ext/jobqueue.py
jh0ker: oh yeah i see it now, thanks 👍

telegram/ext/jobqueue.py
jh0ker: Awesome! I think it would be great if we had a test case like `test_run_daily` but with a timezone-aware time. Would you mind adding that here as well, to prevent future regressions in regards to timezone handling?

plammens:
> Are timezone-aware datetimes in the job queue supported with this? 

Yes; namely, `when` in `run_once`, `first`  in `run_repeating`, and `time` in `run_daily` now all accept timezone-aware objects (and use that information adequately). It all boils down to `_datetime_to_float_timestamp`, which now supports aware datetimes, so I believe everything else in the repository that handles date/time objects by converting them to timestamps also supports aware datetimes now.

https://github.com/plammens/python-telegram-bot/blob/56466ac449028e4c5eb0470c919ec2b7466d6737/telegram/utils/helpers.py#L58-L61

> Were they supported before? 

No; for example you would get a `TypeError` when passing an aware datetime to `when` in `run_once`.

> There seem to be no tests for this yet.

Will add in the near future.

jh0ker:
Thank you very much! This looks very nice to me, but timezones confuse me so i requested a review from @tsnoam :wink:

telegram/ext/jobqueue.py
plammens: > I think it would be great if we had a test case like `test_run_daily` but with a timezone-aware time.

I included that in the same commit; it's a bit flaky though, suggestions for improvement are welcome :wink:

jh0ker: Hmm that test seems to only test for the correct selection of the weekday, while `test_run_daily` compares the specified time specifically. One might imply the other but as i said, timezones confuse me ^^

telegram/utils/helpers.py
```diff
@@ -40,54 +44,148 @@ def get_signal_name(signum):
...
 
-    return int(_timestamp(dt_obj))
+    if isinstance(t, dtm.datetime):
+        return _datetime_to_float_timestamp(t)
 
+    raise TypeError('Unable to convert {} object to timestamp'.format(type(t).__name__))
 
-def from_timestamp(unixtime):
+
+def to_timestamp(t, *args, **kwargs):
```
tsnoam: We prefer not to use `*args, **kwargs`. Please change it to the same signature of `to_float_timestamp`.

Additionally, it breaks backward compatibilty. Please rename `t` to `dt_obj`.

telegram/utils/helpers.py
```diff
@@ -40,54 +44,148 @@ def get_signal_name(signum):
...
+    """
+    Converts a time object to an integer UNIX timestamp.
+    Returns the corresponding float timestamp truncated down to the nearest integer.
+    See the documentation for :func:`to_float_timestamp` for more details.
+    """
+    return int(to_float_timestamp(t, *args, **kwargs)) if t is not None else None
+
+
+def from_timestamp(timestamp):
```
tsnoam: This is a breaking change, in case a user is using keyword args. Please revert it to the old argument name.
```suggestion
def from_timestamp(unixtime):
```

telegram/utils/helpers.py
```diff
@@ -40,54 +44,148 @@ def get_signal_name(signum):
...
 
+    raise TypeError('Unable to convert {} object to timestamp'.format(type(t).__name__))
 
-def from_timestamp(unixtime):
+
+def to_timestamp(t, *args, **kwargs):
+    """
+    Converts a time object to an integer UNIX timestamp.
+    Returns the corresponding float timestamp truncated down to the nearest integer.
+    See the documentation for :func:`to_float_timestamp` for more details.
```
tsnoam: Please change the docstring to:
```suggestion
    Wrapper over :func:`to_float_timestamp` which returns an integer (trunacted down to the nearest integer).

    See the documentation for :func:`to_float_timestamp` for more details.
```

telegram/ext/jobqueue.py
```diff
@@ -70,30 +71,25 @@ def __init__(self):
...
-            next_t = next_t.total_seconds()
+        Args:
+            job (telegram.ext.Job): job to enqueue
+            next_t (optional):
+                Time for which the job should be scheduled. The precise semantics of this parameter
+                depend on its type (see :func:`telegram.ext.JobQueue.run_repeating` for details).
+                Defaults to now + ``job.interval``.
+            previous_t (optional):
+                Time at which the job last ran (``None`` if it hasn't run yet).
+        """
```
tsnoam: Please move the empty line to before the `"""`.

telegram/ext/jobqueue.py
```diff
@@ -372,20 +373,22 @@ def __init__(self,
                  repeat=True,
                  context=None,
                  days=Days.EVERY_DAY,
+                 tzinfo=UTC,
```
tsnoam: Please move `tzinfo=UTC` to be the last kwarg, in order to maintain backward compatibility.

telegram/ext/jobqueue.py
tsnoam: Actually, I don't understand where it is being set now. Or something is wrong with my diff?

tsnoam: oic. yes, it's ok.

telegram/utils/helpers.py
```diff
@@ -40,54 +44,148 @@ def get_signal_name(signum):
...
+if hasattr(dtm, 'timezone'):
     # Python 3.3+
-    def _timestamp(dt_obj):
+    def _datetime_to_float_timestamp(dt_obj):
+        if dt_obj.tzinfo is None:
+            dt_obj = dt_obj.replace(tzinfo=UTC)
         return dt_obj.timestamp()
+
+    UtcOffsetTimezone = dtm.timezone
+    UTC = dtm.timezone.utc
```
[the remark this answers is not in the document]
plammens: Agree; for some reason I hadn't thought of the `helpers` module as public API

telegram/utils/helpers.py
```diff
@@ -40,54 +44,148 @@ def get_signal_name(signum):
...
+            If ``t`` is ``None``, ``None`` is returned (to facilitate formulating HTTP requests
+            when the object to be serialized has a ``date`` which is ``None`` without having to
+            check explicitly).
     """
-    if not dt_obj:
+
+    if reference_timestamp is None:
+        reference_timestamp = time.time()
+
+    if t is None:
```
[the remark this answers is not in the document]
plammens: I was also keen to remove that, but then I saw that a good chunk of calls to `to_timestamp`, i.e. when making a dict from a `telegram` API object (to be then transformed into JSON), rely on this behaviour (so that it's less verbose: `to_timestamp(x)` rather than `to_timestamp(x) if x is not None else None`. For example, see https://github.com/python-telegram-bot/python-telegram-bot/blob/965ad17af810df0a4c04ce37f335e78d406a2f69/telegram/chatmember.py#L156-L159

So, to keep it consistent, I kept the behaviour with `to_float_timestamp`  too. 

Should I make a separate function that wraps `to_timestamp` with this `None` check, e.g. `maybe_to_timestamp`, and remove it from `to_timestamp` and `to_float_timestamp`?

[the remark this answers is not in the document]
tsnoam: that sounds like a good solution. lets to do it with a wrapper as you've suggested

telegram/utils/helpers.py
```diff
@@ -40,54 +44,148 @@ def get_signal_name(signum):
...
+        reference_dt = dtm.datetime.fromtimestamp(reference_timestamp, tz=t.tzinfo) if t.tzinfo \
+            else dtm.datetime.utcfromtimestamp(reference_timestamp)  # assume UTC for naive
+        reference_date, reference_time = reference_dt.date(), reference_dt.timetz()
+        if reference_time > t:  # if the time of day has passed today, use tomorrow
+            reference_date += dtm.timedelta(days=1)
+        t = dtm.datetime.combine(reference_date, t)
 
-    return int(_timestamp(dt_obj))
+    if isinstance(t, dtm.datetime):
+        return _datetime_to_float_timestamp(t)
```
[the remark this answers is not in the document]
plammens: Yes, I should definitely add this

telegram/ext/jobqueue.py
plammens: Fixed in 6ed29bb

telegram/ext/jobqueue.py
plammens: Fixed in 6ed29bb . I actually didn't know this was a PEP 257 thing! I've been doing my docstrings wrong all along... :laugh:

telegram/utils/helpers.py
plammens: Fixed in 6ed29bb

telegram/utils/helpers.py
plammens: Fixed in 7806071

telegram/utils/helpers.py
```diff
@@ -40,54 +44,148 @@ def get_signal_name(signum):
...
+    if t is None:
         return None
+    elif isinstance(t, dtm.timedelta):
+        return reference_timestamp + t.total_seconds()
+    elif isinstance(t, Number):
+        return reference_timestamp + t
+    elif isinstance(t, dtm.time):
+        reference_dt = dtm.datetime.fromtimestamp(reference_timestamp, tz=t.tzinfo) if t.tzinfo \
+            else dtm.datetime.utcfromtimestamp(reference_timestamp)  # assume UTC for naive
+        reference_date, reference_time = reference_dt.date(), reference_dt.timetz()
```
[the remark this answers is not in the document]
plammens: Fixed in 6ed29bbf

telegram/utils/helpers.py
```diff
@@ -40,54 +44,148 @@ def get_signal_name(signum):
...
+
+    if t is None:
         return None
+    elif isinstance(t, dtm.timedelta):
+        return reference_timestamp + t.total_seconds()
+    elif isinstance(t, Number):
+        return reference_timestamp + t
+    elif isinstance(t, dtm.time):
+        reference_dt = dtm.datetime.fromtimestamp(reference_timestamp, tz=t.tzinfo) if t.tzinfo \
+            else dtm.datetime.utcfromtimestamp(reference_timestamp)  # assume UTC for naive
```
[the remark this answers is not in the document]
plammens: Fixed in 6ed29bbf

telegram/utils/helpers.py
```diff
@@ -40,54 +44,148 @@ def get_signal_name(signum):
...
+        reference_date, reference_time = reference_dt.date(), reference_dt.timetz()
+        if reference_time > t:  # if the time of day has passed today, use tomorrow
+            reference_date += dtm.timedelta(days=1)
+        t = dtm.datetime.combine(reference_date, t)
```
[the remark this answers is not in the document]
plammens: Fixed in 6ed29bbf

telegram/utils/helpers.py
```diff
@@ -40,54 +44,148 @@ def get_signal_name(signum):
...
 
-    return int(_timestamp(dt_obj))
+    if isinstance(t, dtm.datetime):
```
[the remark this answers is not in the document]
plammens: Fixed in 6ed29bbf

telegram/utils/helpers.py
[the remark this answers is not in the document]
plammens: Fixed in cc40d63a. I moved that at the top with the `None` check for `reference_timestamp`, so that we don't have to store an `original_reference_timestamp`. Is that okay?

telegram/utils/helpers.py
plammens: Fixed in 78060716

plammens:
@tsnoam Before merging I think I should add:

- [ ] Tests to check that timezone aware datetimes work
- [ ] A general note somewhere (where?) in the documentation stating that all naive datetimes will be assumed to be in UTC

Anything else/any thoughts?

telegram/utils/helpers.py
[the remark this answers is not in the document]
plammens: Ok. Technically this is a breaking change, since the name of `to_timestamp` suggests it's public API; here's hoping that nobody relied on that behaviour. Should I name the wrapper with a leading underscore? And is `_maybe_to_timestamp` a good name? I can't think of another one.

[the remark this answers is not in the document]
tsnoam: You've got a good point regarding the naming and bkwd-compat. Thanks.
 Looking at the code as it is now, admitting past mistakes but yet considering backward compatibility, I think we should stick with:

1. New function: `to_float_timestamp()` - will raise `TypeError` if `t` is `None`. (the same `raise` you've put at the end of the function.
2. Older functions: `to_timestamp()` & `from_timestamp()` - since both had previously returned `None` for `None` input, their names will remain the same and they will continue accepting `None`. (so we won't be using `maybe_XXX_timestamp()` afterall, even though I really liked it).

telegram/utils/helpers.py
```diff
@@ -138,6 +142,9 @@ def to_float_timestamp(t, reference_timestamp=None):
 
     if reference_timestamp is None:
         reference_timestamp = time.time()
+    else:
+        if isinstance(t, dtm.datetime):
```
tsnoam: nitick: this can be:
```python
elif isinstance...
```

telegram/ext/jobqueue.py
```diff
@@ -70,30 +71,25 @@ def __init__(self):
...
+                Time for which the job should be scheduled. The precise semantics of this parameter
+                depend on its type (see :func:`telegram.ext.JobQueue.run_repeating` for details).
+                Defaults to now + ``job.interval``.
+            previous_t (optional):
+                Time at which the job last ran (``None`` if it hasn't run yet).
+        """
 
-        next_t += last_t or time.time()
+        # get time at which to run:
+        next_t = to_float_timestamp(next_t or job.interval, reference_timestamp=previous_t)
```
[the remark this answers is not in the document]
plammens: Actually, I've come to think that if both `next_t` and `job_interval` are `None`, we should raise an error, since we have no idea when to schedule this job. We cannot enqueue a job with priority (time) `None`—it'll raise an obscure (from the perspective of `_put`'s caller) error about `None` not being comparable to `float` if the queue is not empty, or if it gets past this, the [comparison between `t` and `now` in `JobQueue.tick()`](https://github.com/python-telegram-bot/python-telegram-bot/blob/bbcff96804bc8dc40ad948dabbce264a7f83538c/telegram/ext/jobqueue.py#L258) will break for the same reason.

telegram/utils/helpers.py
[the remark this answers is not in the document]
plammens: Change made in 82fe342.

telegram/utils/helpers.py
plammens: Fixed in 82fe342.

plammens:
@tsnoam I've addressed the remaining points and added tests. I tested the `helpers` thoroughly for aware datetimes, but I only did that for `run_once` in `JobQueue`. Should I make a test for every single `run_*` to check support for aware datetimes?

 Sorry for the delay!

tsnoam:
@plammens 
Thank you very much for your contribution. Also you patience with us is highly appreciated.
I've pushed a minor cosmetic fix, and now we're just waiting for unitests to pass (no reason they shouldn't) before merging.

tsnoam:
I don't know why pre-commit test failed. On my machine it passes successfully.
Merging anyway.

PR #1464, opened 2019-08-05
API 4.4 changes

Bibo-Joshi:
Attempt to incorporate API 4.4 changes.

**Stuff I'm not sure I handled it correctly:**
- `de_json` methods (`Chat` and `ChatPermission`)
- Deprecation warning for `all_members_are_administrators` in `Chat`
- new parameter sytax for `restrict_chat_members` in `Bot`
- Does something need to be changed for the new webhooks?
- Preparation for API 4.5, which brings nested `MessageEntities`: I adjusted `Message._parse_html` and `Message._parse_markdown` to allow for this. However, this only allows something like `_bold *within* italic_` and not `*bold _overlapping* italic_`. The announcement is not every clear, if that is also allowed …

Did I miss anything?

PR #1508, opened 2019-09-06
API 4.5

### LINKED ISSUE #1689: 4.5 bot api update arrived
Aint much, but its honest work
- [x]  Two new messageEntities, strikethrough and underline
- [x] nested messageEntities, we have a PR for that somewhere
- [x] New parsemode, MardownV2
- [x] New file_unique_id for Animation, Audio, Document, PassportFile, PhotoSize, Sticker, Video, VideoNote, Voice, File
- [x] small_file_unique_id and big_file_unique_id to the object ChatPhoto.
- [x] custom_title for ChatMember
- [x] new method setChatAdministratorCustomTitle
- [x] new slow_mode_delay for Chat object

Bibo-Joshi:
PR for the API 4.5 changes. Closes #1689 

- [x]  Two new messageEntities, strikethrough and underline
- [x] nested messageEntities
- [x] New parsemode, MardownV2
- [x] New file_unique_id for Animation, Audio, Document, PassportFile, PhotoSize, Sticker, Video, VideoNote, Voice, File
- [x] small_file_unique_id and big_file_unique_id to the object ChatPhoto.
- [x] custom_title for ChatMember
- [x] new method setChatAdministratorCustomTitle
- [x] new slow_mode_delay for Chat object 
- [x] Before merge, merge master and revert #1738 
- [ ] When merged, update this [wiki page](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Code-snippets#message-formatting-bold-italic-code-).

--------

**Original description:**

Cherry-picked from #1464 since we're not sure yet how Telegram will actually handle nested entities in API 4.5.

This only allows something like `_bold *within* italic_` and not `*bold _overlapping* italic_`. The announcement is not every clear, if that is also allowed …

Leaving this here as a reminder, that we already have some code for that case.


Bibo-Joshi:
The PR is ready for review now. The test failures seem unrelated (`TestBot.test_pin_and_unpin_message` and `TestConversationHandler.test_per_message_false_warning_is_only_shown_once` are nothing we fiddled with)

telegram/message.py
```diff
@@ -1157,10 +1289,24 @@ def text_markdown_urled(self):
         """
         return self._parse_markdown(self.text, self.parse_entities(), urled=True)
 
+    @property
+    def text_markdown_urled_v2(self):
```
dmytrohoi: Maybe need to use `text_markdown_v2_urled` name, for insight of `v2` relation to `markdown` rather than `urled`?

Bibo-Joshi:
> 1. Should we add a shortcut for `bot.set_chat_administrator_custom_title()` to ChatMember class from chatmember.py?
> 2. I think because of this description, `This is an obsolete mode saved for backward compatibility. To use this mode, pass Markdown to the parse_mode field from` [chapter "Markdown" in Telegram Bot API documentation](https://core.telegram.org/bots/api#markdown-style), we need to use `MarkdownV2` as * default * option in all related to markdown functions.
> 3. Instead of the `markdown_smt_v2` statements, perhaps we need to use the` markdown_v2_smt`? (one of them is commented in this review)


1. and 3. sound good to me. Could do it in the next days, but if you want to, please go ahead.

2. There is no default value for `parse_mode` anywhere, there is only `Message.reply_markdown` (correct me, if I'm missing something). For the latter, there is `Message.reply_markdown_v2` now. For the former, #1490 is the PR, but it's left to the user to choose a default value, which makes sense because just putting `MarkdownV2` as default would break backwards compitibility since it requires more characters to be escaped than v1

Bibo-Joshi:
@dmytrohoi, I did the renaming, but I realized, that `ChatMember` does not have a `chat(_id)` attribute. So adding a `set_custom_title` method to `ChatMember` would really just save 1 of 3 arguments of `bot.set_chat_administrator_custom_title()`. That doesn't seem much of a win to me and I left it out.

Bibo-Joshi:
Apparently `file_id`, when sending a file via the id, the file id in the resulting message doesn't need to be the same any more. Hence, I changed the `_id_attrs` to use the new `file_unique_id` only.

Unfortunately, since API 4.6 is out already, test_official fails …

telegram/bot.py
```diff
@@ -2841,6 +2841,44 @@ def set_chat_permissions(self, chat_id, permissions, timeout=None, **kwargs):
...
 
+    @log
+    def set_chat_administrator_custom_title(self,
+                                            chat_id,
+                                            user_id,
+                                            custom_title,
+                                            timeout=None,
+                                            **kwargs):
+        """
+        Use this method to set custom title for owner or administrators. The bot must be an
```
Eldinnie: According to the official docs, bot's can only use this for admins promoted by the bot. Our docs should refelct this

telegram/chat.py
```diff
@@ -96,6 +100,7 @@ def __init__(self,
                  invite_link=None,
                  pinned_message=None,
                  permissions=None,
+                 slow_mode_delay=None,
```
Eldinnie: please add at then end of the constructor. It's annoying, I know, but people still use args...

telegram/chatmember.py
```diff
@@ -112,7 +115,7 @@ class ChatMember(TelegramObject):
     RESTRICTED = 'restricted'
     """:obj:`str`: 'restricted'"""
 
-    def __init__(self, user, status, until_date=None, can_be_edited=None,
+    def __init__(self, user, status, custom_title=None, until_date=None, can_be_edited=None,
```
Eldinnie: please add at the end.

telegram/files/animation.py
```diff
@@ -26,6 +26,9 @@ class Animation(TelegramObject):
 
     Attributes:
         file_id (:obj:`str`): Unique file identifier.
```
Eldinnie: From https://core.telegram.org/bots/api#animation
```file_id - Identifier for this file, which can be used to download or reuse the file```
I think we should use that here too. (also for all the others)

File: tests/test_animation.py [hunk 6]
```diff
@@ -72,7 +75,9 @@ def test_send_all_args(self, bot, chat_id, animation_file, animation, thumb_file
 
         assert isinstance(message.animation, Animation)
         assert isinstance(message.animation.file_id, str)
+        assert isinstance(message.animation.file_unique_id, str)
         assert message.animation.file_id != ''
+        assert message.animation.file_unique_id != ''
```
Eldinnie: shouldn;t we actually test it is the expected unique_id? (also for the others I suppose

tests/test_animation.py
```diff
@@ -102,14 +107,17 @@ def test_send_animation_url_file(self, bot, chat_id, animation):
...
+
         assert message.animation.duration == animation.duration
+        assert message.animation.file_name == animation.file_name
         assert message.animation.mime_type == animation.mime_type
         assert message.animation.file_size == animation.file_size
 
     @flaky(3, 1)
     @pytest.mark.timeout(10)
-    @pytest.mark.skip(reason='Doesnt work without API 4.5')
```
Eldinnie: can you enlighten me why this is in in the first place? As far as I understood. This should just work without 4.5...

tests/test_message.py
```diff
@@ -144,6 +168,8 @@ def test_dict_approach(self, message):
         assert message['chat_id'] == message.chat_id
         assert message['no_key'] is None
 
+    # START WORKING HERE
```
Eldinnie: ??

Eldinnie:
Overall looking good. Just some questions and remarks added

Eldinnie:
Codecov, mainly is unhappy because of the added checks for py2 (in parse_entities) which are never hit, because we don't test 2.7 anymore. You can ignore it

tests/test_animation.py
Bibo-Joshi: Yes, that would probably be better. I thinks this was just copy-pasted from the exsting tests for `file_id`. I'll try and adjust everywhere

tests/test_animation.py
Bibo-Joshi: Since API 4.5 any file may have multiple `file_id` s and the only valid comparison is `file_unique_id`. The latter is not present in master and hence tests will fail.

tests/test_message.py
Bibo-Joshi: mh, must have missed that. will remove

Poolitzer:
@Bibo-Joshi Since we want to get this done asap, I guess I can chime in this afternoon for me, if thats alright with you

tests/test_animation.py
Poolitzer: wait, arent we creating a new file here?

Poolitzer: we would need to send an animation via file_id to test if its the same unique file id no?

Bibo-Joshi: Oh you're right. Then testing for `file_(unique_)id == self.file_(unique_)id` doesn't make sense.

Bibo-Joshi:
Latest commit closes #1654 . Put this here since we're fiddling with `_parse_markdown` anyway and we would have merge conflicts else …

gpchelkin:
Just a friendly reminder, I think it better belongs here. Please bump supported Bot API version number in https://github.com/python-telegram-bot/python-telegram-bot/blob/master/README.rst

It says 4.1 now, but one could tell from the changelogs, that actually it is 4.4 now. I think it might be repelling or misleading for new users of this great wrapper :-)

Bibo-Joshi:
pytest failures seem unrelated (`test_conversation_timeout_cancel_conflict`)

PR #1510, opened 2019-09-07
Support all_members_are_administrators only by kwargs

Bibo-Joshi:
As discussed offline, this removes `all_members_are_administrators` from `Chat` s parameters. I didn't remove it from `test_chat` since Telegram still returns it, so I guess we should still test it.


tsnoam:
@Bibo-Joshi That&#x27;s a good idea regarding the test. That way we&#x27;d get test failures when telegram will stop sending this parameter and we&#x27;ll know to remove it from our code base.

PR #1525, opened 2019-10-01
Fix for Bot.to_dict returning the wrong value for first_name (#1519)

### LINKED ISSUE #1519: Bot dict return wrong value for first_name
### Steps to reproduce
1. ``from telegram import Bot``

2. ``bot: Bot = Bot(token='708626524:abc-token')``

3. ``print(bot.to_dict())``

### Expected behaviour
``{'id': 708626524, 'username': 'for_testx_bot', 'first_name': 'For Test'}``

### Actual behaviour
``{'id': 708626524, 'username': 'for_testx_bot', 'first_name': 'for_testx_bot'}``

### Configuration
Microsoft Windows [Version 10.0.16299.15]
(c) 2017 Microsoft Corporation. All rights reserved.

(venv) E:\Working\Running_Projects\abc>python -m telegram
python-telegram-bot 12.1.1
certifi 2019.09.11
future 0.17.1
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 21:26:53) [MSC v.1916 32 bit (Intel)]

### More Info
bot_issue

clonex10100:
Closes #1519

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -3439,7 +3439,7 @@ class Bot(TelegramObject):
         return Poll.de_json(result, self)
 
     def to_dict(self):
-        data = {'id': self.id, 'username': self.username, 'first_name': self.username}
+        data = {'id': self.id, 'username': self.username, 'first_name': self.first_name}
 
         if self.last_name:
             data['last_name'] = self.last_name


tsnoam:
@clonex10100 1. We have a very fragile logic with username/first name/last name when it comes to user details.
We should make sure it&#x27;s either consistent or the difference makes sense.

2. Unitests seems to fail due to (what I think is) a change in the shell scripting behaviour of Travis. The <code>if... then...</code> logic should be modified to have <code>else true</code> (this is very common behaviour with Makefile)

Poolitzer:
Hey @tsnoam , what do you mean?
1. The bots first name should be the bots first name. Its always given, it should be correctly returned in the to_dict function.

2. Unitest should not fail because of this change. If there is actually an issue in the file, we should move it to a separate issue and PR, but this one is fine imo

tsnoam:
@Poolitzer 
I would expect `test_get_me_and_properties()` to fail after this change. 
I need to understand why it didn't fail before or why it doesn't fail now.

tsnoam:
uh. of course. we need a new test for to_dict

Poolitzer:
@tsnoam because that is not affected by to_dict right? They are only using de_json

PR #1528, opened 2019-10-03
Bot dic returns first_name for first_name, not username

MaestroJolly:
Fixed the to_dict function to return the correct first_name in the response.


Poolitzer:
Hey, thanks for your PR. Why did you change the gitignore file though? Can you revert that? Then we would be happy to merge :)

Issue #1029, opened 2018-03-03
ConversationHandler is not thread safe


tsnoam:
The following are initial findings, but it's a good place to start:

Members which need a mutex protecting them:
- [x] `conversations`
- [x] `timeout_jobs`

Members which should be removed. Instead `check_update()` should probably return a context to be passed to `handle_updates()`:
- [x]  `current_conversation`
- [x]  `current_handler`

- [x] All other members must be set internal (i.e. renamed to be prefixed with `__`) so they won't be modified after `__init__()`.

jh0ker:
`current_conversation` and `current_handler` could also be in a `threading.local`

bomjacob:
The issues with `current_conversation` and `current_handler`  have been fixed by implementing context based callbacks (#1100) (see `ConversationHandler.check_update`).

PR #1533, opened 2019-10-09
Add mutex protection on ConversationHandler

SnowyCoder:
Advances #1029 adding thread safety on the mutable variables of the class (`conversations` and `timeout_jobs`).
The original issue also asked to internalize all the other properties but it would incur breaking changes and a lot of refactoring, I'll leave this task to someone more proficient with the codebase.

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -20,6 +20,7 @@
 
 import logging
 import warnings
+from threading import Lock
 
 from telegram import Update
 from telegram.ext import (Handler, CallbackQueryHandler, InlineQueryHandler,
@@ -162,7 +163,9 @@ class ConversationHandler(Handler):
         Set by dispatcher"""
 
         self.timeout_jobs = dict()
+        self.timeout_jobs_lock = Lock()
         self.conversations = dict()
+        self.conversations_lock = Lock()
 
         self.logger = logging.getLogger(__name__)
 
@@ -240,7 +243,8 @@ class ConversationHandler(Handler):
             return None
 
         key = self._get_key(update)
-        state = self.conversations.get(key)
+        with self.conversations_lock:
+            state = self.conversations.get(key)
 
         # Resolve promises
         if isinstance(state, tuple) and len(state) == 2 and isinstance(state[1], Promise):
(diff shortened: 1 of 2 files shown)


Poolitzer:
Hey man, this looks good to me, but there should be definitly someone else to look over this *cough cough* @eldin? *cough cough*. Can you remove the #Todo line though and either put it in this conversation or in the related issue? Then codacy should be fine with it.

telegram/ext/conversationhandler.py
```diff
@@ -319,46 +324,62 @@ def handle_update(self, update, dispatcher, check_result, context=None):
...
 
-        if timeout_job is not None:
-            timeout_job.schedule_removal()
-        if self.conversation_timeout and new_state != self.END:
-            self.timeout_jobs[conversation_key] = dispatcher.job_queue.run_once(
-                self._trigger_timeout, self.conversation_timeout,
-                context=_ConversationTimeoutContext(conversation_key, update, dispatcher))
+        with self.timeout_jobs_lock:
+            # Remove the old timeout job (if present)
+            # TODO: should this be called before the child handle_update? the timeout could be called while the child
```
jh0ker: Yes, I believe you are correct, this would be more consistent.

SnowyCoder:
I've addressed the todo, should I squash the commits?
Sorry for the trouble, I wanted to help you a bit but It's my first time doing a PR in a big repo :sweat_smile:.

Poolitzer:
No need to, we will before merging into master. Just go ahead and make as many commits as you want :)

jh0ker:
@SnowyCoder Now you have the chance to get the true _**big repo** experience™️_ 😉 
It would be great to have a test for that last change you made regarding the removal of the timeout job. Basically a test that demonstrates that the timeout job will be removed and not executed, even if you have a child handler that starts before the timeout would run out and ends after.

Refer to https://github.com/python-telegram-bot/python-telegram-bot/blob/master/tests/test_conversationhandler.py line 414 and following for inspiration.

telegram/ext/conversationhandler.py
```diff
@@ -162,7 +163,9 @@ def __init__(self,
         Set by dispatcher"""
 
         self.timeout_jobs = dict()
+        self.timeout_jobs_lock = Lock()
         self.conversations = dict()
+        self.conversations_lock = Lock()
```
jh0ker: Please prefix those locks with a `_` so we're not committed to this API

SnowyCoder:
I've been studying the test code and I noticed something strange: the assignment at line 494 (that have moved to 552 in the master branch) is duplicated, it seems like a merge error or a typo.

jh0ker:
@SnowyCoder Interesting find ^^

jh0ker:
Thanks a lot for your contribution @SnowyCoder!

SnowyCoder:
Thank you for the support!

PR #1553, opened 2019-10-15
Implement github actions CI

jsmnbom:
Should be as functional as travis and appveyor but now it's all in one place.

Note: Travis and Appveyor CI will probably fail since they will use fallback tokens.

@Poolitzer did most of the initial work getting pytest to work, and i think he's working on more github actions stuff.


Poolitzer:
@jsmnbom Yeah go on, steal my work, pfffffffff :(

PR #1556, opened 2019-10-15
Implement github actions ci

jsmnbom:
Should be as functional as travis and appveyor but now it's all in one place.

Note: Travis and Appveyor CI will probably fail since they will use fallback tokens.

@Poolitzer did most of the initial work getting pytest to work, and i think he's working on more github actions stuff.


jsmnbom:
This should be ready for review.

Some notes:
- Not testing on PyPy. Does it even run on windows? Also we have 9 bots right now and we're using 8 of them.
- Not testing on macOS, do we need to?
- Not testing on 3.8 for now. Waiting for https://github.com/actions/setup-python/issues/30, or maybe we can do something with a container job?

jsmnbom:
Okay nvm, still some issues. Looks like 'continue-on-error' doesn't do quite what i thought, so right now the builds never actually fail.

Issue #768, opened 2017-08-03
[FeatureRequest] Split too long messages


sprnza:
Hi there!
Thanks for the project! It's really awesome how it's easy to set up a bot with it!
The only thing I'm missing migrating from php implementation of a bot is a feature that splits outgoing messages automaticaly if it exceeds `MAX_MESSAGE_LENGTH`.

What I currently do is grepping a file line by line and if there is a pattern found I add the line to a variable that's being sent. And sometimes it's exceeds `MAX_MESSAGE_LENGTH` obviously.

Do you have any advice how to send message once variable reaches `MAX_MESSAGE_LENGTH` and then to proceed executing my process of composing text?

bomjacob:
Thank you for the suggestion, we're currently working on a solution (that also takes into account the maximum number of message entity and that takes care to not split in the middle of an entity).

sprnza:
Yay! Cool! Thanks!

JosXa:
```
    def send_message(bot, chat_id, text: str, **kwargs):
        text = text.encode('utf-8').decode('utf-8')
        if len(text) <= constants.MAX_MESSAGE_LENGTH:
            return self.bot.send_message(chat_id, text, **kwargs)

        parts = []
        while len(text) > 0:
            if len(text) > constants.MAX_MESSAGE_LENGTH:
                part = text[:constants.MAX_MESSAGE_LENGTH]
                first_lnbr = part.rfind('\n')
                if first_lnbr != -1:
                    parts.append(part[:first_lnbr])
                    text = text[first_lnbr:]
                else:
                    parts.append(part)
                    text = text[constants.MAX_MESSAGE_LENGTH:]
            else:
                parts.append(text)
                break

        msg = None
        for part in parts:
            msg = bot.send_message(chat_id, part, **self._set_defaults(kwargs))
            time.sleep(1)
        return msg  # return only the last message
```

This is a snippet I used in one of my projects. Perhaps it's of help to anyone

JosXa:
@bomjacob What's the status on this?

codewhizz:
> ```
>     def send_message(bot, chat_id, text: str, **kwargs):
>         if len(text) <= constants.MAX_MESSAGE_LENGTH:
>             return self.bot.send_message(chat_id, text, **kwargs)
> 
>         parts = []
>         while len(text) > 0:
>             if len(text) > constants.MAX_MESSAGE_LENGTH:
[...]

PR #1557, opened 2019-10-15
Splits long messages into smaller messages (solves #768)

hanlog-4:
Added so that when a too long message is sent it is split up into smaller messages as describe in issue #768.
I first try to split it on a newline, if there is none, I try to split it on a new sentence. If there is none, it will just split it where the MAX_MESSAGE_LENGTH is.

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -34,6 +34,7 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from future.utils import string_types
 
+from telegram import constants
 from telegram import (User, Message, Update, Chat, ChatMember, UserProfilePhotos, File,
                       ReplyMarkup, TelegramObject, WebhookInfo, GameHighScore, StickerSet,
                       PhotoSize, Audio, Document, Sticker, Video, Animation, Voice, VideoNote,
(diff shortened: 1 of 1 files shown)


Poolitzer:
Hey there.
First of all, thanks so much for your PR. It is much appreciated.
We are going to reject it though, and thats just partly related to what you did.

We as the maintainers are not that intruged by the whole idea of this issue (someone wanted to post on that topic there but didnt do it yet smh). Not only is it a lot of work to implement (I will come to that part), but we would also have to maintain this feature by ourself, which would be a lot of work. If implemented, we would at least need a opt-in/out option, since it may break some bots which are expecting long message to just fail and want to work with that. And we would also have to think about providing it for other limited areas like captions, or maybe even files - thats where the lot of work to implement comes into place,

Now to your PR. Please dont take this as an attack, which it isn't. I just want to give you some criticism, so you can improve yourself if you want to.
First of all, this implementation is pretty basic. You changed it at one point in the whole code. There is at least editing of messages which would run into the same problem and sending an InlineQueryResult with InputTextMessageContent.
Then we have [line 46](https://github.com/python-telegram-bot/python-telegram-bot/blob/2e58d176b567a4f79576f81de52b6edf14ed319f/telegram/bot.py#L46), which you changed. Probably by mistake, but it breaks the whole code. If you have a last look at the files changed section when you create a PR, you can spot those mistakes :)
There is also a behaviour we have in this project we call tests. You might not find them everywhere else (or in your own projects - I dont do them :D), but they are here. Whenever there is a new feature someone implements, there has to be tests for it. You didnt include them, so we would have asked for that at one point before we'd have merged your PR.

Thanks again for trying to improve our little project here, and dont be too disappointed please. Everybody has to start somewhere, you just picked a not so good issue for that :D

PR #1560, opened 2019-10-16
Add MAX_PHOTOSIZE_UPLOAD size limit constant

revolter:


jh0ker:
Can you add a link to the source of this information in the docstring? ie https://core.telegram.org/bots/api#sending-files

revolter:
Is it ok like this? The one I added is not the only one from the API page.

jh0ker:
Nice, thanks for adding this :+1:

PR #1564, opened 2019-10-17
Fix formatting in issue template

revolter:


tsnoam:
We like the issue template as it is. Thank you for your suggestion.

PR #1591, opened 2019-10-31
fix download without path arguments

### LINKED ISSUE #1590: [BUG] Cannot Download a File if it doesnt have a `file_path` or if `custom_path` is not provided to `download()`
### Steps to reproduce
1. Create an operation with the above mentioned aspects

Ex:
```
def run_bot(token):
    def make_pasta(update, context):
        msg = update.message

        if msg.reply_to_message is None:
            msg.reply_text(responses.NOT_A_REPLY)
            return
        if msg.reply_to_message.document is None:
            msg.reply_text(responses.NOT_A_DOC)
            return
        telegram_file = File(msg.reply_to_message.document.file_id)
        telegram_file.download() 

    updater = Updater(token, use_context=True)
    dp = updater.dispatcher

    dp.add_handler(CommandHandler('hello', hello))
    dp.add_handler(CommandHandler('make_pasta', make_pasta))

    print('Log: Seu bot iniciou! (:')
    updater.start_polling()
    updater.idle() 
```
(I know this is not exactly a MWE, sorry)

### Expected behaviour
According to the documentation, it should download the file directly to my current working directory `Download this file. By default, the file is saved in the current working directory with its original filename as reported by Telegram.`

### Actual behaviour
on `telegram/files/file.py`, on the download function, we get a type error since `self.file_path` is NoneType, not str or os.PathLike object

### Configuration
Ubuntu 18.04

python-telegram-bot 12.2.0
certifi 2019.09.11
future 0.18.1
Python 3.7.3 (default, Apr  3 2019, 19:16:38)  [GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]

GabrielSimonetto:
fix #1590

### DIFF
--- a/telegram/files/file.py
+++ b/telegram/files/file.py
@@ -19,6 +19,7 @@
 """This module contains an object that represents a Telegram File."""
 from base64 import b64decode
 from os.path import basename
+from os import getcwd
 
 from future.backports.urllib import parse as urllib_parse
 
@@ -116,8 +117,10 @@ class File(TelegramObject):
         else:
             if custom_path:
                 filename = custom_path
-            else:
+            elif file_path:
                 filename = basename(self.file_path)
+            else:
+                filename = getcwd()
 
             buf = self.bot.request.retrieve(url, timeout=timeout)
             if self._credentials:
--- a/telegram/files/file.py
+++ b/telegram/files/file.py
@@ -19,7 +19,7 @@
 """This module contains an object that represents a Telegram File."""
 from base64 import b64decode
 from os.path import basename
-from os import getcwd
+import os
 
 from future.backports.urllib import parse as urllib_parse
 
@@ -120,7 +120,7 @@ class File(TelegramObject):
             elif self.file_path:
                 filename = basename(self.file_path)
             else:
-                filename = getcwd()
+                filename = os.path.join(os.getcwd(), self.file_path)
 
             buf = self.bot.request.retrieve(url, timeout=timeout)
             if self._credentials:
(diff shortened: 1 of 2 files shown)


Poolitzer:
thanks for this.

After a short discussion, we came to the conclusion that you are actually on something here. Telegram does define file_name, on which file_path is based, as optional. If you want to include this, we ask to change this though:
[x] Make a proper check if file_path exists
[] If not, set the file_path to the string of the file_id
[] implement a unitest for this case

This way, the actual filename will be unique. We dont think that this is ever going to happen, but better be save then sorry

GabrielSimonetto:
Ok, it will be done soon enough

GabrielSimonetto:
So... I have a problem.

I don't have experience with tests, I've read some of them on `tests/test_file.py`, but I don't understand how they work. So I would need to do a test that build a `FIle` with the constructor using `file_id`, and then download and see if an error ocurs, but on this module we don't have a valid file_id to do this, so I don't know how that could be done. Can someone point me in some direction?

Bibo-Joshi:
Hi, sorry for the late reply. In `test_file.py`, a fixture `file` is defined ([Line 29](https://github.com/python-telegram-bot/python-telegram-bot/blob/2a3169a22f7227834dd05a35f90306375136e41a/tests/test_file.py#L29)). This acts just like an ordinary Telegram File. You can pass it to your new test like 

```
def test_new_test(self, bot, file):
    file.file_id = 5                # Just an example
    assert file.file_id == 5
```

Within that test you can change the attributes of `file` as needed for your test. If you have more questions, please don't hesitate to ask :)

Bibo-Joshi:
@GabrielSimonetto I took the liberty to go ahead and implement the test case ;)

GabrielSimonetto:
Thanks @Bibo-Joshi, sorry for disappearing, I've been a bit busy this month.

Eldinnie:
Thanks you for your contribution

Issue #1595, opened 2019-11-01
Custom Filters: Wiki and code are in conflict


ema-pe:
About custom filters, the [Wiki](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Extensions-%E2%80%93-Advanced-Filters#custom-filters) says that the `filter` method accepts a `Message` instance, but in the [source code it accepts](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/filters.py#L114) only an `Update` instance.

What is right? Should the Wiki be updated (and can I update the accused page)?

jh0ker:
You are correct, good catch!
The wiki is wrong, this has been changed in https://github.com/python-telegram-bot/python-telegram-bot/pull/1221
Feel free to update the wiki page :slightly_smiling_face:

ema-pe:
@jh0ker it is not easy, because in the [source code](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/filters.py#L73) `filter` can accepts also a `Message`. Why there are two equivalent options?

Eldinnie:
@ema-pe Well. It&#x27;s not wrong are it is...
Because we long as you don&#x27;t set the class attribute <code>update</code> on your custom, only accepting a message is fine

ema-pe:
> @ema-pe Well. It's not wrong are it is...
> Because we long as you don't set the class attribute `update` on your custom, only accepting a message is fine

You're right, I did not see the class attribute. But `filter`'s method documentation should be updated.

Eldinnie:
@ema-pe I agree on that. Would you be comfortable creating a PR to fix it?

ema-pe:
> @ema-pe I agree on that. Would you be comfortable creating a PR to fix it?

Yes, I'm creating a PR

ema-pe:
Here is the PR: #1596

Bibo-Joshi:
#2032 clarifies this. As it's already merged into the v13 branch, I'll go ahead and close this issue.

PR #1596, opened 2019-11-01
filter accepts a Message by default, not an Update

ema-pe:
The method `BaseFilter.filter()` accepts a `Message` by default. Only if
class attribute `update_filter` is set to `True` the method accepts an
`Update`.

This pull request provides a commit that emphasizes this behaviour, because one can misunderstand it.

Closes: #1595

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -103,15 +103,13 @@ class BaseFilter(object):
             self.name = self.__class__.__name__
         return self.name
 
-    def filter(self, update):
+    def filter(self, message):
         """This method must be overwritten.
 
-        Note:
-            If :attr:`update_filter` is false then the first argument is `message` and of
-            type :class:`telegram.Message`.
-
         Args:
-            update (:class:`telegram.Update`): The update that is tested.
+            message (:class:`telegram.Message`): The message that is tested. If
+            :attr:`update_filter` is True update is `update` of type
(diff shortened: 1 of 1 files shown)


telegram/ext/filters.py
```diff
@@ -103,15 +103,13 @@ def __repr__(self):
...
         """This method must be overwritten.
 
-        Note:
-            If :attr:`update_filter` is false then the first argument is `message` and of
-            type :class:`telegram.Message`.
-
         Args:
-            update (:class:`telegram.Update`): The update that is tested.
+            message (:class:`telegram.Message`): The message that is tested. If
+            :attr:`update_filter` is True update is `update` of type
```
Poolitzer: you mean, message is `update` of type update right? And you missed a comma there ;P

Bibo-Joshi:
@ema-pe I merged master so tests will succeed and changed the last line of the PR description so it will automatically close #1595 on merge :)

ema-pe:
Good!
--
Emanuele Petriglia (ema-pe)

Bibo-Joshi:
After some discussion, we're actually a bit unsure about this change. Hence, I move this PR to Milestone 13. We'll need some more internal discussion before making a decision.

Bibo-Joshi:
We decided to go another way in #2032, so I'll be closing this. Still, thank you for PRing :)

PR #1603, opened 2019-11-08
Proposal: Refactor msg_in filter to become classmethod of regex

Bibo-Joshi:
As discussed offline this proposes building the `message_in` filter on top of `Filters.regex` instead of it being a standalone filter. Also ads the optional argument `caption` to `Filters.regex` to allow to check the caption instead of the text.

If we decide, not to have `message_in` in the code base at all, I'll put this in the snippet section of the wiki.


Bibo-Joshi:
Closing in favour of #1625

PR #1611, opened 2019-11-15
Fix error when decode() function convert something to utf8

rikonaka:
According to the Python3.8 Unicode [doc](https://docs.python.org/3/howto/unicode.html).

And the original program did not make any feasible remedies for utf-8 decoding failure.

### DIFF
--- a/telegram/utils/request.py
+++ b/telegram/utils/request.py
@@ -169,15 +169,17 @@ class Request(object):
 
         try:
             decoded_s = json_data.decode('utf-8')
+            data = json.loads(decoded_s)
         except UnicodeDecodeError:
             try:
                 decoded_s = json_data.decode('utf-8', 'backslashreplace')
+                data = json.loads(decoded_s)
             except UnicodeDecodeError:
                 logging.getLogger(__name__).debug(
                     'Logging raw invalid UTF-8 response:\n%r', json_data)
                 raise TelegramError('Server response could not be decoded using UTF-8')
-        try:
-            data = json.loads(decoded_s)
+            except ValueError:
+                raise TelegramError('Invalid server response')
         except ValueError:
             raise TelegramError('Invalid server response')
 
--- a/telegram/utils/request.py
+++ b/telegram/utils/request.py
@@ -168,12 +168,15 @@ class Request(object):
         """
 
         try:
-            decoded_s = json_data.decode('utf-8', 'backslashreplace')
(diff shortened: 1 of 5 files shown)


telegram/utils/request.py
```diff
@@ -171,9 +171,15 @@ def _parse(json_data):
...
-            raise TelegramError('Server response could not be decoded using UTF-8')
+            try:
+                decoded_s = json_data.decode('utf-8', 'ignore')
+                data = json.loads(decoded_s)
+            except UnicodeDecodeError:
+                logging.getLogger(__name__).debug(
+                    'Logging raw invalid UTF-8 response:\n%r', json_data)
+                raise TelegramError('Server response could not be decoded using UTF-8')
+            except ValueError:
+                raise TelegramError('Invalid server response')
```
tsnoam: 1. no need to decode twice. just call:
```python
decoded_s = json_data.decode('utf-8', 'replace')
```
2. please use `replace` instead of `ignore`.
3. the `except UnicodeDecodeError` is not needed any longer

tests/bots.py
```diff
@@ -49,7 +49,10 @@
 BOTS = os.getenv('BOTS', None)
 JOB_INDEX = os.getenv('JOB_INDEX', None)
 if GITHUB_ACTION is not None and BOTS is not None and JOB_INDEX is not None:
-    BOTS = json.loads(base64.b64decode(BOTS).decode('utf-8'))
+    try:
+        BOTS = json.loads(base64.b64decode(BOTS).decode('utf-8'))
+    except UnicodeDecodeError:
+        BOTS = json.loads(base64.b64decode(BOTS).decode('utf-8', 'ignore'))
```
tsnoam: please revert. not needed in this context.

tests/test_official.py
```diff
@@ -130,7 +130,11 @@ def check_object(h4):
     cert_reqs='CERT_REQUIRED',
     ca_certs=certifi.where())
 request = http.request('GET', 'https://core.telegram.org/bots/api')
-soup = BeautifulSoup(request.data.decode('utf-8'), 'html.parser')
```
tsnoam: please revert, not needed in this context

telegram/utils/request.py
rikonaka: Got it.😋

tests/bots.py
rikonaka: Got it.😋

PR #1614, opened 2019-11-15
github workflow: Add python-3.8

tsnoam:


tsnoam:
* workflow fails on py3.8 linux. need to investigate
* workflow hangs on py3.8 windows. need to investigate as well.

vergeev:
I also wanted to note that it fails on macOS Catalina with an error that basically reads
```
from telegram import Update
ImportError: cannot import name 'Update' from 'telegram' (unknown location)
```
Only with python 3.8.

Bibo-Joshi:
GH Actions now provides py3.8 (https://github.com/actions/setup-python/issues/30), but `test_updater` hangs on windows. Will need to investigate

Bibo-Joshi:
> GH Actions now provides py3.8 ([actions/setup-python#30](https://github.com/actions/setup-python/issues/30)), but `test_updater` hangs on windows. Will need to investigate

Apparently `urlopen()` doesn't return for some reason. I remember having a similar problem in #1791, but the networking backend is beyond me 🙈

Bibo-Joshi:
@Bibo-Joshi @tsnoam can you maybe have a look?

Bibo-Joshi:
Found the issue at [tornadoweb/tornado#2608](https://github.com/tornadoweb/tornado/issues/2608) and a workaround [here](https://github.com/ipython/ipykernel/pull/456/).

As far as I can see, `urlopen` is only used explicitly in `test_updater` (and in the vendored `urllib3` …), so I guess we don't have to make changes elsewhere.

tsnoam:
@Bibo-Joshi Good catch!!!

This is a problem  related to the webhook Updater.
It&#x27;s the only component using Tornado.

The workaround is good and should probably be applied in webhookhandler.py

We should also consider moving all imports of tornado to the init of the webhook instead of top of the file.

Bibo-Joshi:
@tsnoam @tsnoam Added the workaround to <code>WebhookHandler</code>, but left the imports at the top of the file, since <code>WebhookServer</code> and <code>WebhookAppClass</code> in the same file use them, too.

Bibo-Joshi:
CI fails:
* py3.7 is unrelated (`set_game_score`)
* codecov: Don't know how to get the coverage up …

Bibo-Joshi:
@Bibo-Joshi @tsnoam @tsnoam @tsnoam

I think I&#x27;ve done all I can 😀

sorry for bothering you. sorry for bothering you. sorry for bothering you.

PR #1615, opened 2019-11-15
add a cron job

Poolitzer:
this is in place in order to test if an update of our dependencies breaks our library.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,6 +3,9 @@ on:
   pull_request:
     branches:
       - master
+  schedule:
+    - cron: 0 2 * * *
+  
 
 jobs:
   pytest:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -3,6 +3,9 @@ on:
   pull_request:
     branches:
       - master
+  schedule:
+    - cron: 0 2 * * *
```
tsnoam: please use a non exact our.
like 2:07, or 2:03 or 2:22

times like 2:00 are prone to errors as many frequent tasks are scheduled for these times.

.github/workflows/test.yml
```diff
@@ -3,6 +3,9 @@ on:
...
+
```
tsnoam: redundant empty line

.github/workflows/test.yml
Poolitzer: @tsnoam yeah, that was a copy paste error

PR #1616, opened 2019-11-16
Fix error when decode() function convert something to utf8 [final version]

rikonaka:
Use the `decoded_s = json_data.decode('utf-8', 'replace')` instead of `decoded_s = json_data.decode('utf-8')`


rikonaka:
I don't want to break the original code structure, so I still keep `except UnicodeDecodeError` to avoid some unknown errors. I try to make the smallest change.

PR #1617, opened 2019-11-16
Fix error when decode() function convert something to utf8 again

rikonaka:
- remove the `except UnicodeDecodeError`
- use `decoded_s = json_data.decode('utf-8', 'replace')` instead of `decoded_s = json_data.decode('utf-8')`


rikonaka:
I will do it again after reorganizing. THX.😝

PR #1619, opened 2019-11-18
add private /c links to message.links object

Poolitzer:


telegram/message.py
```diff
@@ -338,10 +338,14 @@ def chat_id(self):
 
     @property
     def link(self):
-        """:obj:`str`: Convenience property. If the chat of the message is a supergroup or a
-        channel and has a :attr:`Chat.username`, returns a t.me link of the message."""
-        if self.chat.type in (Chat.SUPERGROUP, Chat.CHANNEL) and self.chat.username:
-            return "https://t.me/{}/{}".format(self.chat.username, self.message_id)
+        """:obj:`str`: Convenience property. If the chat of the message is not
+        a private chat, returns a t.me link of the message."""
+        if self.chat.type is not Chat.PRIVATE:
```
Bibo-Joshi: At least on my machine, for some reason only `if self.chat.type != Chat.PRIVATE` will work. With `is not` it will also return a link in private chats.

telegram/message.py
```diff
@@ -338,10 +338,14 @@ def chat_id(self):
...
+            if self.chat.username:
+                to_link = self.chat.username
+            else:
+                to_link = "c/{}".format(self.chat.id)
```
Bibo-Joshi: As far as my testing goes, this should be
 ```
to_link = "c/{}".format(abs(int(self.chat.id)))
```
The `int` just in case, `chat.id` is a string and the `abs` because Telegram redirects to telegram.org, if you insert the negative chat id of groups.

Comment (Poolitzer) [reply telegram/message.py]: actually, we have to remove the -100 for groups.

File: tests/test_message.py [hunk 3]
```diff
@@ -304,15 +304,31 @@ def test_chat_id(self, message):
         assert message.chat_id == message.chat.id
 
     def test_link(self, message):
```
tsnoam: This unitest contains lot of internal changes of the `message` object.
Personally it makes me confused on what is the desired test.

Can you please split it into multiple unitests, each one testing one desired behaviour?

telegram/message.py
```diff
@@ -338,10 +338,19 @@ def chat_id(self):
...
-        if self.chat.type in (Chat.SUPERGROUP, Chat.CHANNEL) and self.chat.username:
-            return "https://t.me/{}/{}".format(self.chat.username, self.message_id)
+        """:obj:`str`: Convenience property. If the chat of the message is not
+        a private chat, returns a t.me link of the message."""
+        if self.chat.type != Chat.PRIVATE:
+            if self.chat.username:
+                to_link = self.chat.username
+            else:
+                # this way, we remove the potential minus
+                id_to_link = abs(int(self.chat.id))
```
Eldinnie: I find this logic confusing.
There isn't a potential minus, there is one. If it's a group, the id is negative.
So I would propose the entire login to change to:
```python
if self.chat.type != Chat.GROUP:
  id_to_link = self.chat.id[3:]
else:
  id_to_link = self.chat.id[1:]
```

Bibo-Joshi: Done

Eldinnie:
Even fixed my mistake thanks :D
Merging

PR #1620, opened 2019-11-18
Add missing return to IQH

Bibo-Joshi:


Bibo-Joshi:
Actually, without the `return False`, `None` will be returned, which is [fine](https://python-telegram-bot.readthedocs.io/en/stable/telegram.ext.handler.html#telegram.ext.Handler.check_update). Stupid me.

Issue #1613, opened 2019-11-15
Add tz kwarg to from_timestamp()


Bibo-Joshi:
See 5. [here](https://github.com/python-telegram-bot/python-telegram-bot/pull/1506#pullrequestreview-307544004).

@plammens I'll try and create a PR on sunday. But if your in the mood and don't see anything related to this issue, please go ahead - you're much more familiar with the changes than I am ;)

Bibo-Joshi:
Closed by #1621

PR #1621, opened 2019-11-18
Add tz kwarg to from_timestamp()

Bibo-Joshi:
Closing #1613 as addendum to #1506 

@tsnoam @plammens Is this, what was missing or am I overlooking half of it? :D


plammens:
With this the default is an aware `datetime`, while before it was naive—e.g. `telegram.Message.date` will return an aware `datetime`. If that's okay, great; if not, I would also change the calls in the `de_json` of some `telegram` classes (e.g. https://github.com/python-telegram-bot/python-telegram-bot/blob/93449443b2293472c005675d22a95f695c19210a/telegram/message.py#L355) to  `data['date'] = from_timestamp(data['date']).replace(tzinfo=None)`, to avoid the breaking change.

Otherwise I think the only problem is handling `None`: passing `None` to `tzinfo` will return a naive datetime in *local* rather than UTC. Maybe something like this instead:
```python
if tzinfo is not None:
    return dtm.datetime.fromtimestamp(unixtime, tz=tzinfo)
else:
    return dtm.datetime.utcfromtimestamp(unixtime)
```

Bibo-Joshi:
> With this the default is an aware `datetime`, while before it was naive—e.g. `telegram.Message.date` will return an aware `datetime`. If we want `Message.date` and the like to keep being naive, the calls in the `de_json` of some `telegram` classes, e.g.
> 
> https://github.com/python-telegram-bot/python-telegram-bot/blob/93449443b2293472c005675d22a95f695c19210a/telegram/message.py#L355
> 
> would also need to be changed to `data['date'] = from_timestamp(data['date']).replace(tzinfo=None)`. Or the default for `from_timestamp` could be kept as naive UTC to avoid the breaking change.

Good point. Let's wait, what tsoam says.

> Otherwise I think the only problem is handling `None`: passing `None` to `tzinfo` will return a naive datetime in _local_ rather than UTC. Maybe something like this instead:
> 
> ```python
> if tzinfo is not None:
>     return dtm.datetime.fromtimestamp(unixtime, tz=tzinfo)
> else:
>     return dtm.datetime.utcfromtimestamp(unixtime)
> ```

Oh, shuh, you're absolutely right. Will update tonight.

Bibo-Joshi:
Updated as suggested. About the failing tests:

*  Py2.7 fails because `datetime` has no `timezone`. But since this PR is only due for v12.5, which I reckon won't be released bevore 2020, when we drop Py2.7, I guess we don't really *have* to adjust for that. Although, I wonder why it didn't fail on the first run …
* As usual, I have no clue, what codecov wants from me 🤔

Poolitzer:
agree with the 2.7 part

tests/test_helpers.py
```diff
@@ -113,8 +113,16 @@ def test_to_timestamp_none(self):
         # this 'convenience' behaviour has been left left for backwards compatibility
         assert helpers.to_timestamp(None) is None
 
-    def test_from_timestamp(self):
-        assert helpers.from_timestamp(1573431976) == dtm.datetime(2019, 11, 11, 0, 26, 16)
+    def test_from_timestamp_naive(self):
+        datetime = dtm.datetime(2019, 11, 11, 0, 26, 16, tzinfo=dtm.timezone.utc)
```
plammens: Using `helpers._UTC` here instead of `dtm.timezone.utc` will fix the Python 2 issue—`helpers._UTC`  was made to emulate `dtm.timezone.utc` in Python 2 (in Python 3 it is equivalent to the actual `dtm.timezone.utc`).

tests/test_helpers.py
```diff
@@ -113,8 +113,16 @@ def test_to_timestamp_none(self):
...
+        assert helpers.from_timestamp(1573431976) == datetime
```
plammens: Here I would test for conversion when `tzinfo=None`, i.e.
```python
datetime = dtm.datetime(2019, 11, 11, 0, 26, 16)
assert helpers.from_timestamp(1573431976, tzinfo=None) == datetime
```
since the other test already takes care of aware datetimes with explicit timezones.

tests/test_helpers.py
```diff
@@ -114,8 +114,8 @@ def test_to_timestamp_none(self):
         assert helpers.to_timestamp(None) is None
 
     def test_from_timestamp_naive(self):
-        datetime = dtm.datetime(2019, 11, 11, 0, 26, 16, tzinfo=dtm.timezone.utc)
-        assert helpers.from_timestamp(1573431976) == datetime
+        datetime = dtm.datetime(2019, 11, 11, 0, 26, 16, tzinfo=_UTC)
+        assert helpers.from_timestamp(1573431976, tzinfo=None) == datetime
```
plammens: And thus (see comment above) here we should test for the naive conversion:
```python
datetime = dtm.datetime(2019, 11, 11, 0, 26, 16, tzinfo=None)
assert helpers.from_timestamp(1573431976, tzinfo=None) == datetime
```

Bibo-Joshi:
@plammens Lost in translation … Thanks for clarifying!

telegram/utils/helpers.py
```diff
@@ -196,22 +162,27 @@ def to_timestamp(dt_obj, reference_timestamp=None):
...
+def from_timestamp(unixtime, tzinfo=dtm.timezone.utc):
     """
-    Converts an (integer) unix timestamp to a naive datetime object in UTC.
+    Converts an (integer) unix timestamp to a timezone aware datetime object.
     ``None`` s are left alone (i.e. ``from_timestamp(None)`` is ``None``).
 
     Args:
         unixtime (int): integer POSIX timestamp
+        tzinfo (:obj:`datetime.tzinfo`, optional): The timezone, the timestamp is to be converted
+            to. Default to UTC.
```
tsnoam, line 173:
```suggestion
            to. Defaults to UTC.
```

tsnoam:
actually, i just fixed the docstring myself.

you can merge from master and then you can merge this PR.

PR #1622, opened 2019-11-19
Add healtcheck endpoint

### LINKED ISSUE #1438: Add health endpoint - 405 Method not allowed
Before upgrading to version 12, a basic HTTP GET request to the running bot would return 200 OK with an empty response body. Now, I get 405 Method Not Allowed. I used this for my monitoring to regularly check if the bot is responsive to requests. Ideally, there would be some kind of health endpoint, which only returns 200 OK if the bot is really ok to process incoming messages from Telegram servers, but an endpoint that returns OK all the time (maybe there is already one?) would help me fix my monitoring, too. 

``$ python -m telegram
python-telegram-bot 12.0.0b1
certifi 2018.11.29
future 0.17.1
Python 3.6.7 (default, Oct 22 2018, 11:32:17)  [GCC 8.2.0]``

reimarstier:
* User option to define custom healthchecks
* Example checking for communication errors with telegram api

Fixes #1438

### DIFF
--- a/examples/healthcheck_echobot.py
+++ b/examples/healthcheck_echobot.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This program is dedicated to the public domain under the CC0 license.
+
+"""
+Simple Bot to reply to Telegram messages.
+
+First, a few handler functions are defined. Then, those functions are passed to
+the Dispatcher and registered at their respective places.
+Then, the bot is started and runs until we press Ctrl-C on the command line.
+
+Additional features:
+- Configuration via environment variables allows to quickly
+  check the connectivity in a certain test environment
+- Added health check when run with webhook
+  - check if the webhook url is properly set as expected
+  - check if webhook is overridden by another bot instance
+  - ask telegram api for communication errors with the bot within the last few minutes
+
+Example set of environment variables to run with webhook:
+TELEGRAM_AUTH_TOKEN=1234:AAbbCCddEEffGGhh
+TELEGRAM_WEBHOOK=true
+TELEGRAM_WEBHOOK_URL=https://bot.mysite.org
+TELEGRAM_WEBHOOK_PORT=443
+TELEGRAM_LISTEN_PORT=8443
+LOG_LEVEL=DEBUG
+SSL_CERT=webhook_cert.pem
+SSL_KEY=webhook_key.pem
+
+Usage:
+Basic Echobot example, repeats messages.
+Press Ctrl-C on the command line or send a signal to the process to stop the
+bot.
+"""
+import datetime
+import logging
+import os
+import uuid
+import requests
+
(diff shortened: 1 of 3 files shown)


Poolitzer:
Hey, thanks for your PR. Before you continue developing it though, I have to tell you that we really, really don't like adding requirements to our project.

While it might make things easier, the exact same result could be achieved with tornado. So I don't think me or the other maintainers can approve this. I am sorry :(

reimarstier:
Hey @Poolitzer, no worry, I did not expect this PR to accepted. At least not the way it is now. I intended to start exactly this discussion. What are the requirements for doing this? The first would be "not adding any requirements" which I don't think is necessary. This kind of requirement might be worth documenting in a guide for contributing to the project. My PR is just a quick shot at what could be done without reinventing the wheel. 

One more requirement that I think should apply: 
1.) So I'd expect the health checks to reside in the base code and could be enabled as needed. 
These are the suggested health checks in my example:

2.) Now what should these health checks do? In my example I implemented a possibility to check if the registered web hook is the one that the bot has set and it was not overwritten by another instance of the bot started somewhere else. 
3.) Then the Telegram API supports checking whether Telegram had any issues communicating with the bot and is telling the most recent time when the error occurred and how many messages are pending to be sent to the bot. 

Now this is the external view of the bots health. What would be an appropriate check within the bot to tell if it is functioning as expected? What kind of health check would apply when not run with a web hook?

Poolitzer:
Hey.

I am so sorry, I have no idea why your reply went by me. Let me comment now, an in depth reply will be guaranteed by this sunday :)

> "not adding any requirements" which I don't think is necessary

I do think so though. We are very hesitant with requirements. It add a potential of bugs out of our control. And a health checkpoint should be perfectly doable with Tornado.

>This kind of requirement might be worth documenting in a guide for contributing to the project

Yes. Agreed.

>health checks to reside in the base code and could be enabled as needed

looks like too much for me for a simple health check as suggested by the issue

>Now what should these health checks do?

The original idea was very simple: Check if the bot is running. Imo that would mean that I call an URL from an external app and get a 200 response. Thats it. I like where you went though.

>What would be an appropriate check within the bot to tell if it is functioning as expected? What kind of health check would apply when not run with a web hook?

I would say we leave this up to the endusers, kind of like you did. I would expect a very simple set-to-True argument to the dispatcher though, where you can either set it to True or to a custom URL, then use this URL (or get a generated one from PTB from the logs) and use that for the 200 check. And then users could implement their own healthchecks on top of it, kinda like your example did.

----

Just to underline this: This is my personal, spontaneous view on this matter. We will talk about this whole issue as the maintainers and give you a proper, longer response on what we as a group can agree on :D

Bibo-Joshi:
@reimarstier Are you still persuing this pr?

Bibo-Joshi:
Closing due to inactivity

Issue #1625, opened 2019-11-20
Refactor Filters.msg_in


Bibo-Joshi:
As discussed offline: We'll remove `Filters.msg_in` and instead build `__call__()` methods for `Filters.text` and `Filters.caption` so that we can allow for `Filters.text` and `Filters.text(<SOME ITERABLE>)`.

Bibo-Joshi:
Closed by #1631

PR #1631, opened 2019-11-20
Refactor msg_in

Bibo-Joshi:
As discussed offline, this refactors the `Filters.msg_in` filter to work as `__call__` of `Filters.text`, thus closing #1625 
I encountered two (and a half) obstacles:

* `BaseFilter`s `__call__` is used to perform the update checks, i.e. it executes `filter()`. To allow for `Filters.text({'foo', 'bar'})` anyway, `Filters.text`s `__call__` now checks, if the argument is of type `Update`. If it's not, it returns an instance of a subclass that works like the former `msg_in`. An alternative way would be to do something like

    ```
    def __call__(self, update=None, iterable=None):
        if update:
            self.filter(update)
        else:
            return _TextIterable(iterable)
    ```

    However calling `Filters.text(iterable={'foo', 'bar'})` seemed neither convenient nor intuitive for me. Also, usually you won't have to call `Filter.text(update)` manually anyway …
* To recreate the caption filter functionality, I needed to add the base class `Filters.caption` that allows only messages that have a caption. Including caption filtering into `Filters.text` like

    ```
    Filters.text({'foo', 'bar'}, caption=True)
    ```

    didn't seem very natural to me …
    If you feel like `Filters.caption` is unneeded, we can just remove it. The main usecase for this PR is `Filters.text(buttons)` anyway …

* I wasn't very original in naming the subclasses (`_TextIterable` and `_CaptionIterable`). If you have a better naming idea, please tell :)

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -909,39 +909,6 @@ officedocument.wordprocessingml.document")``-
             return message.from_user.language_code and any(
                 [message.from_user.language_code.startswith(x) for x in self.lang])
 
-    class msg_in(BaseFilter):
-        """Filters messages to only allow those whose text/caption appears in a given list.
-
-        Examples:
-            A simple usecase is to allow only messages that were send by a custom
-            :class:`telegram.ReplyKeyboardMarkup`::
-
-                buttons = ['Start', 'Settings', 'Back']
-                markup = ReplyKeyboardMarkup.from_column(buttons)
-                ...
-                MessageHandler(Filters.msg_in(buttons), callback_method)
-
-        Args:
-            list_ (List[:obj:`str`]): Which messages to allow through. Only exact matches
-                are allowed.
-            caption (:obj:`bool`): Optional. Whether the caption should be used instead of text.
-                Default is ``False``.
-
-        """
-
-        def __init__(self, list_, caption=False):
-            self.list_ = list_
-            self.caption = caption
-            self.name = 'Filters.msg_in({!r}, caption={!r})'.format(self.list_, self.caption)
-
-        def filter(self, message):
-            if self.caption:
-                txt = message.caption
-            else:
-                txt = message.text
-
-            return txt in self.list_
-
     class _UpdateType(BaseFilter):
         update_filter = True
 
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -22,7 +22,7 @@ import re
 
 from future.utils import string_types
 
-from telegram import Chat
+from telegram import Chat, Update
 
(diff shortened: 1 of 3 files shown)


Bibo-Joshi:
About failing tests:

* Py2: Something about pinning messages. Seems unrelated. will restart.
* Codecov: :man_shrugging:

telegram/ext/filters.py
```diff
@@ -236,11 +236,81 @@ def filter(self, message):
...
+                    return self.filter(update.effective_message)
+            else:
+                return self._CaptionIterable(update)
+
+        def filter(self, message):
+            return bool(message.caption)
+
+    caption = _Caption()
+    """Messages with a caption. If an iterable of strings is passed, it filters messages to only
+    allow those whose caption is appearing in the given iterable.
```
Eldinnie: same

telegram/ext/filters.py
```diff
@@ -236,11 +236,81 @@ def filter(self, message):
...
         def filter(self, message):
             return bool(message.text and not message.text.startswith('/'))
 
     text = _Text()
-    """Text Messages."""
+    """Text Messages. If an iterable of strings is passed, it filters messages to only allow those
+    whose text is appearing in the given iterable.
+
+    Examples:
+        A simple usecase is to allow only messages that were send by a custom
```
Eldinnie: I would [prefer it if there's also an example for the simplest usecase (being onlt `Filters.text` for text messages.

Bibo-Joshi: Added

telegram/ext/filters.py
Bibo-Joshi: Same

telegram/ext/filters.py
```diff
@@ -307,6 +310,9 @@ def filter(self, message):
     """Messages with a caption. If an iterable of strings is passed, it filters messages to only
     allow those whose caption is appearing in the given iterable.
 
+    Examples:
+        ``MessageHandler(Filters.user(1234), callback_method)``
```
Eldinnie: ehm?

Eldinnie: I think you mistkanly copied a wrong line here

Bibo-Joshi:
@Eldinnie Are you okay with the changes? And do you have any clue, what codecov wants?

Eldinnie:
Codecov is telling us that we did not test all branches:
image
It's telling you that the test if it's an update_filter is obsolete, because it never is (at least never tested). That's why the diff_target is not 100%

Anyway I'll allow it for now :D

PR #1634, opened 2019-11-23
Make Filters.user an update filter

Bibo-Joshi:
This allows to filter any `Update` with an `effective_user` to be filtered using `Filters.user`.

- [ ] As discussed in the dev chat, the release notes should contain a warning that this *may* lead to some handlers being called for more updates than before (although `Filters.user` probably was just used with `MessageHandler` until now b/c it only worked with messages).

About Codacy: It wants me to add a docstring to `Filters.user.filter` but that won't happen ;)


Bibo-Joshi:
After internal discussion, we decided that filters should be be reserved for MessageHandlers. Leaving this note here for future reference.

PR #1636, opened 2019-11-23
Rename internal list of PrefixHandler

### LINKED ISSUE #1507: rename Prefixhandlers internally constructed list to `_commands`
Note to self: do this

Bibo-Joshi:
Closes #1507 
I hope this is what you meant, @Eldinnie ?

### DIFF
--- a/telegram/ext/commandhandler.py
+++ b/telegram/ext/commandhandler.py
@@ -317,7 +317,7 @@ class PrefixHandler(CommandHandler):
             self.command = [command.lower()]
         else:
             self.command = command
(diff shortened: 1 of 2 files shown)


Eldinnie:
It is exactly what I meant, but looking at it, I think we should maybe make the `command` and `prefix` attributes setable attributes in such a way that `_commands` will be recreated when they're set.  What do you think?

Bibo-Joshi:
While I don't think that this will be needed very often, it sure makes it safer. Will do.

Issue #1609, opened 2019-11-15
[BUG]How can I make the bot ignore the invalid UTF-8 response?


rikonaka:
### Steps to reproduce
1. I think it is a small probability event, so I don't know how to reproduce it.

### Expected behaviour
Run perfectly.

### Actual behaviour
Bot raise error.
```
2019-11-15 12:50:38,048 - telegram.ext.updater - ERROR - Error while getting Updates: Server response could not be decoded using UTF-8
2019-11-15 12:50:38,048 - xxbot.__main__ - WARNING - Update None caused error Server response could not be decoded using UTF-8
2019-11-15 12:50:39,220 - telegram.ext.updater - ERROR - Error while getting Updates: Server response could not be decoded using UTF-8
2019-11-15 12:50:39,221 - xxbot.__main__ - WARNING - Update None caused error Server response could not be decoded using UTF-8
```
Those error message has been constantly appearing and repeating, causing the program to not work properly.

And I change the `logging` module form `INFO` to `DEBUG`, the more error details show like this.
```
2019-11-15 11:02:44,710 - telegram.ext.dispatcher - DEBUG - Setting singleton dispatcher as <telegram.ext.dispatcher.Dispatcher object at 0x7ff7b973f358>
2019-11-15 11:02:44,714 - JobQueue - DEBUG - Putting job repeat_submission with t=1573815764.714072
2019-11-15 11:02:44,714 - JobQueue - DEBUG - Putting job repeat_submission_list_clear with t=1573815764.714721
2019-11-15 11:02:44,714 - JobQueue - DEBUG - Putting job repeat_table_name with t=1573815764.714840
2019-11-15 11:02:44,714 - JobQueue - DEBUG - Putting job repeat_table_name_ugly with t=1573815764.714952
2019-11-15 11:02:44,715 - JobQueue - DEBUG - Putting job repeat_database_process with t=1573815764.715042
2019-11-15 11:02:44,715 - JobQueue - DEBUG - Putting job repeat_repeat_list_delete_one with t=1573815764.715124
2019-11-15 11:02:44,715 - JobQueue - DEBUG - Putting job repeat_appreciate_list_delete_one with t=1573815764.715253
2019-11-15 11:02:44,715 - JobQueue - DEBUG - Putting job repeat_rest_list_delete_one with t=1573815764.715346
[...]

PR #1623, opened 2019-11-19
Fix Server response could not be decoded using UTF-8

rikonaka:
- Remove the handling of `UnicodeDecodeError` errors beacause of using the `decoded_s = json_data.decode('utf-8', 'replace')`.

- Remove test for invalid JSON data cause there is no more error caused by decoding utf-8..

issues [#1609](https://github.com/python-telegram-bot/python-telegram-bot/issues/1609)

PR #1651, opened 2019-11-28
increase update_id by 1 to skip bad server response

isjerryxiao:
A possible fix for #1609.  
We actually should not ignore unicode decode error or json decode error in order to make it work, my solution is to increase last_update_id by 1 until that bad update is skipped.  
This may cause the bot to **lose some of its updates**, which is a little bit of a problem. However this is way better than **looping and throwing exceptions endlessly**.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -42,6 +42,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `JASON0916 <https://github.com/JASON0916>`_
 - `jeffffc <https://github.com/jeffffc>`_
 - `Jelle Besseling <https://github.com/pingiun>`_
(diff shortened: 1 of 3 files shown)


rikonaka:
Hi, guys, thanks for comment in my reply, and I had read your pull right now, but I think your pull [
84fd9b2](https://github.com/python-telegram-bot/python-telegram-bot/pull/1651/commits/84fd9b28a981bc853d687b2e175b80c1f14d59f0) is just a clone of my pull [
84a2a1](https://github.com/python-telegram-bot/python-telegram-bot/pull/1623/commits/84a2a14cece7b32e9753e6335a46b03a9a1561c3).

isjerryxiao:
However, 98e15ea is the main part.

Issue #1652, opened 2019-11-28
[BUG] ConversationHandler.TIMEOUT handlers get called with deprecated callback signature


tobiaswicker:
### Steps to reproduce
1. Create a `ConversationHandler` with a `ConversationHandler.TIMEOUT` state.

2. Add any handler to that state with a callback that has `(update: Update, context: CallbackContext)` as signature.

3. Watch it fail as `(bot, update)` are passed along to that callback.

### Expected behaviour
When `use_context=True` is set on the updater, it should call the callback with `(update: Update, context: CallbackContext)`.

### Actual behaviour
The callback is called with `(bot, update)` instead.

### Configuration
**Operating System:**
Any

**Version of Python, python-telegram-bot & dependencies:**
``$ python -m telegram``
python-telegram-bot 12.0.0
certifi 2019.06.16
future 0.17.1
Python 3.6.4

Issue #1509, opened 2019-09-06
Use CallbackContext in tests


Bibo-Joshi:
Tests should be updated to use `CallbackContext`, I guess. Probably more tiresome than complicated.

Poolitzer:
I thought about that, yeah, we should ;P

Bibo-Joshi:
Closing in favor of #1817

PR #1653, opened 2019-11-28
Fix wrong signature call for ConversationHandler.TIMEOUT handlers (issue #1652)

tobiaswicker:
Fix for issue #1652

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -393,8 +393,10 @@ class ConversationHandler(Handler):
         self.logger.debug('conversation timeout was triggered!')
 
         # Backward compatibility with bots that do not use CallbackContext
+        callback_context = None
         if isinstance(context, CallbackContext):
             job = context.job
+            callback_context = context
 
         context = job.context
 
@@ -409,5 +411,5 @@ class ConversationHandler(Handler):
         for handler in handlers:
             check = handler.check_update(context.update)
             if check is not None and check is not False:
-                handler.handle_update(context.update, context.dispatcher, check)
+                handler.handle_update(context.update, context.dispatcher, check, callback_context)
         self.update_state(self.END, context.conversation_key)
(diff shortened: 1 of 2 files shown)


tobiaswicker:
Forced push was necessary to trigger the checks to run again, because one failed for unrelated reasons.

Eldinnie:
`Thanks for making this fix.
Can we explore adding a test to make sure this (keeps) work(ing/s) as intended?

tobiaswicker:
@Eldinnie PTB does have tests for the `ConversationHandler` and there is even a `test_conversation_handler_timeout_state()` test case that specifically tests the `ConversationHandler.TIMEOUT` state. I think the issue you are talking about, is that those tests don't `use_context`, right?

So would you like to see the existing tests migrated to use context or rather have separate tests? I'd opt for the latter for backward compatibility for now.

Poolitzer:
@tobiaswicker Switching all the tests to contexts is actually an issue, #1509

tobiaswicker:
@Poolitzer OK, thanks! 

@Eldinnie does the mentioned issue satisfy your request for a test or have you had a test in mind that wouldn't be covered by migrating any of the existing tests?

Poolitzer:
@tobiaswicker I am very sure that solving that issue will satisfy his testing needs

Bibo-Joshi:
@Poolitzer Note, that #1509 is currently assigned only to the v13,0 milestone, which will probably not be closed too soon.

Poolitzer:
@Bibo-Joshi I mean, we are talking about a test for a callback here. I dont think it will break soon enough that we will have a problem with adding this test later, do you?

Bibo-Joshi:
@Poolitzer I guess not. Just wanted to mention it ;)

Poolitzer:
@Bibo-Joshi Fair enough

Eldinnie:
@tobiaswicker IMO the new PR should be paired with a test. One that would fail without this fix, and will pass with this fix.

tsnoam:
@tobiaswicker thankyou for your contribution

PR #1679, opened 2019-12-29
Add base_file_url support for telegram.ext.Updater

billchenchina:
This commit is for when sometimes we need to specify a Telegram Bot API file URL when using telegram.ext.Updater.


tsnoam:
the failure is unrelated. merging.

tsnoam:
@billchenchina thank you for your contribution

PR #1680, opened 2019-12-29
Make Filters.text accept leading slash

### LINKED ISSUE #1678: Fix Filters.text (docs)
As a legacy `filters.text` has a check that the message does not start with `/`. In order to not conflict with `Commandhandler`
https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/filters.py#L246
This is no longer needed because Commandhandler now uses entities to check for a command. In any case this bahavior is not documented.
A user can well be confused by `MessageHandler(Filters.text, callback)` will not respond to a message that starts with a `/`. At least the docs should be fixed, but I prefer it to remove the check and just pass any message with text to the user.

Bibo-Joshi:
Closes #1678

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -243,7 +243,7 @@ class Filters(object):
                 self.name = 'Filters.text({})'.format(iterable)
 
             def filter(self, message):
-                if message.text and not message.text.startswith('/'):
+                if message.text:
                     return message.text in self.iterable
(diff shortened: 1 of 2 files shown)


Eldinnie:
@Bibo-Joshi Test

Eldinnie:
@Bibo-Joshi test

Issue #1676, opened 2019-12-26
[QUESTION] job execution datetime and forcing/postponing jobs


Andrej730:
1. I need to obtain the exact time when the job will be executed. To do so i'm currently accessing job_queue._queue.queue. Is there some other function for this purpose that i'm missing?

2. I need to be able to force job execution or postpone it. Is there any more simple way to do it?
To force job execution I do (To postpone execution I do similar thing but without step 1 of course):
- create a new job with job_queue.run_once that will run the callback right now
- disable the current job and set it's interval to (base_interval - (t - now)). (t is time when the job supposed to be executed according to it's original interval)
- create a job that will run in (t - now) seconds and will enable the forced job and set it's interval to base_interval (so the job would keep working with the original interval as it was before forcing it).

And the problem with my method is that some time after the job was postponed or forced I cannot access the time it will be executed with job_queue._queue.queue and I need to calculate it and store it somewhere else.

I think postponing, forcing the jobs and accessing the time job will be executed should be more straightforward and it should be handled by the library. Am I wrong?

Eldinnie:
1. jobqueue.jobs

Andrej730:
> 1. jobqueue.jobs

It returns list of current jobs but how do I get the time when they will be executed?

Bibo-Joshi:
- To force execution, you can just run

    ```
    job.run(dispatcher)
    job.schedule_removal() #If you want to delete the job
    ```
    (see the [docs](https://python-telegram-bot.readthedocs.io/en/stable/telegram.ext.job.html#telegram.ext.Job.run))

- If you want to postpone a `run_once` job to a given time `T`, just create a new `run_once` job with interval `T - now` and delete the original one.
- If you want to postpone *a single* execution of a `run_repeating` job `j` to a given time `T`, I suggest deactivating `j` and creating a `run_once` job with interval `T-now` that also reactivates `j`.
- If you want to postpone *every* execution of a `run_repeating` job, I'd also suggest to delete it after creating a new job with the appropriate interval.

> And the problem with my method is that some time after the job was postponed or forced I cannot access the time it will be executed with job_queue._queue.queue and I need to calculate it and store it somewhere else.

I don't really get, what you mean by that

> I think postponing, forcing the jobs and accessing the time job will be executed should be more straightforward and it should be handled by the library. Am I wrong?

[...]

PR #1685, opened 2020-01-01
next_t property is added to Job class

Andrej730:
Added new property to Job class - next_t, it will show the datetime when the job will be executed next time.The property is updated during JobQueue._put method, right after job is added to queue.

Related to #1676


telegram/ext/jobqueue.py
```diff
@@ -376,7 +377,8 @@ class Job(object):
             Only optional for backward compatibility with ``JobQueue.put()``.
         tzinfo (:obj:`datetime.tzinfo`, optional): timezone associated to this job. Used when
             checking the day of the week to determine whether a job should run (only relevant when
-            ``days is not Days.EVERY_DAY``). Defaults to UTC.
+            ``days is not Days.EVERY_DAY``); also used when :attr:`next_t` is retrieved.
```
Bibo-Joshi: Personally I'd think this goes without saying. Can be removed IMO.

telegram/ext/jobqueue.py
```diff
@@ -103,6 +103,7 @@ def _put(self, job, time_spec=None, previous_t=None):
         # enqueue:
         self.logger.debug('Putting job %s with t=%f', job.name, time_spec)
         self._queue.put((next_t, job))
+        job._next_t = next_t
```
Bibo-Joshi: Accessing a protected member upsets Codacy. Please add a setter for `Job.next_t` and use that one here.

Andrej730: Added a setter. But now setter for `Job.next_t` requires timestamp value and getter returns value in a different format, `datetime`, which is kind of counterintuitive. Is it a problem?

telegram/ext/jobqueue.py
Andrej730: Sure, fixed.

telegram/ext/jobqueue.py
Bibo-Joshi: Good point. Could you make the setter accept both timestamp and datetime by converting datetime to timestamp using `helpers.to_float_timestamp()`? And raise a `TypeError` if a different type is passed.
The setter won't be used often, but still …

tests/test_jobqueue.py
```diff
@@ -325,3 +326,29 @@ def test_context_based_callback(self, job_queue):
...
+
+                if job.removed:
+                    assert job.next_t is None
+                else:
+                    assert job.next_t == t
+
+        sleep(0.03)
+
+        assert job1.next_t is None
+        assert job2.next_t is None
```
Bibo-Joshi: Could you add assertions checking that `self.result` (updated by `self.job_run_once`) has the correct value after each job execution? Just for completenes sake …

Andrej730: Added.

telegram/ext/jobqueue.py
Andrej730: Added.

tests/test_jobqueue.py
Andrej730: I also noticed that setters Job.days and Job.interval raise ValueError when TypeError is more appropriate - changed it to TypeError in 6b1fce5 and 8231df1

telegram/ext/jobqueue.py
```diff
@@ -485,6 +485,14 @@ def next_t(self):
 
     @next_t.setter
     def next_t(self, next_t):
+        if isinstance(next_t, datetime.datetime):
+            # Set timezone to UTC in case datetime is in local timezone.
+            next_t = next_t.replace(tzinfo=datetime.timezone.utc)
```
Bibo-Joshi: I think we'd rather want `next_t.astimezone(UTC)` here, looking at the [docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone). Do you agree?

tests/test_jobqueue.py
```diff
@@ -348,7 +349,14 @@ def test_job_next_t_property(self, job_queue):
...
                     assert job.next_t == t
 
-        sleep(0.03)
+        assert self.result == 1
+        sleep(0.02)
 
+        assert self.result == 2
         assert job1.next_t is None
         assert job2.next_t is None
+
```
Bibo-Joshi: Can you put this in a separate `test_job_next_t_setter`? That way the output of the pytest is more precise.

tests/test_jobqueue.py
```diff
@@ -348,7 +349,14 @@ def test_job_next_t_property(self, job_queue):
...
+        t = dtm.datetime.now(tz=dtm.timezone(dtm.timedelta(hours=12)))
+        job1.next_t = t
+        job1.tzinfo = dtm.timezone.utc
+        assert job1.next_t == t.replace(tzinfo=job.tzinfo)
```
Bibo-Joshi: See above

telegram/ext/jobqueue.py
Andrej730: Sure, it will be more readable.

tests/test_jobqueue.py
Andrej730: Moved to separated test.

Andrej730:
Because after this PR `Job.timetz` will be used more often, I've added option to define `Job.timetz` from `run_repeating` and `run_once` by passing `timetz` in one of their `datetime` arguments (discussed in #1696).

telegram/ext/jobqueue.py
```diff
@@ -130,6 +131,9 @@ def run_once(self, callback, when, context=None, name=None):
                   job should run. This could be either today or, if the time has already passed,
                   tomorrow.
 
+                If ``when`` is :obj:`datetime.datetime` or :obj:`datetime.datetime` type
```
tsnoam, line 134:
```suggestion
                If ``when`` is :obj:`datetime.datetime` or :obj:`datetime.time` type
```

telegram/ext/jobqueue.py
```diff
@@ -170,6 +181,9 @@ def run_repeating(self, callback, interval, first=None, context=None, name=None)
                   job should run. This could be either today or, if the time has already passed,
                   tomorrow.
 
+                If ``first`` is :obj:`datetime.datetime` or :obj:`datetime.datetime` type
```
tsnoam, line 184:
```suggestion
                If ``first`` is :obj:`datetime.datetime` or :obj:`datetime.time` type
```

telegram/ext/jobqueue.py
```diff
@@ -205,6 +222,7 @@ def run_daily(self, callback, time, days=Days.EVERY_DAY, context=None, name=None
                 or change it to a repeating job.
             time (:obj:`datetime.time`): Time of day at which the job should run. If the timezone
                 (``time.tzinfo``) is ``None``, UTC will be assumed.
+                ``time.tzinfo`` also defines ``Job.tzinfo``.
```
tsnoam, line 225:
```suggestion
                ``time.tzinfo`` will implicitly define ``Job.tzinfo``.
```

telegram/ext/jobqueue.py
```diff
@@ -226,10 +244,10 @@ def run_daily(self, callback, time, days=Days.EVERY_DAY, context=None, name=None
                   interval=datetime.timedelta(days=1),
                   repeat=True,
                   days=days,
-                  tzinfo=time.tzinfo,
                   context=context,
                   name=name,
-                  job_queue=self)
+                  job_queue=self,
+                  tzinfo=time.tzinfo)
```
tsnoam, line 250: There is no real change here, only confuses the reader or possible code merges.
Please revert.

telegram/ext/jobqueue.py
```diff
@@ -387,7 +406,7 @@ def __init__(self,
                  days=Days.EVERY_DAY,
                  name=None,
                  job_queue=None,
-                 tzinfo=_UTC):
+                 tzinfo=None):
```
tsnoam, line 409: Please rebase/merge with master.
It will make the diff smaller.

telegram/ext/jobqueue.py
```diff
@@ -469,6 +490,28 @@ def interval_seconds(self):
...
+        """
+        ::obj:`datetime.datetime`: Datetime for the next job execution.
+            Datetime is localized according to :attr:`tzinfo`.
+            If job is removed or already ran it equals to ``None``.
+
+        """
+        return datetime.datetime.fromtimestamp(self._next_t, self.tzinfo) if self._next_t else None
+
+    @next_t.setter
+    def next_t(self, next_t):
```
tsnoam, line 504: Setting `next_t` should not be accessible to users. It is an internal implementation detail which we merely give the option to peek at.
Setting `next_t` will most like not have the affect a user would expect (it would probably do nothing and only change the value which can be "peeked" at).

So, what I'm suggesting is:
1. This shouldn't be a property setter.
2. Rename this function to: `def _set_next_t(self, next_t):`
3. Line 309 (part of the JobQueue code) will do:
```python
job._set_next_t(next_t)
```
Despite looking "ugly" (as `JobQueue` is accessing a "private" method of `Job`), it is not. The way I see it, `Job` is an implementation detail of `JobQueue` and as such, `JobQueue` is allowed to access such a "private" method.

Bibo-Joshi:
Codacy fails as expected, test fail is unrelated. Merging.

Thank you for your contribution, @Andrej730!

Issue #1538, opened 2019-10-10
drop python2 support


Poolitzer:
That means to remove it from:

- [ ] README.rst
- [ ] CONTRIBUTING.rst
- [ ] .pre-commit-config.yaml
- [ ] .travis.yml
- [ ] appveyor.yml
- [ ] setup.py
- [ ] tox.ini
- [ ] all tests (hihi)

We can even think about dropping all the python2 workarounds we have in the library itself:

- [ ] remove `future` from requirements and all the imports
- [ ] `_UTC` in `telegram.utils.helpers`
- [ ] Workaround in `telegram.InputFile.__init__()`
- [ ] `from builtins import str` in `telegram.utilts.request`
- [ ] `test_send_on_server_emoji`
- [ ] imports in `test.updater`

(This list is not comprehensive)

Poolitzer:
@jh0ker funny how you write the most work required part. But yes. We should.

Bibo-Joshi:
Took the liberty to add some ToDos to the description ;) We can update that list, if we stumble across some workaround while coding something else …

septatrix:
Hey, if possible I would like to work on this.
I already started a bit and would create a WIP merge request tomorrow...

Bibo-Joshi:
Sure, please go ahead :)

septatrix:
Pull request is now live #1715...

Poolitzer:
#1731 removed 2.7 from the workflow to reduce testing time

Bibo-Joshi:
Closed by #1715

PR #1686, opened 2020-01-01
Updated readme to comply with use of python3

NilsDeckert:
As of 01.01.2020 python 2.7 is deprecated


Poolitzer:
Hey there.

While appreciated and needed, we will do all of the depreciate python two change in one issue/PR, not split them up between several ones :)

See issue #1538

PR #1696, opened 2020-01-07
Fix for tzinfo issues in JobQueue

### LINKED ISSUE #1693: [BUG] Job_queue is skipping job if timezone is not provided for job.run_daily
For example, my local timezone is +3 UTC and I want to run job every working day at 23:00 UTC.

`launch_time.tzinfo` is `None` so it's expected to be interpreted as 23:00 UTC.
```python
launch_time = dtm.time(hour=23)
job = job_queue.run_daily(run_once, launch_time,
                          days=(0, 1, 3, 4))
```

Since `launch_time.tzinfo` was `None`, `job.tzinfo` is also equals `None` because of the following code from `JobQueue.run_daily` https://github.com/python-telegram-bot/python-telegram-bot/blob/3d42df3366d7cff030c3ce06d6a95259b3bdf62e/telegram/ext/jobqueue.py#L225-L232

Let's assume now is 23:00 UTC Friday. 

The following code from `JobQueue.tick` won't run the job because 
1) `datetime.datetime.now(None)` returns local time and local time now is 02:00 Saturday.
2) then `datetime.datetime.now(job.tzinfo).date().weekday()` will return `5`
3) `any(day == current_week_day for day in job.days)` will return `False` because we're looking for `(0, 1, 3, 4)`
So job won't be run until 23:00 UTC Sunday.
https://github.com/python-telegram-bot/python-telegram-bot/blob/3d42df3366d7cff030c3ce06d6a95259b3bdf62e/telegram/ext/jobqueue.py#L275-L280

And one more thing - when #1685 will be merged, in the described case `Job.next_t` will be returning local time when utc time is expected:
```python
    @property
    def next_t(self):
        return datetime.datetime.fromtimestamp(self._next_t, self.tzinfo) if self._next_t else None

```

Andrej730:
Closes #1693

Changes:
1. Added optional argument `tzinfo` for `run_once`, `run_repeating`.
2. Made sure `job.tzinfo` will be set as UTC by default and not `None`. 
3. Added test that checks that all methods by default set `job.tzinfo` as UTC.

First I've added `tzinfo` argument to `run_daily` too but it's seems excessive. For example, to make sure that job will be ran every monday and tuesday on  14:00 in desired timezone you'd had to pass timezone twice - as `time.tzinfo` and as `tzinfo`. 
```python
target_tzinfo = dtm.timezone(dtm.timedelta(hours=3))
target_time = dtm.time(hour=14).replace(tzinfo=target_tzinfo)
job_queue.run_daily(
    job_callback,
    time=target_time,
    days=(0, 1),
    tzinfo=target_tzinfo)
```
And I think there is no need to use different timezones for `time.tzinfo` and `tzinfo` because result may be counterintuitive. So, I've removed `tzinfo` from `run_daily`.

### DIFF
--- a/telegram/ext/jobqueue.py
+++ b/telegram/ext/jobqueue.py
@@ -275,7 +275,7 @@ class JobQueue(object):
             if job.enabled:
                 try:
                     current_week_day = datetime.datetime.now(job.tzinfo).date().weekday()
-                    if any(day == current_week_day for day in job.days):
+                    if current_week_day in job.days:
                         self.logger.debug('Running job %s', job.name)
                         job.run(self._dispatcher)
 
@@ -387,7 +387,7 @@ class Job(object):
                  days=Days.EVERY_DAY,
                  name=None,
                  job_queue=None,
-                 tzinfo=_UTC):
+                 tzinfo=None):
 
         self.callback = callback
         self.context = context
@@ -400,7 +400,7 @@ class Job(object):
 
         self._days = None
         self.days = days
-        self.tzinfo = tzinfo
+        self.tzinfo = tzinfo or _UTC
 
         self._job_queue = weakref.proxy(job_queue) if job_queue is not None else None
 
--- a/tests/test_jobqueue.py
+++ b/tests/test_jobqueue.py
@@ -28,7 +28,7 @@ from flaky import flaky
 
 from telegram.ext import JobQueue, Updater, Job, CallbackContext
 from telegram.utils.deprecate import TelegramDeprecationWarning
-from telegram.utils.helpers import _UtcOffsetTimezone
+from telegram.utils.helpers import _UtcOffsetTimezone, _UTC
 
 
 @pytest.fixture(scope='function')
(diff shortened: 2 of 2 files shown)


Andrej730:
You're right in your argumentation - `tzinfo` is only really used in `run_daily`. My thought was to add `tzinfo` to all methods because it will be helpful for #1685 ( `next_t` uses `Job.tzinfo` to return datetime of the next job execution) - because after adding `next_t` property `Job.tzinfo` will be used more often so it will be convenient to provide option to define `tzinfo` during job initiation instead of:

```python
job = job_queue.run_repeating(self.job_run_once, 10)
job.tzinfo = dtm.timezone(dtm.timedelta(hours=3))
```

Bibo-Joshi:
Okay, I see your point. However, I think it would be more appropriate then to move the addition of the `tzinfo` argument to #1685, because this is where it's acutally needed. Might seem very nitpicky, but a logical changelog often is crucial in debugging :D
Two more thoughts:
* `tzinfo` should only be taking into account in the `run_...` methods, if there is no other way to pass it. E.g. for `run_repeating`, if `first` is passed as `(date)time`, this should be aware instead of passing an additional `tzinfo`. Please add corresponding checks that raise `ValueError`s
* We will also need test cases for the new argument. Both for the `Job` class and to make sure that the `run_...` methods use it correctly

Andrej730:
Agree, I reverted changes except actual fix for #1693

Bibo-Joshi:
CI fail is unrelated

Bibo-Joshi:
Thank you for you contribution!

PR #1721, opened 2020-01-23
Improve handling of CallbackQuery in Examples

### LINKED ISSUE #1712: Answer CallbackQueryies in Examples
Not answering the queries may lead to problems on some clients even if the keyboard changes.

Bibo-Joshi:
* Answer `CallbackQueries` even if message is edited
* Use `CQs` shortcut for `edit_message_text`

Closes #1712


examples/inlinekeyboard.py
```diff
@@ -29,6 +29,10 @@ def start(update, context):
 def button(update, context):
     query = update.callback_query
 
+    # CallbackQueries need to be answered, even if no notification to the user is needed
+    # Some clients may have trouble otherwise. See https://core.telegram.org/bots/api#callbackquery
+    query.answer()
+
```
Poolitzer: I tend to answer last, after everything else is done. But won't matter.

examples/inlinekeyboard2.py
```diff
@@ -65,9 +66,7 @@ def start_over(update, context):
     # Instead of sending a new message, edit the message that
     # originated the CallbackQuery. This gives the feeling of an
     # interactive menu.
-    bot.edit_message_text(
-        chat_id=query.message.chat_id,
-        message_id=query.message.message_id,
+    query.edit_message_text(
```
Poolitzer: good catch, way better

examples/nestedconversationbot.py
```diff
@@ -66,6 +66,7 @@ def start(update, context):
 
     # If we're starting over we don't need do send a new message
     if context.user_data.get(START_OVER):
+        update.callback_query.answer()
```
Poolitzer: maybe we should do query = update.callbackquery in this file as well?

Bibo-Joshi:
@Poolitzer I understood from @EchteEldin that the Keyboard wouldn&#x27;t change for him because the CallbackQuery wasn&#x27;t answered before editing the message. That&#x27;s why I answer before editing.

Bibo-Joshi:
CI fails unrelated, since I only changed the examples

PR #1723, opened 2020-01-23
4.6 API update

Poolitzer:
docs need to be built and checked. Tests should run through now.

- [x] When #1508 is merged to master, merge master and update `Message._parse_*` to handle the `MessageEntities.language` attribute correctly. Doesn't actually do anything yet, but we should still add it to `Message.text_html` and friends
- [x] Make `test_official` fail again (i.e. partly revert #1716)

closes #1722

PR #1724, opened 2020-01-23
Make (most) objects comparable

### LINKED ISSUE #1708: [FEATURE] Add _id_attrs
`InlineKeyboardMarkup` and `InlineKeyboardButton` objects are currently not really testable for equality because the `_id_attrs` are not set. We should add all attributes of `IKB` to the `_id_attr` and `inline_keyboard` for `IKM`.

#### Additional context
We already did that for `MessageEntity` iirc. Probably should go through the lib and do that for all the telegram objects.

#### Current workaround
Use `__dict__` for everything

Bibo-Joshi:
Closes #1708

Added `_id_attrs` to most of the telegram objects (more or less) by the following paradigm:
1. If object has some sort of ID, just use that
2. If object has only optional args, use all of them
3. If object has required arg, use that
4. If object hast required arg and exactly one of the optionals must be given, use all

Also added the corresponding tests.

On the way I encountered some difficulties:
1. Objects with nested lists as attributes, which are not hashable: I had to overwrite `__hash__` to use a tuple representation of the lists, e.g. `InlineKeyboardMarkup`
2. `ReplyKeyboardMarkup` accepts both strings and `KeyboardButtons` while converting stings to `KB` only on `to_dict`. So I overwrote `__equal__` to cope with this
3. There was no test for `WebhookInfo`. I added one
4. I couldn't come up with a clever way to test `InputFile.input_media_content` for equality (in case an actual file handler is passed, that is), so I left that and `InputMedia…` out
5. I have no clue of the passport module. All objects that have an ID already had the `_id_attrs` set and it seemed to me that all the other classes have no stand alone tests. So I left them out as well …

Needs to be updated, when 

- [x] API 4.5 #1508 
- [x] API 4.6 #1723
- [x] API 4.7 #1858 
- [x] API 4.8

are merged


Bibo-Joshi:
equality test for `InlineKeyboardMarkup` fails on Py2.7. Don't know why, but don't really care 🤷‍♂

Poolitzer:
thanks for asking me to review, but I gonna decline, this is too much for me :D

Bibo-Joshi:
`test_set_game_score_1` is the one again …

Bibo-Joshi:
Just a thought: should we make `TelegramObject.__eq__` raise an error or warning, if it's comparing two objects, one of which has an emtpy `_id_attrs` tuple, i.e. it's not comparable?

telegram/base.py
```diff
@@ -78,6 +80,9 @@ def to_dict(self):
 
     def __eq__(self, other):
         if isinstance(other, self.__class__):
+            if self._id_attrs == ():
```
tsnoam, line 83: What about checking `other._id_attrs`?

Bibo-Joshi: Mh … you're right. Will add.

telegram/forcereply.py
```diff
@@ -49,3 +49,5 @@ def __init__(self, force_reply=True, selective=False, **kwargs):
         self.force_reply = bool(force_reply)
         # Optionals
         self.selective = bool(selective)
+
+        self._id_attrs = (self.force_reply, self.selective)
```
Bibo-Joshi, line 53: Adding `selective` to `_id_attrs` differs from the other classes with required args, where no optionals where added. However, `selective` is quite a difference `ForceReply`, actually the only relevant setting (as `force_reply` is always `True`, i.e. the API docs state its Type as `True`, not `Boolean`)

telegram/inline/inputcontactmessagecontent.py
```diff
@@ -48,3 +48,5 @@ def __init__(self, phone_number, first_name, last_name=None, vcard=None, **kwarg
         # Optionals
         self.last_name = last_name
         self.vcard = vcard
+
+        self._id_attrs = (self.phone_number, self.first_name)
```
Bibo-Joshi, line 52: One *could* argue, that `last_name` should be added. That reduces the risk of evaluating to people with the same (first) name as equal.

telegram/inline/inputlocationmessagecontent.py
```diff
@@ -43,3 +43,5 @@ def __init__(self, latitude, longitude, live_period=None, **kwargs):
         self.latitude = latitude
         self.longitude = longitude
         self.live_period = live_period
+
+        self._id_attrs = (self.latitude, self.longitude, self.live_period)
```
Bibo-Joshi, line 47: As keeping live locations up to dateis not easy as far as I understood from the recent discussion in the user group, one could arge that one could skip `live_period` here

telegram/inline/inputtextmessagecontent.py
```diff
@@ -55,3 +55,5 @@ def __init__(self,
         # Optionals
         self.parse_mode = parse_mode
         self.disable_web_page_preview = disable_web_page_preview
+
+        self._id_attrs = (self.message_text,)
```
Bibo-Joshi, line 59: `text='*bold*', parse_mode='Markdown'`  is arguably not the same as `text='*bold*', parse_mode=None`. Add `parse_mode`?

telegram/poll.py
```diff
@@ -43,6 +43,8 @@ def __init__(self, text, voter_count, **kwargs):
         self.text = text
         self.voter_count = voter_count
 
+        self._id_attrs = (self.text,)
```
Bibo-Joshi, line 46: `voter_count` is required. should be added, as polls might have duplicate options.

telegram/inline/inputcontactmessagecontent.py
Bibo-Joshi: Then again, for `Contact`, we only use phone number. That seems cleaner to me, as the same phone number shouldn't appear twice.

Bibo-Joshi:
* Added documentation, also for the classes, who already had id attrs.
* Addressed some of the above mentioned stuff
* In the process, I added `message.chat` to the id attrs, as `message_id` is not unique across chats. Strictly speaking this is somewhat breaking. Then again, the behaviour wasn't documented yet, so I'd sleep well with that …
* Coday complains: Access to a protected member _id_attrs of a client class: `if other._id_attrs == ():` That's okay imo.
* py3.6 fails on `set_game_score` …

telegram/inline/inputcontactmessagecontent.py
Poolitzer: Agreed. Phone number should be enough to uniquely identify someone, first name on top is nice to have, but definitely enough

telegram/inline/inputtextmessagecontent.py
Poolitzer: No Id say plain text should be enough here. Doesnt matter how markup changes it in the end on the client.

telegram/payment/orderinfo.py
```diff
@@ -24,6 +24,10 @@
 class OrderInfo(TelegramObject):
     """This object represents information about an order.
 
+    Objects of this class are comparable in terms of equality. Two objects of this class are
+    considered equal, if their :attr:`name`, :attr:`phone_number`, :attr:`email` and
+    :attr:`shpping_address` are equal.
```
Poolitzer, line 29:
```suggestion
    :attr:`shipping_address` are equal.
```

Poolitzer:
Sorry, I forgot the actual issue I have with this: Documentation. I like that you wrote down what gets compared where, but I dont like that being before the attributes of the object. I feel like most users will want to read about attributes rather then what gets compared if I compare this object. What about putting the what-gets-compared paragraph at the end, after the parameters?

Bibo-Joshi:
> What about putting the what-gets-compared paragraph at the end, after the parameters?

I fear that it's heard to find it there, especially in the docs of classes with many attributes. And it's only 1-3 lines, so it doesn't take up too much space, imo … The alternative would be to have an explicit docstring for `__eq__()`, but I don't think that I like that …

telegram/files/venue.py
```diff
@@ -24,6 +24,9 @@
 class Venue(TelegramObject):
     """This object represents a venue.
 
+    Objects of this class are comparable in terms of equality. Two objects of this class are
+    considered equal, if their :attr:`location` and :attr:`title` are equal.
```
Bibo-Joshi, line 28: `address` is required, too. We should add it to the id_attrs

Bibo-Joshi:
Some concerns about this PR were voiced offline. I'll try to summarize in hopes of not missing the point:

* Implementing `__eq__()` to compare only a subset of a classes attributes will be unintuitive for users
* This is especially true for TG objects that can change, e.g. two `Message` objects will be equal when testing for `message_id` only, while maybe having different text
* Custom comparison by attributes in easy to achive, so there is no reason to implement `__eq__()` 

While I don't agree with this argumentation, I find the arguments important and like to have this properly discussed before going for merge. For completeness sake, here is my line of thought:

* We already have `__eq__()` implemented for objects with some kind of ID. So we could a) remove that capability, b) leave it as is, or c) extend it to all/most objects
* The default implementation of `__eq__()` is useless - I'd rather use `is`, if I want to compare by object identity. So giving objects a meaningful comparison is crucial imo, which is why I went for c)
* Currently this PR aims to set `_id_attrs` to the attrs that uniquely identify an object, i.e. this comparison is the "correct" one in most use cases
* This PR adds documentation, so `==` is no black box and if users need a different comparison, they can still implement that

Bibo-Joshi:
We discussed the PR in the dev chat again and decided to merge.
I went over the changes once again ad made some minor corrections.
py3.8 fail unrelated, codacy has v13 issues. Merging.

PR #1725, opened 2020-01-24
Add --with-urllib3 option to remove vendored version

### LINKED ISSUE #1583: [FEATURE] Make vendored urllib3 optional during setup.py
This is following the request in #1568 

We need to be able to identify when this is the case, thus `python -m telegram` should provide adequate information.

pingiun:
Fixes #1583 

To use this feature, install using:
```bash
python setup.py install --with-urllib3
```


pingiun:
It seems the failing tests are not of my wrongdoing, I can rebase whenever those are fixed/removed

PR #1730, opened 2020-01-26
Add strikethrough to available types

dpaleino:
While developing a Telegram bot, I saw that parse_entities() filtered "strikethrough" entities out, i.e. they were available in the message sent from the API, but wasn't there in the parse_entities() output.

"strikethrough" entities are pieces of text surrounded by ~~ (i.e. ~~like this~~).

Thank you for considering this patch.


Bibo-Joshi:
@dpaleino Hi. That&#x27;s because the API 4.5 changes have not yet been implemented. we have #1508 for that, but it might take a but until it&#x27;s merged. Thanks for the PR nonetheless :)

dpaleino:
Ah, I completely missed that!

Will wait for an official update, then!

Thank you,
David

PR #1741, opened 2020-01-31
fixing message.link

Poolitzer:
the good eyes of @Bibo-Joshi saw that message.link was wrongfully generated for private groups. This PR fixes that as well as expanding tests for this case

### DIFF
--- a/telegram/message.py
+++ b/telegram/message.py
@@ -339,8 +339,8 @@ class Message(TelegramObject):
     @property
     def link(self):
         """:obj:`str`: Convenience property. If the chat of the message is not
-        a private chat, returns a t.me link of the message."""
-        if self.chat.type != Chat.PRIVATE:
+        a private chat or normal group, returns a t.me link of the message."""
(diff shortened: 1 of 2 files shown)


telegram/message.py
```diff
@@ -339,8 +339,8 @@ def chat_id(self):
     @property
     def link(self):
         """:obj:`str`: Convenience property. If the chat of the message is not
-        a private chat, returns a t.me link of the message."""
-        if self.chat.type != Chat.PRIVATE:
+        a private chat or normal group, returns a t.me link of the message."""
+        if self.chat.type not in [Chat.PRIVATE, Chat.GROUP]:
             if self.chat.username:
                 to_link = self.chat.username
             else:
```
Bibo-Joshi: You can shorten this if clause. if there is no username, now it has to be supergroup and we just need to skip the leading -100

Poolitzer: right

Poolitzer:
fail unrelated.

PR #1744, opened 2020-01-31
Make Filters.command only accept MessageEntity commands

### LINKED ISSUE #1755: Make `filters.command` work as expected
Now `filters.text` allows messages starting with a `/`, not excluding bot commands, we need a fast and simpler way to stop those entering `filters.text` if you want to exclude them.

I propose to change `filters.command` to the following behavior:
- invoking `filters.command` will filter any messages with `entity.type='bot_command`. Basically a mirror of `filters.entity(['bot_command']) which we can just invoke this way)
- invoking `filters.command(only_start=True) | filters.command(True)` (name is open for discussion) filters any messages that **start** with a bot command.

this will allow for easily using `filters.text & ~filters.command(true)` to not get any command updates with your text filter.

Optionally we could make `filters.command` a `data_filter` and add the found commands to CallbackContext` as the `context.commands` (list). Although maybe that is outside this scope and then should also be added for commandhandler to add.

Bibo-Joshi:
Since `CommandHandler` is only accepting commands marked by a `MessageEntity` since v12.0.0 (#1114), I thought it fitting for `Filters.command` to do the same.
However, this is a breaking change, which is why I put this into the v13 Milestone.

Closes #1755

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -249,10 +249,7 @@ class Filters(object):
 
         def __call__(self, update):
             if isinstance(update, Update):
-                if self.update_filter:
-                    return self.filter(update)
-                else:
-                    return self.filter(update.effective_message)
+                return self.filter(update.effective_message)
             else:
                 return self._TextIterable(update)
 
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
Added changes as discussed in #1755 . Implementation is similar to the one of `filters.text(iterable)` to make sure, that nobody as to change `filters.command` and `filters.command()`

Bibo-Joshi:
CI fail unrelated (`TestUpdater.test_webhook_invalid_posts`)

PR #1750, opened 2020-02-03
Rename Test suite …

Bibo-Joshi:
… so that the badge on `README` doesn't say "Test your PR" anymore.
`README` doesn't show it yet, but afaics this is because there has been no scheduled test with the new name yet …

Fail unrelated, see #1752


Eldinnie:
Unrelated CI problems

PR #1751, opened 2020-02-03
Bump v12.4

Bibo-Joshi:
The PRs currently still open in M12.4 are already added.

Correct date needs to be inserted

Fails unrelated b/c I only changed docs


tsnoam:
merged manually

PR #1756, opened 2020-02-06
Make ConversationHandler attributes immutable

### LINKED ISSUE #1029: ConversationHandler is not thread safe

Bibo-Joshi:
This should close #1029

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -165,23 +165,23 @@ class ConversationHandler(Handler):
                  persistent=False,
                  map_to_parent=None):
 
-        self.entry_points = entry_points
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
CI Fail unrelated (`test_webhook_invalid_posts`)

PR #1757, opened 2020-02-06
Allow updating user_ids/usernames of a running filters.user

### LINKED ISSUE #1562: [BUG] Change condition in Filters.user to avoid error with empty lists
Consider changing this condition
https://github.com/python-telegram-bot/python-telegram-bot/blob/88eccc6608d141004d231b38df76492e1c5f5b93/telegram/ext/filters.py#L796
 to something like
```python
if (user_id is None) == (username is None):
```
To avoid throwing the `ValueError` _"One and only one of user_id or username must be used"_ when an empty list of `user_id`s **or** `username`s is issued.

Similarly for `Filters.chat`.

PS This should also be 3 to 4 times faster.

Bibo-Joshi:
1. Makes `filters.user.{user_ids, usernames}` editable by:
    1. Making `user_ids` and `usernames` properties
    2. Adding locks for thread safety
2. Adds tests for editing the attributes

If I got that right, this closes #1562


Bibo-Joshi:
Maybe a comment on the change to sets is in order: I did this because we need only user id/name only once and that way removing a user from the filter is easier.
This kind of breaks backwards compitibility, but only kind of, as `Filters.user` was not really designed to be mutable.
Then again, I'm not very passionate about that change and can revert it.

telegram/ext/filters.py
```diff
@@ -850,29 +851,68 @@ class user(BaseFilter):
...
-                self.usernames = [username.replace('@', '')]
-            else:
-                self.usernames = [user.replace('@', '') for user in username]
+
+            self._user_ids_lock = Lock()
+            self._usernames_lock = Lock()
+
+            # Initialize in a way that will not fail the first setter calls
+            self._user_ids = user_id
+            self._usernames = username
```
tsnoam, line 878: Maybe initialize both to sets, saving some extra validation further down the line?

Correct me if I'm wrong, but for the usecase of this filter, a None value and an empty set has equivalent meaning.

Having this filter with two empty sets (usernames, user_ids) will mean it will allow all messages (my suggestion. you may decide exactly the opposite. you can even suggest controlling it with a variable)

```suggestion
            if isinstance(user_id, int):
                user_id = {user_id}
            elif not isinstance(user_id, set):
                user_id = set(user_id)
            self._user_ids = user_id
            
            # and similarly for `self._usernames`, `usernames` and compare to str & set
```

p.s. you should unify the if...elif....actually_set code with the property setters down below

telegram/ext/filters.py
```diff
@@ -850,29 +851,68 @@ class user(BaseFilter):
...
+            # Actually initialize
+            self.user_ids = user_id
+            self.usernames = username
+
+        @property
+        def user_ids(self):
+            with self._user_ids_lock:
+                return self._user_ids
```
tsnoam, line 886: If you return the (mutable) object itself, it might be changed again after the lock is released. or even the caller might change it. you need to return a copy of the object (preferably frozenset, as it won't give any illusions to the caller).

So:
```suggestion
            with self.__lock:
                return frozenset(self._user_ids)
```

telegram/ext/filters.py
```diff
@@ -850,29 +851,68 @@ class user(BaseFilter):
...
+
+        @user_ids.setter
+        def user_ids(self, user_id):
+            if (user_id is None) == (self.usernames is None):
+                raise ValueError('One and only one of user_id or username must be used')
```
tsnoam, line 891: 1. Take the lock before looking at privates, not just before you're about to write to them.
2. Thinking of Locks, you may want to consider a RwLock, but for the simplicity of first implementation lets stick to a plain Lock.
3. Since I'm considering an empty Set and None to be equivalent, code may be simplified a bit:
```suggestion
          with self.__lock:
              if user_id and self.usernames:
                    raise RuntimeError("Can't set user_id in conjunction with (already set) usernames")
```

telegram/ext/filters.py
```diff
@@ -850,29 +851,68 @@ class user(BaseFilter):
...
+            with self._user_ids_lock:
+                if user_id is None:
+                    self._user_ids = None
+                elif isinstance(user_id, int):
+                    self._user_ids = set([user_id])
+                else:
+                    self._user_ids = set(user_id)
```
tsnoam, line 898: Unity this code with what I suggested during `__init__()`

telegram/ext/filters.py
```diff
@@ -850,29 +851,68 @@ class user(BaseFilter):
...
+
+        @property
+        def usernames(self):
```
tsnoam, line 901: Similar comments like with `user_id` getter

telegram/ext/filters.py
```diff
@@ -850,29 +851,68 @@ class user(BaseFilter):
...
+            with self._usernames_lock:
+                return self._usernames
+
+        @usernames.setter
+        def usernames(self, username):
```
tsnoam, line 906: Similar comments like with `user_id` setter

telegram/ext/filters.py
Bibo-Joshi: I tried to incorporate most of your suggestions. As to the ``None`` question:

The way this filter was designed before you could only user *either* usernames *or* user_ids. I guess because it's not clear what do to, if a user is allowed by username but not by id or vice versa. Personally, I wouldn't wanna change that. But that's why from my perspective ``None`` is not equal to an empty set here. So I kept if ``None`` checks.

tsnoam: 1. I understand why not have both types (usernames/groups).
2. The only question is what considered an as no value: None, empty set or both?

I claim that an empty set means: no value was set at all.

Where is a potential problem? When both usernames and userids are empty sets.
However, this is now an allowed scenario: set may start empty, or will be cleared further down the line.

So how do I suggest to approach this:
In the filter itself perform following checks:

```
if self._usernames:
   return filter_by_username()
if self._user_ids:
   Return filter_by_user_ids()
Return default_value

```
Default value can be denied for backward compat but we should allow control to the user to use a different default value (allow)

telegram/ext/filters.py
Bibo-Joshi: Just a thought: This way something like `filter.user_ids.add(chat_id)` won't work and instead one would have to do

```python
current_ids = set(filer.user_ids)
current_ids.add(chat_id)
filters.user_ids = current_ids
```

but that's probably the only really save way to do it?

telegram/ext/filters.py
Bibo-Joshi: Okay, I think I get it now. I'll update.

Bibo-Joshi:
Two more things:

1. I just noticed, that `Filters.chat` currently does pretty much the same as `Filters.user`. I think it would be good to update it as well. I'll try and set up a base class for the two of them in order to reduce code duplication
2. As it wasn't really discussed yet: We're okay with switching from lists to sets for usernames/ids? Users shouldn't have fiddled with them anyway, so if they if that's their problem? :D

telegram/ext/filters.py
tsnoam: you are correct * 2
in order to be thread safe, we must control the modification of the internal objects ourselves.
also, the fact we're using set is an implementation detail. we shouldn't be exposing it as an API.

tsnoam:
> * I just noticed, that `Filters.chat` currently does pretty much the same as `Filters.user`. I think it would be good to update it as well. I'll try and set up a base class for the two of them in order to reduce code duplication

good idea. go for it.

> 
> * As it wasn't really discussed yet: We're okay with switching from lists to sets for usernames/ids? Users shouldn't have fiddled with them anyway, so if they if that's their problem? :D
as i see it, whether it's a list or set is an implementation detail. not exposed to the user. (the user may pass anything which we can iterate. we'll convert it into a set)

so what i'm saying, sets are good and for large numbers will provide better performance. in small numbers it wouldn't matter if it's a list or set.

telegram/ext/filters.py
```diff
@@ -865,73 +865,83 @@ class user(BaseFilter):
...
-                if user_id is None:
-                    self._user_ids = None
-                elif isinstance(user_id, int):
-                    self._user_ids = set([user_id])
-                else:
-                    self._user_ids = set(user_id)
+            with self.__lock:
+                if user_id and self._usernames:
+                    raise RuntimeError(('Can\'t set user_id in conjunction with (already set) '
+                                        'usernames.'))
```
tsnoam, line 919: Why the extra `(`? it won't be treated as a tuple (as you didn't add a comma at the end)

Also, being the nitpicker I am, it really hurts my eyes to see backspace: `Can\'t` 
This is a legitimate place to use double quotes so you won't have to escape the `'`.

telegram/ext/filters.py
```diff
@@ -865,73 +865,83 @@ class user(BaseFilter):
...
+
         @property
         def user_ids(self):
-            with self._user_ids_lock:
-                return self._user_ids
+            with self.__lock:
+                return frozenset(self._user_ids) if self._user_ids is not None else None
 
         @user_ids.setter
         def user_ids(self, user_id):
```
tsnoam, line 915: Documentation needs to clearly specify that this will completely replace the set of user_ids.

Thinking of it, we might need to add a proper function to add to user_ids (and user_names) and another to remove:

`def add_user_id(self, user_id):`
`def remove_user_id(self, user_id):`

Rationale:
If a user "polls" the current user_ids and creates a new Set to set, by this time, a different thread might have updated the user_ids... (race condition).
So we need to provide granular operations which will be useful to users.

telegram/ext/filters.py
```diff
@@ -865,73 +865,83 @@ class user(BaseFilter):
...
 
         @property
         def usernames(self):
-            with self._usernames_lock:
-                return self._usernames
+            with self.__lock:
+                return frozenset(self._usernames) if self._usernames is not None else None
 
         @usernames.setter
         def usernames(self, username):
```
tsnoam, line 928: similar comments as for user_ids

telegram/ext/filters.py
```diff
@@ -865,73 +865,83 @@ class user(BaseFilter):
...
             """"""  # remove method from docs
-            if self.user_ids is not None:
-                return bool(message.from_user and message.from_user.id in self.user_ids)
-            else:
-                # self.usernames is not None
-                return bool(message.from_user and message.from_user.username
-                            and message.from_user.username in self.usernames)
+            if bool(message.from_user):
+                if self.user_ids:
+                    return bool(message.from_user.id in self.user_ids)
```
tsnoam, line 939: that's already a bool. you don't need to wrap it with another bool...

(and yes, i know it was like that before. doesn't meed we need to keep doing the same thing ;-) )

telegram/ext/filters.py
```diff
@@ -865,73 +865,83 @@ class user(BaseFilter):
...
+                if self.usernames:
+                    return bool(message.from_user.username
+                                and message.from_user.username in self.usernames)
+                return self.allow_empty
```
tsnoam, line 943: I guess that this is the place where we should be using `Defaults`. 
Does filters support `Defaults`? (if not, we should open an issue for that)

tsnoam:
> > * I just noticed, that `Filters.chat` currently does pretty much the same as `Filters.user`. I think it would be good to update it as well. I'll try and set up a base class for the two of them in order to reduce code duplication
> 
> good idea. go for it.

on a second thought, having a general solution might be over engineering. 
IMHO, you can just copy/paste the code.

telegram/ext/filters.py
Bibo-Joshi: Fair enough. Resolving.

telegram/ext/filters.py
```diff
@@ -865,73 +865,83 @@ class user(BaseFilter):
...
 
         Raises:
             ValueError: If chat_id and username are both present, or neither is.
 
         """
 
-        def __init__(self, user_id=None, username=None):
-            if (user_id is None) == (username is None):
-                raise ValueError('One and only one of user_id or username must be used')
+        def __init__(self, user_id=None, username=None, allow_empty=False):
```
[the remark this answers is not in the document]
Bibo-Joshi: Personally, I don't think this would be a good fit for defaults, copying from dev chat:

> Right now those are only used for arguments for a) bot methods, which b) appear in multiple different methods c) are well established

But maybe I change my mint for a next iteration

telegram/ext/filters.py
```diff
@@ -865,73 +865,83 @@ class user(BaseFilter):
...
-            self._usernames_lock = Lock()
+            self._user_ids = {}
+            self._usernames = {}
 
-            # Initialize in a way that will not fail the first setter calls
-            self._user_ids = user_id
-            self._usernames = username
-            # Actually initialize
             self.user_ids = user_id
             self.usernames = username
```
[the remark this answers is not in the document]
Bibo-Joshi: But … This *calls* the property …

telegram/ext/filters.py
```diff
@@ -865,73 +865,83 @@ class user(BaseFilter):
...
+            else:
+                self._user_ids = set(user_id)
+
+        def _set_usernames(self, username):
+            if username is None:
+                self._usernames = {}
+            elif isinstance(username, str):
+                self._usernames = {username.replace('@', '')}
+            else:
+                self._usernames = set([user.replace('@', '') for user in username])
```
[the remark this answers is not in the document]
Bibo-Joshi: I understand the documentation as "If the username has a leading @, the @ will be ignored". Dropping the `replace` would be breaking, too …

telegram/ext/filters.py
```diff
@@ -865,73 +865,83 @@ class user(BaseFilter):
...
-            if (user_id is None) == (username is None):
-                raise ValueError('One and only one of user_id or username must be used')
+        def __init__(self, user_id=None, username=None, allow_empty=False):
+            self.allow_empty = allow_empty
+            self.__lock = Lock()
 
-            self._user_ids_lock = Lock()
-            self._usernames_lock = Lock()
+            self._user_ids = {}
+            self._usernames = {}
```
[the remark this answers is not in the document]
Bibo-Joshi: That way we would have to do another check `if user_id and username`. By calling the setters, they do that for us.

telegram/ext/filters.py
Bibo-Joshi: Not really.  See comment on #1757

telegram/ext/filters.py
Bibo-Joshi: For some reason I thought without the extra `(` that would be invalid syntax … removing.

telegram/ext/filters.py
Bibo-Joshi: Sounds good. will do.

Bibo-Joshi:
> > > * I just noticed, that `Filters.chat` currently does pretty much the same as `Filters.user`. I think it would be good to update it as well. I'll try and set up a base class for the two of them in order to reduce code duplication
> > 
> > 
> > good idea. go for it.
> 
> on a second thought, having a general solution might be over engineering.
> IMHO, you can just copy/paste the code.

Okay. But then I'll wait until we're happy with `Filters.user`, so it's a one time jobe :D

telegram/ext/filters.py
```diff
@@ -850,29 +851,68 @@ class user(BaseFilter):
...
+
+            self._user_ids_lock = Lock()
+            self._usernames_lock = Lock()
+
+            # Initialize in a way that will not fail the first setter calls
+            self._user_ids = user_id
+            self._usernames = username
+            # Actually initialize
+            self.user_ids = user_id
+            self.usernames = username
```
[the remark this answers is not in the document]
tsnoam: Oh... I understand now.
You're using the setter from inside `__init__()`.
Personally, I wouldn't do it as it might confuse others (it confused me!), but it works. Ok, we can leave it as it is.

telegram/ext/filters.py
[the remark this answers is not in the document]
tsnoam: yeah.. commented above...
it's confusing, but it works. we can keep it as is.

telegram/ext/filters.py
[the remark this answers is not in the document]
tsnoam: ok. so lets make the documentation more clear about that.

telegram/ext/filters.py
```diff
@@ -892,38 +893,153 @@ class user(BaseFilter):
...
+
+                user_id = self._parse_user_id(user_id)
+                username = self._parse_username(username)
+
+                for uid in user_id:
+                    self._user_ids.add(uid)
+                for name in username:
+                    self._usernames.add(name)
+
+        def remove_users(self, user_id=None, username=None):
```
tsnoam, line 1007: any reason not to make it a dedicated method for remove_usernames and dedicated method for remove_user_ids?

telegram/ext/filters.py
```diff
@@ -892,38 +893,153 @@ class user(BaseFilter):
...
+        @property
+        def usernames(self):
+            with self.__lock:
+                return frozenset(self._usernames)
+
+        @usernames.setter
+        def usernames(self, username):
+            self._set_usernames(username)
+
+        def add_users(self, user_id=None, username=None):
```
tsnoam, line 981: any reason not to make it a dedicated method for add_usernames and dedicated method for remove_user_ids?

telegram/ext/filters.py
tsnoam: Left some comments on that code

telegram/ext/filters.py
Bibo-Joshi: Not really. Changed.

telegram/ext/filters.py
Bibo-Joshi: Not really. Changed.

telegram/ext/filters.py
```diff
@@ -978,57 +978,73 @@ def usernames(self):
...
+                if self._user_ids:
                     raise RuntimeError("Can't set username in conjunction with (already set) "
                                        "user_ids.")
 
-                user_id = self._parse_user_id(user_id)
                 username = self._parse_username(username)
-
-                for uid in user_id:
-                    self._user_ids.add(uid)
                 for name in username:
```
tsnoam, line 995: self._usernames |= username

telegram/ext/filters.py
```diff
@@ -978,57 +978,73 @@ def usernames(self):
...
             """
-            Adds a user(s) to the allowed users.
+            Adds (a) user(s) to the allowed users by name.
 
             Args:
-                user_id(:obj:`int` | List[:obj:`int`], optional): Which user ID(s) to allow
-                    through.
                 username(:obj:`str` | List[:obj:`str`], optional): Which username(s) to allow
-                    through. If username starts with '@' symbol, it will be ignored.
+                    through. Leading '@'s in usernames will be discarded.
```
tsnoam, line 987: Similar comment

telegram/ext/filters.py
```diff
@@ -978,57 +978,73 @@ def usernames(self):
...
-                user_id = self._parse_user_id(user_id)
                 username = self._parse_username(username)
-
-                for uid in user_id:
-                    self._user_ids.add(uid)
                 for name in username:
                     self._usernames.add(name)
 
-        def remove_users(self, user_id=None, username=None):
+        def add_user_ids(self, user_id):
```
tsnoam, line 998: Similar comments to add usernames

telegram/ext/filters.py
```diff
@@ -978,57 +978,73 @@ def usernames(self):
...
-                if username and self._user_ids:
+
+                user_id = self._parse_user_id(user_id)
+
+                for uid in user_id:
+                    self._user_ids.add(uid)
+
+        def remove_usernames(self, username):
+            """
+            Removes (a) user(s) from the allowed users by name.
```
tsnoam, line 1018: Remove one or more users from allowed usernames

telegram/ext/filters.py
```diff
@@ -978,57 +978,73 @@ def usernames(self):
...
+
+            Args:
+                username(:obj:`str` | List[:obj:`str`], optional): Which username(s) to disallow
+                    through. Leading '@'s in usernames will be discarded.
```
tsnoam, line 1022: Similar comment

telegram/ext/filters.py
```diff
@@ -978,57 +978,73 @@ def usernames(self):
...
+            """
+            with self.__lock:
+                if self._user_ids:
                     raise RuntimeError("Can't set username in conjunction with (already set) "
                                        "user_ids.")
 
-                user_id = self._parse_user_id(user_id)
                 username = self._parse_username(username)
+                for name in username:
```
tsnoam, line 1030: self._usernames -= username

telegram/ext/filters.py
```diff
@@ -978,57 +978,73 @@ def usernames(self):
...
+                    self._usernames.discard(name)
+
+        def remove_user_ids(self, user_id):
+            """
+            Removes (a) user(s) from the allowed users by id.
```
tsnoam, line 1035: Similar comments to remove usernames

telegram/ext/filters.py
```diff
@@ -892,38 +893,169 @@ class user(BaseFilter):
...
+        def _parse_username(username):
             if username is None:
-                self.usernames = username
-            elif isinstance(username, string_types):
-                self.usernames = [username.replace('@', '')]
-            else:
-                self.usernames = [user.replace('@', '') for user in username]
+                return set()
+            if isinstance(username, str):
+                return {username.replace('@', '')}
```
tsnoam, line 946: I've just realized that this will replace all @ from the username.
Maybe we had a (minor) bug until now...

What do you think about removing only a single leading @ ?

```python
if username.startswith('@'):
    username = username[1:]
```

Bibo-Joshi: "Allowed are a-z, 0-9 and underscores." Still, that might change in the future, so I changed it and also added it to the test cases.

tsnoam:
LGTM

Bibo-Joshi:
Fails unrelated. Merging.

tsnoam:
@Bibo-Joshi 💥

PR #1760, opened 2020-02-08
Make sure PP can read files that dont have bot_data

### LINKED ISSUE #1761: [BUG] v12.4 breaks PicklePersistence
### Steps to reproduce
1. Have a bot using PicklePersistence with singlefile=True

2. Upgrade to v12.4

3. restart bot

### Expected behaviour
pickled file is read correctly

### Actual behaviour
key error `bot_data` is thrown

### Current workaround:
Add an empty dict `bot_data` to the file manually. Quick and dirty script:
```
import pickle

filename = 'my_pickle_persistence_file'

with (open(filename, 'rb')) as file:
    data = pickle.load(file)

data['bot_data'] = {}

with open(filename, 'wb') as f:
    pickle.dump(data, f)
```

Will be closed by #1760

Bibo-Joshi:
This assures that a pickle file that doesn't contain bot_data is read correctly. Tests are added as well.
Not necessary for DictPersistence because it doesn't store anything.

closes #1761

### DIFF
--- a/telegram/ext/picklepersistence.py
+++ b/telegram/ext/picklepersistence.py
@@ -85,9 +85,7 @@ class PicklePersistence(BasePersistence):
                 self.user_data = defaultdict(dict, all['user_data'])
                 self.chat_data = defaultdict(dict, all['chat_data'])
                 # For backwards compatibility with files not containing bot data
-                self.bot_data = all.get('bot_data')
-                if self.bot_data is None:
-                    self.bot_data = {}
+                self.bot_data = all.get('bot_data', {})
                 self.conversations = all['conversations']
         except IOError:
             self.conversations = {}
--- a/telegram/ext/picklepersistence.py
+++ b/telegram/ext/picklepersistence.py
@@ -84,7 +84,10 @@ class PicklePersistence(BasePersistence):
                 all = pickle.load(f)
                 self.user_data = defaultdict(dict, all['user_data'])
                 self.chat_data = defaultdict(dict, all['chat_data'])
-                self.bot_data = all['bot_data']
+                # For backwards compatibility with files not containing bot data
+                self.bot_data = all.get('bot_data')
+                if self.bot_data is None:
+                    self.bot_data = {}
                 self.conversations = all['conversations']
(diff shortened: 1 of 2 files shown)


tests/test_persistence.py
```diff
@@ -335,6 +335,23 @@ def good_pickle_files(user_data, chat_data, bot_data, conversations):
         os.remove(name)
 
 
+@pytest.fixture(scope='function')
+def pickle_files_wo_bot_data(user_data, chat_data, conversations):
+    all = {'user_data': user_data, 'chat_data': chat_data, 'conversations': conversations}
```
tsnoam: Isn't `all` a built-in function?

tests/test_persistence.py
```diff
@@ -335,6 +335,23 @@ def good_pickle_files(user_data, chat_data, bot_data, conversations):
...
+    with open('pickletest_user_data', 'wb') as f:
```
tsnoam: Better use `tempfile` module here. You can never know when you'll have same tests running in parallel on the same machine.

telegram/ext/picklepersistence.py
```diff
@@ -84,7 +84,10 @@ def load_singlefile(self):
                 all = pickle.load(f)
                 self.user_data = defaultdict(dict, all['user_data'])
                 self.chat_data = defaultdict(dict, all['chat_data'])
-                self.bot_data = all['bot_data']
+                # For backwards compatibility with files not containing bot data
+                self.bot_data = all.get('bot_data')
```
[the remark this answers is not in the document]
Bibo-Joshi: It is. Didn't come up with the naming here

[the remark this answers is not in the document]
Bibo-Joshi: sure, that's better

tests/test_persistence.py
Bibo-Joshi: Don't know about tempfile. But the fixture is basically just copy-paste from the others. So I'd rather open an issue to change them all and leave this as is for now, if that's okay with you

PR #1764, opened 2020-02-08
correct_option_id validated with None

sharunkumar:
when trying to send a poll with correct option id 0 it was failing. Now None check is done so that even when 0 is passed it is assigned.


Poolitzer:
There we go, thanks

Issue #1373, opened 2019-03-26
Type-hinting for IDEs


TiiFuchs:
Hi there,

I like to use the Message.reply_* functions but it's kind of annoying to always look up the parameters manually. There is a lack of type-hinting here. And with python stub files (or type hinting in the python files directly this would not be a problem at all.

The Problem is, if I have a update instance and I call `update.message` it doesn't know that this is a Message instance. If I tell my IDE using `message = update.message # type: Message` it works so far. But when I then type `message.reply_text(` there is no type-hinting (or variable hinting for that matter) for the named parameters. This is annoying.

I started to create python stub files like this:
![message-type-hint](https://user-images.githubusercontent.com/1958744/54982287-ca628680-4faa-11e9-9a7e-4e4f1f7a2cb4.png)

But it would be awesome if type-hinting a la [PEP-484](https://www.python.org/dev/peps/pep-0484/) would be integrated in the project itself. 

I know that python 2 compatibility (at least until EOL) is still a thing, but type-hinting with comments like described above (and at some place in PEP-484) is always possible for both python versions.
![update-type-hint](https://user-images.githubusercontent.com/1958744/54982440-26c5a600-4fab-11e9-801d-ecea47505dca.png)

tiirex9:
As far as I understand it, python stub files can have python 3 syntax type-hinting since they don't get executed but parsed by the IDE that does the type-hinting. 
If that is correct a type-hinting like this would work too:
```
class Update(TelegramObject):
    message: Message
```

Eldinnie:
Yes, as far as I can see stub files can be made for the library (quite a lot of work) and a PR towards this is very welcome.

Placing comments in the code for type hinting would be a half-baked (no signature types) and intermediate solution. I definitely would prefer stub files.

@tiirex9 would you like to do this?

tiirex9:
I will have a look into this, yes. :)

JosXa:
Those stub files can be put in Typeshed which is where the major IDEs pull their stubs from.

tiirex9:
Is that general consent that this should be added to typeshed instead of the library itself?

I think it’s easier to add it to the library and this makes it also easier to handle different versions and different method signatures.

JosXa:
It can be both indeed

tiirex9:
So it's a lot of work if I would create all stub files manually and it would get very difficult to keep in track with the code. But since all the information is in the docstrings, we **could** generate it.
[...]

PR #1767, opened 2020-02-08
Type hinting added for properties of Update object

igorsereda:


Bibo-Joshi:
Hi. Thanks for the PR :)

However, I fear that if we want to add type hinting, it should be done for everything *at once*. I.e. maybe not in a single PR, but at least we should work on a dedicated branch and then merge to master. Therefore, I'll be closing this for now.

If you feel like starting on adding type hinting for the entire library (you are very welcome!), I suggest to comment on #1373 to make sure we're in sync on how we approach this.

PR #1787, opened 2020-02-20
updating pre-commit config file

Poolitzer:
as discussed in the chat

I changed sha to rev (which is the way to go) and also changed the deprecated url from flake8 to the official one. I picked 3.7.1 because thats the version we took from the old collection


tsnoam:
@Poolitzer Have you tested it on a clean clone?
If you have, we can merge.

Poolitzer:
I did a reinstall... Is that enough? ;p

Issue #1766, opened 2020-02-08
Don't use builtin function names as variables


Bibo-Joshi:
There is e.g. `all` in `PicklePersistence` and I think also `type` somewhere. Also `id`

tsnoam:
@Bibo-Joshi &quot;type&quot; might be forced upon us because of it is derived from telegram API.

Poolitzer:
@tsnoam Actually, I think most of our usage of built-in function names will be where we take the naming from the API, and I dont think its worth changing that. Nevertheless, I can have a look later and see if I find issues of that in other places

Issue #1765, opened 2020-02-08
Use tempfile module in tests


Bibo-Joshi:
E.g. in test_persistence, but also check the other tests. Better than actually writing to disk.

Bibo-Joshi:
Mh, with `tempfile` module there is no way to specify the exact file name of the created temporary files. But `PicklePersistence` will look for specific file names and I wouldn't fancy monkey patching the load methods of `PicklePersistence` in the tests …

tsnoam:
@Bibo-Joshi What about using tempfile to create a temp directory?

PR #1789, opened 2020-02-21
Roles

### LINKED ISSUE #1151: Feature suggestion: Authentication
I wrote an authentication function to my private bots, to control who can use the bot.
I thought to add it to this project, but there are 2 problems:

1. The bot should save a file, DB, to remember which user are authenticated.
2. The manager- the first user, should insert his id directly to code or DB.

Bibo-Joshi:
This PR adds the functionality of restricting access to handlers to

* specified users (by id)
* specified chats (by id)
* chat admins
* chat creators

Builds upon #1757 (see below)
Closes #1151, #1562 

## Example usage:

```python

updater = Updater('TOKEN')
dp = updater.dispatcher
roles = updater.dispatcher.roles

roles.add_admin('authors_user_id')
roles.add_role(name='my_role_1')
my_role_1 = roles['my_role_1']
roles.add_role(name='my_role_2', parent_roles=roles['my_role_1'])
my_role_2 = roles['my_role_2']

def add_to_my_role_2(update, context):
    user = update.effective_user
    context.roles['my_role_2'].add_member(user.id)
    
def add_to_my_role_1(update, context):
    user_id = update.message.text
    context.roles['my_role_1'].add_member(user_id)

....

# Anyone can add themself to my_role_2
dp.add_handler(TypeHandler(Update, add_to_my_role_2))

# Only the admin can add users to my_role_1
dp.add_handler(MessageHandler(Filters.message, add_to_my_role_1, roles=roles.ADMINS))

# This will be accessable by my_role_2, my_role_1 and the admin
dp.add_handler(SomeHandler(..., roles=my_role_2))

# This will be accessable by my_role_1 and the admin
dp.add_handler(SomeHandler(..., roles=my_role_1))

# This will be accessable by anyone except my_role_1 and my_role_2
dp.add_handler(SomeHandler(..., roles=~my_role_1))

# This will be accessable by users, who are in my_role_2 bot not in my_role_1
dp.add_handler(SomeHandler(..., roles=~my_role_1 & my_role_2))

# This will be accessable only by the admins of the group the update was sent in
dp.add_handler(SomeHandler(..., roles=roles.CHAT_ADMINS))

# This will be accessable only by the creator of the group the update was sent in
dp.add_handler(SomeHandler(..., roles=roles.CHAT_CREATOR))

# You can compare the roles regarding hierarchy:
roles.ADMINS >=  roles['my_role_1'] #True
roles.ADMINS >=  roles['my_role_2'] #True
roles['my_role_1'] < roles['my_role_2'] #False
roles.ADMINS >= Role(...) #False, since neither of those is a parent of the other
```

#### Role

* Inherits from `Filters.user` (This is, where #1757 comes in. Can be changed to inherinting from `BaseFilter` if the reviewer want that)
* Can be combined by bitwise operations (This is why I'm Inheriting from `Filters`)
* Can be compared w.r.t. hierarchy. E.g. in above example: `my_role_1 > my_role_2`
* Dynamically add/kick members

#### Roles

A container for multiple roles

* Inherits from `dict` to allow accessing the roles by `roles['role_name']`
* `Roles.ADMINS` is parent to every role
* `Roles.CHAT_ADMINS`, `Roles.CHAT_CREATOR` as shortcuts for restricting acces in group chats
* Dynamically add/remove admins and roles

### Integration with handlers

The base class `Handler` now has a `__new__` method that makes sure, that `check_update` is called only, if the roles checked out. Alternatives approaches would be

* Make `Handler.check_update` check the roles directly and call `super().check_updater` in all the specific handlers. This means that custom handlers would have to change their `check_update` implementation. Then again, they may have to be adjusted to accept the `roles` argument anyway
* Add a metho `Handler.check_update_with_roles` that calles `Handler.check_update` after checking the roles. In the current code base, replace all calls to `check_update` with `check_update_with_role`. This would leave current implementation of `check_update` untouched.

### Integration with dispatcher

* `Dispatcher.roles` is the `Roles` instance managing the roles
* `Dispatcher.roles` is available as `context.roles` to allow editing roles in callbacks
* `store_roles` is added to persistence to allow keeping track of roles over restarts. I know that adding stuff to persistence should be done with care. But if we want dynamic access control (and I do), we need persistence here.

## ToDo

- [x] Add roles to new `PollAnswerHandler`. As `Poll` updates have neither user nor chat, we don't need roles with `PollHandler`
- [x] Cache chat admins and creators (timeout is customizable via `roles.CHAT_ADMINS.timeout = ...`)
- [x] Handle private chats correctly with `roles.CHAT_ADMINS/CHAT_CREATOR` (this is as easy if checking `chat.id == user.id`)
- [ ] When merged, we should create a wiki page. Important points to mention there:

  * Hard coding `dp.roles.add_admin(id)` will *always* add the user to admins on start up with persistence, even if they are removed by dynamic access control. Thus, only the author of the bot should be hard coded as admin.
  * Admins should *only* be in `roles.ADMINS`, else they might be excluded by `roles=~roles['my_role']`, if they are in `my_role`
  * Hard coding `dp.roles.add_role(name='my_group')` will lead to `ValueError` on restart with persistence because after restoring the data, the role `my_group` will already be set. Use `'my_group' in dp.roles` to check for that before adding the group.

- [ ] Depending on how long it will take to merge this:
    - [x] Don't use `all` in persistence (see #1766)
    - [x] Use `tempfile` module in persistence tests (see #1765)
    - [ ] Add type hinting (see #1373)

### DIFF
--- a/docs/source/telegram.ext.chatadminsrole.rst
+++ b/docs/source/telegram.ext.chatadminsrole.rst
@@ -1,0 +1,6 @@
+telegram.ext.ChatAdminsRole
+===========================
+
+.. autoclass:: telegram.ext.ChatAdminsRole
+    :members:
+    :show-inheritance:
--- a/docs/source/telegram.ext.chatcreatorrole.rst
+++ b/docs/source/telegram.ext.chatcreatorrole.rst
@@ -1,0 +1,6 @@
+telegram.ext.ChatCreatorRole
+============================
+
+.. autoclass:: telegram.ext.ChatCreatorRole
+    :members:
+    :show-inheritance:
--- a/docs/source/telegram.ext.role.rst
+++ b/docs/source/telegram.ext.role.rst
@@ -1,0 +1,6 @@
+telegram.ext.Role
+=================
+
+.. autoclass:: telegram.ext.Role
+    :members:
+    :show-inheritance:
--- a/docs/source/telegram.ext.roles.rst
+++ b/docs/source/telegram.ext.roles.rst
@@ -1,0 +1,6 @@
+telegram.ext.Roles
+==================
+
+.. autoclass:: telegram.ext.Roles
+    :members:
+    :show-inheritance:
--- a/docs/source/telegram.ext.rst
+++ b/docs/source/telegram.ext.rst
@@ -13,6 +13,8 @@ telegram.ext package
     telegram.ext.delayqueue
     telegram.ext.callbackcontext
     telegram.ext.defaults
+    telegram.ext.role
+    telegram.ext.roles
 
 Handlers
 --------
--- a/docs/source/telegram.ext.rst
+++ b/docs/source/telegram.ext.rst
@@ -46,3 +46,13 @@ Persistence
     telegram.ext.basepersistence
     telegram.ext.picklepersistence
     telegram.ext.dictpersistence
+
+Authentication
+--------------
+
+.. toctree::
+
+    telegram.ext.role
+    telegram.ext.roles
+    telegram.ext.chatadminsrole
+    telegram.ext.chatcreatorrole
--- a/telegram/ext/__init__.py
+++ b/telegram/ext/__init__.py
@@ -44,7 +44,7 @@ from .messagequeue import DelayQueue
 from .pollanswerhandler import PollAnswerHandler
 from .pollhandler import PollHandler
 from .defaults import Defaults
-from .roles import Role, Roles
+from .roles import Role, Roles, ChatAdminsRole, ChatCreatorRole
 
 __all__ = ('Dispatcher', 'JobQueue', 'Job', 'Updater', 'CallbackQueryHandler',
            'ChosenInlineResultHandler', 'CommandHandler', 'Handler', 'InlineQueryHandler',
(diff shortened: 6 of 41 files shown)


Bibo-Joshi:
Codacy mainly complains about missing docs for methods which are intended for internal use only. IMHO we can ignore that.

Bibo-Joshi:
As suggested by @tsnoam I had a look at third party ACL and RBAC implementations. Though there are some libraries that seem well maintained, they mostly seem

* tailored for website applications
* not to allow for easy dynamic editing of roles
* have rather involved persistence integrations

My impression may well be based on lack of expertise, but to me it seems cleaner to have a custom implementation tailored to our needs than to rely on those third party implementations.

tsnoam:
just a thought:
would it be possible to implement roles using decorators?
user who wants a handler with a certain role, can just decorate it with something to do that magic.

classic for a contrib repo, btw ;-)

Bibo-Joshi:
> classic for a contrib repo, btw ;-)

decorators? yeah, sounds about right

> just a thought:
> would it be possible to implement roles using decorators?
> user who wants a handler with a certain role, can just decorate it with something to do that magic.

one would need to decorate every callback … We just made a point not to use decorators on callbacks with deprecating @run_async :D
but one could think of a Handler-Wrapper:

```
class RolesHandlerWrapper:
    def __init__(base_class, roles, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.roles = roles

    def check_update(update):
       if self.roles(update):
            return super().check_update(update)
       return False
```

tsnoam:
@Bibo-Joshi 
I don't like decorators. 
But contrib? That can be a jungle AFAIC.

I like your idea for a HandlerWrapper. That can be contrib as well.

Bottom line: contrib.

Bibo-Joshi:
Closing to be revisited after #1912 is done. v13 changed so much, I can probably start from scratch anyway :D

PR #1808, opened 2020-03-04
Use temporary directories in test_persistence

### LINKED ISSUE #1765: Use tempfile module in tests

Bibo-Joshi:
Closes #1765 
Uses `pytest` s `tmp_path`


Bibo-Joshi:
Fails on ubuntu. Need to investigate.

Bibo-Joshi:
Codecov complains about JobQueue and utils.requests. dunno, why, I'll just ignore it …

PR #1811, opened 2020-03-05
Append parameters passed as kwargs to data

jondef:
This allows the programmer to pass additional parameters to control the bot. For instance, you could do the following: bot.sendPoll(chat_id=chat_id, question=pollQuestion, options=pollOptions, allows_multiple_answers=false, is_anonymous=false).

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -169,6 +169,9 @@ class Bot(TelegramObject):
             else:
                 data['media'].parse_mode = None
 
+        # Append parameters passed as kwargs to data
+        data.update(kwargs)
+
         result = self._request.post(url, data, timeout=timeout)
 
         if result is True:


Poolitzer:
You mean, parameters like in this tree? https://github.com/python-telegram-bot/python-telegram-bot/tree/4.6-update

jondef:
No, I mean [telegram API parameters](https://core.telegram.org/bots/api#sendpoll). When you call `bot.sendPoll(..., **kwargs)`, the kwargs are passed to `Bot._message(...,**kwargs)` which does nothing with them. My suggestion is to append them to the data dict to pass them to the Telegram API.

The reason for this is because I need to make polls with multiple answers that are not anonymous. I'm not sure if this API currently supports that. I currently create the link manually, which allows me to create the polls I want.

Poolitzer:
Yes, we are on 5.4 on pip :)

PR #1817, opened 2020-03-08
Deprecation

Bibo-Joshi:
In v13 we want to deprecate some stuff:

* non-context based callbacks (i.e. all the `pass_*` arguments)
* `allow_edited` argument of `CommandHandler`
* non-filter arguments of `MessageHandler`
* `RegexHandler`
* `bot` argument of `JobQueue`

This removes the corresponding code and docs


Bibo-Joshi:
Looks like I overshot a bit here. Actually removing legacy-style callbacks is only due for v14. I'll close this one and open another smaller PR for v13 in time.

PR #1820, opened 2020-03-08
Add stale bot

Bibo-Joshi:
The bot will mark question issues as stale after 5 days and close them after 2 more, if there is still no reaction (so 7 days in total). My impression with question issues is that they are either resolved quickly or become stale, that's why I chose those values.

If we feel like we need to add the feature for more labels, that's no problem.


Bibo-Joshi:
CI failures obviously unrelated

PR #1824, opened 2020-03-10
Add Py3.8 to docs

Bibo-Joshi:
The README already says python 3.5+, so we're good there.
Added py38 to setup.py
If I understand correctly, the pypi badge will state py3.8 as soon as we release with the new setup.py

@Poolitzer, do we need to do something with the pytest yml?

```yml
matrix:
  python-version: [3.5, 3.6, 3.7, 3.8]
  os: [ubuntu-latest, windows-latest]
  include:
    - os: ubuntu-latest
      python-version: 3.7
      test-build: True
    - os: windows-latest
      python-version: 3.7
      test-build: True
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,5 +64,5 @@ with codecs.open('README.rst', 'r', 'utf-8') as fd:
               'Programming Language :: Python :: 3.5',
               'Programming Language :: Python :: 3.6',
               'Programming Language :: Python :: 3.7',
-              'Programming Language :: Python :: 3.8'
+              'Programming Language :: Python :: 3.8',
           ],)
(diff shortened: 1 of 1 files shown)


Poolitzer:
it already has 3.8 what more do you want? ;P

Bibo-Joshi:
No, I mean the `include: ...` part

setup.py
```diff
@@ -63,5 +63,6 @@ def requirements():
               'Programming Language :: Python :: 3',
               'Programming Language :: Python :: 3.5',
               'Programming Language :: Python :: 3.6',
-              'Programming Language :: Python :: 3.7'
+              'Programming Language :: Python :: 3.7',
+              'Programming Language :: Python :: 3.8'
```
tsnoam: one more comma at eol please

Poolitzer:
Okay, no, the include part is for running the build test, which I think just tests if it build correctly, which is enough if we do it once per platform, doesnt really matter which version its on.

https://github.com/python-telegram-bot/python-telegram-bot/blob/43bfebb150920f852c0f16e33fd8a56f9ac2a813/tests/test_meta.py#L37-L39 this is the test itself, if you want to check it out

Issue #1759, opened 2020-02-07
[BUG] make test bots un/pin messages in different chats to make tests more stable


Bibo-Joshi:

artemrys:
Came here from a conversation in https://github.com/python-telegram-bot/python-telegram-bot/pull/1826, looks like that the issue is bigger than I was thinking about :(

So the main issue is the state of the bot in the tests. Example (the same bot is chosen): I run `test_set_chat_photo` test locally, someone else run `test_delete_chat_photo` test locally and if I run `test_delete_chat_photo` locally - it will fail.

The doable option that I can see so far - use different bots for the CI tests but let developers still use the existing ones. It will reduce the flakiness in the tests by not damaging the state of the bots from local tests. And reordering of some tests can help.

Example of the tests for the master branch (https://github.com/python-telegram-bot/python-telegram-bot/commit/28ded6718ecaf2541b2dd25a14bad0a7c3068c7b commit):
```python
    @flaky(3, 1)
    @pytest.mark.timeout(10)
    def test_delete_chat_photo(self, bot, channel_id):
        assert bot.delete_chat_photo(channel_id)

    @flaky(3, 1)
    @pytest.mark.timeout(10)
    def test_set_chat_photo(self, bot, channel_id):
        with open('tests/data/telegram_test_channel.jpg', 'rb') as f:
            assert bot.set_chat_photo(channel_id, f)
```
py.test executes tests in the order they are given in the file unless something else is specified.
In this case - it is an issue. The first test tries to delete a chat photo and the second one actually set its up.

PR #1826, opened 2020-03-10
Pass context of last valid update to timeout handlers

### LINKED ISSUE #1823: [BUG] context.user_data saved in a conversation is lost upon timeout.
### Steps to reproduce
1. In the entry function of a conversation save some data in context.user_data

2. Set the conversation_timeout of the conversation to some value other than 0. I used it as 10

3. In the "ConversationHandler.TIMEOUT" state define a callback handler and in that try reading the user_data from context.

### Expected behaviour
Should be able to fetch the saved data using context.user_data.get method.

### Actual behaviour
I get the following error: AttributeError: 'NoneType' object has no attribute 'get'. In all other states I am able to read the user_data except in the TIMEOUT state.

### Configuration
**Operating System:**
Windows 10

Version of Python 3.7.5
Version of python-telegram-bot 12.4.2

Bibo-Joshi:
Closes #1823

Also added a test case that makes sure that the last update & context are the same as the ones in the timeout handler.


Bibo-Joshi:
`test_set_game_score` scores again …

artemrys:
@Bibo-Joshi regarding your comment about `test_set_game_score` test I have prepared a draft PR with a possible workaround for this issue. PTAL https://github.com/artemrys/python-telegram-bot/commit/3734b01898b724f41f20f27281b4b643160b0d19 It works for me locally.
Does it worth to create a separate bug for this issue and start discussing it there?

Bibo-Joshi:
> @Bibo-Joshi regarding your comment about test_set_game_score test I have prepared a draft PR with a possible workaround for this issue. PTAL artemrys@3734b01 It works for me locally.
Does it worth to create a separate bug for this issue and start discussing it there?

The failing tests are a more general issue, see also #1759. So, yes, please go ahead and pr, so we can see, if the reordering also works in the CI. Please make sure to comment, *why* the reorder should help.

PR #1827, opened 2020-03-10
Improve Persistence

### LINKED ISSUE #1671: [FEATURE] Persistence inside of Jobs
The job runner are amazing. But they lack some documentation in terms of persistence. For me it turned out to do something along the lines of this:

```python
persistence = PicklePersistence(DB_FILE)
updater: Updater = Updater('token', use_context=True, persistence=persistence)

def update_updater_data():
    updater.dispatcher.user_data = persistence.user_data
    updater.dispatcher.update_persistence()
    persistence.flush()

def my_job():
    persistence.user_data......do stuff and save it
    update_updater_data()
```

which works, but took me a lot of trial and error.
Is there a better approach to this? (maybe I'm missing something)
It might be a good idea to add a little section either in the persistence or jobqueue wiki site :)

Bibo-Joshi:
Originally intended to close #1671 , this now does 3 things:

1. Unify the two methods of dispatcher used to update the persistence. This makes sure that
    1. If persistence is updated after handling an update, only the corresponding user/chat_data is updated
    2. Errors that appear when updating persistence on shutting down the bot will be handled correctly
2. Fixing a minor bug in Pickle/DictPersistence: When calling `update_*_data` before calling `get_*_data`, we need to check that the internal attributes are not `None`
3. Actually close #1671 by updating the persistence after running a job

### DIFF
--- a/telegram/ext/dispatcher.py
+++ b/telegram/ext/dispatcher.py
@@ -323,53 +323,6 @@ class Dispatcher(object):
 
         """
 
-        def persist_update(update):
-            """Persist a single update.
-
-            Args:
-            update (:class:`telegram.Update`):
-                The update to process.
-
-            """
-            if self.persistence and isinstance(update, Update):
-                if self.persistence.store_bot_data:
-                    try:
-                        self.persistence.update_bot_data(self.bot_data)
-                    except Exception as e:
-                        try:
-                            self.dispatch_error(update, e)
-                        except Exception:
-                            message = 'Saving bot data raised an error and an ' \
-                                      'uncaught error was raised while handling ' \
-                                      'the error with an error_handler'
-                            self.logger.exception(message)
-                if self.persistence.store_chat_data and update.effective_chat:
-                    chat_id = update.effective_chat.id
-                    try:
-                        self.persistence.update_chat_data(chat_id,
-                                                          self.chat_data[chat_id])
-                    except Exception as e:
-                        try:
-                            self.dispatch_error(update, e)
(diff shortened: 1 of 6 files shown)


Bibo-Joshi:
CI:

* Py3.7 fails on `test_game_score`. Unrelated
* Codecov mainly complains that the parts

```python
'Saving user data raised an error and an ' \
'uncaught error was raised while handling ' \
'the error with an error_handler'
```
are not tested. But as far as I can see, they weren't tested before …

telegram/ext/dispatcher.py
```diff
@@ -471,18 +424,62 @@ def remove_handler(self, handler, group=DEFAULT_GROUP):
...
+                    try:
+                        self.persistence.update_chat_data(chat_id, self.chat_data[chat_id])
+                    except Exception as e:
+                        try:
+                            self.dispatch_error(update, e)
+                        except Exception:
+                            message = 'Saving chat data raised an error and an ' \
+                                      'uncaught error was raised while handling ' \
+                                      'the error with an error_handler'
+                            self.logger.exception(message)
```
Poolitzer, line 470: are we loosing the dispatcher exception here? Or does logging take care of that?

Bibo-Joshi: No, I think we're loosing it.

Bibo-Joshi: Tested it - the error thrown in `disptach_error` and the traceback are in the log.

PR #1833, opened 2020-03-18
Minor readability improvement and test for valid Bot token

artemrys:
Improvements:

 * use a wildcard instead of naming an unused variable
 * add test for valid Bot token

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -142,7 +142,7 @@ class Bot(TelegramObject):
         if any(x.isspace() for x in token):
             raise InvalidToken()
 
-        left, sep, _right = token.partition(':')
+        left, sep, _ = token.partition(':')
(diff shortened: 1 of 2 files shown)


artemrys:
Py3.5 fails on `test_delete_chat_photo`. Unrelated to the change.

artemrys:
As far as I researched the code - it is connected with flakiness.
Decorator `pytest.flaky` is trying to solve this issue.
But something it is not enough retries to test pass.

Bibo-Joshi:
Hey. After some internal discussion, we decided against this PR. I.e. valid bot tokens are already tested implicity by actually running the test suite and the wildcard thingy, while valid, maybe doesn't need it's own PR. Thank you for taking the time, nonetheless!

artemrys:
No problems :)

PR #1841, opened 2020-03-21
Non anonymous polls support

zhelnio:
added new arg to  **send_poll**(..., **is_anonymous=True**, ...)
to support non anonymous polls


Poolitzer:
Hi, this is already implemented in our [4.6 branch](https://github.com/python-telegram-bot/python-telegram-bot/tree/4.6-update), since it is a 4.6 feature.

PR #1844, opened 2020-03-23
 Arbitrary callback data

### LINKED ISSUE #709: make callbackquery safe
[botogram ](https://botogram.pietroalbini.org/)implemented a safe way to handle callback queries as you can read in the documentation (https://botogram.pietroalbini.org/docs/dev/buttons/#buttons-security)

the problem is that the callback_data is not sent by telegram, but the clients. So if a bot is admin of a group and it lets ban spamming users with buttons and the callback_data containts parameters (for example ban_user:123456789) a maliciuos client could change the parameter (for example ban_user:987654321) despite the text of the callback could be the same

Bibo-Joshi:
**This description is quite outdated, please see the discussion below**

# ToDo

- [ ] When merged, whe should create a wiki page
- [ ] Open an issue for the v14 milestone as reminder to make `Persistence.store_callback_data` default to `True` and make `BP.get/update_callback_data` abstract methods. Currently they're not b/c backwards compatibility …
- [x] Add/Update `.. versionadded:: directives`
- [ ] Add a note to #2347 to remove `Defaults` from `tg.Bot` completey and move them to `tg.ext.Bot`. We may have to think about how to handle signatures with `DefaultVaule`, then, i.e. if `_insert_defaults` should stay in `tg.Bot` or should be moved to `tg.ext.Bot`.
- [x] should probably be merged after #2262, so that we can update the added persistence logic according to https://github.com/python-telegram-bot/python-telegram-bot/pull/2262#discussion_r632762276, see also https://github.com/python-telegram-bot/python-telegram-bot/pull/2262/commits/39a9f833104556047506541bf5bcb4873a1443cf
- [x] check that the "edit on GH" thingies redirect to the correct file

This PR adds the functionality to

* pass arbitrary objects as `callback_data` to `InlineKeyboardButtons`
* in accordance to that, the `pattern` argument for `CallbackQueryHandler` may now be a callable accepting the `callback_data` as argument.
* validate that the `callback_data` received in an update is the same as the sent one

This is opt-in, so we're not forcing anyone to use it.
Closes #709

# How does it work

## Passing arbitrary objects as `callback_data`

This  is achieved by replacing them by their `id()` and keeping a dictionary that maps the ids to the objects. I decided to use `id()` over `uuid` or similar because

* Python guarantees that the `id()` is unique, so that's covered already and
* immutables won't be stored multiple times. Since most callback data will probably still consist of strings, this seemed clever to me.

Of course, this ties in with persistence. As for the roles PR #1789: Adding stuff to persistence should be reserverd for major features. IMHO, this is one of them.

## Signing the data

The callback data sent by the bot will read `signature data`. `signature` is generated using `hmac` from the standard library, where the key consists of the bots token and username and the message consits of the `data` and the `chat_id`, the message was sent to.

Upon receiving an update, `CallbackQuery.de_json` will check, that the data is valid by checking the signature. Malicious updates will be discarded and logged.

Generating the signature and validation done by methods added to the `helpers` module.

Users can decide, not to validate the data. This can be useful, when the token has been revoked and they want the old buttons to still work.

# Wiki Page: Callback Data

## Arbitrary objects as `callback_data`

Telegrams Bot API only accepts strings with length up to 64 bytes as `callback_data` for `InlineKeyboardButtons`, which sometimes is quite a limitation.

With PTB you are able, to pass *any* object as `callback_data`. This is achieved by storing the object and passing a unique id to Telegram. When a `CallbackQuery` is recieved, the id in the `callback_data` is replaced with the stored object. To use this feature, set the parameter `arbitrary_callback_data` for your `Updater` or `Bot` instance to `True`.

This means two thigs for you:

1. If you don't use persistence, buttons won't work after restarting your bot, as the stored updates are lost. More precicely, the `callback_data` you will recieve is `None`. If you don't need persistence otherwise, you can set `store_callback_data` to `True` and all the others to `False`.
2. When using the `CallbackQueryHandler`, the `pattern` argument can now be either

    * a regex expression, which will be used, if the `callback_data` is in fact a string
    * a callable accepting the `callback_data` as only argument. You can perform any kinds of tests on the `callback_data` and return `True` ore `False` accordingsly

## Security of InlineKeyboardButtons

Callback updates are not sent by Telegram, but by the client. This means that they can be manipulated by a user. (While Telegram inofficially does try to prevent this, they don't want Bot devs to rely on them doing the security checks).

Most of the time, this is not really a problem, since `callback_data` often just is `Yes`, `No`, etc. However, if the callback data is something like `delete message_id 123`, the malicious user could delete any message sent by the bot.

To increase security of callbacks, PTB signs outgoing callback data. The signature not only makes sure that the `callback_data` is valid, but also that the update comes from the chat, the message with the `InlineKeyboardButton` was sent in. Incoming updates with an invalid signature will be rejected.

The signature is based on your bots token and username. This means that old buttons will no longer be accepted, if you revoke your token. If you need to revoke your token, e.g. during developement, and want the buttons to still work, you can pass

```python
validate_callback_data = False
```

to either your `Updater` or `Bot` instance. Make sure to switch it to `True` after some time.


telegram/ext/callbackqueryhandler.py
```diff
@@ -76,9 +76,13 @@ class CallbackQueryHandler(Handler):
             :class:`telegram.ext.JobQueue` instance created by the :class:`telegram.ext.Updater`
             which can be used to schedule new jobs. Default is ``False``.
             DEPRECATED: Please switch to context based callbacks.
-        pattern (:obj:`str` | `Pattern`, optional): Regex pattern. If not ``None``, ``re.match``
-            is used on :attr:`telegram.CallbackQuery.data` to determine if an update should be
-            handled by this handler.
+        pattern (:obj:`str` | `Pattern` | :obj:`callable`, optional): Regex pattern. If not
```
Bibo-Joshi, line 79: Actually, allowing a type here could also be good, making `CQH` act similar to `TypeHandler` when we have custom types as callback data.

telegram/utils/helpers.py
```diff
@@ -438,3 +443,79 @@ def __bool__(self):
...
+
+def sign_callback_data(chat_id, callback_data, bot):
+    """
+    Prepends a signature based on :meth:`telegram.utils.helpers.get_callback_data_signature`
+    to the callback data.
+
+    Args:
+        chat_id (:obj:`str` | :obj:`int`): The chat the :class:`telegram.InlineKeyboardButton` is
+            sent to.
+        callback_data (:obj:`str`): The callback data.
```
GauthamramRavichandran, line 476: `callback_data` here is `id` of the data stored in dict, not the actual `callback_data`. Could we change the name of the argument, I was going through this and got confused why are we returning the actual data

Bibo-Joshi: Mh, theoretically you could pass string-valued `callback_data`, I guess reformulating to

```
callback_data (:obj:`str`): The callback data/the id of the callback data object.
```
or sim. is appropriate.

Bibo-Joshi:
IISC this doesn't handle inline keyboards attached to inline result messages yet. If we revisit this PR, that's one issue to think about

Bibo-Joshi:
We just discussed this and will be going forward with it. A few notes:

* Switch to UUID instead of `id()`
* built the data storage as LRU cache limiting memory usage
• store the timestamp along with the timestamp so if the user passes `maxsize=None` they can take care of deleting based on time stamps on their own.
• Still provide `CallbackQuery/Context/Bot.drop_callback_data` for fine grained control

sketch for a LRU data class:

```python
class LRUDict:
    data = {}
    maxsize = 1000

def put(self, key, object):
    if len(self.data) == maxsize:
        self.data.pop(next(iter(self.data)))  # dicts are only guaranteed to be ordered since py3.7, so maybe use OrderedDict instead …
    self.data[key] = object
```

Bibo-Joshi:
Merging master was surprisingly easy :D  
Adapted the LRU idea. Namings can ofc be changed.
Tests run fine locally, let's see how they perform in CI.

Bibo-Joshi:
After internal discussion, using the uuid should be enough, so the latest commit removes the signing procedure.
However I became aware of two flaws in the current setup:

* the data in the keyboard attached to the incoming `CallbackQuery` is currently not replaced
* pressing a button does *not* automatically mean that the keyboard will be deleted. The bot might just call `answer_callback_query(text='hint')` to display some info, without changing the message or the keyboard. So we musn't delete the data on incoming `CallbackQueries`. This means that
  1. `CallbackDataCache` needs to be properly converted to a LRU-cache by updating the order of the items in the doubly linked queue
  2. We should provide convenience methods to delete the objects corresponding to the incoming query or the complete attached keyboard. Probably samething like `CallbackQuery.{clear_data, clear_keyboard_data}`, `InlineKeyboardMarkup.clear_data`, `InlineKeyoardButton.clear_data()`  and `Message.clear_keyboard_data`

Bibo-Joshi:
The latest commits introduce `tg.ext.Bot` as discussed offline to separate pure API from tg.ext further. Also introduces `tg.ext.utils`, which is a bit of duplication from #2324, but I guess both of those should be released in the same release anyway … Also tried to already include CSI-style comments (see #2333), which will hopefull help in a review to a) follow my logic and b) find logic-flaws.

Things that I'd like a review to consider are:

* insertion-logic: Are there updates/return values where we need to insert the actual data, that I missed?
* User interface: Is the structure usable enough or would you expect e.g. methods to be available elsewhere?
* placement of `CallbackDataCache/InvalidCallbackData`: I put them into `tg.ext.utils`, but one could argue that at least `InvalidCallbackData` should be available directly throguh `tg.ext`. What do you think?

Below I attach a proposed wiki page. Maybe read this before reviewing, hopefully it makes it a bit clearer, what's going on.

# Wiki Page: Callback Data

## Arbitrary objects as `callback_data`

The Telegrams Bot API only accepts strings with length up to 64 bytes as `callback_data` for `InlineKeyboardButtons`, which sometimes is quite a limitation.

With PTB you are able, to pass *any* object as `callback_data`. This is achieved by storing the object in a cache and passing a unique identifier for that object to Telegram. When a `CallbackQuery` is recieved, the id in the `callback_data` is replaced with the stored object. To use this feature, set the parameter `arbitrary_callback_data` for your `Updater` or `ext.Bot` instance to `True`. The cache that holds the stored data has limited size (more details on memory usage below). If the cache is full and objects from a new `lnlineKeyboardMarkup` need to be stored, it will discard the data for the least recently used keyboard.

This means two thigs for you:

1. If you don't use persistence, buttons won't work after restarting your bot, as the stored updates are lost. More precicely, the `callback_data` you will recieve is an instance of  `telegram.ext.utils.callbackdatacache.InvalidCallbackData`. If you don't need persistence otherwise, you can set `store_callback_data` to `True` and all the others to `False`.
2. If you have a number of keyboards that need to stay valid for a very long time, you might need to do some tweaking manually (see below)
3. When using the `CallbackQueryHandler`, the `pattern` argument can now be either

    * a regex expression, which will be used, if the `callback_data` is in fact a string
    * a callable accepting the `callback_data` as only argument. You can perform any kinds of tests on the `callback_data` and return `True` ore `False` accordingsly
    * a type. In that case the `CallbackQuery` will be handled, if the `callback_data` is an instance of that type. Btw: This allows you to inform users, when a buttons data has been dropped from cache. With `CallbackQueryHandler(callback, pattern=InvalidCallbackData)` you can e.g. call `update.callback_query.answer(text='Button is no longer valid', show_alert=True)` to inform the user.

## Memory Usage

PTB stores the callback data objects in memory. Additionally to that, it stores a mapping of `CallbackQuery.id` to the corresponding UUID. By default both storages contain a maximum number of 1024 items. You can change the size by passing an integer to the `aribtrary_callback_data` argument of `Updater/ext.Bot`.

As PTB can't know when the stored data is no longer needed, it uses a LRU (Least Recently Used) cache, i.e. when the cache is full, it will drop the keyboard that has been not used for the longest time. However, if you want to keep memory usage low, you have 
additional options to drop data:

* on receiving a `CallbackQuery`, you can call `context.drop_callback_data(callback_query)`. This will delete the data associated with the keyboard attached to the message that originated the `CallbackQuery`. Of course, for inline messages, no message is attached to the `CallbackQuery` and it won't work in that case. Calling `context.drop_callback_data` is safe in any case where you change the keyboard, i.e. `callback_query.edit_message_text/reply_markup/…`
* To drop more data from memory, you can call `bot.callback_data.clear_callback_queries()` or `bot.callback_data.clear_callback_data()`, which will drop the mapping of `CallbackQuery` ids to the associated UUID and the mapping of UUIDs to data, respectively. Both methods also accept a `time_cutoff`, allowing you to delete only entries older than a specified time.

## Security of InlineKeyboardButtons

Callback updates are not sent by Telegram, but by the client. This means that they can be manipulated by a user. (While Telegram inofficially does try to prevent this, they don't want Bot devs to rely on them doing the security checks).

Most of the time, this is not really a problem, since `callback_data` often just is `Yes`, `No`, etc. However, if the callback data is something like `delete message_id 123`, the malicious user could delete any message sent by the bot.

When using `arbitrary_callback_data` as described above, PTB relpaces the outgoing `callback_data` with a [UUID](https://docs.python.org/3/library/uuid.html), i.e. a random unique identifier. This makes the `callback_data` safe: If a malicious client alters the sent `CallbackQuery`, the invalid UUID can't be resolved. In this case `CallbackQuery.data` will be an instance of `telegram.ext.utils.callbackdatacache.InvalidCallbackData`. Note that this is also the case, when the UUID *was* valid, but the data has already been dropped from cache - PTB can't distinguishe between the two cases.

telegram/ext/picklepersistence.py
```diff
@@ -286,5 +330,8 @@ def flush(self):
                 self.dump_file("{}_chat_data".format(self.filename), self.chat_data)
             if self.bot_data:
                 self.dump_file("{}_bot_data".format(self.filename), self.bot_data)
+            if self.callback_data:
+                print('flushing cd')
```
recepsirin: Might you have forgotten that?

Bibo-Joshi: Thanks for the catch :)

Poolitzer (changes requested):
Looks very good, just found a print statement.

Regarding tests, maybe we should discuss including userbots in our tests? Some will benefit from it.

Bibo-Joshi:
> Looks very good, just found a print statement.

Thanks for the finding!

> Regarding tests, maybe we should discuss including userbots in our tests? Some will benefit from it.

What do you have in mind that you would test?

Poolitzer:
> What do you have in mind that you would test?

We should move this in the dev group or in a separate issue, but I can remember at least 3 tests which would need an user "interaction" to be properly tested. It came up because of [this changed line](https://github.com/python-telegram-bot/python-telegram-bot/blob/7868cb550d2a7f934666d184fc62b9a04477edd9/tests/test_bot.py#L684-L686)

telegram/ext/callbackqueryhandler.py
```diff
@@ -79,9 +85,20 @@ class CallbackQueryHandler(Handler[Update]):
...
             which can be used to schedule new jobs. Default is :obj:`False`.
             DEPRECATED: Please switch to context based callbacks.
-        pattern (:obj:`str` | `Pattern`, optional): Regex pattern. If not :obj:`None`, ``re.match``
-            is used on :attr:`telegram.CallbackQuery.data` to determine if an update should be
-            handled by this handler. If :attr:`telegram.CallbackQuery.data` is not present, the
+        pattern (:obj:`str` | `Pattern` | :obj:`callable` | :obj:`type`, optional):
+            Pattern to test :attr:`telegram.CallbackQuery.data` against. If a string or a regex
+            pattern is passed, :meth:`re.match` is used on :attr:`telegram.CallbackQuery.data` to
+            determine if an update should be handled by this handler. If your bot allows arbitrary
+            objects as ``callback_data``, non-strings will not be accepted. To filter arbitrary
```
harshil21, line 92: non-strings will not be accepted?

tests/test_bot.py
```diff
@@ -1929,3 +2019,285 @@ def test_copy_message_with_default(self, default_bot, chat_id, media_message):
...
+            assert inline_keyboard[0][0] == replace_button
+            keyboard = list(bot.callback_data_cache._keyboard_data)[0]
+            data = list(bot.callback_data_cache._keyboard_data[keyboard].button_data.values())[0]
+            assert data == 'replace_test'
+        finally:
+            bot.arbitrary_callback_data = False
+            bot.callback_data_cache.clear_callback_data()
+            bot.callback_data_cache.clear_callback_queries()
+
+    # def test_replace_callback_data_reply_to_m
```
harshil21, line 2111: comment

telegram/ext/callbackdatacache.py
```diff
@@ -0,0 +1,423 @@
...
+            time_cutoff (:obj:`float` | :obj:`datetime.datetime`, optional): Pass a UNIX timestamp
+                or a :obj:`datetime.datetime` to clear only entries which are older.
+                For timezone naive :obj:`datetime.datetime` objects, the default timezone of the
+                bot will be used.
+
+        """
+        with self.__lock:
+            self.__clear(self._keyboard_data, time_cutoff)
+
+    def clear_callback_queries(self) -> None:
```
harshil21, line 400: wiki states that this accepts a time_cutoff

harshil21 (changes requested):
Things that need to be changed:
  
- [ ]  There's still traces of `InvalidButtonData` all over in the docs, needs to be updated (I gave up after a point)
- [ ] All instances of `Any` to `object` 
- [ ] The TODO section

There's also [TTLCache ](https://cachetools.readthedocs.io/en/stable/#cachetools.TTLCache)from cachetools which would mean that we wouldn't need the `access_time` logic, as we can just use their implementation. Maybe that can help simplifying the code?

Regarding slots, should we make a new branch based on this and then merge here? The implementation detail should be kept separate from this imo. This PR is big enough already

Finally I think an example for this would help users

telegram/callbackquery.py
```diff
@@ -53,6 +53,10 @@ class CallbackQuery(TelegramObject):
           until you call :attr:`answer`. It is, therefore, necessary to react
           by calling :attr:`telegram.Bot.answer_callback_query` even if no notification to the user
           is needed (e.g., without specifying any of the optional parameters).
+        * If you're using :attr:`Bot.arbitrary_callback_data`, :attr:`data` may be be an instance
```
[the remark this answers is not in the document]
Bibo-Joshi: I updated only for the Attribute, b/c if you manually instantiate a `CallbackQuery` (why the heck would anyone do that?), you probably shouldn't pass an arbitrary object.
btw, I left the type hint for `data` as `str`, b/c even if we put `object`, users would have to `cast` and this way you don't have to if you don't use arbitrary callback data. will include this in an example.

telegram/ext/basepersistence.py
```diff
@@ -337,6 +363,17 @@ def get_bot_data(self) -> Dict[object, object]:
...
+    def get_callback_data(self) -> Optional[CDCData]:
+        """Will be called by :class:`telegram.ext.Dispatcher` upon creation with a
+        persistence object. If callback data was stored, it should be returned.
+
+        Returns:
+            Optional[:class:`telegram.utils.types.CDCData`:]: The restored meta data as three-tuple
+                of :obj:`int`, dictionary and :class:`collections.deque` or :obj:`None`, if no data
+                was stored.
+        """
+        raise NotImplementedError
```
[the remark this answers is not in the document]
Bibo-Joshi: `get/update_callback_data` are always behind an `if store_callback_data`, which defaults to `False`, so we should be good.

telegram/ext/callbackdatacache.py
```diff
@@ -0,0 +1,423 @@
...
+    def persistence_data(self) -> CDCData:
+        """:obj:`telegram.ext.utils.types.CDCData`: The data that needs to be persisted to allow
+        caching callback data across bot reboots.
+        """
+        # While building a list/dict from the LRUCaches has linear runtime (in the number of
+        # entries), the runtime is bounded by maxsize and it has the big upside of not throwing a
+        # highly customized data structure at users trying to implement a custom persistence class
+        with self.__lock:
+            return list(data.to_tuple() for data in self._keyboard_data.values()), dict(
+                self._callback_queries.items()
```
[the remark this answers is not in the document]
Bibo-Joshi: not quite - it's an instance of `LRUCache`, which is a subclass of `Cache`, which is a subclass of `MutableMapping`. A user advanced enough to implement a persistence class should probably be able to handle that, but it feels cleaner to hand them a built-in type and it eliminates the possibility of the user doing something stupid with the cache …

telegram/ext/callbackdatacache.py
```diff
@@ -0,0 +1,423 @@
...
+        Args:
+            message (:class:`telegram.Message`): The message.
+
+        """
+        with self.__lock:
+            self.__process_message(message)
+
+    def __process_message(self, message: Message) -> Optional[str]:
+        """
+        As documented in process_message, but as second output gives the keyboards uuid, if any.
```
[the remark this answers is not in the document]
Bibo-Joshi: My bad. In an earlier version, this method also returned the message even though it is in-place. Now it doesn't return the message anymore, so changing to 

> As documented in process_message, but returns the uuid of the attached keyboard, if any, which is relevant for process_callback_query.

telegram/ext/picklepersistence.py
```diff
@@ -200,6 +211,26 @@ def get_bot_data(self) -> Dict[object, object]:
             self._load_singlefile()
         return deepcopy(self.bot_data)  # type: ignore[arg-type]
 
+    def get_callback_data(self) -> Optional[CDCData]:
+        """Returns the callback data from the pickle file if it exists or :obj:`None`.
+
+        Returns:
+            Optional[:class:`telegram.utils.types.CDCData`:]: The restored meta data as three-tuple
```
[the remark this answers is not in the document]
Bibo-Joshi: Oh boy, so many iterations for this PR 🙈 You have eagle eyes!

telegram/ext/callbackdatacache.py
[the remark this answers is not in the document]
harshil21: oh right. I saw the `.items()` and thought it was a regular python dict

telegram/ext/callbackdatacache.py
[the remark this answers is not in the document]
harshil21: alright ✔️

telegram/ext/updater.py
```diff
@@ -52,8 +52,8 @@ class Updater:
 
     Note:
         * You must supply either a :attr:`bot` or a :attr:`token` argument.
-        * If you supply a :attr:`bot`, you will need to pass :attr:`defaults` to *both* the bot and
-          the :class:`telegram.ext.Updater`.
+        * If you supply a :attr:`bot`, you will need to pass :attr:`arbitrary_callback_data`,
+          and :attr:`defaults` to the bot instead of the :class:`telegram.ext.Updater`.
```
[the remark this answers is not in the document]
Bibo-Joshi: I added an "… to the bot instead of the :class:`telegram.ext.Updater`. In this  case, you'll have to use the class :class:`telegram.ext.ExtBot`." to this note

Docs & type hints don't need to be updated IMO because `ExtBot` is a subclass of `Bot` …

telegram/ext/utils/webhookhandler.py
```diff
@@ -127,6 +127,11 @@ def post(self) -> None:
         update = Update.de_json(data, self.bot)
         if update:
             self.logger.debug('Received Update with ID %d on Webhook', update.update_id)
+            # handle arbitrary callback data, if necessary
+            # we can't do isinstance(self.bot, telegram.ext.Bot) here, because that class
+            # doesn't exist in ptb-raw
```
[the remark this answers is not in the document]
Bibo-Joshi: that is a very good point! :D

telegram/ext/callbackqueryhandler.py
```diff
@@ -79,9 +85,20 @@ class CallbackQueryHandler(Handler[Update]):
...
+            objects you may pass
+
+                * a callable, accepting exactly one argument, namely the
+                  :attr:`telegram.CallbackQuery.data`. It must return :obj:`True`, :obj:`False` or
+                  :obj:`None` to indicate, whether the update should be handled.
```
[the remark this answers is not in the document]
Bibo-Joshi: Yes, tried to make it more clear with 

> It must return :obj:`True` or  :obj:`False`/:obj:`None` to indicate, whether the update should be handled.

telegram/ext/callbackqueryhandler.py
Bibo-Joshi: 🤦

telegram/ext/callbackdatacache.py
Bibo-Joshi: Maybe my thinking was that it doesn't make as much sense for deleting callback query data … But whatever, to have is better than to need  :D

Bibo-Joshi: Ah, lol, found the reason: for `_callback_queries` I currently don.t even store the access time. Must have another look, why that is …

Bibo-Joshi: Okay, I think it's just the fact that CallbackQuery.id is unique and the query will only be handled once in contrast to a button which can be pressed repeatedly. So there is no need to differentiate between "old" and "new" callbackqueries - once you've handled them, you can forget about them.

Bibo-Joshi:
Thanks for the thorough review @harshil21, you found a lot of things to improve 😃 So far I've gone through your comments and marked all as resolved that where either clear or where you reacted with 👍 to my comments. I will get to the TODO section soonish. Regarding the other comments:

> There's also [TTLCache ](https://cachetools.readthedocs.io/en/stable/#cachetools.TTLCache)from cachetools which would mean that we wouldn't need the `access_time` logic, as we can just use their implementation. Maybe that can help simplifying the code?

Unfortunately this doesn't allow for the entries not to expire :/

> Regarding slots, should we make a new branch based on this and then merge here? The implementation detail should be kept separate from this imo. This PR is big enough already

Upsie, read that too late :D already merged that

> Finally I think an example for this would help users

*stay tuned* 🎶

Bibo-Joshi:
I added an example that can be reviewed.
Also I just cherry-picked the stuff from https://github.com/python-telegram-bot/python-telegram-bot/pull/2262#discussion_r632762276 + made according adjustments here, so all the open todo items in the description are "clean-up" after merge.
Only thing left from my side is to try & up the coverage a bit when the latest tests are finished

harshil21 (changes requested):
Okay example looks good. However, I noticed one thing while running the example - When a user spam clicks a button (say '1'), it becomes an instance of `InvalidCallbackData`, and it fails. Of course when you remove the `drop_callback_data` line, this doesn't happen anymore. But maybe such behaviour should be at least documented somewhere? 

Also can we remove the `cast()` in the example and ask mypy to ignore them? makes it a little complicated for beginners.

Bibo-Joshi:
> Okay example looks good. However, I noticed one thing while running the example - When a user spam clicks a button (say '1'), it becomes an instance of `InvalidCallbackData`, and it fails. Of course when you remove the `drop_callback_data` line, this doesn't happen anymore. But maybe such behaviour should be at least documented somewhere?

Ah, yes, that's right. In fact it could be considered a feature, see #2539 :D Anyway, I've updated the wiki above as follows:

> However, if you want to keep memory usage low, you have 
> additional options to drop data:
> 
> * on receiving a `CallbackQuery`, you can call `context.drop_callback_data(callback_query)`. This will delete the data associated with the keyboard attached to the message that originated the `CallbackQuery`. Calling `context.drop_callback_data` is safe in any case where you change the keyboard, i.e. `callback_query.edit_message_text/reply_markup/…`
>   **Note:** If the user clicks a button more than one time fast enough, but you call `context.drop_callback_data` for the first resulting `CallbackQuery`, the second one will have `InvalidCallbackData` as `callback_data`. However, this is usually not a problem, because one only wants one button click anyway.

Note that I also scraped the sentence about `drop_callback_data` not working for inline messages, b/c that's BS (commit incoming) :D 

> Also can we remove the `cast()` in the example and ask mypy to ignore them? makes it a little complicated for beginners.

I'm not too sure about this, tbh. If we drop `cast`, we'll have to add a `type: ignore`, which probably also confusing for beginners. I'd like to keep at least the cast in 

```python
# Get the data from the callback_data.
    number, number_list = cast(Tuple[int, List[int]], query.data)
```

because *if* you use type hinting, you'll have to do that cast in any case when using arbitrary callback data, so I think the example should showcase it. I could live with droppen the other two casts, though ….

harshil21 (approved):
LGTM!

PR #1845, opened 2020-03-26
Fix request timeout

AlexxIT:
Hi. Can you release this fix v11.1.1?
Long polling does not work without it (timeout always 5 sec).

Need to repair Home Assistant:
https://github.com/home-assistant/core/issues/18271#issuecomment-566763489

The update to the 12th version of the library will require a large rewrite of the Home Assistant component.


tsnoam:
@AlexxIT We do not support old branches.
This change is already part of later releases, please upgrade.

PR #1851, opened 2020-03-28
Prepare bumping version

Bibo-Joshi:
API 4.6 PR #1723 is already included
- [x] Release date needs to be inserted


Bibo-Joshi:
Merged manually

PR #1867, opened 2020-04-04
assortify examples

khode-mohsen:


Bibo-Joshi:
Hi. Thanks for your PR. However, after some interal discussion we decided against merging, as we

1. don't see a huge benefit of putting the examples in directories
2. don't like to dead links, which we would risk.

PR #1874, opened 2020-04-06
[Documentation] InlineQueryResultAudio doc fix

harshil21:


telegram/inline/inlinequeryresultaudio.py
```diff
@@ -48,9 +48,9 @@ class InlineQueryResultAudio(InlineQueryResult):
         id (:obj:`str`): Unique identifier for this result, 1-64 bytes.
         audio_url (:obj:`str`): A valid URL for the audio file.
         title (:obj:`str`): Title.
-        performer (:obj:`str`, optional): Caption, 0-200 characters.
-        audio_duration (:obj:`str`, optional): Performer.
-        caption (:obj:`str`, optional): Audio duration in seconds.
+        performer (:obj:`str`, optional): Optional. Performer.
+        audio_duration (:obj:`str`, optional): Audio duration in seconds.
+        caption (:obj:`str`, optional): Caption, 0-200 characters.
```
Bibo-Joshi, line 53: By now Telegram allows 0-1024 characters. Same holds for the other InlineQueryResult* classes

telegram/inline/inlinequeryresultaudio.py
```diff
@@ -48,9 +48,9 @@ class InlineQueryResultAudio(InlineQueryResult):
         id (:obj:`str`): Unique identifier for this result, 1-64 bytes.
```
Bibo-Joshi, line 48: In the base class we actually call `str(id)`, so `id (:obj:`str` | :obj:`int`)` would be okay for the argument. Just noticed while reviewing. Also for the other InlineQueryResult* classes

harshil21: Alright I'll change that in all InlineQueryResult* classes

telegram/inline/inlinequeryresultaudio.py
harshil21: Ah yes, I see it in the Bot API docs too. I'll update that.

PR #1913, opened 2020-04-20
Remove NullHandlers to unsuppress lastResort loging handler

n5y:
This removes the requirement to call `logging.basicConfig()` to see errors raised in handlers.

Users that already configure logging are unaffected by this change. Users that rely on python default logging configuration will now see warning+ level logs from PTB in stderr.

Suppressing default logging handler is not a good way control whether or not PTB logging is enabled, since users that actually do want to disable PTB logs might need to use basicConfig to configure something else. For this reason I removed all NullHandlers and not just the ones in dispatcher and updater.

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -44,8 +44,6 @@ from telegram.error import InvalidToken, TelegramError
 from telegram.utils.helpers import to_timestamp, DEFAULT_NONE
 from telegram.utils.request import Request
 
-logging.getLogger(__name__).addHandler(logging.NullHandler())
-
 
 def info(func):
     @functools.wraps(func)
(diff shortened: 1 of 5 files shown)


Poolitzer:
Sounds good to me.

tsnoam:
@Poolitzer Need to be carefully reviewed.
Everything related to logging is very fragile.

Poolitzer:
@tsnoam But in my opinion still a very important fix to have

n5y:
I'm reasonably sure NullHandlers were only needed to avoid "no handlers could be found for logger" error with py2, as lastResort handler was only added in py3.2. Since PTB no longer supports py2 it should be fine to remove them.

PR #1917, opened 2020-04-24
Api 4.8

Bibo-Joshi:
#### Pure API changes
- [x] `Bot.send_poll`: `explanation`, `explanation_parse_mode`, 
- [x] `Poll.{explanation, explanation_entities, open_period, close_date}`
- [x] `Bot.send_dice`: `emoji` (for whatever reason, sending a 'darts' is considered a die now …
- [x] `Dice.emoji`

#### Convenience stuff
- [x] `Poll.parse_explanation_entity/ies`
- [x] `Filters.dice.dice` and `Filters.dice.darts` for filtering only the corresponding emojis. Allow the `Filters.dice.*([1, 2])` syntax for both

### Todo
- [ ] Release notification should maybe include a statemant, that `Dice` a second positional argument.
- [x] Telegram seems to have changed something with  the photos. Getting `telegram.error.BadRequest: Type of file mismatch`. Need to investigate

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -3790,7 +3790,7 @@ class Bot(TelegramObject):
         Args:
             chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target private chat.
             emoji (:obj:`str`, optional): Emoji on which the dice throw animation is based.
-                Currently, must be one of “🎲” or “🎯”. Defauts to “🎲”
+                Currently, must be one of “🎲” or “🎯”. Defaults to “🎲”
             disable_notification (:obj:`bool`, optional): Sends the message silently. Users will
                 receive a notification with no sound.
             reply_to_message_id (:obj:`int`, optional): If the message is a reply, ID of the
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -3633,6 +3633,10 @@ class Bot(TelegramObject):
                   reply_to_message_id=None,
                   reply_markup=None,
                   timeout=None,
+                  explanation=None,
+                  explanation_parse_mode=DEFAULT_NONE,
+                  open_period=None,
+                  close_date=None,
                   **kwargs):
         """
         Use this method to send a native poll.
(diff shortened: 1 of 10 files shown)


Poolitzer:
@Bibo-Joshi Last time I faced this error in the support groups I suggested checking the dimensions of it. Maybe try uploading said photo as document, if that works, that might be it?

Bibo-Joshi:
@Poolitzer Problem seems to be sending a photo via

<pre>with open(&#x27;tests/data/thumb.jpg&#x27;, &#x27;rb&#x27;) as f:
    bot.send_photo(chat_id, f)</pre>

Problem persists, if I swat <code>tests/data/telegram.jpg</code> with <code>telegram.png</code> or <code>thumb.jpg</code>. However, I have no problems sending the to myself with a personal bot(token). Could that be a type of flood control? We have been getting some flooding errors in the tests recently … @tsnoam maybe we can add new/replace the existing tests bots?

telegram/dice.py
```diff
@@ -23,21 +23,33 @@
...
     """
-    This object represents a dice with random value from 1 to 6. (The singular form of "dice" is
-    "die". However, PTB mimics the Telegram API, which uses the term "dice".)
+    This object represents a dice with random value from 1 to 6 for currently supported base eomji.
+    (The singular form of "dice" is "die". However, PTB mimics the Telegram API, which uses the
+    term "dice".)
+
+    Note:
+        If :attr:`emoji` is "🎯", a value of 6 represents a bullseye, while a value of 1 indicates
+        that the dartboard was missed. However, this behaviour is undocumented and might be changed
```
Poolitzer, line 32:
```suggestion
        If :attr:`emoji` is "🎯", a value of 6 currently represents a bullseye, while a value of 1 indicates
        that the dartboard was missed. However, this behaviour is undocumented and might be changed
```

Poolitzer:
One thing: What about adding the emojis from Dice as constants? We normally do that. Right now we only have a list.

telegram/dice.py
Bibo-Joshi: Fair enough, will do.

Bibo-Joshi:
> One thing: What about adding the emojis from Dice as constants? We normally do that. Right now we only have a list.

Added that. Still kept the `ALL_EMOJI` list, though

Poolitzer:
@Bibo-Joshi Sure, didnt complain about that ;P

Bibo-Joshi:
CI Fail: `test_max_caption_length` gives an Internal server error for some reason. seems unrelated.
Coverage: can't test utf-16 encoding, that's basically it. Still, codecov is satisfied.

telegram/bot.py
```diff
@@ -3749,13 +3781,16 @@ def send_dice(self,
...
+                  emoji=None,
                   **kwargs):
         """
         Use this method to send a dice, which will have a random value from 1 to 6. On success, the
         sent Message is returned.
 
         Args:
             chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target private chat.
+            emoji (:obj:`str`, optional): Emoji on which the dice throw animation is based.
+                Currently, must be one of “🎲” or “🎯”. Defauts to “🎲”
```
tsnoam, line 3793:
```suggestion
                Currently, must be one of “🎲” or “🎯”. Defaults to “🎲”
```

PR #1920, opened 2020-04-25
Type hinting

### LINKED ISSUE #1373: Type-hinting for IDEs

Bibo-Joshi:
Preparations for type hinting:

* Mypy as static type checker is added to pre-commit and make-file
* Python 3.5 support is dropped, to allow typing for variables
* Two make sure that new PRs include typing, I tried two approaches:
  1. Use mypys report functionality together with `diff-cover` to compute the coverage of the changed lines only and let the test fail, if its < 100%. Unfortunately, mypy will mark any line with an "Any" as uncovered, even if that's allowed during type checking. But especially in `telegram.ext` we have a lot of stuff that can be "Any" (CH states, entries of `user_data` and so on). So, even after adding typing for the current code base, we would either
      1. have failing tests all the time or
      2. Allow the test to fail and manually look at the logs. grrr.
  2. Use mypys settings to disallow function definitions with no or incomplete typing. Upside compared to the `diff-cover` approach: its easier to see, if we're satisfied or not. Downside: When a file is changed *all* functions in that file will be checked and *all* classes/functions that are imported or inherited from, too. We can either
      1. type the complete base before merging
      2. try to single out the mypy result from the `pre-commit` test and allow it to fail until the whole base is typed
      3. Just live with the fact the the first few PRs after merging will either fail the pre-commit test or will have to do a lot of type hinting

  Personally, I like ii. better that i., which is why this is the current state of this pr. More precisely, I think I'd vote for ii.c)

Closes #1373


Bibo-Joshi:
Okay, I prepared for going for ii.a) :D

TBH, I didn't try to be as specific as possible. I rather tried to annotate most almost everything on a reasonably specific level and made sure that Mypy checks out. Many things can surely be improved, but IMO they can be revisited if

1. We're touching the code anyway
2. People have issues with the hints I wrote,

i.e. we can gradually improve the hints.

Some more specific notes:

* Telegram classes that have class methods using the bot, would need a `if self.bot is None: raise RuntimeError` or something, if we're strict. I found that unnecessary and just told mypy to not check optionals strictly in those classes (see `setup.cfg`)
* Not using `Literals` just yet as this is the first iteration  on typing and they're natively supported only for py>=3.8. They could be useful for things like `Poll.type` where only specific strings are allowed
* No `TypedDicts` yet as this is the first iteration. These *could* be used for `to_dict`, `de_json`. However the need is not too big, as those are internal methods
* reduced code duplication of `de_list` on the fly
* Noticed thet `Message.from_user` was a positional argument, which is wrong. Fixed. This needs a Release note, as this might be breaking. On the other hand, manually instantiating `Messages` doesn't happen too often, mostly in tests …
* I put some `type: ignore` where I didn't think doing it 100% correct would be worth the trouble
* `Request.stop(): Item "AppEngineManager" of "Union[PoolManager, AppEngineManager, SOCKSProxyManager, ProxyManager]" has no attribute "clear"` is that a bug?
* Passports are scary. I scipped some typing there
* Values returned by requests are kind of a trust thingy. I'm ignoring all return value warnings from there right now as we kind of have to expect Telegram to return what they are promising.
* Handlers: callback is typed as `update, context` as old api is deprecated
* Similar for Job callbacks
* `telegram.ext` is a bit harder to type for several reasons:
  * We're doing some hacky stuff, e.g. calling `Filters.filter` with `Update` *or* `Message`
  * Handlers are a bit tricky, as they *usually* accept `Update`s but for advanced handlers, `str` or even `object` (for `TypeHandler`) are okay
  * The docs are usually written w.r.t. to handling `Updates`, partly ignoring the other cases
  * The internals of `Updater` are a black box for me. I just marked a bunch of functions os "no type check" here

  That's why I've been a bit more lax on `telegram.ext.` Big parts of the typing usually isn't relevant for the user most times, anyway …

telegram/utils/typing.py
```diff
@@ -0,0 +1,35 @@
...
+# You should have received a copy of the GNU Lesser Public License
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+"""This module contains custom typing aliases."""
+from typing import Union, Any, Dict, TYPE_CHECKING, IO, Tuple, Optional
+
+if TYPE_CHECKING:
+    from telegram import InputFile, Update
+
+FileLike = Union[IO, 'InputFile']
+"""Either an open file handler or in :class:`telegram.InputFile`."""
```
septatrix, line 26: It did not look at this PR in depth but was curious what was in this file. If this just contains types I would prefer to call this while file `types.py` instead of `typing.py` as this is not contain any other type related stuff like `cast` etc. Furthermore `InputFile` seems to be casted to `IO` a lot. I did not look into the exact use cases but maybe it would be a good alternative to make `InputFile` inherit from `IO` and implement the same interface?

.pre-commit-config.yaml
```diff
@@ -18,3 +18,8 @@ repos:
         args:
         - --errors-only
         - --disable=import-error
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: 'v0.770'
+    hooks:
+    -   id: mypy
+        files: ^telegram/.*\.py$
```
septatrix, line 25: Missing newline :D

setup.cfg
```diff
@@ -40,3 +40,18 @@ omit =
...
+disallow_untyped_decorators = True
+show_error_codes = True
+
+[mypy-telegram.vendor.*]
+ignore_errors = True
+
+# Disable strict optional for telegram objects with class methods
+# We don't want to clutter the code with 'if self.bot is None: raise RuntimeError()'
+[mypy-telegram.callbackquery,telegram.chat,telegram.message,telegram.user,telegram.files.*,telegram.inline.inlinequery,telegram.payment.precheckoutquery,telegram.payment.shippingquery,telegram.passport.passportdata,telegram.passport.credentials,telegram.passport.passportfile,telegram.ext.filters]
+strict_optional = False
```
septatrix, line 57: Newline again ;D

telegram/__init__.py
```diff
@@ -124,6 +123,7 @@
                                    SecureData,
                                    FileCredentials,
                                    TelegramDecryptionError)
+from .bot import Bot
```
septatrix, line 126: Unrelated?

Bibo-Joshi:
> I did not find which sphinx version you use in the `requirements-dev.txt` though you might consider using the [autodoc_typehints](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints) option to display the type hints next to the parameter description instead of in the signature though that is probably personal preference.

Thanks for pointing it out :)
Iirc, RTD builds the docs with Sphinx 1.8.5, so `autodoc_typehints` is not present. Also, I think that I had tried that and some things I didn't quite work as expected, so it would have been a half-baked solution …

telegram/__init__.py
Bibo-Joshi: I think this was to avoid circular imports

setup.cfg
Bibo-Joshi: There wasn't one before

telegram/utils/typing.py
Bibo-Joshi: Renaming sounds reasonable. About casting `InputFile`: You probably mean those snippets in the `send_<media_type>` methods?

```python
        elif InputFile.is_file(video_note):
            video_note = cast(IO, video_note)
            video_note = InputFile(video_note)
```
`InputFile.is_file` just checks, if the input has a `read()` methods so that `InputFile` can handle the file on init. But it's hard to explain to mypy that this is valid, so I casted.

septatrix: Yes I meant those snippets. Hm that sound tricky - I just tested and `isinstance(open("somefile"), typing.IO)` weirdly returns `False` (though checking with `io.IOBase` yields `True`). `Protocols` would be perfect though they only exist in Python 3.8 and later. Can't think of a better solution right now

telegram/__init__.py
septatrix: I think it was just moved a few lines but I guess that could avoid some rare circular import so I am fine with it :D

telegram/__main__.py
```diff
@@ -21,11 +21,13 @@
 
 import certifi
 
+from typing import Optional
+
```
JosXa, line 25:
```suggestion
```

telegram/base.py
```diff
@@ -17,36 +17,64 @@
...
 
 import warnings
 
+from telegram.utils.types import JSONDict
+from typing import Tuple, Any, Optional, Type, TypeVar, TYPE_CHECKING, List
+
+if TYPE_CHECKING:
+    from telegram import Bot
+
+TO = TypeVar('TO', bound='TelegramObject')
```
JosXa, line 33: covariant=True?

telegram/callbackquery.py
```diff
@@ -144,7 +149,8 @@ def edit_message_text(self, text, *args, **kwargs):
             return self.bot.edit_message_text(text, chat_id=self.message.chat_id,
                                               message_id=self.message.message_id, *args, **kwargs)
 
-    def edit_message_caption(self, caption, *args, **kwargs):
+    def edit_message_caption(self, caption: str, *args: Any,
+                             **kwargs: Any) -> Union[Message, bool]:
```
JosXa, line 153: strange line break

telegram/callbackquery.py
```diff
@@ -172,7 +178,8 @@ def edit_message_caption(self, caption, *args, **kwargs):
                                                  message_id=self.message.message_id,
                                                  *args, **kwargs)
 
-    def edit_message_reply_markup(self, reply_markup, *args, **kwargs):
+    def edit_message_reply_markup(self, reply_markup: 'InlineKeyboardMarkup', *args: Any,
```
JosXa, line 181: Tbh I'd just add black to pre-commit.yml aswell if every single file is being changed anyway...

```
-   repo: https://github.com/psf/black
    rev: 20.8b1
    hooks:
    -   id: black
```

telegram/chat.py
```diff
@@ -86,32 +91,32 @@ class Chat(TelegramObject):
...
+                 bot: 'Bot' = None,
+                 photo: ChatPhoto = None,
+                 description: str = None,
+                 invite_link: str = None,
+                 pinned_message: 'Message' = None,
+                 permissions: ChatPermissions = None,
+                 sticker_set_name: str = None,
+                 can_set_sticker_set: bool = None,
+                 slow_mode_delay: int = None,
+                 **kwargs: Any):
```
JosXa, line 119: I'd generally prefer `**kwargs: Dict[str, Any]`

telegram/base.py
Bibo-Joshi: I'm not too familiar with the details, but from what I understand I guess you're right. E.g. in any place, where `Message` is valid, any subclass of `Message` should be valid

telegram/callbackquery.py
Bibo-Joshi: I think code formatting is worth its own discussion and PR

telegram/chat.py
Bibo-Joshi: https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values

JosXa: Uiiii TIL, thanks!

JosXa (approved):

Poolitzer (changes requested):
Monumental PR! I eagerly look forward to have static typing supported in our library. Just a couple of small discussion points, most of them repeat themself. Then I am all good with it. Thank you so much for putting the time into it.

telegram/bot.py
```diff
@@ -211,77 +235,77 @@ def _validate_token(token):
...
+        """:obj:`bool`: Bot's supports_inline_queries attribute."""
 
-        return self.bot.supports_inline_queries
+        return self.bot.supports_inline_queries  # type: ignore
 
-    @property
+    @property  # type: ignore
     @info
-    def commands(self):
+    def commands(self) -> List[BotCommand]:
```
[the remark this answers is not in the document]
septatrix: If you mean that an empty list is valid as an instance of a `List[BotCommand]` then yes

telegram/chat.py
```diff
@@ -86,32 +91,32 @@ class Chat(TelegramObject):
...
-                 description=None,
-                 invite_link=None,
-                 pinned_message=None,
-                 permissions=None,
-                 sticker_set_name=None,
-                 can_set_sticker_set=None,
-                 slow_mode_delay=None,
-                 **kwargs):
+                 id: int,
+                 type: str,
```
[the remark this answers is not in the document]
septatrix: A Union of Literals might work. However that might require Python 3.7 (or 3.8 I am not sure) - alternatively the `typing_extensions` package might be used for older versions

telegram/ext/conversationhandler.py
```diff
@@ -442,19 +454,26 @@ def check_update(self, update):
...
             check_result: The result from check_update. For this handler it's a tuple of key,
                 handler, and the handler's check result.
             update (:class:`telegram.Update`): Incoming telegram update.
             dispatcher (:class:`telegram.ext.Dispatcher`): Dispatcher that originated the Update.
+            context (:class:`telegram.ext.CallbackContext`, optional): The context as provided by
+                the dispatcher.
 
         """
-        conversation_key, handler, check_result = check_result
+        update = cast(Update, update)  # for mypy
```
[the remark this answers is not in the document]
septatrix: I think the comments are superfluous. A cast is a noop operation anyway so I cannot think of a reason why one would need a cast for anything but mypy

telegram/ext/defaults.py
```diff
@@ -76,71 +77,71 @@ def __init__(self,
         self._tzinfo = tzinfo
 
     @property
-    def parse_mode(self):
+    def parse_mode(self) -> Optional[str]:
         return self._parse_mode
 
     @parse_mode.setter
-    def parse_mode(self, value):
+    def parse_mode(self, value: Any) -> None:
```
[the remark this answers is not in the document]
septatrix: There is a `NoReturn` or similar type IIRC. It signals function which never return (infinite loop) or which always raise an exception

telegram/bot.py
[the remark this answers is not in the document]
Bibo-Joshi: Yes, the list can be empty. `List[BotCommand]` also covers `[]`

telegram/bot.py
```diff
@@ -1636,8 +1664,42 @@ def answer_inline_query(self,
...
+                        next_offset_int * MAX_INLINE_QUERY_RESULTS]
+                else:
+                    effective_results = results[current_offset_int * MAX_INLINE_QUERY_RESULTS:]
+        else:
+            effective_results = results
+
+        for result in effective_results:
+            _set_defaults(result)
+
+        results_dicts = [res.to_dict() for res in effective_results]
```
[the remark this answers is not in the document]
Bibo-Joshi: No, it shouldn't … But it will be gone, after I rebase one more time before I merge …

telegram/chat.py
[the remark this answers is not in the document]
Bibo-Joshi: > * Not using Literals just yet as this is the first iteration on typing and they're natively supported only for py>=3.8. They could be useful for things like Poll.type where only specific strings are allowed

We have two options:

1. Add a third party lib just to type hint literals
2. Live with not being able to hint them until 2023-06, when Py3.7 dies

IMHO the need to type hint literals is not very high in our lib. We have class attributes for basically all those input values (`Chat.{PRIVATE, GROUP, …}`, poll types, dice emojis, parse modes, you name it) and using those is faster anyway than typing manually and running the type checker …

telegram/ext/conversationhandler.py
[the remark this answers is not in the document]
Bibo-Joshi: I don't understand what @Poolitzer want's to say … :/

telegram/ext/defaults.py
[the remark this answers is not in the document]
Bibo-Joshi: uh, that's sweet. Will change!

telegram/files/inputmedia.py
```diff
@@ -129,15 +136,19 @@ class InputMediaPhoto(InputMedia):
...
-    def __init__(self, media, caption=None, parse_mode=DEFAULT_NONE):
+    def __init__(self,
+                 media: Union[str, FileLike, PhotoSize],
+                 caption: str = None,
+                 parse_mode: Union[str, DefaultValue] = DEFAULT_NONE):
         self.type = 'photo'
 
         if isinstance(media, PhotoSize):
-            self.media = media.file_id
+            self.media: Union[str, InputFile] = media.file_id
```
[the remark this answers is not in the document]
Bibo-Joshi: this is just the annotation for `self.media`. The `InputFile` case happens in both of the other two clauses

telegram/bot.py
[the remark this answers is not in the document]
Bibo-Joshi: uh, no, that's just an ugly diff because I moved the setting of the defaults to the local function. noam would probably scold me for it but imho it doesn't really matter with this PR xD

telegram/bot.py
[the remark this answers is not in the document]
Poolitzer: great!

telegram/bot.py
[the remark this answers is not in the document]
Poolitzer: Not only noam *looks at you disapprovingly*

telegram/chat.py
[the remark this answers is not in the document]
Poolitzer: okay, fair enough. I will set myself a reminder for then then ;P

telegram/files/inputmedia.py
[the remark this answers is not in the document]
Poolitzer: hmmm, kinda ugly if you ask me, but okay.

Poolitzer (approved):
Looks good to me now! If there actually are errors, we can fix them in a quick release after 13. I would merge.

Bibo-Joshi:
Codecov is still failing, but that's mostly due to this PR

* annotating code that wasn't tested before
* adding some `if`-clauses that are just there to make mypy happy

So we can just up the coverage the time we change some code … Merging :)

(Codacy fails, because the v13 branch is not tracked …)

PR #1931, opened 2020-05-01
Stabelizing CI

Bibo-Joshi:
* adds `expect_bad_request` to deal with `Type of file mismatch` errors (makes them xfail)
* Makes flood control errors xfail. Actually respecting them, i.e. retrying later, doesn't play well with `pytest.mark.timeout`, unfortunately …
* Deletes stickers from sets, if it gets full.

### DIFF
--- a/tests/bots.py
+++ b/tests/bots.py
@@ -21,6 +21,9 @@ import json
 import base64
 import os
 import random
+import pytest
+from telegram.utils.request import Request
+from telegram.error import RetryAfter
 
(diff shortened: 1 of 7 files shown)


Bibo-Joshi:
Fails:

* test_official b/c no api 4.8 yet
* py3.7: test_set_gamescore:
* py3.8: timeout in get_updates. don't think that's related …

Bibo-Joshi:
@Bibo-Joshi @tsnoam would you like to review?

PR #1940, opened 2020-05-03
Doc fixes

### LINKED ISSUE #1932: Doc strings on Message.edit_
They currently state

> You can only edit messages that the bot sent itself, therefore this method can only be used on the return value of the `bot.send_*` family of methods.

However, channel-posts can be edited, too. Sadly, Telegram is not very specific on which messages can be edited, so this behaviour is not documented …

### LINKED ISSUE #1938: Doc strings for Thumbs
Our docs on `send_document` suggest that thumbs can only be passed as url or file_id.
The official api states that thumbs can not be reused and must be uploaded as new file.

* correct the doc string for `send_document`
* check all other `thumb` doc strings
* check that the thumbs are acutally handled correctly

Possibly related to #1709

Bibo-Joshi:
* Updates the notes on which messages can be edited for `Message.edit_*`, closes #1932
* Updates the docstring on the `thumb` argument of `Bot.send_*` and `InputMedia*`, closes #1938 
* Shortens the description of `InputMedia*` attributes: I.e. the limitations should be listet for the arguments, but not for the attributes
* Corrects the type statements of `InputMedia.{media, thumb}`: After initializing both are either `string`, `Audio/Video/...` or `InputFile`
* Corrects attribute docstring of `Updater.user_sig_handler`
* Improves rendering of `CallbackContexts` attributes ("Optional" goes in the description, not in the type)

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -549,9 +549,10 @@ class Bot(TelegramObject):
                 JSON-serialized object for an inline keyboard, custom reply keyboard, instructions
                 to remove reply keyboard or to force a reply from the user.
             thumb (`filelike object`, optional): Thumbnail of the file sent; can be ignored if
-                thumbnail generation for the file is supported server-side. The thumbnail should
-                be in JPEG format and less than 200 kB in size. A thumbnail's width and height
-                should not exceed 320. Ignored if the file is not is passed as a string or file_id.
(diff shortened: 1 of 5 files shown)


Bibo-Joshi:
Fails unrelated, as this are only doc fixes. Merging.

PR #1955, opened 2020-05-14
Add Basketball Dice

code1mountain:
There is a new Dice Variation for the Basketball (🏀) emote, that only returns values from 1-5.

### DIFF
--- a/telegram/dice.py
+++ b/telegram/dice.py
@@ -31,13 +31,17 @@ class Dice(TelegramObject):
         If :attr:`emoji` is "🎯", a value of 6 currently represents a bullseye, while a value of 1
         indicates that the dartboard was missed. However, this behaviour is undocumented and might
         be changed by Telegram.
+        
+        If :attr:`emoji` is "🏀", a value of 4 or 5 currently score a basket, while a value of 1 to 3
+        indicates that the basket was missed. However, this behaviour is undocumented and might
+        be changed by Telegram.
 
     Attributes:
         value (:obj:`int`): Value of the dice.
         emoji (:obj:`str`): Emoji on which the dice throw animation is based.
 
     Args:
-        value (:obj:`int`): Value of the dice, 1-6 for DICE and DARTS, 1-5 for BASKETBALL
+        value (:obj:`int`): Value of the dice
         emoji (:obj:`str`): Emoji on which the dice throw animation is based.
     """
     def __init__(self, value, emoji, **kwargs):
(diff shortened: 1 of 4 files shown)


telegram/dice.py
```diff
@@ -37,7 +37,7 @@ class Dice(TelegramObject):
         emoji (:obj:`str`): Emoji on which the dice throw animation is based.
 
     Args:
-        value (:obj:`int`): Value of the dice, 1-6.
+        value (:obj:`int`): Value of the dice, 1-6 for DICE and DARTS, 1-5 for BASKETBALL
```
Bibo-Joshi, line 40: This should rather go in the `Note` above. It would also be good to describe the different values for the basketball. What I found out so far:

1 bounces off directly
2 circles & then bounces off
3 get's stuck between ring and the back thingy
4 circles, but get's in the net (stuck)
5 get's in the net directly
6 doesn't seem to exist.

Just use proper formulations instead :D
Please make sure that it's clear that Telegram didn't document that.

telegram/dice.py
```diff
@@ -55,6 +55,8 @@ def de_json(cls, data, bot):
...
     DARTS = '🎯'
     """:obj:`str`: '🎯'"""
-    ALL_EMOJI = [DICE, DARTS]
-    """List[:obj:`str`]: List of all supported base emoji. Currently :attr:`DICE` and
-    :attr:`DARTS`."""
+    BASKETBALL = '🏀'
+    """:obj:`str`: '🏀'"""
+    ALL_EMOJI = [DICE, DARTS, BASKETBALL]
+    """List[:obj:`str`]: List of all supported base emoji. Currently :attr:`DICE`,
+    :attr:`DARTS` and :attr: `BASKETBALL`."""
```
Bibo-Joshi, line 62:
```suggestion
    :attr:`DARTS` and :attr:`BASKETBALL`."""
```
Else the docs won't render correctly. Please see the contribution guide on how to build the docs locally.

telegram/dice.py
code1mountain: Thanks for the review, I added the requested changes and fixed the docs.

code1mountain:
The requested changes are implemented now.

Bibo-Joshi:
Thanks for the update. Still, after internal discussion we decided not to merge, before we get an official announcement from Telegram (that might i.e. even have info about the missing 6 for basketball).

tsnoam:
@Bibo-Joshi In general, we prefer not to merge undocumented features because they&#x27;re prone to change.

Poolitzer:
Thanks for that, I will merge it in my 4.9 branch. Do you want to be included in the author file as well?

PR #1962, opened 2020-05-22
Fix one typo

starry69:


Bibo-Joshi:
Thanks for pointing this out! :)

starry69:
👍👍

Issue #1797, opened 2020-02-27
[FEATURE] Extensions of Defaults


Bibo-Joshi:
Possible candidates to add to the Defaults concept:

## tzinfo
We should use Defauts.tzinfo as default instead of UTC … That would mean:

* When a job is created without setting tzinfo explicitly, use Defaults.tzinfo
* When calling JobQueue._put() with a naive `when`, use `when.replace(tzinfo=defaults.tzinfo)`
  This would basically not change the current set up, but just put it under the hood
* Probably `Message.date` and all the friends should also use `defaults.tzinfo`, so basically every `datetime` instance occuring in the lib.

## Arguments of Filters

* Filters.command: allow commands *anywhere* in the text or not
* Filters.{user, chat}: *allow_empty* (only relevant after #1757)

How to make it work: We would need to override those attributes on `dispatcher.add_handler` somehow. that's pretty much the only time, the filters and the dispatcher can see each other …

## Dice.emoji

* Only relevant after #1917. Also discussable, if this really is a good idea, as the emoji attribute of `Dice` is positional and not optional

## ReplyKeyboardMarkup

* `resize_keyboard`
* `one_time_keyboard`

# Redesign of `default_quote`

Right now, `default_quote` get's passed from getting updates through all objects, whose attributes have a `Message` attribute and is set as attribute in `Message`. But as the `Message.reply_*` methods only work with a bot anyway, we could just access `Message.bot.defaults.default_quote`.

See #1965

Bibo-Joshi:
On second thought, we should use `Defauts.tzinfo` as default instead of `UTC` then, so it's probaby worth it already …

Bibo-Joshi:
closed by #2042 and #1965

(We decided that additional things should probably only be added to defaults only, when users actually request it  …)

JosXa:
> Filters.command: allow commands anywhere in the text or not

what the heck? ^^

Bibo-Joshi:
> > Filters.command: allow commands anywhere in the text or not
> 
> what the heck? ^^

Read as: Add default setting for making `Filters.command` only filter for commands at the start of the message or anywhere in the message. The filter has a parameter for that since … 12.4?
But as stated above, the consensus in dev chat was to not add more defaults too quickly. See [here](https://t.me/c/1494805131/5336) and above

PR #1965, opened 2020-05-26
Refactor handling of default_quote

Bibo-Joshi:
Right now, `bot.defaults.quote` is passed through the json data when decoding an update for it to be stored in `Message.default_quote`. As @n5y pointed out offline, it would be cleaner to just use `Message.bot.defaults.quote`. This is, what this PR does.

To not break backwards compatibility, I left some optional `default_quote` args in the Webhook classes untouched and also kept `default_quote` as argument for `Message` (with a warning that it will override the bots settings). That's discussable though, as nobody should have actually used that …

Closes parts of #1797


Bibo-Joshi:
Latests commits  

* Remove `Message.default_quote` all along, making it a breaking change. Hence, changed PR to the new v13 branch. Also, this way, unpickling is no problem anymore.
* Makes sure that a bots default values are un/pickled along with it.
  * upside: the unpickled bot of an unpickled `Message` still has the correct default `quote` setting
  * downside: the unpickled bot is not the same as the updaters one. If the defaults for the updater are changed, the unpickled bots don't change accordingly. Then again, this is not really new, i.e. the behaviour is the same for the token. maybe that should at least be documented somewhere …

Bibo-Joshi:
We should postpone this after #1994

Bibo-Joshi:
Also, we could/should raise a warning, if both `bot` and `defaults` are passed to `Updater`, as after this the defaults passed to the updater have no effect.

telegram/ext/updater.py
```diff
@@ -116,6 +118,12 @@ def __init__(self,
                  dispatcher=None,
                  base_file_url=None):
 
+        if defaults and bot:
+            warnings.warn('Passing defaults to an Updater has no effect, if a Bot is passed, '
+                          'too. Pass it to the Bot instead.',
```
Poolitzer, line 123:
```suggestion
            warnings.warn('Passing defaults to an Updater has no effect when a Bot is passed'
                          'as well. Pass them to the Bot instead.',
```

Poolitzer:
Here is a lot of changes which I reviewed in different PRs, like the refactor of the jobqueue, the api kwargs or the persistence part. Is that intended?

PR #1980, opened 2020-06-05
API 4.9 update

Poolitzer:
Basketball was added by @code1mountain, rest done by me.
Open questions:

- [x] How to test for via_bot attribute of a Message
- [ ] Do we want to offer the mime types of the InlineQueryResults thumb urls as constants? If yes, where?

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -3787,13 +3787,14 @@ class Bot(TelegramObject):
                   emoji=None,
                   **kwargs):
         """
-        Use this method to send a dice, which will have a random value from 1 to 6. On success, the
+        Use this method to send an animated emoji, which will have a random value. On success, the
         sent Message is returned.
 
         Args:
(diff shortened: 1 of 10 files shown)


Bibo-Joshi:
Thanks for the quick PR! Will have a closer look later today/on the weekend

> * [ ]  How to test for via_bot attribute of a Message

We're currently only testing `answer_inline_query` by mocking. But as setting the attribute is not our thing to do, but TGs, personally I'd say test manually and that's it.

> * [ ]  Do we want to offer the mime types of the InlineQueryResults thumb urls as constants? If yes, where?

Can you test, if TG servers infere that automatically? If yes: I don't really see a need for someone setting that manually and hence, for adding the corresponding constants

Poolitzer:
@Bibo-Joshi The docs say telegrams server automatically assume JPEG. I can test it if you want

PR #1981, opened 2020-06-05
Refactor JobQueue

### LINKED ISSUE #1936: Refactor JobQueue
Maintaining `JobQueue` is not easy and sometimes error prone, see e.g. #1338 

We could refactor `JobQueue` to use a third party library for the scheduling, keeping the `run_*` methods as wrappers. We should probably also add a `run_cron` as suggested in #1337 as convenience wrapper for all the other cases we can't all cover (like run a job every time mars and saturn align …)

Just implement `run_cron` would be a first step.

### LINKED ISSUE #2004: [BUG] Errors in Job callbacks are not passed to error handler
### Steps to reproduce
1. Register an errory_handler
2. Set up a job

3. Make that job fail on execution

### Expected behaviour
Error should be handled by the error handler

### Actual behaviour
It's not, as the JobQueue doesn't handle that correctly (see [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/15268acb2729411ab8a64ef6127933b1d100c27d/telegram/ext/jobqueue.py#L571), this should call `dispatcher.dispatch_error`)

### Logs
https://pastebin.com/LhFTBb6S

Bibo-Joshi:
When ready, closes #1936 and on the fly closes #2004 

Introduces APScheduler as 3rd party lib for scheduling tasts. The PR is not yet really polished, especially I'll need to increase coverage. But before getting started on that, I'd like to get feedback on the general design.

My main concern is the `Job` (already mentioned offline). Unfortunately, simply subclassing `aps.Job` to add a `context` argument doesn't work - we'd have to patch `aps.Job` to do that. I don't fancy that.
On the other hand, having a custom `Job` class with custom attributes allows to make switching to another lib a bit less breaking, if we have to do that at some point.
So I added a `job` attribute to our `Job` class, which holds the corresponding `aps.Job`.

Besides the fact that rescheduling jobs must now be done directly through the `aps` interface, the main breaking change is, that `JobQueue ` now can only handle `pytz` timezones. But personally, I don't see a problem in that. Switching from fixet utc-offset timezones should be easy enough and people who need weird custom timezones should be capable enough of handling the change …

Reragrding tests:
* test_official currently fails b/c API 4.9
* made the timeout tests for `ConversationHandler` a bit more generous on the timing, `aps` seems to like that
* switched to `pytzs` everywhere in the tests
* besides that, I only made the existing tests run. as mentioned above, I'll need to improve coverage, once we're okay with the general design.

EDIT:
That are a lot more failing tests than I'd hoped for … Will need to check on an actual OS locally (yes, I developed on Windows, shame on me …)

EDIT 2:
* `the updater.idle()` test just wasn't executed on windows … fixed it.
* APS is a bit pickier on doing things "right now". That's one of the things, we should mention on the release notes. So I just removed `test_run_repeating_first_immediate` and `test_run_repeating_first_timezone` is now also adapted for APS' needs


telegram/ext/jobqueue.py
```diff
@@ -195,10 +199,21 @@ def run_repeating(self, callback, interval, first=None, context=None, name=None)
                 then ``first.tzinfo`` will define ``Job.tzinfo``. Otherwise UTC will be assumed.
 
                 Defaults to ``interval``
+            last (:obj:`int` | :obj:`float` | :obj:`datetime.timedelta` |                        \
+                   :obj:`datetime.datetime` | :obj:`datetime.time`, optional):
+                Latest possible time for the job to run. This parameter will be interpreted
+                depending on its type. See ``first`` for details.
+
+                If ``last`` is :obj:`datetime.datetime` or :obj:`datetime.time` type
+                and ``last.tzinfo`` is :obj:`None` UTC will be assumed.
```
Poolitzer, line 208:
```suggestion
                and ``last.tzinfo`` is :obj:`None`, UTC will be assumed.
```

tests/test_conversationhandler.py
```diff
@@ -742,8 +735,7 @@ def test_conversation_handler_timeout_state_context(self, cdp, bot, user1):
         message.text = '/startCoding'
         message.entities[0].length = len('/startCoding')
         cdp.process_update(Update(update_id=0, message=message))
-        sleep(0.5)
-        cdp.job_queue.tick()
+        sleep(0.6)
```
Poolitzer, line 738: doesnt it make sense to put this in a constant?

telegram/ext/jobqueue.py
```diff
@@ -63,12 +61,38 @@ def __init__(self):
...
+            return None
+        if isinstance(time, (int, float)):
+            return self._tz_now() + datetime.timedelta(seconds=time)
+        if isinstance(time, datetime.timedelta):
+            return self._tz_now() + time
+        if isinstance(time, datetime.time):
+            dt = datetime.datetime.combine(
+                datetime.datetime.now(time.tzinfo or self.scheduler.timezone).date(), time)
+            if dt.tzinfo is None:
+                dt = dt.replace(tzinfo=self.scheduler.timezone)
```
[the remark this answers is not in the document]
jh0ker: Hmm I don't really understand what that means, @Poolitzer
We need to make sure we are getting the correct date for the provided timezone and also that the resulting `datetime` has the correct `tzinfo`.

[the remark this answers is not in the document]
Bibo-Joshi: I guess what @Poolitzer means is that the `if dt.tzinfo is None` check is unnecassary, as it's already set in line 88. Don't know why I doubled that; will check if the tests run without it.

tests/test_conversationhandler.py
Bibo-Joshi: Not sure, if that's a good idea. Have one of the tests failing in the future, fiddle with the constant, suddenly everthing fails and it's not obvious why …

tests/test_jobqueue.py
```diff
@@ -89,17 +91,11 @@ def test_run_once_timezone(self, job_queue, timezone):
         """
```
jh0ker, line 91: I don't think this docstring is still applicable

telegram/ext/jobqueue.py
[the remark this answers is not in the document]
jh0ker: Hmm i don't think so, cause a `date` object doesn't have timezone information. `combine` takes the `tzinfo` from the `time` argument, if available.

telegram/ext/jobqueue.py
```diff
@@ -63,12 +61,36 @@ def __init__(self):
...
+    def _parse_time_input(self, time, shift_day=False):
+        if time is None:
+            return None
+        if isinstance(time, (int, float)):
+            return self._tz_now() + datetime.timedelta(seconds=time)
+        if isinstance(time, datetime.timedelta):
+            return self._tz_now() + time
+        if isinstance(time, datetime.time):
+            dt = datetime.datetime.combine(datetime.datetime.now().date(), time,
+                                           tzinfo=time.tzinfo or self.scheduler.timezone)
```
jh0ker, line 88: I don't know about this... Are we dropping 3.5 support (or did we already)? The third argument to `combine` has been added in 3.6.

Apart from that, couldn't `datetime.datetime.now().date()` return the wrong date this way?

Bibo-Joshi: Uh, you're right. Couldn't tell, because TG messed up our tests again …. But we're dropping py3.5 an #1920 anyway, so we might as well do here.

And right once more. I'll add it back.

Bibo-Joshi:
py3.7 fail unrelated. Codacy can't check for some reason, but I ran pre-commit locally again and it's fine.
As discussed offline we'll probably need to do some clean up after release anyway, but we need people to actaully use the new JobQueue for that. Merging.

PR #1983, opened 2020-06-07
Remove error handlers from examples

### LINKED ISSUE #1921: [FEATURE] Remove error handlers from examples
In the examples, we have a very simple error handler, that hides the traceback. This leads to python beginners having difficulties seeing the origin of their error.

Drop the error handler from examples and link to the snippet by @Poolitzer or maybe add a new example just for an error handler.

Make the current errors handlers at least show the traceback.

#### Additional context
@n5y suggested this and might even want to pr

n5y:
This should have been done a long time ago. A new example for a custom error handler can be added later.

Closes #1921

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -34,5 +34,8 @@ A basic example of a bot that can accept payments. Don't forget to enable and co
 ### [`persistentconversationbot.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/persistentconversationbot.py)
 A basic example of a bot store conversation state and user_data over multiple restarts.
 
+### [`errorhandlerbot.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/errorhandlerbot.py)
+A simple example of a bot that has a custom error handler.
+
 ## Pure API
 The [`echobot.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/echobot.py) example uses only the pure, "bare-metal" API wrapper.
--- a/examples/error_handler.py
+++ b/examples/error_handler.py
@@ -6,6 +6,7 @@
 This is a very simple example on how one could implement a custom error handler
 """
 import html
+import json
 import logging
 import traceback
 
@@ -40,7 +41,7 @@ def error_handler(update: Update, context: CallbackContext):
         '<pre>context.user_data = {}</pre>\n\n'
         '<pre>{}</pre>'
     ).format(
-        html.escape(update.to_json()),
+        html.escape(json.dumps(update.to_dict(), indent=2, ensure_ascii=False)),
         html.escape(str(context.chat_data)),
         html.escape(str(context.user_data)),
         html.escape(tb)
@@ -50,7 +51,7 @@ def error_handler(update: Update, context: CallbackContext):
     context.bot.send_message(chat_id=DEVELOPER_CHAT_ID, text=message, parse_mode=ParseMode.HTML)
 
 
-def bad_command(udpate: Update, context: CallbackContext):
+def bad_command(update: Update, context: CallbackContext):
     """Raise an error to trigger the error handler."""
     context.bot.wrong_method_name()
 
--- a/examples/error_handler.py
+++ b/examples/error_handler.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This program is dedicated to the public domain under the CC0 license.
+
+"""
+This is a very simple example on how one could implement a custom error handler
+"""
+import logging
+
+from telegram import Update, ParseMode
+from telegram.ext import Updater, CallbackContext
+
+import traceback
+
+logging.basicConfig(format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+                    level=logging.INFO)
+
+logger = logging.getLogger(__name__)
+# this can be your own ID, or one for a developer group/channel
+BOT_DEV = 208589966
+
+
+def on_error(update: Update, context: CallbackContext):
+    # lets log the error before we do anything else, so this appears at least in the logs
+    logger.error(msg="The following error happened while handling an update",
+                 exc_info=context.error)
+    # now we collect chat and user data
+    chat_data = str(context.chat_data)
+    user_data = str(context.user_data)
+    # format_tb returns the traceback in a list. We take the traceback from the error object
+    tb_list = traceback.format_tb(context.error.__traceback__)
+    # now we turn the list into a string
+    tb = ''.join(tb_list)
+    # build the message with some markdown
+    message = 'Update <pre>{}</pre> caused the error <i>{}</i>. The current chat data is ' \
+              '<i>{}</i>, the user data <i>{}</i>. The traceback is as follows:\n<pre>{}</pre>'
+    # and send it away
+    context.bot.send_message(chat_id=BOT_DEV,
+                             text=message.format(update, context.error, chat_data, user_data, tb),
+                             parse_mode=ParseMode.HTML)
+
+
+def main():
+    """Start the bot."""
+    # Create the Updater and pass it your bot's token.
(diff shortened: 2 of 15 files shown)


Poolitzer:
@n5y Kind of related to our refactoring to drop the need to import and call logging. I wanted to get (up) to (date on) that, soonTM

Bibo-Joshi:
Hi. Thanks for the PR. After short internal discussion we'd like to add a dedicated error handler example to the examples section in the same move. Would you care to add one? You can base it on the snippet in the [wiki](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Code-snippets#an-good-error-handler).

n5y:
I don't think lack of custom error handler example justifies keeping traceback-hiding error handlers in the examples any longer. Especially if there is a wiki entry about custom error handlers.

I also don't think the wiki snippet can be used as is. It's rather complex, but lacks some required safeguards and can hide original exception with a new one. It re-rises exception at the end which I don't think works too well with how PTB handles this. It uses f-string which are not supported by python 3.5.

Perhaps the snippet can just continue to live in the wiki where it can keep f-strings and is easier to modify. It's also easier to show multiple handler variants on the wiki. Or perhaps we can make another feature request issue specifically for an error handler example.

Poolitzer:
Yeah the snippet isnt perfect, agreed. Nonetheless, a simple error handler example isnt too hard to do, if you dont want to, I will happily throw one together.

Im not sure I get _It re-rises exception at the end which I dont think works too well with how PTB handles this._ This re raising works perfectly for me.

Poolitzer:
okay I added a simple error handler imo, is that ok?

examples/error_handler.py
```diff
@@ -0,0 +1,83 @@
...
+    # Build the message with some markup and additional information about what happened.
+    # You might need to add some logic to deal with messages longer than the 4096 character limit.
+    message = (
+        'An exception was raised while handling an update\n'
+        '<pre>update = {}</pre>\n\n'
+        '<pre>context.chat_data = {}</pre>\n\n'
+        '<pre>context.user_data = {}</pre>\n\n'
+        '<pre>{}</pre>'
+    ).format(
+        html.escape(update.to_json()),
```
Bibo-Joshi, line 43: Can we format the json outputs here for better readability? `json.dumps(update.to_dict()/chat_data, indent=2/4)` should do the trick.

n5y: We can do it for the update, but not for user/chat data, since it can be not json serializable. We can use pprint, but I don't like how it looks, and the huge indents that can quickly consume the 4096 message length limit. In the example chat data is empty and I imagine it's usually a lot shorter than the update, so I suggest we leave it unformatted.

n5y: Just noticed reason `.to_dict`/`.to_json` includes `_effective_*` for some reason.

Bibo-Joshi: I see. I think I'll vote for `pprint` then. The length limit is covered by the comment IMO. And as the error handler should be informative, a flat representation of a dict doesn't fulfill that goal imo. If only one of the `chat_data` s values is a dict/list, it quickly becomes a headache to read that.
As for the private attributes in `to_dict()`: Thanks for opening an issue. Already commented. We can ignore that for this PR, though.

examples/error_handler.py
```diff
@@ -0,0 +1,83 @@
...
+    # Create the Updater and pass it your bot's token.
+    # Make sure to set use_context=True to use the new context based callbacks
+    # Post version 12 this will no longer be necessary
+    updater = Updater("TOKEN", use_context=True)
+
+    # Get the dispatcher to register handlers
+    dp = updater.dispatcher
+
+    # Register the command...
+    dp.add_handler(CommandHandler('bad_command', bad_command))
```
Bibo-Joshi, line 68: Please make that listen to `'start'` to be consistent with the other examples.
Thought I had added this yesterday. Probably forgot to hit "Add to review" …

examples/errorhandlerbot.py
```diff
@@ -0,0 +1,88 @@
...
+        html.escape(json.dumps(update.to_dict(), indent=2, ensure_ascii=False)),
+        html.escape(str(context.chat_data)),
+        html.escape(str(context.user_data)),
+        html.escape(tb)
+    )
+
+    # Finally, send the message. You can replace this with something like
+    # context.bot.send_message(chat_id=DEVELOPER_CHAT_ID, text=message, parse_mode=ParseMode.HTML)
+    # to send the message to the bot developer rather than the user.
+    update.effective_message.reply_text(text=message, parse_mode=ParseMode.HTML)
```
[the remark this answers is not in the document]
n5y: I just don't like that this example requires two edits 40 lines apart in order to run, and if you forget the first one things will break only after you use a command instead of immediately.

Would you rather
1) Leave it as it was, perhaps replacing Poolitzer's chat id with 123456789.
1) Change developer chat line to raise a an error if not modified e.g. to `DEVELOPER_CHAT_ID = ` (unfinished line) or `DEVELOPER_CHAT_ID = 0123456789` (leading zero is a syntax error in python 3)
1) Make bot token a module level constant so all the edits are in one place. I prefer this option.
1) Leave things as they are now, with assumption that the user will change the line to suit their needs.
1) Change the code to use update.effective_user to answer most updates and ignore the ones it can't.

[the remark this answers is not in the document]
Bibo-Joshi: 3 it is, then

n5y:
I also added chat id in start command response to make it easier for people who don't know how to see their chat id.

n5y:
Should I squash this branch into a single "Remove error handlers from examples and add a dedicated error handler example" commit? And reabase it onto current master.

Bibo-Joshi:
> Should I squash this branch into a single "Remove error handlers from examples and add a dedicated error handler example" commit? And reabase it onto current master.

No need. I'll squash on merge. LGTM now.

n5y:
There are a few other examples missing from the readme, but it should be fixed in a different PR

Bibo-Joshi:
> There are a few other examples missing from the readme, but I think it should be fixed in a separate PR, rather than this one.

Thanks for pointing that out. Updated right now, giving you a MC ;P. But @Poolitzer wants to review anyway.

examples/README.md
```diff
@@ -34,5 +34,8 @@ A basic example of a bot that can accept payments. Don't forget to enable and co
 ### [`persistentconversationbot.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/persistentconversationbot.py)
 A basic example of a bot store conversation state and user_data over multiple restarts.
 
+### [`errorhandlerbot.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/errorhandlerbot.py)
+A simple example of a bot that has a custom error handler.
```
Poolitzer, line 38:
```suggestion
A simple example for a custom error handler.
```

n5y: Both the file change name and the addition of start command were implying it's not example of just an error handler (+ minimal code required to run it)

n5y: Doesn't mean I'm specifically against this change, I just wanted to point this thing-example vs bot-example parity even earlier, when these two changes were requested.

Poolitzer:
@Poolitzer see

examples/README.md
Poolitzer: hmm. The original sentence kinda focuses on the bot part imho, for which we have more suited examples (now). This is mainly for the custom error one. Well, I dont know, it was a mere suggestion, nothing more.

n5y: I'm going cast a 0.25 vote against this change. But I'm obviously biased because I wrote the original text. Feel free to hit that "commit suggestion" button if you think it's better.

Bibo-Joshi: The file name change was simply to fit with the other examples (didn't notice that we three without trailing `bot`, though, but doesn't really matter). I'll just rephrase to `A basic example on how to set up a costum error handler.` without overthinking and merge :D

PR #1978, opened 2020-06-04
add new predefined filter 'Filters.Date' to filter on message age.

ikkemaniac:
This allows for filtering messages based on the date they were send by specifying the maximum time difference between 'now' and when the message was send.

I've added some basis tests. Pls verify these as I'm not familiar with pytest and can't seem to get them to fail (ex. change 'seconds_ago' to 15 in test_filters_date_init doesn't throw an error in testing)

PR #1987, opened 2020-06-09
clean pending update based on Timedelta or datetime

ikkemaniac:
Follow-up of #1978 

Add option to pass a `timedelta`  or `datetime` as var `clean` to `.start_polling()` or `.start_webhook()`. 
Updates with a timestamp before `now - timedelta` will be ignored. 
Updates before `datetime` will be ignored.

I'm not familiar with the test-module to write these from scratch as this is not a 'direct' function to test. Can put in some time if y'all can point me to similar examples?

Maybe needs some cleanup but need the gh test bot results for that.


Bibo-Joshi:
Hi. Thanks for the PR. TBH, I haven't looked in too much detail, yet.

Regarding tests: I'm a bit surprised that we don't test clean at all. But If I see correctly, it should be enough to mock `Bot.get_updates` to return a custom list of updates. You can add `monkeypatch` to the args of the test function and use `monkeypatch.setattr(bot, 'get_updates', 'some_func_to_call_instead_of_get_updates')` for that. We use `monkeypatch` a lot in the tests, so you should be able to find examples :)

A few other things I noticed:
* It would be desirable to also allow `datetime` input for `clean` (naive datetimes should be interpreted as UTC)
* After #1621 `message.date` is already timezone aware, so you shouldn't have to replace the timezone
* In fact, we'll need to make sure that different timezones are handled correctly, especially for `datetime`. We have a custom fixture `timezone` for that, that you can use as arg for the test functions

If you have more questions, please don't hesitate and ask :)

ikkemaniac:
Ok, I'll look into it. Is the use of `**kwargs` like this accepted?

Bibo-Joshi:
> Ok, I'll look into it. Is the use of `**kwargs` like this accepted?

Mh … I guess it would be cleaner to either add an explicit argument `action_db_kwargs` or use `functools.partial` to pass the `timedelta/datetime` to `bootstrap_clean_time…`

Poolitzer:
No but we have dev requirements did you install these?

ikkemaniac:
yes, tried both `pip` and `pip3`.

the `pip` one complains about not finding timezone as part of datetime:
`telegram/utils/helpers.py:165: in <module>
    def from_timestamp(unixtime, tzinfo=dtm.timezone.utc):
E   AttributeError: 'module' object has no attribute 'timezone'
`

`pip3`:
`/home/me/.local/lib/python3.6/site-packages/pytest/__init__.py", line 9, in <module>
    from _pytest.config import ExitCode
ImportError: cannot import name 'ExitCode'`

I'll start over with a container.

Bibo-Joshi:
Try following our [Contribution guide](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/.github/CONTRIBUTING.rst) and make sure to run on Py>=3.5. Using a virtual environment to install the packages to will surely be helpful.

ikkemaniac:
_this started as a short reply but turned into my diary trying to get r going, srry_

> Try following our [Contribution guide](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/.github/CONTRIBUTING.rst) and make sure to run on Py>=3.5. Using a virtual environment to install the packages to will surely be helpful.

started from scratch:
```
FROM python:3.6.10-buster

WORKDIR /home/bot/
RUN git clone https://github.com/python-telegram-bot/python-telegram-bot.git --recursive
WORKDIR ./python-telegram-bot
RUN git checkout v12.7
RUN python setup.py install
RUN pip install -r requirements.txt
RUN pip install -r requirements-dev.txt
RUN pip install -r docs/requirements-docs.txt
```
on `requirement-dev.txt`:
```
ERROR: virtualenv 20.0.21 has requirement importlib-resources<2,>=1.0; python_version < "3.7", but you'll have importlib-resources 2.0.0 which is incompatible.
```
Note that pip exits with code 0.... smh

`# make -C docs html` : 4 warnings, no error => ok

`# make test`
```
================================= 1 failed, 1539 passed, 178 skipped, 9 xfailed, 2 xpassed in 155.69 seconds =================================
make: *** [Makefile:32: test] Error 1

```
got an error in game_test but no longer import errors.

importlib-resources fix:
```
# pip check
virtualenv 20.0.21 has requirement importlib-resources<2,>=1.0; python_version < "3.7", but # you have importlib-resources 2.0.0.
# pip uninstall importlib-resources
# pip install importlib-resources==1.5.0
# pip check
No broken requirements found.
```

Let's retry:
`# make -C docs html` : 4 warnings, no error =>  no change

`# make test`
```
===================================== 1539 passed, 178 skipped, 12 xfailed, 2 xpassed in 158.82 seconds ===========================
```

Succes!

lets try test_official:
```
# export TEST_OFFICIAL=true
# make test
================================= 4 failed, 1702 passed, 11 skipped, 9 xfailed, 3 xpassed in 152.48 seconds ==================================
make: *** [Makefile:32: test] Error 1
```
:( but good enough for me.

*ikkemaniac adds Dockerfile to repo
```
FROM python:3.6.10-buster

WORKDIR /home/bot/

RUN apt update
RUN apt install less nano
RUN git clone https://github.com/python-telegram-bot/python-telegram-bot.git --recursive
WORKDIR ./python-telegram-bot
RUN git checkout v12.7
RUN python setup.py install
RUN pip install -r requirements.txt
RUN pip install -r requirements-dev.txt
RUN pip install -r docs/requirements-docs.txt
RUN pip uninstall -y importlib-resources
RUN pip install importlib-resources==1.5.0
```

telegram/ext/updater.py
```diff
@@ -251,11 +255,54 @@ def start_polling(self,
...
 
+        Raises:
+            ValueError: If :attr:`clean` is set as :obj:`datetime.timedelta` and is < 1 second.
+
         """
         with self.__lock:
             if not self.running:
                 self.running = True
 
+                self.logger.debug('clean:               "%s"', type(clean))
```
Bibo-Joshi, line 266: Logging should suffice in the `bootstrap_*` methods

telegram/ext/updater.py
```diff
@@ -251,11 +255,54 @@ def start_polling(self,
...
+                self.logger.debug('clean:               "%s"', clean)
+                self.logger.debug('now:                 "%s"', datetime.now(tz=timezone.utc))
+
+                if isinstance(clean, timedelta):
+                    if clean.total_seconds() < 1:
+                        raise ValueError('Clean as timedelta needs to be >= 1 second')
+                    else:
+                        a=(datetime.now(tz=timezone.utc) - clean)
+                        self.logger.debug('now - delta:         "%s"', a)
```
Bibo-Joshi, line 275: see above

telegram/ext/updater.py
```diff
@@ -251,11 +255,54 @@ def start_polling(self,
...
 
+                self.logger.debug('clean:               "%s"', type(clean))
+                self.logger.debug('clean:               "%s"', clean)
+                self.logger.debug('now:                 "%s"', datetime.now(tz=timezone.utc))
+
+                if isinstance(clean, timedelta):
+                    if clean.total_seconds() < 1:
+                        raise ValueError('Clean as timedelta needs to be >= 1 second')
+                    else:
+                        a=(datetime.now(tz=timezone.utc) - clean)
```
Bibo-Joshi, line 274: Please use descriptive variable names. Also this is computed twice (Line 279)

telegram/ext/updater.py
```diff
@@ -251,11 +255,54 @@ def start_polling(self,
...
         with self.__lock:
             if not self.running:
                 self.running = True
 
+                self.logger.debug('clean:               "%s"', type(clean))
+                self.logger.debug('clean:               "%s"', clean)
+                self.logger.debug('now:                 "%s"', datetime.now(tz=timezone.utc))
+
+                if isinstance(clean, timedelta):
+                    if clean.total_seconds() < 1:
```
Bibo-Joshi, line 271: Maybe consider `abs(clean.total_seconds())` here and subtract the absolute value below. People might not be sure whether they need to pass a positive or a negative timedelta. Also add a note to the doc string stating that the sign doesn't matter.

telegram/ext/updater.py
```diff
@@ -251,11 +255,54 @@ def start_polling(self,
...
+                    if (
+                            clean.tzinfo is None or
+                            (clean.tzinfo is not None and clean.tzinfo.utcoffset(clean) is None)
+                        ):
+                            # we need a tz as msg.date is aware and we want to compare
+                            # no tz passed so we assume it's UTC
+                            self.logger.warning('"clean" passed as datetime object ("%s") but'
+                                                ' has no timezone set, we\'re assuming it\'s UTC!'
+                                                , clean)
+                            clean=clean.replace(tzinfo=timezone.utc)
```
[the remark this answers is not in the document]
Bibo-Joshi: Naive datimetimes should simply be treated as UTC. Don't log a warning for that, just add it to the doc string. See the docstring of `JobQueue` for similar formulations.

telegram/ext/updater.py
```diff
@@ -251,11 +255,54 @@ def start_polling(self,
...
+                        # convert to datetime
+                        clean = datetime.now(tz=timezone.utc) - clean
+                        self.logger.debug('clean as datetime:   "%s"', clean)
+                elif isinstance(clean, datetime):
+
+                    self.logger.debug('clean tz:            "%s"', clean.tzname())
+
+                    if (
+                            clean.tzinfo is None or
+                            (clean.tzinfo is not None and clean.tzinfo.utcoffset(clean) is None)
```
Bibo-Joshi, line 287: What is the second check good for? If the datetime is aware it shouldn't matter what the UTC offset is?

telegram/ext/updater.py
```diff
@@ -475,6 +527,23 @@ def bootstrap_clean_updates():
                 updates = self.bot.get_updates(updates[-1].update_id + 1)
             return False
 
+        def bootstrap_clean_updates_datetime(datetime_cutoff):
+            self.logger.debug('Cleaning updates from Telegram server with datetime "%s"',
+                                    datetime_cutoff)
+            updates = self.bot.get_updates()
+
+            # reversed as we just need to find the first msg that's too old
```
Bibo-Joshi, line 535: Can we be sure that the updates are always ordered? I just don't know tbh.

ikkemaniac:
> Hi. I left you some comments. Some general things I noticed:
> 
>     * Please reduce the logging to one clean entry in the `bootstrap_` method. Something along the line of `Cleaning updates older than {}`

I'll do that as part of the cleanup once y'all accept the basic code/model. I that that for the initial pull as well but had to put it all back to change it up... i don't like rework.

> 
>     * Why did you add the 1 second restriction?

copy paste from the original 'clean as bool' implementation. Didn't know why it was there so just left it in. 

> 
>     * If none of the pending updates has a date, what do we do?

Good point...

telegram/ext/updater.py
ikkemaniac: part of cleanup once y'all are happy with the code itself. I'll wrap it in a comment to make it clearer this is just for debugging while developing

telegram/ext/updater.py
ikkemaniac: see above

telegram/ext/updater.py
ikkemaniac: part of cleanup once y'all are happy with the code itself. I'll wrap it in a comment to make it clearer this is just for debugging while developing

telegram/ext/updater.py
ikkemaniac: I will do because you request it but under formal protest :) imhoo the user should use timedelta correctly and be warned if they don't.

telegram/ext/updater.py
ikkemaniac: per [PY 3.6.10 documentation (scroll down just a little to the datetime.timezone section](https://docs.python.org/3.6/library/datetime.html?highlight=datetime#available-types) the requirement to be 'aware' is to have both conditions. If UTC offset is `None` datetime is naive.

[](url)
> An object of type time or datetime may be naive or aware. A datetime object d is aware if d.tzinfo is not None and d.tzinfo.utcoffset(d) does not return None. If d.tzinfo is None, or if d.tzinfo is not None but d.tzinfo.utcoffset(d) returns None, d is naive. A time object t is aware if t.tzinfo is not None and t.tzinfo.utcoffset(None) does not return None. Otherwise, t is naive.

ikkemaniac: [PY 3.8.3 documentation](https://docs.python.org/3.8/library/datetime.html?highlight=datetime#determining-if-an-object-is-aware-or-naive) has even dedicated a paragraph on it

telegram/ext/updater.py
```diff
@@ -475,6 +527,23 @@ def bootstrap_clean_updates():
...
+            for up in reversed(updates):
+                self.logger.debug('cutoff:              "%s"', datetime_cutoff)
+                self.logger.debug('msg date:            "%s"', up.message.date)
+                self.logger.debug('msg date < cutoff:   "%s"', (up.message.date < datetime_cutoff))
+                if up.message and (up.message.date < datetime_cutoff):
```
Bibo-Joshi, line 540: this needs to be `effective_message` as `channel_post`, and `edited_*` are messages as well

telegram/ext/updater.py
ikkemaniac: :facepalm: assumption is the mother of f..... i got a sequential list every time i ran my tests (and a ran a lot of m... *mumbling something about  dates and time*) 
I'll have look.

Bibo-Joshi:
> I'll do that as part of the cleanup once y'all accept the basic code/model. I that that for the initial pull as well but had to put it all back to change it up... i don't like rework.

It's hard to properly review if one does not know what's going to change again anyway ;)

> copy paste from the original 'clean as bool' implementation. Didn't know why it was there so just left it in.

There is no timing logic right now. `clean` currently is only based on the `update_id`

> Good point...

Personally I'd say to let them through in that case and document that.

telegram/ext/updater.py
ikkemaniac: Good point

ikkemaniac:
> > I'll do that as part of the cleanup once y'all accept the basic code/model. I that that for the initial pull as well but had to put it all back to change it up... i don't like rework.
> 
> It's hard to properly review if one does not know what's going to change again anyway ;)

I fully understand, will wrap everything in a comment block to make it clear.

> 
> > copy paste from the original 'clean as bool' implementation. Didn't know why it was there so just left it in.
> 
> There is no timing logic right now. `clean` currently is only based on the `update_id`

I copied this block of code as basis, this has the `sleep` in it.:
https://github.com/python-telegram-bot/python-telegram-bot/blob/ac7cc7fe5e376e26ce1b38b587b9d4c14a515bee/telegram/ext/updater.py#L506

> 
> > Good point...
> 
> Personally I'd say to let them through in that case and document that.

10-4

Bibo-Joshi:
> I copied this block of code as basis, this has the `sleep` in it.:

AFAICS that's just to wait for the cleaning process is done.

ikkemaniac:
> > I copied this block of code as basis, this has the `sleep` in it.:
> 
> AFAICS that's just to wait for the cleaning process is done.

ok, wanna leave it in for this or remove it?

telegram/ext/updater.py
ikkemaniac: I thought about changing it to
`clean.tzinfo.utcoffset(clean) is None`
but since the documentation is so specific on both requirements I don't think it's a good idea.

telegram/ext/updater.py
ikkemaniac: resolved with 33410d3 and bd21699

telegram/ext/updater.py
ikkemaniac: resolved with cb7bb3b

telegram/ext/updater.py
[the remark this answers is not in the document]
ikkemaniac: resolved with 99dd377

telegram/ext/updater.py
ikkemaniac: - telegram bot API does not mention it is ordered
- i've looked at `get_updates()` but don't see ordering
- i've looked at `de_json` but don't see ordering

Note that the clean as bool part also assumes an ordered by update_id list.

we could do
`updates.sort(key=lambda x: x.update_id, reverse=True)`

maybe even better to put this in `get_updates()` before it returns the list

ikkemaniac:
> > * If none of the pending updates has a date, what do we do?
>
> Personally I'd say to let them through in that case and document that.

resolved in 6c44b67bb92123e6937bab43631d456ef68d461b

telegram/ext/updater.py
Bibo-Joshi: We don't check `utcoffset` in other places, too… If people set non-working stuff as `tzinfo`, they must expect to get errors. Also the refactoring of `JobQueue` will make people use `pytz` with PTB anyway, so they might as well use it here

telegram/ext/updater.py
Bibo-Joshi: Mh … Trusting TG to send updates with incremental `update_id` is okay. They implicity guarantee that by 
>In order to avoid getting duplicate updates, recalculate offset after each server response.
Trusting that the date necessarily increasing with the `update_id` is another thing. I don't see, why it shouldn't, but still … for the moment, just assume it is

Bibo-Joshi:
> ok, wanna leave it in for this or remove it?

Please remove

ikkemaniac:
> > ok, wanna leave it in for this or remove it?
> 
> Please remove

done, 907e092

telegram/ext/updater.py
Bibo-Joshi: Doesn't cover `edited_*`, though

Bibo-Joshi:
Hi again. TBH, I'm having some seconds thoughts about this PR. Let me summarize the problematic situations I see arising:

1. None of the pending updates has a date. We can skip all, handle all or add a second parameter to let the user decide.
2. Updates are not sorted by date. If the last pending update has a date that we want to skip, but all updates before have no date, can we be sure that they must be skipped, too?
3. Updates with and without dates are ordered like: [skip_date, no date, dont_skip_date]. What do we do with the middle update?

1 is simply a decision and 2. is basically overcoming trust issues. But 3. is tricky and I don't see how we could make a clever decision here.
The easiest solution I can currently think of for everything would be to move the time logic in a seperate argument `date_cutoff` or whatever and have it just handle date-updates (i.e. messages). Then `clean` could a) still apply for all other updates b) apply for all other updates, for which we can't make a good decision (i.e. 3.)
Basically, I fear that we might construct a very specific solution for a complex use case, which might lead to it either being not very useful and/or people wanting to add more and more options to it in order to adapt to their wishes.
But maybe you see things differently and/or have a clever solution for the things a mentioned. Anyway, I'd like to hear what you think :)

Apart from that:
* The tests are failing due to some indentation error. Did you run the tests locally as described in the costribution guide?
* What I meant to remove was the logic of rasing errors when time-diff < 1 sec, not the `sleep(1)` ;)
* I don't understand the formulation of the `Note:` part in the docstring of `clean`

ikkemaniac:
> Hi again. TBH, I'm having some seconds thoughts about this PR. Let me summarize the problematic situations I see arising:
> 
>     1. None of the pending updates has a date. We can skip all, handle all or add a second parameter to let the user decide.
> 
>     2. Updates are not sorted by date. If the last pending update has a date that we want to skip, but all updates before have no date, can we be sure that they must be skipped, too?
> 
>     3. Updates with and without dates are ordered like: [skip_date, no date, dont_skip_date]. What do we do with the middle update?
> 
> 
> 1 is simply a decision and 2. is basically overcoming trust issues. But 3. is tricky and I don't see how we could make a clever decision here.
> The easiest solution I can currently think of for everything would be to move the time logic in a seperate argument `date_cutoff` or whatever and have it just handle date-updates (i.e. messages). Then `clean` could a) still apply for all other updates b) apply for all other updates, for which we can't make a good decision (i.e. 3.)
> Basically, I fear that we might construct a very specific solution for a complex use case, which might lead to it either being not very useful and/or people wanting to add more and more options to it in order to adapt to their wishes.
> But maybe you see things differently and/or have a clever solution for the things a mentioned. Anyway, I'd like to hear what you think :)
> 

Good point, didn't consider it as i was just looking at from my personal use case. I'm starting to think you're right and it's too specific or either to loosely implemented if we want to keep it simple. Let me think about it for a couple of days to see if I have a good brain fart.

> Apart from that:
> 
>     * The tests are failing due to some indentation error. Did you run the tests locally as described in the costribution guide?
> 

caught redhanded, missing space. (have to run tests manually in a contianer as sphinx keeps throwing a fit on my dev machine)

>     * What I meant to remove was the logic of rasing errors when time-diff < 1 sec, not the `sleep(1)` ;)
>

haha, 10-4
 
>     * I don't understand the formulation of the `Note:` part in the docstring of `clean`

if clean is defined as either of the date objects, and an effective_message is found w/ date = none while we're looping through updates ('before condition is met' part) all updates will pas through.

ikkemaniac:
> Basically, I fear that we might construct a very specific solution for a complex use case, which might lead to it either being not very useful and/or people wanting to add more and more options to it in order to adapt to their wishes.
> But maybe you see things differently and/or have a clever solution for the things a mentioned. Anyway, I'd like to hear what you think :)
> 

I haven't been able to come up with a all encompassing solution ... i guess you are right, it's too specific and not worth the effort.

I'll close this PR.

Bibo-Joshi:
Okay. Still, thanks for taking the time and making the effort. It's always appreciated :)

PR #1994, opened 2020-06-11
Refactor Bot persistence

### LINKED ISSUE #1992: [BUG] Persistence of Bots
### Steps to reproduce
1. Add an object with a bot attribute to e.g. `chat_data` and persist it

2. Change the bots token and restart the updater

### Expected behaviour
The reloaded bot should have the new token.

### Actual behaviour
The bot in the loaded `chat_data` still has the old token. Similar for the other bot attributes

### Proposed solution
1. Remove `__reduce__` from `Bot` , as we don't want to encourage pickling bots
2. Add static methods `replace_bots(object)` and `set_bots(objects, bot)` (or other naming) to `BasePersistence`, where `replace_bots` implements the logic to replace every `Bot` instance in an object, be it a dict, a list, a custom object … with some special placeholder `BOT_INSTANCE` or whatever and `set_bots` in turn replaces `BOT_INSTANCE` with the given bot. To make `get_*_data` and `update_*_data` use those, we have several options:
    1. Just call them in `get/update_`. Custom subclasses of `BasePersistence` will have to adjust.
    2. Add a `__new__` method to make the `get/update_` methods call the methods under the hood. that way custom subclasses don't need to change anything
    3. Add methods `get/update_*_replace_bot`, which call `set/replace_bots` and before/after `get/update_*`. Use those in `Updater`. This way, custom classes also don't need to be changed.

    don't know, whats better …
3. Alternatively to passing the bot to `set_bots`, we could add `set_bot` to `BasePersistence` so that we can use `self.bot` in `set_bot`. That would save us from having to pass `bot` to `get_*` as well

Bibo-Joshi:
Closes #1992

I went for the `__new__` approach, but that's up for discussion.

Handling arbitrary collections that may contain bots is really not easy. For example, how would one replace a non-mutable `Mapping` subclass that contains a bot with an corresponding object, where the bot is replaced, if one does not know how it was initialized? Therefore I'm currently only handling the built-in types (`(default)dict`, `list`, `tuple`, `(frozen)set`). IMO that should suffice, as the more important use case is a bot as ordinary attribute (which is far easier to handle via `__dict__` and `__slots__`). Still, we'll need a proper documentation for that, if that's approved.


Bibo-Joshi:
py3.7 fail unrelated and codacy has problems with the v13 branch (we'll see any problems in #1982, though). Merging.

PR #1997, opened 2020-06-14
Do not override builtin method help()

OctoNezd:
The echobot2 example overrides builtin method `help()`. This PR renames the `help(update, context)` function to `help_command(update, context)`. 

Perhaps change `start` to `start_command` too?

### DIFF
--- a/examples/echobot2.py
+++ b/examples/echobot2.py
@@ -33,7 +33,7 @@ def start(update, context):
     update.message.reply_text('Hi!')
 
 
-def help(update, context):
+def help_command(update, context):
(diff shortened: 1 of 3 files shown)


Bibo-Joshi:
Personally, I don't see a need to change `start`, as that's not builtin. But would you mind checking the other examples, too?

OctoNezd:
The only other examples that used help() are inllinebot and inlinekeyboard. Both updated now.

Bibo-Joshi:
Thanks for you contribution!

PR #2002, opened 2020-06-17
added test for 'clean' argument passed to 'start_polling()'

ikkemaniac:
Looking for feedback, first attempt at this.

Kinda feels wrong to raise an error to jump out of the 'update' def when there is success.


ikkemaniac:
> Hi. Thanks for the PR.
> TBH, I don't really understand, what's going on. IISC, What we would want to test is
Trying to test clean. Clean will cause all updates to be ignored. It does that by getting all updates and than calling get_updates() with the highest update_id. This results in all updates with a lower id to be discarded.

> 
>     * On the first call of `get_updates()`, that returns a list of updates with ids
I mocked get_updates() to return a list of fake updates when called the 1 first time. The second time it's called we will pass an in update_id from the bootstrap ____clean function. We want to know what id that is and verify it's the 'highest'.

> 
>     * Make sure that for every such id `i`, `get_updates()` is called with `offset=i`

bootstrap___clean doesn't behave like this. It just assumes that the updates are chronological and takes update_id from the last item in the list.

> 
> I don't really get, how `self.update_id` achieves that and what the "infinity loop protection" does 

My logic is that if we pass 3 id's, and we know the highest id = 3 than we can validate clean by checking which update_id it will call get_updates() with the second time.

If get_updates() behaviour changes, it won't be caught with this test. if we want to test for that we shouldn't mock get_updates().

>
> … Could you try to make the code more intuitive to read? Descriptive namings and precise comments are helpful for that ;)
10-4

tests/test_updater.py
```diff
@@ -331,6 +330,56 @@ def attempt(*args, **kwargs):
             updater._bootstrap(retries, False, 'path', None, bootstrap_interval=0)
         assert self.attempts == attempts
 
+    @pytest.mark.parametrize(('error', ),
```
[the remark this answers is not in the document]
ikkemaniac: you're right. We didn't even need to use the error at all. Removed.

tests/test_updater.py
```diff
@@ -331,6 +330,56 @@ def attempt(*args, **kwargs):
...
+            # 2. with 1 arg, int => if int == expected_id => test successful
+
+            # case inf loop protection
+            if self.update_id > 10:
+                raise ValueError
+
+            # case 2
+            if len(args) > 0:
+                self.update_id = int(args[0])
+                raise error
```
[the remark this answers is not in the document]
ikkemaniac: i wasn't thinking... i used it to jump out of the function. We don't need to. See latest updates.

tests/test_updater.py
```diff
@@ -331,6 +330,45 @@ def attempt(*args, **kwargs):
...
+                    self.update_id = update_id
+
+            # case 1
+            # return list of obj's
+
+            # build list of fake updates
+            # returns list of 3 objects with
+            # update_id's 1, 2 and 3
+            updates = []
+            for i in range(1, expected_id):
```
Bibo-Joshi, line 362: Actually, this can be shortened even more with list comprehension:

```python
return [FakeUpdate(i) for i in range(1, expected_id)]
```

tests/test_updater.py
```diff
@@ -331,6 +330,45 @@ def attempt(*args, **kwargs):
...
+            # 1. no args, return list of updates
+            # 2. with 1 arg, int => if int == expected_id => test successful
+
+            # case 2
+            # 2nd call from bootstrap____clean
+            # we should be called with offset = 4
+            # save value passed in self.update_id for assert down below
+            if len(args) > 0:
+                self.offset = int(args[0])
+                return []
```
Bibo-Joshi, line 349: Maybe change is to 

```python
return [FakeUpdate(i) for i in range(args[0], expected_id)]
```

and add `assert bot.get_updates == []` at the very end.

ikkemaniac: TL;DR: changing this line as suggested doesn't work and break current test-strategy. I've pushed commits to make this work but needed to rearrange quite some stuff and store the list of updates in self.updates


if we want to assert bot.get_updates() = [] than we will call the fake get_updates for a 3rd time. It's called w/o an argument so we'll fall through case 2 and go down to case 1 and return a list of fake objects.

If you want to do this we need to store the FakeUpdate list in self.updates outside the fake get_updates, before the first call, to ensure it's available for (every) call to the fake get_updates. If called we will return self.updates with 3 fake updates (1st call). We than also need to have the self.updates list modified when get_updates is called with the offset (2nd call, with arg). After all that we can return self.updates on assert (3rd call).

ikkemaniac: ps: please not that this
`return [FakeUpdate(i) for i in range(args[0], expected_id)]
`
returns [] when args[0] > expected_id. This should raise an error in my opinion

Bibo-Joshi: Okay. I think the `expected_id` test is more important than testing `get_updates() = []` afterwards (as that's actually something TG promises), just thought it would be a minimal-effort addition. Guess I was wrong … Please revert. Then we're good to go :)

ikkemaniac: I think this is it than?

Bibo-Joshi:
Thank you for your contribution :)

PR #2008, opened 2020-06-28
Doc fixes

Bibo-Joshi:
* Update docstring of  `restrict_chat_members` argument `chat_permission` following official API docs
* Update docstring of  `send_invoices` argument `provider_data` following official API docs


telegram/bot.py
```diff
@@ -2871,7 +2871,8 @@ def restrict_chat_member(self, chat_id, user_id, permissions, until_date=None,
                 will be lifted for the user, unix time. If user is restricted for more than 366
                 days or less than 30 seconds from the current time, they are considered to be
                 restricted forever.
-            permissions (:class:`telegram.ChatPermissions`): New user permissions.
+            permissions (:class:`telegram.ChatPermissions`): A JSON-serialized object for new user
```
Poolitzer, line 2874: But uhm, in our context, it would be a JSON- deserialized Telegram class/object. Since we convert it from the string to our own object.

Bibo-Joshi: Just followed the convention to copy official docs. We have a bunch of similar docstrings  …

PR #2009, opened 2020-06-28
feat: via_bot filter

Poolitzer:
also fixing a small mistake in the empty parameter of the user filter (and the test for it) and improved docs slightly

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -887,7 +887,7 @@ officedocument.wordprocessingml.document")``-
     """Messages sent in a group chat."""
 
     class user(BaseFilter):
-        """Filters messages to allow only those which are from specified user ID(s) and/or
+        """Filters messages to allow only those which are from specified user ID(s) or
         username(s).
 
         Examples:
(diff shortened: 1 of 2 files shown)


telegram/ext/filters.py
```diff
@@ -887,7 +887,8 @@ def filter(self, message):
     """Messages sent in a group chat."""
 
     class user(BaseFilter):
-        """Filters messages to allow only those which are from specified user ID.
+        """Filters messages to allow only those which are from specified user ID(s) and/or
```
Bibo-Joshi, line 890: just "or", both are not allowed at the same time

telegram/ext/filters.py
```diff
@@ -919,7 +920,7 @@ class user(BaseFilter):
             RuntimeError: If user_id and username are both present.
 
         """
-
+        # The via_bot filter subclasses this one. Make sure changes here don't destroy it
```
Bibo-Joshi, line 923: See below: Please remove

telegram/ext/filters.py
```diff
@@ -1050,11 +1051,34 @@ def filter(self, message):
                 if self.usernames:
                     return (message.from_user.username
                             and message.from_user.username in self.usernames)
+                return False
```
Bibo-Joshi, line 1054: Please revert as discussed offline

telegram/ext/filters.py
```diff
@@ -1050,11 +1051,34 @@ def filter(self, message):
...
+            else:
+                return self.allow_empty
+
+    class via_bot(user):
+        """Filters messages to allow only those which are from specified via_bot ID(s) and/or
+        username(s).
+
+        Warning:
```
Bibo-Joshi, line 1062: Please copy the code from `Filters.user` and add explicit doc-strings. The method names of `Filters.user` are not really appropriate here, e.g. `remove_user_ids` should be renamed to `remove_bot_ids`  here. Same applies for the variable names.

Bibo-Joshi:
Thanks for the update @Poolitzer!
I did some cosmetic changes. Didn't look at code and tests too thoroughly, as they're mostly copy-paste. Test succeed, merging :)

PR #2017, opened 2020-07-10
Save bot_data on flush()

### LINKED ISSUE #2016: Pickle Persistence Not Saving Bot_data[BUG]
### Steps to reproduce
1.
In line 282 of picklepersistence.py:
[BUG]   if self.user_data or self.chat_data or self.conversations:
[FIX]  if self.user_data or self.chat_data or self.conversations or self.bot_data:

2.
Add self.bot_data in the if-check
3.

### Expected behaviour
When you register the pickle persistence with (store_user_data=False, store_chat_data=False), the bot should save bot_data to pickle file every time it calls flush().

### Actual behaviour
In the current version, the bot only checks user_data/chat_data/conversation to be non-empty to save the pickle file, which when the case (store_user_data=False, store_chat_data=False), the bot never checks bot_data in order to save the pickle file. In my case, user_data/chat_data are always empty, so bot doesn't save to pickle file

### Configuration
**Operating System:**

``$ python -m telegram``

### Logs

Bibo-Joshi:
Closes #2016

### DIFF
--- a/telegram/ext/picklepersistence.py
+++ b/telegram/ext/picklepersistence.py
@@ -277,7 +277,7 @@ class PicklePersistence(BasePersistence):
         """ Will save all data in memory to pickle file(s).
         """
         if self.single_file:
-            if self.user_data or self.chat_data or self.conversations:
+            if self.user_data or self.chat_data or self.bot_data or self.conversations:
                 self.dump_singlefile()
         else:
             if self.user_data:


Bibo-Joshi:
pytest-fails unrelated, test-pre-commit currently doesn't work and I'm going to ignore codecov … Merging

Poolitzer:
@Bibo-Joshi ?Basically: Im just gonna ignore that all CI tests fail

PR #2028, opened 2020-07-19
Remove depreciated RegexHandler from docs

starry69:
Since RegexHandler is depreciated after v12 and we use MessageHandler for that purpose.


Bibo-Joshi:
Hi. The same change was already proposed in #1973. I merged into #2008 where we gather some doc fixes. Still, thanks for the PR d)

starry69:
Oh nice 👍

PR #2032, opened 2020-07-27
Refactor handling of message vs update filters

### LINKED ISSUE #1595: Custom Filters: Wiki and code are in conflict

Bibo-Joshi:
As mentioned offline:

`update_filter` is a bit unsatisfactory to me. E.g. it leads to irritation with users (see #1595) and makes typing hard (see #1920)

What this PR does is:

* Make `BaseFilter.__call__` an abstract method and implement two subclasses: `MessageFilter` and `UpdateFilter`, each of which will take care of passing the right object to `filter()`
* instead of setting `update_filter`, make the filters inherit from those.
* Also move the abstract filter() method from `BaseFilter` to `Message/UpdateFilter` in order to make typing even more clear. This makes it breaking, but 
  * v13 is breaking anyway
  * switching inheritance from `BaseFilter` to `MessageHandler` is not too bad. Leaving `update_filter = True` won't even hurt …

Offline I also mentioned dropping `data_filter`, but that was nonsense, as we need that for the short circuit logic of the filters

Closes #1595  and #1596

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -32,7 +32,7 @@ __all__ = ['Filters', 'BaseFilter', 'MessageFilter', 'UpdateFilter', 'InvertedFi
 class BaseFilter(ABC):
     """Base class for all Filters.
 
-    Subclassing from this class filters to be combined using bitwise operators:
+    Filters subclassing from this class can combined using bitwise operators:
 
     And:
 
@@ -84,7 +84,7 @@ class BaseFilter(ABC):
 
     @abstractmethod
     def __call__(self, update):
-        ...
+        pass
 
     def __and__(self, other):
         return MergedFilter(self, and_filter=other)
@@ -105,6 +105,8 @@ class BaseFilter(ABC):
 class MessageFilter(BaseFilter, ABC):
     """Base class for all Message Filters. In contrast to :class:`UpdateFilter`, the object passed
     to :meth:`filter` is ``update.effective_message``.
+
+    Please see :class:`telegram.ext.BaseFilter` for details on how to create custom filters.
 
     Attributes:
         name (:obj:`str`): Name for this filter. Defaults to the type of filter.
@@ -134,6 +136,8 @@ class UpdateFilter(BaseFilter, ABC):
     """Base class for all Update Filters. In contrast to :class:`UpdateFilter`, the object
     passed to :meth:`filter` is ``update``, which allows to create filters like
     :attr:`Filters.update.edited_message`.
+
+    Please see :class:`telegram.ext.BaseFilter` for details on how to create custom filters.
 
     Attributes:
         name (:obj:`str`): Name for this filter. Defaults to the type of filter.
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -25,11 +25,12 @@ from threading import Lock
 
 from telegram import Chat, Update, MessageEntity
 
-__all__ = ['Filters', 'BaseFilter', 'InvertedFilter', 'MergedFilter']
+__all__ = ['Filters', 'BaseFilter', 'MessageFilter', 'UpdateFilter', 'InvertedFilter',
+           'MergedFilter']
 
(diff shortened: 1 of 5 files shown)


telegram/ext/filters.py
```diff
@@ -56,38 +57,34 @@ class BaseFilter(ABC):
...
-    update_filter = False
     data_filter = False
 
+    @abstractmethod
     def __call__(self, update):
-        if self.update_filter:
-            return self.filter(update)
-        else:
-            return self.filter(update.effective_message)
+        ...
```
jsmnbom, line 87: Please use `pass` here instead

telegram/ext/filters.py
```diff
@@ -104,13 +101,54 @@ def __repr__(self):
             self.name = self.__class__.__name__
         return self.name
 
+
+class MessageFilter(BaseFilter, ABC):
+    """Base class for all Message Filters. In contrast to :class:`UpdateFilter`, the object passed
+    to :meth:`filter` is ``update.effective_message``.
```
jsmnbom, line 107: Could you maybe add the "If you want to create your own filters create a class inheriting..." text to the MessageFilter and UpdateFilter docstrings too?
Or maybe just "See BaseFilter for details on how to create your own custom filters" or similar

tests/test_filters.py
```diff
@@ -963,7 +963,7 @@ class _CustomFilter(BaseFilter):
             _CustomFilter()
 
     def test_custom_unnamed_filter(self, update):
-        class Unnamed(BaseFilter):
+        class Unnamed(MessageFilter):
```
jsmnbom, line 966: We should probably test that UpdateFilter works as desired too?

telegram/ext/filters.py
```diff
@@ -25,11 +25,12 @@
...
-__all__ = ['Filters', 'BaseFilter', 'InvertedFilter', 'MergedFilter']
+__all__ = ['Filters', 'BaseFilter', 'MessageFilter', 'UpdateFilter', 'InvertedFilter',
+           'MergedFilter']
 
 
 class BaseFilter(ABC):
-    """Base class for all Message Filters.
+    """Base class for all Filters.
 
     Subclassing from this class filters to be combined using bitwise operators:
```
jsmnbom, line 35: I know you didn't change this line, but I just realised it makes like zero sense. Could you maybe fix that? :P

telegram/ext/filters.py
Bibo-Joshi: `...` seems to be standart for abstract methods (according to py docs), but I changed it

telegram/ext/filters.py
Bibo-Joshi: jup

tests/test_filters.py
Bibo-Joshi: Testing with both via fixture, as well in `conftest.py`. Don't know if my fixture is the best … but it works …

telegram/ext/filters.py
Bibo-Joshi: Upsie, didn't even see that :D. Fixed

telegram/ext/filters.py
jsmnbom: ... is just a placeholdder in case you want to put code like "raise NotImplementedError" or "pass" or similar as far as i know.

tests/test_filters.py
jsmnbom: Yea this seems pretty good for now.
We should probably be testing each individual filter, but this is good enough for now :)

PR #2042, opened 2020-08-04
Defaults.tzinfo

### LINKED ISSUE #1797: [FEATURE] Extensions of Defaults

Bibo-Joshi:
Adds `Defaults.tzinfo` so that all naive date/time inputs are interpreted as a specified timezone.
Only handles `pytz`-timezones, as that's pretty much the standard-lib for that and APScheduler also only handles `pytz`.

Closes #1797


telegram/chatmember.py
```diff
@@ -123,7 +123,8 @@ def __init__(self, user, status, until_date=None, can_be_edited=None,
                  can_restrict_members=None, can_pin_messages=None,
                  can_promote_members=None, can_send_messages=None,
                  can_send_media_messages=None, can_send_polls=None, can_send_other_messages=None,
-                 can_add_web_page_previews=None, is_member=None, custom_title=None, **kwargs):
+                 can_add_web_page_previews=None, is_member=None, custom_title=None,
+                 **kwargs):
```
Poolitzer, line 127: why does kwargs get its own line here? Is that a general custom we follow? If yes, why in this PR? You just saw it?

telegram/ext/filters.py
```diff
@@ -62,11 +62,9 @@ class BaseFilter(ABC):
...
 
     If you want to create your own filters create a class inheriting from either
-    :class:`MessageFilter` or :class:`UpdateFilter` and implement a :meth:``filter`` method that
-    returns a boolean: :obj:`True` if the message should be
-    handled, :obj:`False` otherwise.
-    Note that the filters work only as class instances, not
-    actual class objects (so remember to
+    :class:`MessageFilter` or :class:`UpdateFilter` and implement a :meth:`filter` method that
+    returns a boolean: :obj:`True` if the message should be handled, :obj:`False` otherwise.
+    Note that the filters work only as class instances, not actual class objects (so remember to
```
Poolitzer, line 67: also, why in this PR?

Poolitzer (approved):
Looks good to me!

PR #2043, opened 2020-08-04
Add missing shortcuts

### LINKED ISSUE #2034: [FEATURE] Add missing shortcuts
There are some shortcuts missing in `Message`, `CallbackQuery`, `User` and `Chat`. I noticed:

* `send/stop_message_live_location`
* `send_game`
* `set_game_score`
* `get_game_high_scores`

We should check that

* both `User` and `Chat` have shortcuts for all `send_*` methods
* both `Message` and `CallbackQuery` have shortcuts for all methods that can accept `inline_message_id`

Bibo-Joshi:
Turns out, there where a lot more than I expected. 

`Chat` already has a method `send_action`. I added `send_chat_action` for completeness and both to `User`. 
**TL;DR:** Both. Both is good.
I guess the logic was to skip the `chat` similarly to `Message.edit_text` as shortcut for `Bot.edit_message_text`, but the `chat` in `send_chat_action` acutally comes from the class name `ChatAction` … But having both doesn't hurt.

Also, one *could* think of adding shortcuts to `ChosenInlineResult`, which has an `inline_message_id` attribute, but IMHO that doesn't really make sense. It would only be needed, if one would edit the message right after it's posted (why, though?). In all other cases one usually has only the `inline_message_id` and not the `CIR`.

Closes #2034

### DIFF
--- a/telegram/callbackquery.py
+++ b/telegram/callbackquery.py
@@ -175,10 +175,10 @@ class CallbackQuery(TelegramObject):
     def edit_message_reply_markup(self, reply_markup, *args, **kwargs):
         """Shortcut for either::
 
-            bot.edit_message_replyMarkup(chat_id=update.callback_query.message.chat_id,
-                                       message_id=update.callback_query.message.message_id,
-                                       reply_markup=reply_markup,
-                                       *args, **kwargs)
+            bot.edit_message_reply_markup(chat_id=update.callback_query.message.chat_id,
+                                          message_id=update.callback_query.message.message_id,
+                                          reply_markup=reply_markup,
+                                          *args, **kwargs)
 
         or::
 
@@ -204,10 +204,10 @@ class CallbackQuery(TelegramObject):
     def edit_message_media(self, *args, **kwargs):
         """Shortcut for either::
 
-            bot.edit_message_replyMarkup(chat_id=update.callback_query.message.chat_id,
-                                         message_id=update.callback_query.message.message_id,
-                                         media=media,
-                                         *args, **kwargs)
+            bot.edit_message_media(chat_id=update.callback_query.message.chat_id,
+                                   message_id=update.callback_query.message.message_id,
+                                   media=media,
+                                   *args, **kwargs)
 
         or::
 
--- a/telegram/callbackquery.py
+++ b/telegram/callbackquery.py
@@ -309,19 +309,19 @@ class CallbackQuery(TelegramObject):
                                            message_id=self.message.message_id,
                                            *args, **kwargs)
 
-    def get_game_high_score(self, *args, **kwargs):
+    def get_game_high_scores(self, *args, **kwargs):
         """Shortcut for either::
 
-            bot.get_game_high_score(chat_id=update.callback_query.message.chat_id,
-                                    message_id=update.callback_query.message.message_id,
-                                    reply_markup=reply_markup,
-                                    *args, **kwargs)
+            bot.get_game_high_scores(chat_id=update.callback_query.message.chat_id,
+                                     message_id=update.callback_query.message.message_id,
+                                     reply_markup=reply_markup,
+                                     *args, **kwargs)
 
         or::
 
-            bot.get_game_high_score(inline_message_id=update.callback_query.inline_message_id,
-                                    reply_markup=reply_markup,
-                                    *args, **kwargs)
+            bot.get_game_high_scores(inline_message_id=update.callback_query.inline_message_id,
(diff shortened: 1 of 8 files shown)


Poolitzer:
> It would only be needed, if one would edit the message right after it's posted (why, though?)

Agree with that

telegram/callbackquery.py
```diff
@@ -200,3 +200,142 @@ def edit_message_reply_markup(self, reply_markup, *args, **kwargs):
                                                       chat_id=self.message.chat_id,
                                                       message_id=self.message.message_id,
                                                       *args, **kwargs)
+
+    def edit_message_media(self, *args, **kwargs):
+        """Shortcut for either::
+
+            bot.edit_message_replyMarkup(chat_id=update.callback_query.message.chat_id,
```
Poolitzer, line 207:
```suggestion
            bot.edit_message_media(chat_id=update.callback_query.message.chat_id,
```

:)

telegram/chat.py
```diff
@@ -270,6 +258,35 @@ def send_message(self, *args, **kwargs):
         """
         return self.bot.send_message(self.id, *args, **kwargs)
 
+    def send_media_group(self, *args, **kwargs):
+        """Shortcut for::
+
+            bot.send_media_group(Chat.id, *args, **kwargs)
+
+        Where Chat is the current instance.
```
Poolitzer, line 266: All older docstrings use `update.message.chat.id` instead of `Chat.id`. We should decide which one we want to use and stick with through the whole file.

telegram/message.py
```diff
@@ -935,6 +935,86 @@ def edit_reply_markup(self, *args, **kwargs):
...
+
+            bot.edit_message_live_location(chat_id=message.chat_id,
+                                           message_id=message.message_id,
+                                           *args,
+                                           **kwargs)
+
+        Note:
+            You can only edit messages that the bot sent itself (i.e. of the ``bot.send_*`` family
+            of methods) or channel posts, if the bot is an admin in that channel. However, this
+            behaviour is undocumented and might be changed by Telegram.
```
Poolitzer, line 949: While talking about undocumented behaviour: Only one bot can edit a message in a channel, no matter the sender. If a second bot tries it after that, it gets a Bad Request.

Probably too specific for this note, just wanted to mention it.

telegram/user.py
```diff
@@ -206,6 +206,19 @@ def send_photo(self, *args, **kwargs):
         """
         return self.bot.send_photo(self.id, *args, **kwargs)
 
+    def send_media_group(self, *args, **kwargs):
+        """Shortcut for::
+
+            bot.send_media_group(User.id, *args, **kwargs)
+
+        Where User is the current instance.
```
Poolitzer, line 214: Same as with chat, here `update.message.from_user.id` is used

telegram/chat.py
Bibo-Joshi: Thanks for the catch! I went for `update.*` as that's the most common use case and imo with that it's easier to understand what the shortcut does for newbies. However, I changed to `update.effective_chat.id` b/c that should be used in the manual way, too, imho.

telegram/message.py
Bibo-Joshi: Good to know! But I agree, that it's a bit too specific for this note …

telegram/user.py
Bibo-Joshi: Same as above

telegram/chat.py
Poolitzer: best way imo as well

telegram/chat.py
```diff
@@ -364,7 +459,7 @@ def send_video_note(self, *args, **kwargs):
     def send_voice(self, *args, **kwargs):
         """Shortcut for::
 
-            bot.send_voice(Chat.id, *args, **kwargs)
+            bot.send_voice(update.effective_chat.id, *args, **kwargs)
 
         Where Chat is the current instance.
```
harshil21, line 464: 'Where Chat is the current instance' should be removed everywhere since it was changed to `update.effective_chat.id` instead of `Chat.id` right?

harshil21: Same thing in `user.py` too

Bibo-Joshi: Yeah, maybe you're right :) If I remember, I'll add it in #2008. If you want to, you can also PR for that. Just leave a comment here in that case, so we don't do the work twice

Bibo-Joshi: Edit: Done

PR #2047, opened 2020-08-12
pollbot.py example was sending the poll to the effective_user and not…

lrezende:
It's about pollbot.py example. The poll command was sending the poll to the update.effective_user.id... I think it should be update.effective_chat.id so multiple chat members could answer the poll.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -56,6 +56,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Kjwon15 <https://github.com/kjwon15>`_
 - `Li-aung Yip <https://github.com/LiaungYip>`_
 - `Loo Zheng Yuan <https://github.com/loozhengyuan>`_
+- `LRezende <https://github.com/lrezende>`_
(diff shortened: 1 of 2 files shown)


Poolitzer:
Well the answers are only received on a user basis, not chat one, so the "answer" will be send in a private chat anyway. I dont really see a big gain of your PR but I dont mind it either

lrezende:
I think the pollbot example was meant to have some answers in the same poll_id...
```
# Close poll after three participants voted
if context.bot_data[poll_id]["answers"] == 3:
```
If you send the poll to a private chat(user) it will never receive more than one answer.
Correct me if I'm wrong. I'm just starting with ptb and open source contribution.

Poolitzer:
No yeah I guess you are right, never thought about running it in a group, but that change is OK.

Bibo-Joshi:
pytest fails obviously unrelated and we can ignore the examples warning, as there is nothing new in the example. Merging.
Thanks for the contribution @lrezende !

PR #785, opened 2017-08-09
broadcast telegram errors from run_async functions

jh0ker:
fixes #682 
Not sure if this is really the way to do it, IIRC there were some discussions about this in the dev group. Perhaps @Eldinnie can recall more details?

PR #2051, opened 2020-08-18
Refactor @run_async

### LINKED ISSUE #682: Error handlers not working with @run_async
### Steps to reproduce
1. Register any handler, for ex: `CommandHandler`;
2. Write some code to get `TelegramException` (ex: invalid Markdown/HTML);
3. Register error handler;
4. Run it.

Example code:

```python
import logging

from telegram import Update, Bot, ParseMode
from telegram.ext import Updater, CommandHandler
from telegram.ext.dispatcher import run_async

updater = Updater('<telegram_token>')

logging.basicConfig(
    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
    level=logging.INFO
)

dispatcher = updater.dispatcher

msg = "<invalid_tag>Bold line"


def action_start(bot: Bot, update: Update):
    logger.info('/start')
    update.message.reply_text(msg, parse_mode=ParseMode.HTML)


@run_async
def action_async_start(bot: Bot, update: Update):
    logger.info('/async_start')
    update.message.reply_text(msg, parse_mode=ParseMode.HTML)


def _error(bot: Bot, update: Update, e: BaseException):
    logger.error("Boom!")


dp = updater.dispatcher

dp.add_handler(CommandHandler('start', action_start))
dp.add_handler(CommandHandler('async_start', action_async_start))

dp.add_error_handler(_error)

updater.start_polling()
updater.idle()

```

### Expected behavior
Error handler will catch `TelegramException`, so handler could manage it.

### Actual behavior
Error handler is not called.

### Configuration
```
python-telegram-bot 6.0.2
urllib3 1.21.1
certifi 2017.04.17
future 0.16.0
Python 3.5.2 (v3.5.2:4def2a2901a5, Jun 26 2016, 10:47:25)  [GCC 4.2.1 (Apple Inc. build 5666) (dot 3)]
```

### Logs
Commands to send:
```
/start
/async_start
```

Output:
```
<project_path>/venv/bin/python <project_path>/issue.py
2017-06-20 23:24:11,743 - __main__ - INFO - /start
2017-06-20 23:24:11,906 - telegram.ext.dispatcher - WARNING - A TelegramError was raised while processing the Update.
2017-06-20 23:24:11,907 - __main__ - ERROR - Boom!
2017-06-20 23:24:11,907 - __main__ - INFO - /async_start
2017-06-20 23:24:12,188 - telegram.utils.promise - ERROR - An uncaught error was raised while running the promise
  File "<project_path>/venv/src/python-telegram-bot/telegram/utils/promise.py", line 42, in run
    self._result = self.pooled_function(*self.args, **self.kwargs)
  File "<project_path>/issue.py", line 25, in action_async_start
    update.message.reply_text(msg, parse_mode=ParseMode.HTML)
  File "<project_path>/venv/src/python-telegram-bot/telegram/message.py", line 340, in reply_text
    return self.bot.send_message(self.chat_id, *args, **kwargs)
  File "<project_path>/venv/src/python-telegram-bot/telegram/bot.py", line 125, in decorator
  File "<project_path>/venv/src/python-telegram-bot/telegram/bot.py", line 158, in decorator
    return Bot._message_wrapper(self, url, data, *args, **kwargs)
  File "<project_path>/venv/src/python-telegram-bot/telegram/bot.py", line 146, in _message_wrapper
    result = self._request.post(url, data, timeout=kwargs.get('timeout'))
  File "<project_path>/venv/src/python-telegram-bot/telegram/utils/request.py", line 252, in post
    **urlopen_kwargs)
  File "<project_path>/venv/src/python-telegram-bot/telegram/utils/request.py", line 194, in _request_wrapper
    raise BadRequest(message)
telegram.error.BadRequest: Can't parse entities in message text: unsupported start tag "invalid_tag" at byte offset 0

Process finished with exit code 0
```

I suppose, `dispatcher.dispatch_error` should be called in the same thread with async handler.

Bibo-Joshi:
* Deprecates the @run_async decorator
* Instead adds kwargs `update` and `error_handler` to `DP.run_async()` to allow proper error handling
* Allows to specify that a callback should be run asynchronously via `DP.add_handler`/`DP.add_error_handler`

One problem: In the new setting it's currently not possible to make the handlers of a `ConversationHandler` be run asynchronously. Possible solutions:
* When adding a CH via `DP.add_handler(CH, run_asyn=True)` make *all* handlers of the Conversation async
* Set the `run_async` flag as parameter for `Handler` instead of as parameter for `DP.add_handler`
* Somehow build an option for that into CH

Wanted to have that clarified before going for the tests, so WIP.

When ready, closes #682 and superseeds/closes #785


Bibo-Joshi:
I made to additional changes to `Dispatcher.update_persistence`:

1. Use a lock to ensure that the method is not called twice simultanuously
2. convert `dispatcher.user/chat_data.keys()` to lists to prevent errors like [this](https://t.me/pythontelegrambotgroup/363178). That particular one was probably caused by `JobQueue` calling `Dispatcher.update_persistence` while an entry was added to `dispatcher.user_data`, but similar situations can probably also arise when calling `Dispatcher.update_persistence` from the async threads. We don't acutally have to worry about the new entry not being persisted, as `update_persistence()` will be called for the update that lead to the new entry anyway.

GauthamramRavichandran:
Could this `async` be added to job callback as well? I'm using `@run_async` decorator to run jobs in parallel for now

telegram/ext/dispatcher.py
```diff
@@ -448,57 +512,58 @@ def update_persistence(self, update=None):
...
-                        message = 'Saving bot data raised an error and an ' \
-                                  'uncaught error was raised while handling ' \
-                                  'the error with an error_handler'
-                        self.logger.exception(message)
-            if self.persistence.store_chat_data:
-                for chat_id in chat_ids:
+        with self._update_persistence_lock:
+            if self.persistence:
+                chat_ids = list(self.chat_data.keys())
+                user_ids = list(self.user_data.keys())
```
tsnoam, line 518: What is this `list()` change for?

telegram/ext/dispatcher.py
```diff
@@ -512,11 +577,13 @@ def add_error_handler(self, callback):
...
 
                 The error that happened will be present in context.error.
+            run_async (:obj:`bool`, optional): Whether this handlers callback should be run
+                asynchronously using :meth:`run_async`. Defaults to :obj:`False`.
 
         Note:
             See https://git.io/fxJuV for more info about switching to context based API.
         """
-        self.error_handlers.append(callback)
+        self.error_handlers[callback] = run_async
```
tsnoam, line 586: Side effects of using a dict here are, both are changes in behaviour:
1. No guarantees regarding the order of error_handlers (unless it's python3.7 or even 3.8)
2. If the same handler is inserted twice, it will only be used once.

tsnoam (changes requested):

telegram/utils/promise.py
```diff
@@ -41,10 +41,12 @@ class Promise:
 
     """
 
-    def __init__(self, pooled_function, args, kwargs):
+    def __init__(self, pooled_function, args, kwargs, update=None, error_handler=None):
```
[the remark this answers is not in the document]
Bibo-Joshi: Added

telegram/utils/promise.py
```diff
@@ -56,7 +58,6 @@ def run(self):
             self._result = self.pooled_function(*self.args, **self.kwargs)
 
         except Exception as exc:
-            logger.exception('An uncaught error was raised while running the promise')
```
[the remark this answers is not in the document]
Bibo-Joshi: That's covered by this [clause](https://github.com/python-telegram-bot/python-telegram-bot/blob/75447e1c13a8beee8420d13ef999c08ffae5fe0e/telegram/ext/dispatcher.py#L270).

telegram/ext/dispatcher.py
```diff
@@ -164,8 +170,9 @@ def __init__(self,
         """Dict[:obj:`int`, List[:class:`telegram.ext.Handler`]]: Holds the handlers per group."""
         self.groups = []
         """List[:obj:`int`]: A list with all groups."""
-        self.error_handlers = []
-        """List[:obj:`callable`]: A list of errorHandlers."""
+        self.error_handlers = {}
+        """Dict[:obj:`callable`, :obj:`bool`]: A dict, where the keys are error handlers and the
```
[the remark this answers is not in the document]
Bibo-Joshi: We would need to rebase this on v13, then  … but if I understand correctly that would also mean that we'd need to document `IsAsync` and I think that's more confusing for a user than just reading the docstring. Or am I misunderstanding?

telegram/ext/dispatcher.py
```diff
@@ -229,30 +236,84 @@ def _pooled(self):
...
+
             if isinstance(promise.exception, DispatcherHandlerStop):
                 self.logger.warning(
                     'DispatcherHandlerStop is not supported with async functions; func: %s',
                     promise.pooled_function.__name__)
+            elif isinstance(promise.exception, Exception):
+                if promise.pooled_function in self.error_handlers:
+                    self.logger.error('An error was raised while processing the update '
+                                      'and an uncaught error was raised while handling '
+                                      'the error with an error_handler')
```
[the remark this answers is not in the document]
Bibo-Joshi: `return` statement: good idea. wording: I just used the wording that we use everywhere … Should I change everywhere?

telegram/ext/dispatcher.py
```diff
@@ -229,30 +236,84 @@ def _pooled(self):
...
         Warning:
-            If you're using @run_async you cannot rely on adding custom attributes to
-            :class:`telegram.ext.CallbackContext`. See its docs for more info.
+            * If you're using ``@run_async``/:meth:`run_async` you cannot rely on adding custom
+              attributes to :class:`telegram.ext.CallbackContext`. See its docs for more info.
+            * When using :meth:`run_async` within an asynchronously run error handler (see
+              :meth:`add_error_handler`) and ``update`` is passed, you *must* also pass
+              ``error_handler``, as an exception within ``func`` would cause an infinite loop
+              otherwise.
+            * passing ``error_handler`` is not supported for the old style API, i.e.
```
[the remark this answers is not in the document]
Bibo-Joshi: https://t.me/c/1494805131/6006

telegram/ext/dispatcher.py
```diff
@@ -448,57 +512,58 @@ def update_persistence(self, update=None):
...
-                    try:
-                        self.dispatch_error(update, e)
-                    except Exception:
-                        message = 'Saving bot data raised an error and an ' \
-                                  'uncaught error was raised while handling ' \
-                                  'the error with an error_handler'
-                        self.logger.exception(message)
-            if self.persistence.store_chat_data:
-                for chat_id in chat_ids:
+        with self._update_persistence_lock:
```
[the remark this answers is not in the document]
Bibo-Joshi: Great tip! Thanks for that, I adopted it :)

telegram/ext/dispatcher.py
Bibo-Joshi: KeyViews or whatever it's called change, when the dict changes, which is inconvenient here, as we're going to loop over the keys. If we receive an update from an unknown user/chat while doing so, `chat_ids` will change and we get an error like [this](https://t.me/pythontelegrambotgroup/363178) one.
The use of `list()` decouples `chat_ids` from `self.chat_data` and just considers the keys present at call-time. The new entry is ignored and will be handled in the next call to `update_persistence`, which the update from the unknown user triggers.

telegram/ext/dispatcher.py
Bibo-Joshi: I could change to a list of tuples (error_handler, run_async) …

telegram/ext/dispatcher.py
```diff
@@ -229,30 +236,84 @@ def _pooled(self):
...
+                    except Exception:
+                        self.logger.exception('An error was raised while processing the update '
+                                              'and an uncaught error was raised while handling '
+                                              'the error with an error_handler')
+                else:
+                    self.logger.exception(
+                        'No error handlers are set for async function, logging exception.',
+                        exc_info=promise.exception)
+            else:
+                self.update_persistence(update=promise.update)
```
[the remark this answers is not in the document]
Bibo-Joshi: That is so much better than the current state o.0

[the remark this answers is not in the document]
Bibo-Joshi: Uh, except we musn't use `return` as it will end the `while 1` loop. I used continue instead

telegram/ext/dispatcher.py
```diff
@@ -229,30 +236,84 @@ def _pooled(self):
...
             if isinstance(promise.exception, DispatcherHandlerStop):
                 self.logger.warning(
                     'DispatcherHandlerStop is not supported with async functions; func: %s',
                     promise.pooled_function.__name__)
+            elif isinstance(promise.exception, Exception):
+                if promise.pooled_function in self.error_handlers:
+                    self.logger.error('An error was raised while processing the update '
+                                      'and an uncaught error was raised while handling '
+                                      'the error with an error_handler')
+                elif promise.error_handler:
```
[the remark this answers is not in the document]
Bibo-Joshi: > 1. Why is there a `promise.error_handler`: when is it used and why?

`run_async` can not only be used for running handler callbacks asyncronously, but also custom functions. Custom functions will usually have a other arguments than `update, context`, so it might make sense to have special error handlers form them. So the use case is

```python
def custom_func(a, b, c=None):
   raise RuntimeError()

def error_handler(update, context, a, b, c=None):
   print(f'The error {context.error} happend while running custom_func with the arguments {a}, {b}, {c} for update {update}')

    # do some stuff
    context.dispatcher.run_async(custom_func, a, b, update=update, error_handler=error_handler, c=c)
```

> 2. What's the difference between having `promise.update` and not having one?

I was gonna argue about distinguishing handler callbacks from custom functions but acutally I see that it's not really straightforward. I think it would be good to restructure and I have two options in mind:

1. Still allow to set `error_handler`, which can be helpful for custom functions. But the signature can just always be 
    ```python
    def error_handler(update, context, *args, **kwargs):
      ...
    ```
    in case `update` is not passed to `run_async`, it will just be `None` (as is already the case for the usual error handlers called on errors happening in jobs)
    If no `error_handler` is passed, use the dispatchers usual error handlers with the downside of not having access to the parameters of the function that raised the error
2. Remove `error_handler` all together and just use the usual error_handlers. To allow acces to the pooled functions parameters, we could add them to the `context` argument (`context.error_(kw)args`). Maybe even the complete promise (`context.promise`)

Looking at the changes with a bit of distance, I'd acutally go for 2.

telegram/ext/dispatcher.py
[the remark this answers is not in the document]
tsnoam: What about only supporting error handlers which are context aware with `run_async` ?

telegram/ext/dispatcher.py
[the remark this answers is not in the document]
tsnoam: maybe i was shooting too far.
got used to the good life of Rust.

you can ignore that suggestion.

telegram/ext/dispatcher.py
tsnoam: Yes, that's a good idea, just make sure that `remove_error_handler` will provide an expectable semantics (especially as now the caller would need to know what is the `run_async` value).

A twist on your proposal would be (not sure if it really improves):
Wrap the callback into our own object and make it `__eq__` and `__hash__` only on the callback. Something like:
```python
class _PtbErrorHandler:

    def __init__(self, error_handler: Callable, run_async: bool):
       self._cb = error_handler
       self._is_async = run_async

    def __hash__(self):
       return hash(self._cb)

    def __eq__(self, other):
       return self._cb == other._cb

    # And we can even take it a step further (totally optional):
    def __call__(self, *args, **kwargs):
        if self._is_async:
            async_execution(self._cb, *args, **kwargs)
        else:
            self._cb(*args, **kwargs)
```

That might help other parts of the code seamlessly handle the changes of how we internally keep the callback.

telegram/ext/dispatcher.py
tsnoam: Excellent, please add it as a comment on the code.

telegram/ext/dispatcher.py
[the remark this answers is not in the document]
Bibo-Joshi: that's what "passing ``error_handler`` is not supported for the old style API" means?

telegram/ext/dispatcher.py
Bibo-Joshi: Mh … If we go that route, we might as well set up a full blown handler logic for error handlers … Currently, `remove(error_handler)` will just remove the first occurrence of `error_handler` from the list. I'd just do the same for the duples, ignoring the `run_async` part (ofc documenting that). IMHO, if someone registers the same error_handler twice or worse, once synchronously and once asynchronously, they don't deserve us special casing for that …

telegram/ext/dispatcher.py
```diff
@@ -229,30 +236,89 @@ def _pooled(self):
...
             if isinstance(promise.exception, DispatcherHandlerStop):
                 self.logger.warning(
                     'DispatcherHandlerStop is not supported with async functions; func: %s',
                     promise.pooled_function.__name__)
+                continue
+
+            if promise.pooled_function in self.error_handlers:
+                self.logger.error('An error was raised while processing the update '
+                                  'and an uncaught error was raised while handling '
+                                  'the error with an error_handler')
```
tsnoam, line 253: @Bibo-Joshi quoting you (as the original discussion is marked "Outdated":
> return statement: good idea. wording: I just used the wording that we use everywhere, i.e. also in process_update and in JobQueue … Should I change everywhere?

---
In this context (and all the locations it is repeated down in this function) I find the error text to be too long. I prefer the shortest sentence possible which pass the idea.

In this case saying:
`An uncaught error was raised by an error_handler` is still 100% correct and pass the idea.
It's obvious that the error handler wouldn't be called unless there was an error, right?
So why explicitly say that it happened? It just clutters the log (of the user) and makes our code unreadable.

```suggestion
                self.logger.error('An uncaught error was raised by an error_handler')
```

Regarding changing the log text in other places, it doesn't have to happen now. Make sure that the log you issue now makes sense, and if you want open a new issue to fix other logs in the future.

telegram/ext/dispatcher.py
```diff
@@ -229,30 +236,89 @@ def _pooled(self):
...
+                    if promise.update:
+                        promise.error_handler(promise.update, context, *promise.args,
+                                              **promise.kwargs)
+                    else:
+                        promise.error_handler(context, *promise.args, **promise.kwargs)
+                except Exception:
+                    self.logger.exception('An error was raised while processing the update '
+                                          'and an uncaught error was raised while handling '
+                                          'the error with an error_handler')
+            elif promise.update:
```
tsnoam, line 269:
```suggestion
                continue
                
            if promise.update:
```

telegram/ext/dispatcher.py
```diff
@@ -229,30 +236,89 @@ def _pooled(self):
...
+                try:
+                    self.dispatch_error(promise.update, promise.exception)
+                except Exception:
+                    self.logger.exception('An error was raised while processing the update '
+                                          'and an uncaught error was raised while handling '
+                                          'the error with an error_handler')
+            else:
+                self.logger.exception(
+                    'No error handlers are set for async function, logging exception.',
+                    exc_info=promise.exception)
```
tsnoam, line 279: 1. Use `continue`.
2. Shorter log message.

```suggestion
                continue
                
            self.logger.exception('Caught unhandled exception from an async function',
                                  exc_info=promise.exception)
```

telegram/ext/dispatcher.py
```diff
@@ -229,30 +236,89 @@ def _pooled(self):
                 break
 
             promise.run()
```
tsnoam, line 238: @Bibo-Joshi,  you wrote quite a nice explanation on one of the threads, but it is now outdated, so can't continue the discussion there.
The exaplanation related to:


1. Why is there a promise.error_handler: when is it used and why?
2. What's the difference between having promise.update and not having one?

However, even with this explanation I'm still a bit confused. I understand that you want to take a different approach for implementation which is fine.
What I do ask you to do, is to choose the implementation you prefer, but before actually implementing it, write a big nice comment on top of the function detailing:
1. What are the various inputs expected
2. How the input going to be handled.

I believe it will help making sense in this complex function.

(and yes, this is the problem with complex, you need to be very precise on what's going on inside it)

telegram/ext/commandhandler.py
```diff
@@ -301,7 +317,8 @@ def __init__(self,
                  pass_update_queue=False,
                  pass_job_queue=False,
                  pass_user_data=False,
-                 pass_chat_data=False):
+                 pass_chat_data=False,
+                 run_async=False):
 
         self._prefix = list()
         self._command = list()
```
tsnoam, line 324: @Bibo-Joshi quoting your comment because it is now outdated:
> KeyViews or whatever it's called change, when the dict changes, which is inconvenient here, as we're going to loop over the keys. If we receive an update from an unknown user/chat while doing so, chat_ids will change and we get an error like this one.
The use of list() decouples chat_ids from self.chat_data and just considers the keys present at call-time. The new entry is ignored and will be handled in the next call to update_persistence, which the update from the unknown user triggers.

---

i think it's an excellent explanation, but if you put it is a comment on the code itself, it will help whomever reading the code understand it.

telegram/ext/dispatcher.py
tsnoam: Ok. I agree, you may resolve this conversation.

tsnoam (changes requested):

Bibo-Joshi:
As discussed, I restructured a bit. Let me try to summarize:

* The `error_handler` parameter for `run_async` was basically there to avoid infinite loops coming from asynchronously run functions within asynchronously run error handlers
* instead, now there is just a parameter `async_error_handling`. When `False`, no async error_handler will be called
* We only have one set of error handlers, namely those registered with  `add_error_handler`
* new parameter `async_params` for `CallbackContext` so that the params for custom async functions can be accessed from the error handler

telegram/ext/dispatcher.py
```diff
@@ -247,78 +247,51 @@ def _pooled(self):
...
+            # DispatcherHandlerStop nor raised by an error handler. So we can and must handle it
+            try:
+                self.dispatch_error(promise.update, promise.exception, promise=promise)
+            except Exception:
+                self.logger.exception('An uncaught error was raised while handling the error.')
+
+    def run_async(self, func, *args, update=None, error_handling=True, **kwargs):
+        """Queue a function (with given args/kwargs) to be run asynchronously. Exceptions raised
+        by the asynchronously run function will be handled by the error handlers registered with
+        :meth:`add_error_handler`.
```
tsnoam, line 270:
```suggestion
        """
        Queue a function (with given args/kwargs) to be run asynchronously on a different thread. Exceptions raised
        by the  function will be handled by the error handlers registered with
        :meth:`add_error_handler`.
```

telegram/ext/dispatcher.py
```diff
@@ -247,78 +247,51 @@ def _pooled(self):
...
-    def run_async(self, func, *args, update=None, error_handler=None, **kwargs):
-        """Queue a function (with given args/kwargs) to be run asynchronously.
+            # If we arrive here, an exception happened in the promise and was neither
+            # DispatcherHandlerStop nor raised by an error handler. So we can and must handle it
+            try:
+                self.dispatch_error(promise.update, promise.exception, promise=promise)
+            except Exception:
+                self.logger.exception('An uncaught error was raised while handling the error.')
+
+    def run_async(self, func, *args, update=None, error_handling=True, **kwargs):
```
tsnoam, line 267: I think that `error_handling` is an internal implementation detail and needs to be hidden from the user.
When registering error handlers we need to proactive set `error_handling` to `False` instead of leaving it to the hands of the user.

telegram/ext/dispatcher.py
```diff
@@ -247,78 +247,51 @@ def _pooled(self):
...
-            * When using :meth:`run_async` within an asynchronously run error handler (see
-              :meth:`add_error_handler`) and ``update`` is passed, you *must* also pass
-              ``error_handler``, as an exception within ``func`` would cause an infinite loop
-              otherwise.
-            * passing ``error_handler`` is not supported for the old style API, i.e.
-              ``use_context=False``
+            * When using :meth:`run_async` within an error handler (see :meth:`add_error_handler`),
+              you *must* also set :attr:`error_handling` to :obj:`False`, as an exception within
+              :attr:`func` would cause an infinite loop otherwise. Eventual exceptions will then
+              only be logged.
```
tsnoam, line 278: See my comment above regarding hiding `error_handling` implementation detail from the user.

telegram/ext/dispatcher.py
```diff
@@ -247,78 +247,51 @@ def _pooled(self):
...
-                or
-
-                ``def error_handler(update: Update, context: CallbackContext, *args, **kwargs)``
-
-                if ``update`` is passed, where ``*args`` and ``**kwargs`` are the arguments for
-                ``func``.
+                call. If passed, it will be available in the error handlers, in case an exception
+                is raised by :attr:`func`.
+            error_handling (:obj:`bool`, optional): Whether exceptions raised by :attr:`func`
+                may be handled by error handlers. Defaults to :obj:`True`.
```
tsnoam, line 287: See my comment above regarding hiding `error_handling` implementation detail from the user.

telegram/ext/dispatcher.py
```diff
@@ -575,10 +546,11 @@ def __update_persistence(self, update):
 
     def add_error_handler(self, callback, run_async=False):
         """Registers an error handler in the Dispatcher. This handler will receive every error
-        which happens in your bot.
+        which happens in your bot. Each callback can be added as error handler only *once*.
```
tsnoam, line 549:
```python
"""
Registers an error handler in the Dispatcher. This handler will receive every error
which happens in your bot.

Note:
    Multiple attempts to add the same callback will be ignored.
"""

File: telegram/ext/dispatcher.py [hunk 21]
```diff
@@ -593,6 +565,8 @@ def add_error_handler(self, callback, run_async=False):
         Note:
             See https://git.io/fxJuV for more info about switching to context based API.
         """
+        if callback in self.error_handlers:
+            raise ValueError('The callback is already registered as error handler.')
```
tsnoam, line 569: I say just do `return None`.
You can issue a debug log if you feel that you must.

telegram/ext/dispatcher.py
```diff
@@ -604,27 +578,33 @@ def remove_error_handler(self, callback):
...
                                        update=update)
                     else:
                         self.run_async(callback, self.bot, update, error, update=update)
                 else:
                     if self.use_context:
-                        callback(update, CallbackContext.from_error(update, error, self))
+                        callback(update, CallbackContext.from_error(update, error, self,
+                                                                    async_params=async_params))
                     else:
                         callback(self.bot, update, error)
```
tsnoam: This look like hell. (no offence) 
And all of these indents are to make a one liner in the case of use_context.
Also, it is intended for backward compatibility for non-context-callbacks.

I suggest instead:
```python
                cb_ctx = CallbackContext.from_error(update, error, self, async_params=async_params)
                if run_async:
                    if self.use_context:
                        self.run_async(callback,update, cb_ctx, update=update)
                    else:
                        self.run_async(callback, self.bot, update, error, update=update)
                else:
                    if self.use_context:
                        callback(update, cb_ctx)
                    else:
                        callback(self.bot, update, error)
```

This "costs" time only for those who hadn't upgraded to context base callbacks yet (they will have to by next major release!) but looks so much readable and sane.
Once non-context-callbacks will be removed from the code, this will look even cleaner.

tsnoam (changes requested):

telegram/ext/dispatcher.py
```diff
@@ -247,78 +247,51 @@ def _pooled(self):
...
-                    exc_info=promise.exception)
+            # We still don't like infinite loops, but here a different error message is needed
+            if not promise.error_handling:
+                self.logger.error('A promise with deactivated error handling raised an error.')
+                continue
 
-    def run_async(self, func, *args, update=None, error_handler=None, **kwargs):
-        """Queue a function (with given args/kwargs) to be run asynchronously.
+            # If we arrive here, an exception happened in the promise and was neither
+            # DispatcherHandlerStop nor raised by an error handler. So we can and must handle it
```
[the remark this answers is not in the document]
Bibo-Joshi: I don't think I understand what your suggested comment is trying to say …

telegram/ext/dispatcher.py
Bibo-Joshi: I had to revert the order of the `run_async` and `use_context` tests as CC will raise an error when being created with a non-context dispatcher

telegram/ext/dispatcher.py
Bibo-Joshi: I disagree on this one. Wen can't prevent users from calling `run_async` from within error handlers and at some point, someone will do so leading to an infinite loop. So we either document what `error_handling` is good for or *at least* make a note that using `run_async` from within error handlers is forbidden.

telegram/ext/dispatcher.py
```diff
@@ -566,7 +573,8 @@ def add_error_handler(self, callback, run_async=False):
             See https://git.io/fxJuV for more info about switching to context based API.
         """
         if callback in self.error_handlers:
-            raise ValueError('The callback is already registered as error handler.')
+            self.logger.debug('The callback is already registered as error handler. Ignoring.')
```
tsnoam, line 576:
```suggestion
            self.logger.debug('The callback is already registered as an error handler. Ignoring.')
```

Or maybe: `Ignoring already registered error handler.` (I like this one better - shorter text)

telegram/ext/dispatcher.py
```diff
@@ -275,25 +279,22 @@ def run_async(self, func, *args, update=None, error_handling=True, **kwargs):
         Warning:
             * If you're using ``@run_async``/:meth:`run_async` you cannot rely on adding custom
               attributes to :class:`telegram.ext.CallbackContext`. See its docs for more info.
-            * When using :meth:`run_async` within an error handler (see :meth:`add_error_handler`),
-              you *must* also set :attr:`error_handling` to :obj:`False`, as an exception within
-              :attr:`func` would cause an infinite loop otherwise. Eventual exceptions will then
-              only be logged.
+            * :meth:`run_async` must not be used within an error handler, as an exception within
+              :attr:`func` would cause an infinite loop otherwise.
```
tsnoam, line 283: I'd try to make it more clear, maybe something like:
```suggestion
            * Calling a function wrapped with `run_async` from within an error handler can lead to an infinite error handling loop.
```

telegram/ext/dispatcher.py
```diff
@@ -275,25 +279,22 @@ def run_async(self, func, *args, update=None, error_handling=True, **kwargs):
...
 
         Args:
             func (:obj:`callable`): The function to run in the thread.
             *args (:obj:`tuple`, optional): Arguments to ``func``.
             update (:class:`telegram.Update`, optional): The update associated with the functions
                 call. If passed, it will be available in the error handlers, in case an exception
                 is raised by :attr:`func`.
-            error_handling (:obj:`bool`, optional): Whether exceptions raised by :attr:`func`
```
tsnoam: Removing the documentation does not hide the implementation detail from the user.
I suggest doing something like:
```python

# decorator code
def run_async(func):
    def async_func(*args, **kwargs):
        # call the `_` version
        return Dispatcher.get_instance()._run_async(func, *args, update=None, error_handling=False, **kwargs)

class Dispatcher:
    def run_async(self, func, *args, update=None,  **kwargs): 
        self._run_async(func, *args, update, error_handling=True, **kwargs)

    def _run_async(self, func, *args, update=None, error_handling=True, **kwargs): 
       # current run_async() code comes here

```

tsnoam (changes requested):

telegram/ext/dispatcher.py
```diff
@@ -47,6 +47,10 @@ def run_async(func):
 
     Using this decorator is only possible when only a single Dispatcher exist in the system.
 
+    Note:
+        DEPRECATED. Use :attr:`telegram.ext.Dispatcher.run_async` directly instead or the
+        :attr:`Handler.run_async` parameter.
```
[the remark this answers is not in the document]
Bibo-Joshi: parameter. The syntax is now `add_handler(CommandHandler('start', start, run_async=True))`

telegram/ext/dispatcher.py
```diff
@@ -247,12 +251,14 @@ def _pooled(self):
...
                 self.logger.error('An uncaught error was raised while handling the error.')
                 continue
 
-            # We still don't like infinite loops, but here a different error message is needed
+            # Don't perform error handling for a `Promise` with deactivated error handling. This
+            # should happen only via the deprecated `@run_async` decorator or `Promises` created
+            # within error handlers
             if not promise.error_handling:
                 self.logger.error('A promise with deactivated error handling raised an error.')
                 continue
```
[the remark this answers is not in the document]
Bibo-Joshi: added a test

tsnoam (approved):
Great work

PR #2056, opened 2020-08-19
dataclass implementation

hexatester:
This only change class init & sub-classing class constants.
Might wanna install [dacite ](https://pypi.org/project/dacite/) to create an instance of data from dict without making a sweat.

I cant test this locally because the tests always timeout after a few tests on my pc.

https://docs.python.org/3/library/dataclasses.html#post-init-processing
https://stackoverflow.com/questions/52390576/how-can-i-make-a-python-dataclass-hashable


Bibo-Joshi:
Hi! Thanks for the PR.
I ran the tests on my machine and most of them fail because `tg.Base.de_json()` doesn't work with your changes. I rebased your PR to master because that way the automated tests should run, when you add new commits. you can change it back by hitting "Edit" on the top right corner.
If you want to get this concept PR to minimally working, please go ahead. However, I feel obliged to warn you not fine tune every last bit just now, as we'd have to discuss the idea in the maintainers group before preceeding ;)

hexatester:
I see, but __hash__ still working from Base as intended right?

I might change the Base class and give it condition like `isdataclass(cls)` then handle it different way from the others.

But still i didn't find any way to add **kwargs inside init of dataclass.

I'll get this PR minimally working asap, can you just send me failed test logs?

Bibo-Joshi:
uhm, currently the tests don't come further than `test_animation`. I think you should be able to see the log by clickling on "Details".

I also became aware that dataclasses are new in Py3.7 and as 3.6 reaches EOL only at the end of 2021 I don't think that we'll be willing to drop 3.6 before autumn 2021 …

hexatester:
> 
> 
> uhm, currently the tests don't come further than `test_animation`. I think you should be able to see the log by clickling on "Details".
> 
> I also became aware that dataclasses are new in Py3.7 and as 3.6 reaches EOL only at the end of 2021 I don't think that we'll be willing to drop 3.6 before autumn 2021 …

Okay.
Its fine about adding dataclasses to ptb next year, im just testing to see is it possible or not.

hexatester:
I'll close this PR for now. At least we know it's definitely possible to refactor ptb using dataclass.

Thanks, also this really help my project.

Issue #2105, opened 2020-10-02
[BUG] The "TelegramException" (and children) can't be unpickled


Delgan:
### Steps to reproduce

Hi.  I encountered a small issue with an user because it does not seem possible to pickle `TelegramException` exceptions and their descendants. 

```python
import pickle
from telegram.error import TelegramException

error = TelegramException("Foobar")
pickled = pickle.dumps(error)      # OK
unpikcled = pickle.loads(pickled)  # NOK
# TypeError: __init__() missing 1 required positional argument: 'message'
```
This is actually not new, many Python users have encountered this kind of problem by implementing custom exceptions:
* [Cannot unpickle Exception subclass](https://stackoverflow.com/questions/41808912/cannot-unpickle-exception-subclass)
* [How to pickle inherited exceptions?](https://stackoverflow.com/questions/49715881/how-to-pickle-inherited-exceptions)
* [How to make a custom exception class with multiple init args pickleable](https://stackoverflow.com/questions/16244923/how-to-make-a-custom-exception-class-with-multiple-init-args-pickleable)

 
### Expected behaviour
The exception should be deserialized without issue.

### Actual behaviour
A `TypeError` is raised.

### Configuration
**Operating System:** Arch Linux
**Version of Python, python-telegram-bot & dependencies:**
```
python-telegram-bot 12.8 (0.5.3-0-gf31e971)
certifi 2020.06.20
Python 3.8.5 (default, Sep  5 2020, 10:50:12)  [GCC 10.2.0]
```

### Logs
Not relevant.

Bibo-Joshi:
Hi. Thanks for reporting and for including the links to SO. However, I wonder what the use case of pickling exceptions would be. As far as I can see, TG errors should be handled at runtime and not be saved to file.

Delgan:
In this case, the error occurred when trying to send the `TelegramException` to a `multiprocessing.Queue` (which requires that the objects can be (de)serialized using `pickle`). Actually, the `TelegramException` was not passed directly to the queue but rather contained in a `Record` object. Sending this record to a queue allowed parallel processing in a background process.

Surely the `exception` can be removed from the `Record` before sending it to the queue. Now that we figured out the issue, it's easy to find a solution. 

However, since the error occurred at the time of deserialization, it was difficult in our case to identify the source of the problem (since the background process cannot access the `Record`). I guess it may avoid similar incident in the future if exceptions can be safely pickled / unpickled.

[...]

PR #2106, opened 2020-10-02
Fix TypeError while unpickling TelegramError (and children)

Delgan:
Hi.

This is a possible fix for #2105.

Let me know if you think this can be improved in any way.


telegram/error.py
```diff
@@ -51,6 +51,9 @@ def __init__(self, message):
     def __str__(self):
         return '%s' % (self.message)
 
+    def __reduce__(self):
+        return type(self), (self.message,)
```
Bibo-Joshi, line 55: You can use `self.__class__` instead. Don't know if there are any style conventions on that, but that's what we currently do for `Bot` and it doesn't need a function call ... Same for the others

tests/test_error.py
```diff
@@ -87,3 +89,22 @@ def test_retry_after(self):
...
+            TimedOut(),
+            ChatMigrated(1234),
+            RetryAfter(12),
+            Conflict("test message"),
+        ],
+    )
+    def test_pickling(self, exception):
+        pickled = pickle.dumps(exception)
+        unpickled = pickle.loads(pickled)
+        assert type(unpickled) == type(exception)
```
Bibo-Joshi, line 110: Here you should also check that the error message and, if applicable, the additional attributes are deserialized correctly

Bibo-Joshi (changes requested):
Hi. Thanks for the quick PR! I left two comments for you.

tests/test_error.py
Delgan: Sure. Then I need to create one function for each exception instead of relying on `pytest.mark.parametrize()`, as each exception has different attributes. I guess I will create a new `test_errors_pickling.py` file with a new test class then.

telegram/error.py
```diff
@@ -97,6 +109,9 @@ def __init__(self, retry_after):
         super().__init__('Flood control exceeded. Retry in {} seconds'.format(retry_after))
         self.retry_after = float(retry_after)
 
+    def __reduce__(self):
+        return self.__class__, (self.retry_after,), {"message": self.message}
```
Delgan, line 113: @Bibo-Joshi This one is a bit tricky. The first time we initialize a `RetryAfter(12)` exception, the message will display the delay as an integer: `"Retry in 12 seconds"`. If `pickle` tries to re-build the exception, `self.retry_after` being a `float` the message will be changed to `"Retry in 12.0 seconds"`, causing the `assert error.message == unpickled.message` to fail. That's why I specifically update the `message` attribute.

Bibo-Joshi: Mh, that's because retry_after is converted to float only after the message is generated. I think there's no harm in  converting before.

tests/test_error_pickling.py
```diff
@@ -0,0 +1,90 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser Public License for more details.
+#
+# You should have received a copy of the GNU Lesser Public License
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+import pickle
```
Bibo-Joshi, line 19: Please out the tests back in `test_error.py`. if we go for splitting test files, `test_persistence` should go first 😄

Also you can still cover everything with parametrize: pass the list of attributes to check in addition to the exception and use `getattr` to perform the check

Bibo-Joshi (changes requested):

tsnoam:
@Bibo-Joshi If we ever introduce a new exception which inherits from `TelegramError`, would it be covered by the unitests here?

Maybe we also need a unitests which looks at `telegram.error.TelegramError.__subclasses__()` and make sure that all subclasses are known and properly covered?

Bibo-Joshi:
> @Bibo-Joshi If we ever introduce a new exception which inherits from `TelegramError`, would it be covered by the unitests here?
> 
> Maybe we also need a unitests which looks at `telegram.error.TelegramError.__subclasses__()` and make sure that all subclasses are known and properly covered?

It wouldn't. But we might not be directly inheriting from TGError, so we would need a recursive loop. Should be doable, just not with parametrize. I can have a look, when I'm back at a keyboard and had some sleep :D

Bibo-Joshi:
Added the "meta-test" and in the process found that there is a `TelegramDecryptionError` in `passport/credentials.py` :D That one in fact is a bit tricky as it passes `TelegramDecryptionError: {message}` to `super().__init__`. But I didn't want to fiddle with that behaviour, so I just added a `self._msg` that saves the actual message on init and is returned on `__reduce__`.

Bibo-Joshi:
Thank your for your contribution @Delgan :)

PR #2107, opened 2020-10-03
Improving Docs

codewithkapil:


Bibo-Joshi:
Hi. Thanks for the PR but we don't see any benefit of it. The README has the Logo as title and is only relevant on the GitHub and PyPi of PTB anyway, where it is evident what you're supposed to get started on.

PR #2118, opened 2020-10-08
Fixed few type hints

harshil21:


Bibo-Joshi:
Upsie, that should only be `str`. `ChatAction` just provides the correct strings as class attributes … Could you change that and also check the other occurrences?

harshil21:
> Upsie, that should only be `str`. `ChatAction` just provides the correct strings as class attributes … Could you change that and also check the other occurrences?

Oh alright. Just fixed another type hint in send_message()

Bibo-Joshi:
Thanks for the contribution :)

Issue #2124, opened 2020-10-09
[IMPROVEMENT] change our format to f-strings


Poolitzer:
Since we dropped 3.5 support, we can now introduce f-strings. This will be a great first issue for someone new who wants to do manual work, or a great regex kind of task for someone who wants to go crazy.

Bibo-Joshi:
F strings shouldn't be used for logging, though, as that actually doesn't like `format`

NishantPuri99:
Hi @Poolitzer I'd like to take this up

Bibo-Joshi:
@NishantPuri99 I assigned you :)

NishantPuri99:
Thank you so much :)

NishantPuri99:
Hi @Bibo-Joshi and @Poolitzer , I've made the required changes. I hope they are correct and valid, and that the code still works fine.
Thanks and regards

PR #2128, opened 2020-10-10
add supergroup filter

### LINKED ISSUE #2126: [FEATURE] Filter for chat types
We currently have 

* `Filters.private` for private chats
* `Filters.updates.channel_post` for channel posts
* `Filters.group` for normal *and* supergroups

We're missing a standalone supergroup filter. Different ideas listed below. Personally, I'd go for 3.

# Proposal 1: Minimal invasive

* Add `Filters.supergroup`
* Document that `Filters.group` allows both

# Proposal 2: Adding flags

* Add a flag to `Filters.group` to specify, whether supergroups should be allowed or not

# Proposal 3: The full way

* Add `Filters.chat_type` that basically acts like `Filters.chat/user/via_bot`, just with `update.effective_message.chat.type`. We don't need to have `Filters.chat_type` updatetable, though IMHO. I really don't see a usecase for that.
* Add shortcuts for `private, channel, (super)group` as attributes to `Filters.chat_type` (similar to the `Filters.dice`)
* Make `Filters.private, group` shortcuts that work with `Filters.chat_type` and deprecate them (adding `TelegramDeprecationWarning` and documenting it)

By this, we would have two checks for `channel` (`Filters.chat_type.channel` and `Filters.update.channel_post(s)`), but especially is the latter allows to differentiate between original and edited channel post, I thinks both have their value.

GauthamramRavichandran:
- Created a new supergroup filter
- Specified that group filter will accept both normal groups and supergroups

closes #2126

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -304,6 +304,15 @@ class _DiceEmoji(MessageFilter):
         return False
 
 
+class _ChatType(MessageFilter):
+    def __init__(self, chat_type: str = Chat.PRIVATE):
+        self.chat_type = chat_type
+        self.name = 'Filters.chat_type.{}'.format(chat_type)
+
+    def filter(self, message: Message) -> bool:
+        return message.chat.type == self.chat_type
+
+
 class Filters:
     """Predefined filters for use as the ``filter`` argument of
     :class:`telegram.ext.MessageHandler`.
@@ -965,13 +974,22 @@ officedocument.wordprocessingml.document")``-
 
     class _SuperGroup(MessageFilter):
         name = 'Filters.supergroup'
-    
+
         def filter(self, message: Message) -> bool:
             return message.chat.type == Chat.SUPERGROUP
 
     supergroup = _SuperGroup()
     """Messages sent in a supergroup chat."""
-    
+
+    class _ChatTypes(_ChatType):
+        """Filter using chat type"""
+        channel = _ChatType(Chat.CHANNEL)
+        group = _ChatType(Chat.GROUP)
+        supergroup = _ChatType(Chat.SUPERGROUP)
+        private = _ChatType(Chat.PRIVATE)
+
+    chat_type = _ChatTypes()
+
     class user(MessageFilter):
         """Filters messages to allow only those which are from specified user ID(s) or
         username(s).
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -304,15 +304,6 @@ class _DiceEmoji(MessageFilter):
         return False
 
 
-class _ChatType(MessageFilter):
-    def __init__(self, chat_type: str = Chat.PRIVATE):
-        self.chat_type = chat_type
-        self.name = 'Filters.chat_type.{}'.format(chat_type)
-
-    def filter(self, message: Message) -> bool:
-        return message.chat.type == self.chat_type
-
-
 class Filters:
     """Predefined filters for use as the ``filter`` argument of
     :class:`telegram.ext.MessageHandler`.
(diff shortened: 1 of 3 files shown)


GauthamramRavichandran:
Fixes one of #2126

Poolitzer:
Hi.

While it is correct that you implemented proposal 1, we agreed on doing 3 in the linked issue. Can you overhaul your PR to reflect this?

Also, you need to write unitests which test your changes, dont forget that. If you install `pre-commit` hooks as explained in our contribution.md guide, that will be done locally before you commit \o/

GauthamramRavichandran:
- Created `ChatTypes` with channel, group, supergroup and private. Added unit tests for the same

telegram/ext/filters.py
```diff
@@ -304,6 +304,15 @@ def filter(self, message: Message) -> bool:
         return False
 
 
+class _ChatType(MessageFilter):
+    def __init__(self, chat_type: str = Chat.PRIVATE):
+        self.chat_type = chat_type
+        self.name = 'Filters.chat_type.{}'.format(chat_type)
```
Bibo-Joshi, line 310: Please use f-strings, see #2124

telegram/ext/filters.py
```diff
@@ -961,7 +970,25 @@ def filter(self, message: Message) -> bool:
...
+    """Messages sent in a supergroup chat."""
+
+    class _ChatTypes(_ChatType):
+        """Filter using chat type"""
+        channel = _ChatType(Chat.CHANNEL)
+        group = _ChatType(Chat.GROUP)
+        supergroup = _ChatType(Chat.SUPERGROUP)
+        private = _ChatType(Chat.PRIVATE)
+
+    chat_type = _ChatTypes()
```
Bibo-Joshi, line 991: This needs a docstring

telegram/ext/filters.py
```diff
@@ -961,7 +970,25 @@ def filter(self, message: Message) -> bool:
...
+        name = 'Filters.supergroup'
+
+        def filter(self, message: Message) -> bool:
+            return message.chat.type == Chat.SUPERGROUP
+
+    supergroup = _SuperGroup()
+    """Messages sent in a supergroup chat."""
+
+    class _ChatTypes(_ChatType):
+        """Filter using chat type"""
```
Bibo-Joshi, line 985: This class is private, it shouldn't get a docstring

telegram/ext/filters.py
```diff
@@ -304,6 +304,15 @@ def filter(self, message: Message) -> bool:
         return False
 
 
+class _ChatType(MessageFilter):
+    def __init__(self, chat_type: str = Chat.PRIVATE):
```
Bibo-Joshi, line 308: The idea was that you should also be able to pass a list of strings to e.g. filter for private *or* channel

telegram/ext/filters.py
```diff
@@ -961,7 +970,25 @@ def filter(self, message: Message) -> bool:
             return message.chat.type in [Chat.GROUP, Chat.SUPERGROUP]
 
     group = _Group()
-    """Messages sent in a group chat."""
+    """Messages sent in a group or a supergroup chat."""
+
+    class _SuperGroup(MessageFilter):
```
Bibo-Joshi, line 975: please remove. the idea was to move everything to `chat_types`

Bibo-Joshi (changes requested):
Thanks for the updates. Unfortunately, most of the changes discussed in #2126 §3 are still missing. Sorry, if my explanations weren't detailed enough … I tried to clarify what's missing in some comments. The filter that comes closest to what is needed here is the `Filters.dice`, maybe have a look at that :) If you have questions, feel free to reach out on TG.

Finally, recalling @Poolitzer s idea, I think it would be good to add an attribute `Filters.groups` (plural) for both super- and normal groups to `chat_type`

PS: Just writing down a thought as "fyi": I had the idea that one could also make the chat-type filters attriubtes of `Filters.chat`, which filters for specific chat ids. That would have the benefit that we don't have a `Filters.chat_type` which admittetly has a small use case. However, `Filters.chat` is already quite complex and I fear that users might get confused about what it acutally filters for. Keeping filtering for ids and types separate is probably better.

Poolitzer:
+1 for filtering types and ida seperately in two filters please thank you

telegram/ext/filters.py
GauthamramRavichandran: yes, sure. will replace this to line:992

GauthamramRavichandran:
could someone implement the `__call__` method for `_ChatType`, thank you

Bibo-Joshi:
Thinking about it more, passing args to `Filters.chat_type` might be a bit over the top. For now I just added the deprecation warnings.

Poolitzer:
what is stopping us from merging?

telegram/ext/filters.py
```diff
@@ -949,19 +952,103 @@ class _Private(MessageFilter):
...
+        DEPRECATED. Use
+        :attr:`telegram.ext.Filters.chat_type.private` instead.
+    """
 
     class _Group(MessageFilter):
         name = 'Filters.group'
 
         def filter(self, message: Message) -> bool:
+            warnings.warn(
+                'Filters.private is deprecated. Use Filters.chat_type.groups instead.',
```
Poolitzer, line 976:
```suggestion
                'Filters.group is deprecated. Use Filters.chat_type.groups instead.',
```

and now its group :)

telegram/ext/filters.py
```diff
@@ -949,19 +952,103 @@ class _Private(MessageFilter):
...
+                'Filters.private is deprecated. Use Filters.chat_type.private instead.',
+                TelegramDeprecationWarning,
+                stacklevel=2,
+            )
             return message.chat.type == Chat.PRIVATE
 
     private = _Private()
-    """Messages sent in a private chat."""
+    """
+    Messages sent in a group or a supergroup chat.
```
Poolitzer, line 964:
```suggestion
    Messages sent in a private chat.
```
uhm no its send in a private chat?

Poolitzer (changes requested):
Two doc strings are wrong, otherwise great change!

PR #2142, opened 2020-10-13
Don't mark non-universal wheels as universal

michael-k:
Wheels are only universal if they support both Python 2 and 3.


Bibo-Joshi:
Thanks for the catch :)

PR #2147, opened 2020-10-15
Add new constants

### LINKED ISSUE #2104: Overhaul & extension of constants
We have constants in different locations, e.g.

* `telegram.constants`
* class attributes of `telegram.ParseMode`
* class attributes of `telegram.Dice`

That should be unified, i.e.:

- [x] Gather all constants in `telegram.constants` (I guess a regex-search for `SCREAMING_SNAIL_CASE` should do the trick …)
- [x] Repace the class attributes by links to the constant from `telegram.constant`, (e.g. `ParseMode.HTML = constants.PARSE_MODE_HTML`)
- [x] Add additional links, where suitable (e.g. `Message.MAX_LENGTH = constants.MAX_MESSAGE_LENGTH`)

Moreover, in light of the recent TG update, we should add

- [ ] constants for user ID `777000` and `1087968824`, which act as senders for forwarded channel posts and messages from anonymous admins (see [here](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Code-snippets#exclude-messages-from-anonymous-admins))
- [ ] Add properties `is_anonymous_admin` and `is_service_chat` (or similar namings) to `User` and `Chat` that compare the user/chat_id to the IDs menotioned above

*Edit:* there should be notes on the constants, if they are not documented. And going through the docs to see if there are constants not covered already probably won't hurt …

NikitaPirate:
closes #2104


telegram/constants.py
```diff
@@ -38,6 +38,9 @@
 Attributes:
     MAX_MESSAGE_ENTITIES (:obj:`int`): 100 (Beyond this cap telegram will simply ignore further
         formatting styles)
+    ANONYMOUS_ADMIN_ID (:obj:`int`):777000
```
Bibo-Joshi, line 41:
```suggestion
    ANONYMOUS_ADMIN_ID (:obj:`int`): ``777000``
```

telegram/constants.py
```diff
@@ -38,6 +38,9 @@
...
+    CHANNEL_FORWARDER_SERVICE_ID (:obj:`int`): 1087968824 (Telegram service that forwards messages
```
Bibo-Joshi, line 42:
```suggestion
    CHANNEL_FORWARDER_SERVICE_ID (:obj:`int`): ``1087968824`` (Telegram service that forwards messages
```

Also this is the anonymous admin id and 777000 is the service chat ;) I'm not sure about the naming, though. 777000 is also the id of the chat that gives you your login PINs. imo naming ``SERVICE_CHAT_ID`` + the note that this is the one forwarding the messages to discussion groups is more concise.
But I'm not too passionate about it …

telegram/user.py
```diff
@@ -93,6 +97,9 @@ def __init__(
         self.can_read_all_group_messages = can_read_all_group_messages
         self.supports_inline_queries = supports_inline_queries
         self.bot = bot
+        self.is_anonymous_admin = True if self.id == constants.ANONYMOUS_ADMIN_ID else False
```
Bibo-Joshi, line 100:
```suggestion
        self.is_anonymous_admin = self.id == constants.ANONYMOUS_ADMIN_ID
```

telegram/user.py
```diff
@@ -93,6 +97,9 @@ def __init__(
...
+        self.is_channel_forwarder_service = (
+            True if self.id == constants.CHANNEL_FORWARDER_SERVICE_ID else False)
```
Bibo-Joshi, line 102:
```suggestion
        self.is_channel_forwarder_service = self.id == constants.CHANNEL_FORWARDER_SERVICE_ID
```

telegram/user.py
```diff
@@ -49,6 +49,10 @@ class User(TelegramObject):
         supports_inline_queries (:obj:`str`): Optional. :obj:`True`, if the bot supports inline
             queries. Returned only in :attr:`telegram.Bot.get_me` requests.
         bot (:class:`telegram.Bot`): Optional. The Bot to use for instance methods.
+        is_anonymous_admin (:obj:`bool`): :obj:`True`, if this user is an anonymous admin.
+            Undocumented behavior.
```
Bibo-Joshi, line 53:
```suggestion
            This behaviour is undocumented and might be changed by Telegram.
```

Bibo-Joshi (changes requested):
Thanks for the PR!

I did a bit of nitpicking. Please add the `is_*` thingies to `Chat`, too. Also, we'll need some unittests. Have a look at `tests/test_{user/chat}.py` some inspiration :) if you want to run one one single test for checking, if it works, you can use `pytest -k test_function_name`.

Finally, I feel like it would be cleaner to make the `is_*` thingies properties instead of attributes. While no one in the right mind would change the chats id after initialization, properties could still catch that and also that way it would be clearer that it's just a convenience thingy. So IMHO it should be something like [this](https://github.com/python-telegram-bot/python-telegram-bot/blob/3b4559dd95d14a4c9a9ca54237b5fe547386b666/telegram/bot.py#L351).

telegram/chat.py
```diff
@@ -150,6 +150,20 @@ def link(self) -> Optional[str]:
             return "https://t.me/{}".format(self.username)
         return None
 
+    @property
+    def is_anonymous_admin(self) -> bool:
+        """:obj:`bool`: :obj:`True`, if it is chat with an anonymous admin.
```
Bibo-Joshi, line 155:
```suggestion
        """:obj:`bool`: :obj:`True`, if this chat is with is the bot representing anonymous admins.
```

telegram/chat.py
```diff
@@ -150,6 +150,20 @@ def link(self) -> Optional[str]:
...
+            This behaviour is undocumented and might be changed by Telegram. """
+
+        return self.id == constants.ANONYMOUS_ADMIN_ID
+
+    @property
+    def is_service_chat(self) -> bool:
+        """:obj:`bool`: :obj:`True`, if it is chat with the telegram service.
```
Bibo-Joshi, line 162:
```suggestion
        """:obj:`bool`: :obj:`True`, if this chat is the Telegram service chat.
```

telegram/constants.py
```diff
@@ -38,9 +38,8 @@
 Attributes:
     MAX_MESSAGE_ENTITIES (:obj:`int`): 100 (Beyond this cap telegram will simply ignore further
         formatting styles)
-    ANONYMOUS_ADMIN_ID (:obj:`int`):777000
-    CHANNEL_FORWARDER_SERVICE_ID (:obj:`int`): 1087968824 (Telegram service that forwards messages
-        from the channel to the attached chat)
+    ANONYMOUS_ADMIN_ID (:obj:`int`): ``1087968824``
+    SERVICE_CHAT_ID (:obj:`int`): ``777000`` (Telegram service)
```
Bibo-Joshi, line 42:
```suggestion
    SERVICE_CHAT_ID (:obj:`int`): ``777000`` (Telegram service chat, that also acts as sender of channel posts forwarded to discussion groups)
```

Bibo-Joshi (commented):
Thanks, that looks good :) I left some minor reformulation suggestions. But I'd like @Poolitzer to review anyway, so he can decide about those.

telegram/chat.py
```diff
@@ -150,6 +150,20 @@ def link(self) -> Optional[str]:
             return "https://t.me/{}".format(self.username)
         return None
 
+    @property
+    def is_anonymous_admin(self) -> bool:
+        """:obj:`bool`: :obj:`True`, if this chat is with is the bot representing anonymous admins.
+            This behaviour is undocumented and might be changed by Telegram. """
```
Poolitzer, line 156: maybe call it convenience property? We are doing it like that in eg [user.py](https://github.com/python-telegram-bot/python-telegram-bot/blob/264b2c9c72691c5937b80e84e061c52dd2d8861a/telegram/user.py#L101)

telegram/constants.py
```diff
@@ -38,6 +38,9 @@
 Attributes:
     MAX_MESSAGE_ENTITIES (:obj:`int`): 100 (Beyond this cap telegram will simply ignore further
         formatting styles)
+    ANONYMOUS_ADMIN_ID (:obj:`int`): ``1087968824``
```
Poolitzer, line 41: explain where that happens :) Like user id in groups for anon admins

Poolitzer (changes requested):
left two minor comments, thanks for the improvement!

Bibo-Joshi:
Thanks for the contribution!

PR #2149, opened 2020-10-15
Update timerbot example

### LINKED ISSUE #2144: [BUG] Update timerbot example
https://github.com/python-telegram-bot/python-telegram-bot/blob/3b4559dd95d14a4c9a9ca54237b5fe547386b666/examples/timerbot.py#L56-L58

If the job has already run, those lines give an error after v13.0, as APS remvos the job after it has run. But as jobs can't be persisted natively, I think we should just restructure the example to give the jobs the `chat_id` as name and use `get_jobs_by_name` to remove them if necessary

Edit: And we should definetely remove the `pass_*` kwargs, those are deprecated since v12 …

timqsh:
closes #2144

### DIFF
--- a/examples/timerbot.py
+++ b/examples/timerbot.py
@@ -45,11 +45,11 @@ def alarm(context):
 def remove_job_if_exists(name, context):
     """Remove job with given name. Returns whether job was removed."""
     current_jobs = context.job_queue.get_jobs_by_name(name)
-    job_removed = False
+    if not current_jobs:
+        return False
     for job in current_jobs:
         job.schedule_removal()
-        job_removed = True
-    return job_removed
+    return True
 
 
 def set_timer(update, context):
@@ -78,7 +78,7 @@ def unset(update, context):
     """Remove the job if the user changed their mind."""
     chat_id = update.message.chat_id
     job_removed = remove_job_if_exists(str(chat_id), context)
-    text = 'Timer successfully unset!' if job_removed else 'You have no active timer'
+    text = 'Timer successfully cancelled!' if job_removed else 'You have no active timer.'
     update.message.reply_text(text)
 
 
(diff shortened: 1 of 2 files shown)


examples/timerbot.py
```diff
@@ -42,6 +42,16 @@ def alarm(context):
...
 
 
+def remove_job_if_exists(name, context):
+    """Remove job with given name. Returns whether job was removed."""
+    current_jobs = context.job_queue.get_jobs_by_name(name)
+    job_removed = False
+    for job in current_jobs:
+        job.schedule_removal()
+        job_removed = True
+    return job_removed
```
Bibo-Joshi, line 52:
```suggestion
    current_jobs = context.job_queue.get_jobs_by_name(name)
    if not current_jobs:
        return False
    for job in current_jobs:
        job.schedule_removal()
    return True
```

examples/timerbot.py
```diff
@@ -52,30 +62,24 @@ def set_timer(update, context):
...
-        return
-
-    job = context.chat_data['job']
-    job.schedule_removal()
-    del context.chat_data['job']
-
-    update.message.reply_text('Timer successfully unset!')
+    chat_id = update.message.chat_id
+    job_removed = remove_job_if_exists(str(chat_id), context)
+    text = 'Timer successfully unset!' if job_removed else 'You have no active timer'
```
Bibo-Joshi, line 81:
```suggestion
    text = 'Timer successfully cancelled!' if job_removed else 'You have no active timer.'
```

Bibo-Joshi (commented):
Looks and works great :) I left two minor comments

timqsh:
Done =)

Bibo-Joshi:
Thanks for your contribution! :)

PR #2150, opened 2020-10-15
Bug fixes for insert/replace_bot

### LINKED ISSUE #2145: [BUG] Infinite loops with insert/replace_bot
You fiddle with persistence, you introduce bugs … #1994 not introducing bugs would have been too easy :D

```python

class A:
     b = None

class B:
    a = None

a = A()
b = B()
a.b = b
b.a = a

replace/insert_bot(a)  # this exceeds max recursion limit
```

Luckily that can easily resolved by passing a `memo` dict just as `deepcopy` does and while tracking down the bug I already made the required changes to `insert_bot`. I'll try to adjust `repace_bot` accordingly and write some tests in the next days.

Bibo-Joshi:
this does two things:

* pass a `memo` dict to handle recursion, closes #2145 
* if an object can't be copied, they're probably something special and don't contain a `Bot` instance. Therefore, those will now just be skipped.

Note: I noticed the latter while testing the changes on a personal project, where I have `locks` in some objects that have a custom pickle behaviour but not a custom `copy` behaivour

### DIFF
--- a/telegram/ext/basepersistence.py
+++ b/telegram/ext/basepersistence.py
@@ -24,7 +24,7 @@ from copy import copy
 
 from telegram import Bot
 
-from typing import DefaultDict, Dict, Any, Tuple, Optional, cast
+from typing import DefaultDict, Dict, Any, Tuple, Optional, cast, List, Set
 from telegram.utils.types import ConversationDict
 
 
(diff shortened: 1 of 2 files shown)


tsnoam:
I've taken this discussion to the developers group.
I'm not feeling comfortable in fixing "bugs" resulting from bad usage patterns.

We'll post the results of that discussion here.

Bibo-Joshi:
Closing for the reasons mentioned by @tsnoam in favor of #2151

PR #2158, opened 2020-10-22
Fix Black in pre-commit

Bibo-Joshi:
see this [issue](https://github.com/psf/black/issues/1773)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,7 @@ repos:
     -   id: black
         args:
         - --diff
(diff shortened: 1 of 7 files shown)


Bibo-Joshi:
Seems like now I can't get black to respect the in/exclude regexs in pre-commit *sigh *. @JosXa any idea? will try to have a closer look tonight …

Bibo-Joshi:
Let's see what [black/#1778](https://github.com/psf/black/issues/1778) brings to light

PR #2160, opened 2020-10-25
WIP: Refactor MQ

### LINKED ISSUE #1369: Multiple message queues for a bot
I just read this source of the decorator [@messagequeue](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/messagequeue.py)
and I thought what if a bot needs multiple message queues?
because the limits in telegram are different for each type of requests.
for example the limit for editing message texts is too low!

what do you suggest to do?

Thanks.

### LINKED ISSUE #2012: [BUG] Bot hangs when using ctrl + c and MessageQueue
### Steps to reproduce

1. Set the `working` variable to `False`
2. Run the bot using `./main.py`
3. Hit <kbd>ctrl</kbd> + <kbd>c</kbd>

### Expected behaviour

The bot should stop.

### Actual behaviour

The bot hangs.

### Configuration

**Operating System:** 10.15.5 (19F101)

```
python-telegram-bot 12.8
certifi 2020.06.20
Python 3.7.7 (default, Mar 10 2020, 15:43:33)  [Clang 11.0.0 (clang-1100.0.33.17)]
```

### Logs

Working:

```
> ./main.py
2020-07-02 12:59:25,880 - __main__ - INFO - Started polling
2020-07-02 12:59:26,157 - __main__ - INFO - Bot started. Press Ctrl-C to stop.
^C2020-07-02 12:59:27,504 - telegram.ext.updater - INFO - Received signal 2 (SIGINT), stopping...
```

then exits.

Not working:

```
> ./main.py
2020-07-02 12:59:34,283 - telegram.ext.updater - WARNING - Connection pool of Request object is smaller than optimal value (8)
2020-07-02 12:59:34,283 - __main__ - INFO - Started polling
2020-07-02 12:59:34,537 - __main__ - INFO - Bot started. Press Ctrl-C to stop.
^C2020-07-02 12:59:37,574 - telegram.ext.updater - INFO - Received signal 2 (SIGINT), stopping...
```

and hangs.

`main.py`:

```py
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

import logging

logging.basicConfig(format='%(asctime)s - %(name)s - %(levelname)s - %(message)s', level=logging.INFO)

from telegram.ext import Updater

from queue_bot import QueueBot

updater = None


def main():
    logger.info('Started polling')

    updater.start_polling(timeout=0.01)

    logger.info('Bot started. Press Ctrl-C to stop.')

    updater.idle()


if __name__ == '__main__':
    token = 'TOKEN'
    working = True

    if working:
        updater = Updater(token, use_context=True)
    else:
        queue_bot = QueueBot(token)
        updater = Updater(bot=queue_bot, use_context=True)

    main()
```

`queue_bot.py`:

```py
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

from telegram.bot import Bot
from telegram.ext import messagequeue


class QueueBot(Bot):
    def __init__(self, *args, **kwargs):
        super(QueueBot, self).__init__(*args, **kwargs)

        self._is_messages_queued_default = True
        self._msg_queue = messagequeue.MessageQueue()

    def __del__(self):
        try:
            self._msg_queue.stop()
        except:
            pass

    @messagequeue.queuedmessage
    def send_message(self, *args, **kwargs):
        return super(QueueBot, self).send_message(*args, **kwargs)
```

### LINKED ISSUE #2015: [BUG] Defaults not working with MessageQueue
### Steps to reproduce
1. Set up a bot with a MessageQueue as described in the wiki

2. pass a `Defaults` object to the bot

### Expected behaviour
Defaults should be used

### Actual behaviour
They're not. Instead `ValueError("Function has keyword-only parameters or annotations"` is raised

### Why this happens & how to overcome
1. `getargspec` is deprecated in favor of `getfullargspec`. This can easily be resolved
2. This doesn't suffice yet, as the methods of `MQBot` override the bots methods and are decorated. `getfullargspec` can't see through the decorator. using `inspect.getfullargspec(inspect.unwrap(method))` can resolve that
3. This still only works, if we explicitly list the methods args & kwargs while overriden in `MQBot`

After #2012 this is one more reason to tackle MQ in v13.X/v14. Just adding `inspect.getfullargspec(inspect.unwrap(method))` as a more or less quick fix doesn't really seem right to me …

### LINKED ISSUE #2035: [BUG] Context manager not working with MessageQueue
### Steps to reproduce
1. Set up a bot with a MessageQueue as described in the wiki

2. Override methods send_document, send_photo with @messagequeue.queuedmessage decorator

3. Open file with `context manager` and try to send it

### Expected behaviour
Files must be sent to chat as usual, like they do without MessageQueue

### Actual behaviour
Files are not sent. As I found, file is closed at the moment it should have sent.
And we have: `ValueError: read of closed file` of course.

### Why this happens and how to overcome that _(steal that topic)_
Files are closed before actual send, because send_document returns Promise object instead of message, you know, because it's messagequeue. So that Promise results after context manager close this file.
**HotFix:** You should wait for promise.result() if you want to close file properly.
Or InputFile should close file after it is done with it, and you don't have to worry about it.

_Hope I fill everything right._

### LINKED ISSUE #2139: Refactor MessageQueue
# TL;DR:

Known bugs are  #2015, #2012, #2035 and #2392

# Long version:

Using `MessageQueue` currently requires quite some manual work and there also some bugs (see #2015, #2012, #2035, #2392). Time to live up to the wikis statement

> We plan to tightly couple it with telegram.Bot so that it could be used more conveniently (and even implicitly unless other specified).

So here is what my idea for a refactoring is. I'd be happy for feedback before getting started :)

## Refactoring of `MessageQueue` and `DelayQueue`

* `DQ.exc_route`: I'd like to rename to `error_handler` in accordance with `Dispatcher.add_error_handler`.
* If a `Dispatcher` is used, it would make sense to have the execution of the promises to behave like the refactored `Dipsatcher._pooled`, which also includes proper error handling. That can be easily achieved by passing the `Dispatcher` to either the `Bot` or the `MessageQueue`. Not sure yet, what makes more sense …
* I think I'd like to repace the `__call__` methods with `process` or whatever. That's more explicit and easier to read in the code.
* Add `MQ.add/remove_delay_queue` so that users can add custom delay queues in addition to the default & the group one. One should be able to specify a parent (defaulting to the default DQ) for `new_dq`, such that everything going throguh  `new_dq` will also go through the parent. `MQ.add_remove_delay_queue` sohuld also start/stop the `DQ` in case we're adding them at runtime. Edit: This will in fact close #1369 , as I just notice
* Add constants for the default & the group DQ
* Handle the stopping of `MQ` correctly, some hints are already given in #2012

## Integration into Bot

Add a `MessageQueue` to `Bot` by default and rebuild all methods in the fashion

```python

def badly_sketched_messagequeue_wrapper(func, *args, **kwargs):
    delay_queue = kwargs.pop('delay_queue')
    if delay_queue == self.message_queue.DEFAULT_DQ and int(kwargs.get('chat_id')) < 0: 
        return self.message_queue.process(func, self.message_queue.GROUP_DQ)
    elif delay_queue:
        return self.message_queue.process(func, delay_queue)
    return func(*args, **kwargs)

@badly_sketched_messagequeue_wrapper
def send_message(…, delay_queue=None):
    ...
```

such that `delay_queue` allows to select the delay queue the method is passed through, replacing the current `queue` and `isgroup` kwargs

* Together with the changes to `Bot.__new__` as in discussed #2015 , this should resolve #2015 
* for cases where one only needs `Bot`, without `Dispatcher` or `Updater`, one could do something like

```
def badly_sketched_messagequeue_wrapper(func, *args, **kwargs):
    if self.message_queue.is_running:
        # see above
    return func(*args, **kwargs)
```
so users can easily opt-into `MessageQueue` with `bot.mq.start/stop()`

## Integration with `Defaults`

Add `Defaults.delay_queue` defaulting to `None` (I'm not in favor of implicitly always using the `MessageQueue`)

## Handling context managers

I think the approach detailed in #2035 is feasible (build `InputFiles` already in `Promise.__init__`). I can't think of any use case, where a context manager would be used for anything else that we'd to handle …

Edit: Now I can. See #2160 

## Backwards compatibility

All of above changes should be doable without breaking changes.
Hower, given that `MessageQueue` is essentially still in the same quite rough state as when introduced in #537 3.5 years ago, IMHO it would be okay to removethe current usage pattern (i.e. the `@queuedmessage` decorator) in the next major release (which probably will be mostly a "delete all the stuff" release anyway …)

Bibo-Joshi:
Got started on close #2139

Still missing:

- [x] some docs
- [x] tests
- [x] `MQ.add/remove_delay_queue`
- [x] error handling
- [x] handle stopping of the MQ
- [x] use `telegram.constants` instead of hardcoding the defaults for max messages per interval
- [ ] Add tests for ensuring MQ works with defaults and also explicitly test for `Defaults.delay_queue(_per_method)`. Forgot about those before marking as ready for review …
- [x] Use the new versioning directives

#### Just for automatic issue closing:
closes #2015 closes #2012 closes #2035 closes #1369


telegram/bot.py
```diff
@@ -129,6 +129,52 @@ def decorator(self: 'Bot', *args: Any, **kwargs: Any) -> RT:  # pylint: disable=
...
+            if not chat_id:
+                is_group = False
+            elif isinstance(chat_id, str) and chat_id.startswith('@'):
+                is_group = True
+            else:
+                try:
+                    is_group = int(chat_id) < 0
+                except ValueError:
+                    is_group = False
+
```
Bibo-Joshi, line 165: Probably add some debug logs here …

Bibo-Joshi:
I just realized that the handling of streams this PR introduces is immature. It

* makes `Promise` assume that every stream is supposed to be converted into in `InputFile`, isn't necessarily the case for running custom functions through `Dispatcher.run_async`. Then again, using `Dispatcher.run_async` with context managers for reading files into custom functions will not work in either case …
* doesn't pass additional arguments to `InputFile` (or `parse_file_input`, which will be used after #2223)

Possible ways I see:

1. Just don't handle it. opening files is the only context manager situation we can control in the first place, so it may not be worth making an effort for it. We just document that context managers don't work with promises in general.
2. Try to handle at least file-context managers for bot methods. Sketch of an idea:

   ```python
   @mq(input_file_args={'media': {'filename': 'filename', attach=True, 'tg_type': Audio}})
   def bot_method(media, filename, *args):
       …
   ```

   so that the `@mq` decorator can do the handling of file input for the specified arguments, looking up values for `parse_file_input` in the other arguments. Ofc. this leads to some code duplication and is rather hacky.

For the moment I think I'd be fine with 1., i.e. just not handling it.

ChenZehong13 (approved):

Bibo-Joshi:
In fact just not supporting context managers will not lead to any major issues for users once #2230 is implemented as
```pyhon
with open('file', 'rb): as file:
    bot.send_document(chat_id, file)
```
can then be replaced by 
```pyhon
with open('file', 'rb): as file:
    bot.send_document(chat_id, file.read())
```

telegram/ext/messagequeue.py
```diff
@@ -48,21 +52,29 @@ class DelayQueue(threading.Thread):
...
         name (:obj:`str`): Thread's name.
+        error_handler (:obj:`callable`): Optional. A callable, accepting 1 positional argument.
+            Used to route exceptions from processor thread to main thread.
+        dispatcher (:class:`telegram.ext.Disptacher`): Optional. The dispatcher to use for error
+            handling.
 
     Args:
         queue (:obj:`Queue`, optional): Used to pass callbacks to thread. Creates ``Queue``
             implicitly if not provided.
+        parent (:class:`telegram.ext.DelayQueue`, optional): Pass another delay queue to put all
```
Bibo-Joshi, line 64: add a note that the parent should in most cases be the default queue

telegram/ext/messagequeue.py
```diff
@@ -141,130 +181,189 @@ def stop(self, timeout: float = None) -> None:
...
+        if exc_route and error_handler:
+            raise ValueError('Only one of exc_route or error_handler can be passed.')
+        if exc_route:
+            warnings.warn(
+                'The exc_route argument is deprecated. Use error_handler instead.',
+                TelegramDeprecationWarning,
+                stacklevel=2,
+            )
+
+        self._delay_queues: Dict[str, DelayQueue] = {
```
Bibo-Joshi, line 273: this dict should be accessible to make it easier to use existing delay queues as parents for new ones. Either just make `self._delay_queues` public, or even implement `__getitem__` to allow  for `mq['delay_queue_name']`

telegram/ext/messagequeue.py
```diff
@@ -141,130 +181,189 @@ def stop(self, timeout: float = None) -> None:
...
+    def set_dispatcher(self, dispatcher: 'Dispatcher') -> None:
+        """
+        Sets the dispatcher to use for error handling.
+
+        Args:
+            dispatcher (:class:`telegram.ext.Dispatcher`): The dispatcher.
+        """
+        self.dispatcher = dispatcher
+
+    DEFAULT_QUEUE: ClassVar[str] = 'default_delay_queue'
```
Bibo-Joshi, line 363: maybe add convenience properties `MQ.default/group_delay_queue` for accessing `MQ._delay_queue[MQ.DEFAULT/GROUP_QUEUE]`

Bibo-Joshi (commented):
Some more thoughts. Also we should think about #2265

Poolitzer:
ready for review now right? GitHub already complains about a merge confilct you might wanna solve that ;P

Bibo-Joshi:
Closing for now as we'll go a different route (see #2139 for further details). Might still be worth revisiting later.

PR #2163, opened 2020-10-30
Check caption in Filters.regex

marc0777:
Added regex matching for message caption in Filters.regex. 

I noticed that the regex filter only checks inside the message text, so if it's an image or a document it will be ignored, even if inside the caption the content could match.

Since we can already differentiate text messages and other contents using filters, I think the regex filter should work regardless of the type of message.

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -506,11 +506,25 @@ class Filters:
                 match = self.pattern.search(message.text)
                 if match:
                     return {'matches': [match]}
-            elif message.caption:
+            return {}
+
+    class caption_regex (MessageFilter):
+        data_filter = True
+
+        def __init__(self, pattern: Union[str, Pattern]):
+            if isinstance(pattern, str):
+                pattern = re.compile(pattern)
+            pattern = cast(Pattern, pattern)
+            self.pattern: Pattern = pattern
+            self.name = 'Filters.caption_regex ({})'.format(self.pattern)
+
+        def filter(self, message: Message) -> Optional[Dict[str, List[Match]]]:
+            if message.caption:
                 match = self.pattern.search(message.caption)
                 if match:
                     return {'matches': [match]}
             return {}
+
 
     class _Reply(MessageFilter):
         name = 'Filters.reply'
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
Hi. Thanks for the PR! I do see the use case for regexing a caption, but the change you propose would be breaking. I.e. `MessageHandler(Filters.regex(…), …)` currently only accepts text messages and with your PR it would suddently also accept media.

My first idea would be to move it to a standalone `Filters.caption_regex` like for `Filters.caption_entitiy`. Then you could still do `Filters.regex(…) | Filters.caption_regex(…)` to allow both.
The only downside I see is that people might get confused that `Filters.photo & Filters.regex` will not do the expected thing, while `Filters.photo & Filters.caption_regex` does. But I guess I can live with that.

Maybe @Poolitzer has an opinion on that?

In any case, we'd need unittests (see Codecov failure) - maybe have a look at `tests/test_filters.py` for inspiration :)

Poolitzer:
> The only downside I see is that people might get confused that Filters.photo & Filters.regex will not do the expected thing, while Filters.photo & Filters.caption_regex does. But I guess I can live with that.
>
> Maybe @Poolitzer has an opinion on that?

Nope that would be fine for me. We could mention it in Filters.photo doc but we dont do with text either so I think that would be the better solution.

Bibo-Joshi (changes requested):
Thanks for the update! I take it that the tests are basically copy-pasted from Filters.regex, so those should be fine. Two points missing from what I see:

1. docstring for the new filter (you can keep it shorter than the one of `Filters.regex`, just say "works the same" and link to `Filters.regex`)
2. please make sure that pre-commit is happy. have a look at the contribution guide for how to set that up locally.

marc0777:
Of course, next step is docstring! 

I had some trouble configuring the pre-commit on my machine: it did succeed locally, yet the result was different. Now it should be fine (hopefully).

Bibo-Joshi:
> I had some trouble configuring the pre-commit on my machine: it did succeed locally, yet the result was different. Now it should be fine (hopefully).

Maybe merge from master. We recently changed some of the config …

marc0777:
Now everything should be fine, however merging from master made test-official fail, despite me not having touched anything that should trigger it.

Let me know what you think!

Bibo-Joshi:
Yeah, TG just released API 5.0 today, so test official will be failing until we incorporate the changes … nothing to worry about. Looks good, I'll merg :) Thanks for the contribution!

Issue #1881, opened 2020-04-08
Use concrete types instead of iterable


Bibo-Joshi:
Uses of "Iterable" I found:

* Filters.text (+ tests)
* Filters.caption (+ tests)
* Updater.idle

Poolitzer:
@Bibo-Joshi we have iterables at jobs?

PR #2167, opened 2020-10-31
Add `Tuple` option to arguments type hints

### LINKED ISSUE #2125: [BUG] tuples aren't accepted in static typing
Currently, we only accept list where tuples work as well. That should be added to the type hints where applicable.

eIGato:
closes #2125.

Based on https://github.com/python-telegram-bot/python-telegram-bot/pull/2152 with a little fix. closes #2152

### DIFF
--- a/telegram/ext/commandhandler.py
+++ b/telegram/ext/commandhandler.py
@@ -132,7 +132,7 @@ class CommandHandler(Handler):
 
     def __init__(
         self,
-        command: Union[str, List[str]],
+        command: Union[str, List[str], Tuple[str]],
         callback: Callable[[HandlerArg, 'CallbackContext'], RT],
         filters: BaseFilter = None,
         allow_edited: bool = None,
@@ -340,8 +340,8 @@ class PrefixHandler(CommandHandler):
 
     def __init__(
         self,
-        prefix: Union[str, List[str]],
-        command: Union[str, List[str]],
+        prefix: Union[str, List[str], Tuple[str]],
+        command: Union[str, List[str], Tuple[str]],
         callback: Callable[[HandlerArg, 'CallbackContext'], RT],
         filters: BaseFilter = None,
         pass_args: bool = False,
@@ -384,7 +384,7 @@ class PrefixHandler(CommandHandler):
         return self._prefix
 
     @prefix.setter
-    def prefix(self, prefix: Union[str, List[str]]) -> None:
+    def prefix(self, prefix: Union[str, List[str], Tuple[str]]) -> None:
         if isinstance(prefix, str):
             self._prefix = [prefix.lower()]
         else:
@@ -402,7 +402,7 @@ class PrefixHandler(CommandHandler):
         return self._command
 
     @command.setter
-    def command(self, command: Union[str, List[str]]) -> None:
+    def command(self, command: Union[str, List[str], Tuple[str]]) -> None:
         if isinstance(command, str):
             self._command = [command.lower()]
         else:
--- a/telegram/ext/commandhandler.py
+++ b/telegram/ext/commandhandler.py
@@ -385,13 +385,11 @@ class PrefixHandler(CommandHandler):
         return self._prefix
 
     @prefix.setter
-    def prefix(self, prefix: SLT[str]) -> None:
+    def prefix(self, prefix: Union[str, List[str]]) -> None:
         if isinstance(prefix, str):
             self._prefix = [prefix.lower()]
-        elif isinstance(prefix, list):
-            self._prefix = prefix
         else:
-            self._prefix = list(prefix)
+            self._prefix = prefix
         self._build_commands()
 
     @property  # type: ignore[override]
@@ -405,13 +403,11 @@ class PrefixHandler(CommandHandler):
         return self._command
 
     @command.setter
-    def command(self, command: SLT[str]) -> None:
+    def command(self, command: Union[str, List[str]]) -> None:
         if isinstance(command, str):
             self._command = [command.lower()]
-        elif isinstance(command, list):
-            self._command = command
         else:
-            self._command = list(command)
+            self._command = command
         self._build_commands()
 
     def _build_commands(self) -> None:
--- a/telegram/ext/commandhandler.py
+++ b/telegram/ext/commandhandler.py
@@ -49,7 +49,7 @@ class CommandHandler(Handler):
         :class:`telegram.ext.CommandHandler` does *not* handle (edited) channel posts.
 
     Attributes:
-        command (:class:`telegram.utils.types.SLT`[:obj:`str`]):
+        command (:class:`telegram.utils.types.SLT[str]`):
             The command or list of commands this handler should listen for.
             Limitations are the same as described here https://core.telegram.org/bots#commands
         callback (:obj:`callable`): The callback function for this handler.
@@ -83,7 +83,7 @@ class CommandHandler(Handler):
         attributes to :class:`telegram.ext.CallbackContext`. See its docs for more info.
 
     Args:
-        command (:class:`telegram.utils.types.SLT`[:obj:`str`]):
+        command (:class:`telegram.utils.types.SLT[str]`):
             The command or list of commands this handler should listen for.
             Limitations are the same as described here https://core.telegram.org/bots#commands
         callback (:obj:`callable`): The callback function for this handler. Will be called when
@@ -297,9 +297,9 @@ class PrefixHandler(CommandHandler):
         attributes to :class:`telegram.ext.CallbackContext`. See its docs for more info.
 
     Args:
-        prefix (:class:`telegram.utils.types.SLT`[:obj:`str`]):
+        prefix (:class:`telegram.utils.types.SLT[str]`):
             The prefix(es) that will precede :attr:`command`.
-        command (:class:`telegram.utils.types.SLT`[:obj:`str`]):
+        command (:class:`telegram.utils.types.SLT[str]`):
             The command or list of commands this handler should listen for.
         callback (:obj:`callable`): The callback function for this handler. Will be called when
             :attr:`check_update` has determined that an update should be processed by this handler.
--- a/telegram/ext/commandhandler.py
+++ b/telegram/ext/commandhandler.py
@@ -49,9 +49,9 @@ class CommandHandler(Handler):
         :class:`telegram.ext.CommandHandler` does *not* handle (edited) channel posts.
 
     Attributes:
-        command (SLT[:obj:`str`]): The command or list of commands this handler
-            should listen for. Limitations are the same as described here
-            https://core.telegram.org/bots#commands
+        command (:class:`telegram.utils.types.SLT`[:obj:`str`]):
+            The command or list of commands this handler should listen for.
+            Limitations are the same as described here https://core.telegram.org/bots#commands
         callback (:obj:`callable`): The callback function for this handler.
         filters (:class:`telegram.ext.BaseFilter`): Optional. Only allow updates with these
             Filters.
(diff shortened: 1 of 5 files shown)


Bibo-Joshi:
Thanks for the new PR! I edited the PR description, so beth the issue and the old PR will be closed, when merging.
Can you please merge from master? Besides resolving the conflicts that should fix the tests, so we can see better what's going on :)

eIGato:
Okay. I've just rebased the branch to current master and fixed the conflicts.
I see the monthly job tests are fixed already. I've just started to fix them.
The only thing left is to add tests for new "if-else" branches.

Bibo-Joshi (changes requested):
I had a look at the changes and left 2 comments. Some more general notes:

* I think it's worth introducing a shortcut type for `Union[T, List[T], Tuple[T]]` in `telegram.utils.types`. Maybe `STL` for single, tuple, list? that should shave some typing for future changes
* Use of `Iterable`: see my comment
* I don't think that this PR should make any changes to non-signature code or tests. Basically the changes in this PR should not cause mypy to be unhappy. Adding tuple type hints where possible is merely a convenience extension of documentation and should not change the actual implementation.

telegram/ext/filters.py
```diff
@@ -381,8 +401,8 @@ class _Caption(MessageFilter):
         name = 'Filters.caption'
 
         class _CaptionStrings(MessageFilter):
-            def __init__(self, strings: List[str]):
-                self.strings = strings
+            def __init__(self, strings: Iterable[str]):
```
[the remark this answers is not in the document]
eIGato: It cannot lead to problems according to the implementation. Make input requirements as loose as possible and make output requirements as strict as possible. The only requirement here is that i can make a list from the argument. But if you insist, i'll do it.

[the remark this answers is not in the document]
Bibo-Joshi: it can lead to problems: try `list(itertools.count())`. Anyway, we've made the decision before (#1881) and I'd like to stick with it.

telegram/ext/filters.py
```diff
@@ -207,7 +219,10 @@ class MergedFilter(UpdateFilter):
     """
 
     def __init__(
-        self, base_filter: BaseFilter, and_filter: BaseFilter = None, or_filter: BaseFilter = None
+        self,
```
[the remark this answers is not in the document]
eIGato: Oh, just a misclick. I've seen something about too long line 2xx in pre-commit hook and edited a wrong line. I'll rollback this in the evening (UTC).

eIGato:
About changing the implementation. When we started to take `Tuple`, some calls to external libs became mypy-invalid because some of them require only `List`. So i converted tuples to lists where possible.
And do you actually think that a list passed to `__init__()` should ever be preserved as-is when assigned to instance attr? It's understandable that property setters preserve them. But `__init__()` makes a brand-new instance with brand-new attrs. Preserving lists may lead to bad side-effects that would be hard to find.

telegram/ext/filters.py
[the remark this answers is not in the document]
eIGato: Sorry, i lied about the evening.

eIGato:
"STL" sounds too "Std Template Lib", so i chose "SLT".

Bibo-Joshi:
> About changing the implementation. When we started to take `Tuple`, some calls to external libs became mypy-invalid because some of them require only `List`. So i converted tuples to lists where possible.

Frankly, compatibility with 3rd party libs shouldn't be the first priority when making changes. If they need a list, convert the input yourself, that's not PTBs job ;)

> And do you actually think that a list passed to `__init__()` should ever be preserved as-is when assigned to instance attr? It's understandable that property setters preserve them. But `__init__()` makes a brand-new instance with brand-new attrs. Preserving lists may lead to bad side-effects that would be hard to find.

You have a point here. However, changing

```
def __init__(self, some_list: List):
    self.some_list = some_list
```

to 

```
def __init__(self, some_list: List):
    self.some_list = list(some_list)
```
everywhere in the code is a largely independent issue. As of now the aim of this PR is just to make type hinting a bit more generous. So please either only change signatures or go the full way and adjust the behaviour of every `__init__` in the lib ;)

I'll try to have a closer look at the new changes soon.

eIGato:
MyPy check fails on third-party lib calls inside PTB. If you still can merge such PR, so okay.
> or go the full way

You know how to convince people (at least my kind). If you could ignore MyPy or loosen its checks, then say it, and i'll roll the implementation changes back. Even the generator optimization.

Bibo-Joshi:
> MyPy check fails on third-party lib calls inside PTB. If you still can merge such PR, so okay.

Okay, apparently I misunderstood that. I had understood that mypy pecasomes invalid when using the changes with some 3rd party libs used in one of your projects. But in this case, I'll resort to "Basically the changes in this PR should not cause mypy to be unhappy. "

eIGato:
Are there any change request left?

telegram/ext/commandhandler.py
```diff
@@ -49,7 +49,7 @@ class CommandHandler(Handler):
         :class:`telegram.ext.CommandHandler` does *not* handle (edited) channel posts.
 
     Attributes:
-        command (:obj:`str` | List[:obj:`str`]): The command or list of commands this handler
+        command (SLT[:obj:`str`]): The command or list of commands this handler
```
Bibo-Joshi, line 52: Could you change those to ``:obj:`str` | List/Tuple[:obj:`str`]`` or ``:class:`telegram.utils.types.SLT`[:obj:`str`]``? People don't know what `SLT` is and will get confused

telegram/ext/filters.py
```diff
@@ -24,7 +24,19 @@
...
+    FrozenSet,
+    Iterable,
+    List,
+    Match,
+    Optional,
+    Pattern,
+    Set,
+    Tuple,
+    Union,
+    cast,
```
Bibo-Joshi, line 38: IISC this change is only due to the `Iterable` import, which we don't need

telegram/utils/types.py
```diff
@@ -17,11 +17,13 @@
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains custom typing aliases."""
-from typing import IO, TYPE_CHECKING, Any, Dict, Optional, Tuple, Union
+from typing import IO, TYPE_CHECKING, Any, Dict, List, Optional, Tuple, TypeVar, Union
 
 if TYPE_CHECKING:
     from telegram import InputFile, Update
 
+RT = TypeVar("RT")
```
Bibo-Joshi, line 25: please move this down to SLT, where it's acutally used atm

tests/test_commandhandler.py
```diff
@@ -397,13 +397,19 @@ def test_edit_prefix(self):
         handler = self.make_default_handler()
         handler.prefix = ['?', '§']
         assert handler._commands == list(combinations(['?', '§'], self.COMMANDS))
+        handler.prefix = ('!', '^')
```
Bibo-Joshi, line 400: new tests must not be necessary

Bibo-Joshi (changes requested):
> Are there any change requests left?

yes. aside from some minor comments, there is still use of `Iterable` and changes of non-signature code ;)

tests/test_commandhandler.py
eIGato: They ARE necessary. Test coverage check fails without them.

telegram/ext/filters.py
eIGato: And `Tuple` which we need. But without `Iterable` it fits in one line.

eIGato:
The only unnecessary change here was the generator optimization. I've removed it.

tests/test_commandhandler.py
Bibo-Joshi: You're missing the point. As I've tried to explain, this PR should not include changes, that make non-signature code changes (and thus new tests) necessary. Specifically: If changing the signatures of `PrefixHandlers` setters requires code-changes to make MyPy happy, don't change them.

eIGato: I've explained it to you before.
Change type hints
-> MyPy fails on third-party calls
-> convert everything to list
-> CodeCov fails
-> add new tests
-> author of the CodeCov and MyPy configs says it wasn't necessary.
It was you who configured these checks. And now you tell me not to pass them.

Bibo-Joshi: I'm not telling you not to pass the tests. I'm telling you not to change signatures (i.e. change type hints) if that leads to the tests failing.
there is a difference between "not necessary" and "not wanted" ;)

eIGato: The setters were changed in the original PR and it didn't bother you then. Why now?

eIGato: If i roll back all signature changes that require non-signature changes, the PR would be empty.

Bibo-Joshi: > The setters were changed in the original PR and it didn't bother you then. Why now?

citing my comment on #2152:

> I didn't go over i[t] in detail, yet [...]
>
>* Tuples should only be allowed where it really doesn't matter if it's a tuple or a list [...]

regarding:

> If i roll back all signature changes that require non-signature changes, the PR would be empty.

By now I have the impression that you intentionally try to misunderstand me. I'll try to specify my point one more time:

If you change

```python
def foo(var: List[T]):
```
to 

```python
def foo(var: Union[List[T], Tuple[T]]):
```
and the tests still run with that change, it's fine for this PR and this is what I mean by a "signature change" (because you change the signature of a function).
If the tests fail after that change, please don't include the change in this PR.
If you change something that is *not* the signature of a function and not docstring, I consider that a "non-signature change".

Long story very short and imprecise in a first-impression-comment on a PR with failing tests: "Tuples should only be allowed where it really doesn't matter if it's a tuple or a list" (see above)

telegram/ext/filters.py
```diff
@@ -280,8 +281,13 @@ def __init__(self, emoji: str = None, name: str = None):
...
     class _DiceValues(MessageFilter):
-        def __init__(self, values: Union[int, List[int]], name: str, emoji: str = None):
-            self.values = [values] if isinstance(values, int) else values
+        def __init__(
+            self,
+            values: SLT[int],
+            name: str,
+            emoji: str = None,
+        ):
+            self.values = [values] if isinstance(values, int) else list(values)
```
Bibo-Joshi, line 290: .

telegram/ext/filters.py
```diff
@@ -330,8 +336,8 @@ class _Text(MessageFilter):
         name = 'Filters.text'
 
         class _TextStrings(MessageFilter):
-            def __init__(self, strings: List[str]):
-                self.strings = strings
+            def __init__(self, strings: Union[List[str], Tuple[str]]):
+                self.strings = list(strings)
```
Bibo-Joshi, line 340: .

telegram/ext/filters.py
```diff
@@ -381,8 +387,8 @@ class _Caption(MessageFilter):
         name = 'Filters.caption'
 
         class _CaptionStrings(MessageFilter):
-            def __init__(self, strings: List[str]):
-                self.strings = strings
+            def __init__(self, strings: Union[List[str], Tuple[str]]):
+                self.strings = list(strings)
```
Bibo-Joshi, line 391: .

Bibo-Joshi (changes requested):
Did you check if the ``:class:`telegram.utils.types.SLT`[:obj:`int`]`` thingies render correctly when building the docs?

eIGato:
About rendering thingies. Could not find a way to render them as hyperlinks and maintain readability of the remainder.

eIGato:
@Poolitzer could you review this PR too?

Poolitzer (approved):
LGTM

eIGato:
Hi. I'm glad the PRs are merged. But they still don't count for Hacktoberfest for some reason. Could you please add "hacktoberfest-accepted" label to both PRs?

eIGato:
Thank you @Bibo-Joshi. But it didn't work. Hacktoberfest still doesn't count them. May be this is because i submitted these PRs too late (in the evening of 31th of October). Let's pretend this time i contibuted not for the tee)).
image

PR #2156, opened 2020-10-19
Added: Filter for file ending/extension

mlemosf:
# Feature: Added filter for file endings / file extensions on documents

Added class 'file_ending' that filter documents which end in the
extension passed by the user to Filters.document.file_ending;  
Also added name to AUTHORS.rst.

This pull request solves issue #2140

PR #2169, opened 2020-10-31
Add file extension filter

### LINKED ISSUE #2140: [FEATURE] introduce a file ending filter
Users might want to filter incoming updates based on the file endings on top of/instead of the mime type. Both attributes of documents are not safe, but that didn't stop us from introducing a mime type want and shouldn't stop us from a file ending one either. Both are a great way of initially filtering incoming files. We should just mention in the docs, as we do with mime types, that telegram doesnt offer any validation.

This filter can either be a subset of Filters.document or its own imo.

eIGato:
Resolve #2140.
Based on an abandoned PR of another contributor.
Fix #2156.

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -615,6 +615,8 @@ class Filters:
                     you may change it with the flag ``case_sensitive=True``.
                 * Extension should be passed without leading dot
                     unless it's a part of the extension.
+                * Pass :obj:`None` to filter files with no extension,
+                    even empty one. E.g. without a dot in the filename.
 
             Example:
                 ``Filters.document.file_extension("jpg")``
@@ -625,13 +627,11 @@ class Filters:
                     filters files with extension ``".Dockerfile"`` minding the case.
             """
 
-            def __init__(
-                self, file_extension: Optional[str] = None, *, case_sensitive: bool = False
-            ):
+            def __init__(self, file_extension: Optional[str], *, case_sensitive: bool = False):
                 """Initialize the extension you want to filter.
 
                 Args:
-                    file_extension (:obj:`str`, optional):
+                    file_extension (:obj:`str` | :obj:`None`):
                         media file extension you want to filter.
                     case_sensitive (:obj:bool, optional):
                         pass :obj:`True` to make the filter case sensitive.
@@ -640,7 +640,7 @@ class Filters:
                 self.is_case_sensitive = case_sensitive
                 if file_extension is None:
                     self.file_extension = None
-                    self.name = "Filters.document.file_extension()"
+                    self.name = "Filters.document.file_extension(None)"
                 elif case_sensitive:
                     self.file_extension = f".{file_extension}"
                     self.name = (
@@ -656,7 +656,7 @@ class Filters:
                 if message.document is None:
                     return False
                 if self.file_extension is None:
-                    return True
+                    return "." not in message.document.file_name
                 if self.is_case_sensitive:
                     filename = message.document.file_name
                 else:
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -603,31 +603,31 @@ class Filters:
         xml = mime_type('application/xml')
         zip = mime_type('application/zip')
 
-        class file_ending(MessageFilter):
+        class file_extension(MessageFilter):
             """ This filter filters documents by their file ending/extension
 
             Note:
                 This Filter only filters by the file ending/extension of the document,
                     it doesn't check the validity of document.
-                The user can manipulate the mime-type of a message and
+                The user can manipulate the file extension of a document and
                     send media with wrong types that don't fit to this handler.
 
             Example:
-                ``Filters.document.file_ending('jpg')`` filters all files with extension .jpg
+                ``Filters.document.file_extension('jpg')`` filters all files with extension ``.jpg``
             """
 
-            def __init__(self, file_ending: Optional[str]):
-                """Initialize the category you want to filter
+            def __init__(self, file_extension: Optional[str]):
+                """Initialize the extension you want to filter
 
                 Args:
-                    file_ending (str, optional): file ending of the media you want to filter"""
-                self.file_ending = file_ending
-                self.name = "Filters.document.file_ending('{}')".format(self.file_ending)
+                    file_extension (str, optional): file extension of the media you want to filter"""
+                self.file_extension = file_extension
+                self.name = f"Filters.document.file_extenision('{self.file_extension}')"
 
             def filter(self, message : Message) -> bool:
                 """"""  # remove method from docs
                 if message.document:
-                    return message.document.file_name.endswith(self.file_ending)
+                    return message.document.file_name.endswith(self.file_extension)
                 return False
 
         def filter(self, message: Message) -> bool:
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -608,25 +608,25 @@ class Filters:
 
             Note:
                 * This Filter only filters by the file ending/extension of the document,
-                    it doesn't check the validity of document.
+                  it doesn't check the validity of document.
                 * The user can manipulate the file extension of a document and
-                    send media with wrong types that don't fit to this handler.
+                  send media with wrong types that don't fit to this handler.
                 * Case insensitive by default,
-                    you may change this with the flag ``case_sensitive=True``.
+                  you may change this with the flag ``case_sensitive=True``.
                 * Extension should be passed without leading dot
-                    unless it's a part of the extension.
+                  unless it's a part of the extension.
                 * Pass :obj:`None` to filter files with no extension,
-                    i.e. without a dot in the filename.
+                  i.e. without a dot in the filename.
 
             Example:
-                ``Filters.document.file_extension("jpg")``
-                    filters files with extension ``".jpg"``.
-                ``Filters.document.file_extension(".jpg")``
-                    filters files with extension ``"..jpg"``.
-                ``Filters.document.file_extension("Dockerfile", case_sensitive=True)``
-                    filters files with extension ``".Dockerfile"`` minding the case.
-                ``Filters.document.file_extension(None)``
-                    filters files without a dot in the filename.
+                * ``Filters.document.file_extension("jpg")``
+                  filters files with extension ``".jpg"``.
+                * ``Filters.document.file_extension(".jpg")``
+                  filters files with extension ``"..jpg"``.
+                * ``Filters.document.file_extension("Dockerfile", case_sensitive=True)``
+                  filters files with extension ``".Dockerfile"`` minding the case.
+                * ``Filters.document.file_extension(None)``
+                  filters files without a dot in the filename.
             """
 
             def __init__(self, file_extension: Optional[str], case_sensitive: bool = False):
@@ -725,25 +725,25 @@ officedocument.wordprocessingml.document")``-
 
             Note:
                 * This Filter only filters by the file ending/extension of the document,
-                    it doesn't check the validity of document.
+                  it doesn't check the validity of document.
                 * The user can manipulate the file extension of a document and
-                    send media with wrong types that don't fit to this handler.
+                  send media with wrong types that don't fit to this handler.
                 * Case insensitive by default,
-                    you may change this with the flag ``case_sensitive=True``.
+                  you may change this with the flag ``case_sensitive=True``.
                 * Extension should be passed without leading dot
-                    unless it's a part of the extension.
+                  unless it's a part of the extension.
                 * Pass :obj:`None` to filter files with no extension,
-                    i.e. without a dot in the filename.
+                  i.e. without a dot in the filename.
 
             Example:
-                ``Filters.document.file_extension("jpg")``
-                    filters files with extension ``".jpg"``.
-                ``Filters.document.file_extension(".jpg")``
-                    filters files with extension ``"..jpg"``.
-                ``Filters.document.file_extension("Dockerfile", case_sensitive=True)``
-                    filters files with extension ``".Dockerfile"`` minding the case.
-                ``Filters.document.file_extension(None)``
-                    filters files without a dot in the filename.
+                * ``Filters.document.file_extension("jpg")``
+                  filters files with extension ``".jpg"``.
+                * ``Filters.document.file_extension(".jpg")``
+                  filters files with extension ``"..jpg"``.
+                * ``Filters.document.file_extension("Dockerfile", case_sensitive=True)``
+                  filters files with extension ``".Dockerfile"`` minding the case.
+                * ``Filters.document.file_extension(None)``
+                  filters files without a dot in the filename.
     """
 
     class _Animation(MessageFilter):
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -627,7 +627,7 @@ class Filters:
                     filters files with extension ``".Dockerfile"`` minding the case.
             """
 
-            def __init__(self, file_extension: Optional[str], *, case_sensitive: bool = False):
+            def __init__(self, file_extension: Optional[str], case_sensitive: bool = False):
                 """Initialize the extension you want to filter.
 
                 Args:
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -625,11 +625,13 @@ class Filters:
                     filters files with extension ``".Dockerfile"`` minding the case.
             """
 
-            def __init__(self, file_extension: Optional[str], *, case_sensitive: bool = False):
+            def __init__(
+                self, file_extension: Optional[str] = None, *, case_sensitive: bool = False
+            ):
                 """Initialize the extension you want to filter.
 
                 Args:
-                    file_extension (:obj:`str` | :obj:`None`):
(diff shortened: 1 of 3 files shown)


eIGato:
@Bibo-Joshi may you restart failed tests? I don't think they failed because of these changes. More likely, some temporary issue with test env.
Besides, no need to edit keywords in this PR. "Resolve" and "Fix" are fine too.

eIGato:
Thanks.

Bibo-Joshi:
NP. I'll try to review both PRs tomorrow ;)

telegram/ext/filters.py
```diff
@@ -603,6 +603,45 @@ def filter(self, message: Message) -> bool:
...
+                This Filter only filters by the file ending/extension of the document,
+                    it doesn't check the validity of document.
+                The user can manipulate the file extension of a document and
+                    send media with wrong types that don't fit to this handler.
+
+            Example:
+                ``Filters.document.file_extension('jpg')`` filters files with extension ``.jpg``.
+            """
+
+            file_extension: Optional[List[str]] = None
```
Bibo-Joshi, line 619: What is this good for?

telegram/ext/filters.py
```diff
@@ -603,6 +603,45 @@ def filter(self, message: Message) -> bool:
...
+                if message.document is None:
+                    return False
+                if self.file_extension is None:
+                    return True
+                filename_parts = message.document.file_name.split(".")
+                extension_len = len(self.file_extension)
+                if len(filename_parts) - extension_len < 1:
+                    # Exclude extension-only filenames.
+                    return False
+                return filename_parts[-extension_len:] == self.file_extension
```
Bibo-Joshi, line 643: why the hole splitting? `return message.document.file_name.endswith(f'.{self.file_extension}')` should do the trick

tests/test_filters.py
```diff
@@ -425,6 +425,43 @@ def test_filters_document_type(self, update):
         assert Filters.document.category("application/")(update)
         assert Filters.document.mime_type("application/x-sh")(update)
 
+    def test_filters_file_extension(self, update):
```
Bibo-Joshi, line 428: please add a test that makes sure the dot before the extension is respected, i.e. `file.foojpg` should not pass `Filters.file_extension('.jpg')`

telegram/ext/filters.py
```diff
@@ -603,6 +603,45 @@ def filter(self, message: Message) -> bool:
...
+            file_extension: Optional[List[str]] = None
+
+            def __init__(self, file_extension: Optional[str]):
+                """Initialize the extension you want to filter.
+
+                Args:
+                    file_extension (str, optional): media file extension you want to filter.
+                """
+                if file_extension is not None:
+                    file_extension = file_extension.lstrip(".")
```
Bibo-Joshi, line 628: I do see why you want the user to not have to care about the dot, but I'd prefer to not fiddle with it and instead just document if the dot must be passed or not. Image someone for some reason wanting to filter for a `......ext` extension. That wouldn't work in the current setting.

Bibo-Joshi (changes requested):

tests/test_filters.py
eIGato: I've already done it.

telegram/ext/filters.py
eIGato: This is the case i didn't think of. Okay.

telegram/ext/filters.py
eIGato: To not assing the attr in `__init__()` if it's None. Not much good, i'll might edit this.

Bibo-Joshi: please do ;)

telegram/ext/filters.py
eIGato: It mattered when i `lstrip()`ed dots.

eIGato:
I think everything is fixed here.

telegram/ext/filters.py
```diff
@@ -603,6 +603,41 @@ def filter(self, message: Message) -> bool:
...
+            """This filter filters documents by their file ending/extension.
+
+            Note:
+                This Filter only filters by the file ending/extension of the document,
+                    it doesn't check the validity of document.
+                The user can manipulate the file extension of a document and
+                    send media with wrong types that don't fit to this handler.
+
+            Example:
+                ``Filters.document.file_extension('jpg')`` filters files with extension ``.jpg``.
```
Bibo-Joshi, line 616:
```suggestion
                ``Filters.document.file_extension('jpg')`` filters files with extension ``'.jpg'``.
```

telegram/ext/filters.py
```diff
@@ -603,6 +603,41 @@ def filter(self, message: Message) -> bool:
...
+                Args:
+                    file_extension (str, optional): media file extension you want to filter.
+                """
+                if file_extension is None:
+                    self.file_extension = file_extension
+                elif file_extension.startswith("."):
+                    self.file_extension = file_extension
+                else:
+                    self.file_extension = f".{file_extension}"
+                self.name = f"Filters.document.file_extension({self.file_extension!r})"
```
Bibo-Joshi, line 631: why is the `!r` necessary?

telegram/ext/filters.py
```diff
@@ -603,6 +603,41 @@ def filter(self, message: Message) -> bool:
...
+
+                Args:
+                    file_extension (str, optional): media file extension you want to filter.
+                """
+                if file_extension is None:
+                    self.file_extension = file_extension
+                elif file_extension.startswith("."):
+                    self.file_extension = file_extension
+                else:
+                    self.file_extension = f".{file_extension}"
```
Bibo-Joshi, line 630: my point was that this should only be either

* `self.file_extension = file_extension` and the docs should nota that the user needs to care abot the dot
* `self.file_extension = f'.{file_extension}'` and the docs should note that a dot will be added, not matter what

Bibo-Joshi (changes requested):
As a side note: Please don't force push if not necessary. It makes reviewing harder.

telegram/ext/filters.py
eIGato: Because it should differ for `None` and `"None"`.

eIGato:
What about case sensitivity? I've made it insensitive but may be there are cases when it should not be done. For example `*.Dockerfile`. May be make it optional or just score on this?

eIGato:
Sorry for force-pushing. I'm used to Gitlab where one can view diffs of different pushed versions. I just want to bring logically connected changes to one commit.

telegram/ext/filters.py
Bibo-Joshi: but `str(None) == repr(None)`? o.0 shouldn't it rather be `f'{self.f_e if self.f_e else ""}'`, then?

Bibo-Joshi:
> What about case sensitivity? I've made it insensitive but may be there are cases when it should not be done. For example `*.Dockerfile`. May be make it optional or just score on this?

Mh, good point. While intuitively I'd say lower case everything, a quick search tells me that file extension case sensitivity can depend on for operating system and also hard drive encoding or something. So I'd like to go for a flag, defaulting to lowercasing everythng. have a look at the implementation of `Filters.command` for that. Please add something like 

```
Note:
  * adds the dot automatically, so *insert .jpg example here*
  * case insensitive by default, use the flag *whatever it's called*, if you need it
```

to the docstring

> Sorry for force-pushing. I'm used to Gitlab where one can view diffs of different pushed versions. I just want to bring logically connected changes to one commit.

NP. We squash anyway before merging, so that doesn't really matter too much ;)

telegram/ext/filters.py
eIGato: `str("None") != repr("None")` — that's what i mean. Please read the test to understand my intentions. May be i just got the task wrong. In short, `None` and `""` behaviors differ: `""` means filenames ending with a dot and `None` means empty filter that always returns `True` (except if there's no filename at all).

Bibo-Joshi: I do understand that `'None'` and `None` are not the same as input. Still I find `file_extension(None)` resulting in `'Filters.doucment.file_extension(None)'` a bis misleading as name (the difference to `"file_extension('None')"` being marginal and would prefer `'Filters.doucment.file_extension()'`. Anyway, I'm not too passionate about that, but it would maybe be good to explicitly ad tests checking that the name is set correctly

eIGato: Should we accept `None` at all? Argument is not optional, so `None` can only be passed explicitly. Adding `file_extension()` to the name may be misleading.

eIGato: Made it optional and removed `None` from name. But behavior question is still open. I've separated all `None`-related checks in one test, please read.

Bibo-Joshi: No, I think `None` should have to be passed explicitly. Though thinking about it, it would actually make sense to not just return `True` in that case but on the contrary, check that the file has *no* file extension, i.e. `return '.' not in file_extension`. Ofc that should be documented. 
In that case I do agree that `Filters.f_e(None)` would be a more appropriate naming.
I'll try to review this and the types PR later today

telegram/ext/filters.py
```diff
@@ -603,6 +603,66 @@ def filter(self, message: Message) -> bool:
...
+                * This Filter only filters by the file ending/extension of the document,
+                    it doesn't check the validity of document.
+                * The user can manipulate the file extension of a document and
+                    send media with wrong types that don't fit to this handler.
+                * Case insensitive by default,
+                    you may change it with the flag ``case_sensitive=True``.
+                * Extension should be passed without leading dot
+                    unless it's a part of the extension.
+                * Pass :obj:`None` to filter files with no extension,
+                    even empty one. E.g. without a dot in the filename.
```
Bibo-Joshi, line 619:
```suggestion
                    i.e. without a dot in the filename.
```

telegram/ext/filters.py
```diff
@@ -603,6 +603,66 @@ def filter(self, message: Message) -> bool:
...
+            Example:
+                ``Filters.document.file_extension("jpg")``
+                    filters files with extension ``".jpg"``.
+                ``Filters.document.file_extension(".jpg")``
+                    filters files with extension ``"..jpg"``.
+                ``Filters.document.file_extension("Dockerfile", case_sensitive=True)``
+                    filters files with extension ``".Dockerfile"`` minding the case.
+            """
+
+            def __init__(self, file_extension: Optional[str], *, case_sensitive: bool = False):
```
Bibo-Joshi, line 630: I don't think the `*` is necessary

Bibo-Joshi (changes requested):
Thanks for the updates! Just two nitpicks left from my side. @Poolitzer do you want to review, too?

telegram/ext/filters.py
eIGato: Calling `Filter.file_extension('jpg', True)` is just insensible.

Bibo-Joshi: We don't use this kind of signatures anywhere else in the lib and I don't want to add it just in one place. As with `__init__` and lists: either change it everywhere (which is breaking and needs to be discussed), or not at all.
Aside from that, PTB is not about handholding its users.

eIGato:
Cite from your contibuting guidelines: "This gives us the flexibility to re-order arguments and more importantly to add new required arguments".
Why shoul it refer only to calls but not the signatures?

Bibo-Joshi:
> Cite from your contibuting guidelines: "This gives us the flexibility to re-order arguments and more importantly to add new required arguments".
> Why should it refer only to calls but not the signatures?

Well, that style guideline applies only to code within PTB itself and not to how users should use the library. So not having `, *,` in the signatures allows the user to use both.
I do see the advantage of enforcing kwargs in calls, but I also see the philosophy of not handholding the user and the fact that it's a breaking change that should be done consistently throughout the lib.
If it's important to you, please go ahead and open a new issue so that it can be properly discussed, as it's beyond the scope fo this PR.

eIGato:
It's a normal approach: use a good practice for the new code and fix the old code later.
Did as you wish. Any nitpicks left?

Bibo-Joshi:
not from my side, but Poolitzers review is still pending. Same for #2167 btw.

telegram/ext/filters.py
```diff
@@ -603,6 +603,66 @@ def filter(self, message: Message) -> bool:
...
+        class file_extension(MessageFilter):
+            """This filter filters documents by their file ending/extension.
+
+            Note:
+                * This Filter only filters by the file ending/extension of the document,
+                    it doesn't check the validity of document.
+                * The user can manipulate the file extension of a document and
+                    send media with wrong types that don't fit to this handler.
+                * Case insensitive by default,
+                    you may change it with the flag ``case_sensitive=True``.
```
Poolitzer, line 615:
```suggestion
                    you may change this with the flag ``case_sensitive=True``.
```

this because it is a behaviour

telegram/ext/filters.py
```diff
@@ -603,6 +603,66 @@ def filter(self, message: Message) -> bool:
...
+                * Extension should be passed without leading dot
+                    unless it's a part of the extension.
+                * Pass :obj:`None` to filter files with no extension,
+                    i.e. without a dot in the filename.
```
Poolitzer, line 619: I would showcase that in an example

telegram/ext/filters.py
```diff
@@ -603,6 +603,66 @@ def filter(self, message: Message) -> bool:
...
+
+            Example:
+                ``Filters.document.file_extension("jpg")``
+                    filters files with extension ``".jpg"``.
+                ``Filters.document.file_extension(".jpg")``
+                    filters files with extension ``"..jpg"``.
```
Poolitzer, line 625: this is a great example! Good job.

telegram/ext/filters.py
```diff
@@ -603,6 +603,66 @@ def filter(self, message: Message) -> bool:
...
+                        Default: :obj:`False`.
+                """
+                self.is_case_sensitive = case_sensitive
+                if file_extension is None:
+                    self.file_extension = None
+                    self.name = "Filters.document.file_extension(None)"
+                elif case_sensitive:
+                    self.file_extension = f".{file_extension}"
+                    self.name = (
+                        f"Filters.document.file_extension({file_extension!r},"
```
Poolitzer, line 647: sorry what is the !r for?

Poolitzer (changes requested):
Left some minor comments, overall I really like this PR, thanks!

telegram/ext/filters.py
Bibo-Joshi: it adds quotes to the strings. Also make `None` distinguishable from `"None"`. Also asked that :D

Poolitzer:
oh @Bibo-Joshi can you check if you can build docs on this branch, specifically the filters page and this filter appears?

telegram/ext/filters.py
Poolitzer: quick google didnt show that to me :(

Bibo-Joshi: uses `repr()` instead of `str()`

Bibo-Joshi:
> oh @Bibo-Joshi can you check if you can build docs on this branch, specifically the filters page and this filter appears?

It doesn't. @eIGato the docstring of `file_extension` needs to be copied into the docstring of `document`. Please do that and build the docs locally to check that it renders correctly.
thanks for the catch @Poolitzer !

telegram/ext/filters.py
eIGato: Done.

eIGato: Thanks.

telegram/ext/filters.py
eIGato: Thanks.

eIGato:
IDKY the checks fail. New code is fully covered with tests and all tests pass locally: `1716 passed, 187 skipped, 19 xfailed, 1 xpassed in 207.14 seconds`.
I've tried to repush the branch to restart the checks but result did not change.

Bibo-Joshi:
test official fails due to the recent API updates not yet being incorporated and codecov for some reason decided to mark a docstring as not tested, so everthing's fine.

eIGato:
image
IDK how to fix such rendering: the renderer just understands the line feeds literally.

Bibo-Joshi:
I fixed the rendering locally, but I can't push to your branch. did you disable edits by maintainers (very bottom of the panel on the right)? if not:

```python

        file_extension: This filter filters documents by their file ending/extension.

            Note:
                * This Filter only filters by the file ending/extension of the document,
                  it doesn't check the validity of document.
                * The user can manipulate the file extension of a document and
                  send media with wrong types that don't fit to this handler.
                * Case insensitive by default,
                  you may change this with the flag ``case_sensitive=True``.
                * Extension should be passed without leading dot
                  unless it's a part of the extension.
                * Pass :obj:`None` to filter files with no extension,
                  i.e. without a dot in the filename.

            Example:
                * ``Filters.document.file_extension("jpg")`` filters files with extension
                  ``".jpg"``.
                * ``Filters.document.file_extension(".jpg")`` filters files with extension
                  ``"..jpg"``.
                * ``Filters.document.file_extension("Dockerfile", case_sensitive=True)``
                  filters files with extension ``".Dockerfile"`` minding the case.
                * ``Filters.document.file_extension(None)`` filters files without a dot in the
                  filename.
```

Poolitzer (commented):
LGTM with the rendering fixes bibo suggested

PR #2172, opened 2020-11-01
Improve and expand CallbackQuery shortcuts

### LINKED ISSUE #2171: [FEATURE] CallbackQurey.delete_message
missed that in #2043

Bibo-Joshi:
closes #2171

### DIFF
--- a/telegram/callbackquery.py
+++ b/telegram/callbackquery.py
@@ -131,9 +131,7 @@ class CallbackQuery(TelegramObject):
     def edit_message_text(self, text: str, *args: Any, **kwargs: Any) -> Union[Message, bool]:
         """Shortcut for either::
 
-            bot.edit_message_text(text, chat_id=update.callback_query.message.chat_id,
-                                message_id=update.callback_query.message.message_id,
-                                *args, **kwargs)
+            update.callback_query.message.edit_text(text, *args, **kwargs)
 
         or::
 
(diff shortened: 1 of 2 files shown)


Poolitzer (changes requested):
When inline message id is set and a user is calling delete(), the not really helping error `AttributeError: 'NoneType' object has no attribute 'chat_id'` is presented. Why not raise a custom one? At least put that in the docs.

And also, why delete_message()? With the message class, we just call it delete. I prefer that its way shorter (what a shortcut is all about imo)

Bibo-Joshi:
> And also, why delete_message()? With the message class, we just call it delete. I prefer that its way shorter (what a shortcut is all about imo)

I guess you're right. Just copied from the other shortcuts. Actually, if we're at it, we can replace `self.bot.edit_*` calls with `self.message.edit_*`.

If we document `CQ.delete_message()` as shortcut for `callback_query.message.delete()`, would you be fine with leaving the error as is? we don't raise custom errors anywhere else …

Poolitzer:
"anywhere else" can something like this happen, outside of this class? I would be fine not doing it, but I think overall custom errors would be an improvement.

Poolitzer:
okay but you didnt change delete_message to delete, intentional?

Bibo-Joshi:
> okay but you didnt change delete_message to delete, intentional?

uh, yes. I didn't understand your comment as "rename CQ.delete_message to CQ.delete", but as "why are you using `bot.delete_message` instead of `message.delete` xD
 but anyway I wouldn't wanna reanem to `CQ.delete`. Comparing with `CQ.answer` that would impley deleting the CQ … also we named all the other shortcuts `CQ.edit_message_*` and not `CQ.edit_*`. with a decent IDE, the shortcutiness is provided by autocomplete anyway …

Poolitzer (approved):

Poolitzer:
LGTM then

PR #2179, opened 2020-11-04
XOR Filters and make Filters.name a property

### LINKED ISSUE #2176: [FEATURE] XOR for filters
I guess that should be easily doable by

```python
def __xor__(self, other):
    (self & ~other) | (~self & other)
```

Bibo-Joshi:
closes #2176 

in the process refactors `Filters.name` to be a property. this makes it a bit more flexible (alloing it to be non-static) and also make XOR Filters have a proper name


Poolitzer (approved):
LGTM

PR #2183, opened 2020-11-05
API 5.0 WP 1+2 (API Servers & Webhooks)

Bibo-Joshi:


telegram/bot.py
```diff
@@ -2592,18 +2598,26 @@ def set_webhook(
             data['max_connections'] = max_connections
         if allowed_updates is not None:
```
Poolitzer, line 2599: here we use `is not None`, done there you use just `if:`. We probably should agree on one and go with it everywhere

Poolitzer (commented):
LGTM

telegram/bot.py
Bibo-Joshi: I think the important part is, that `[]` evaluates to `False`, but passing an empty list as a different meaning than not passing the value. So, IMO it's okay to use `is not None` if we need to make a difference between `bool(var) is False` and `var is None` (e.g. for booleans) and just use `if` in the other cases …

Poolitzer: but max_connections is an int?

Bibo-Joshi: and `bool(0) is False`, while `0 is not None`, so that fits my argumentation :D

Bibo-Joshi: but you're right that we're not very consistent, e.g. in `send_message`, there's a

```python
        if disable_web_page_preview:
            data['disable_web_page_preview'] = disable_web_page_preview
```

is it really that bad, though? 🤔

Poolitzer: I personally do like consistency. But I dont think it really matters, no. Just saw that and commented.

Bibo-Joshi: that's the right thing to do :) I guess it's not problematic mast times, as `None` and `False` usually mean the same to TG. but I'll try to use `if … is not None` wher it makes sence, as that's safer.

Poolitzer: alright

PR #2185, opened 2020-11-06
API 5.0 WP 3 (Working with Groups)

Bibo-Joshi:


telegram/bot.py
```diff
@@ -2098,6 +2103,9 @@ def unban_chat_member(
         """
         data: JSONDict = {'chat_id': chat_id, 'user_id': user_id}
 
+        if only_if_banned is not None:
```
Poolitzer, line 2106: now `is not None` is fine? :D

Poolitzer (commented):
LGTM

telegram/bot.py
Bibo-Joshi: see WP 1 :)

telegram/chat.py
```diff
@@ -56,6 +59,11 @@ class Chat(TelegramObject):
         sticker_set_name (:obj:`str`): Optional. For supergroups, name of Group sticker set.
         can_set_sticker_set (:obj:`bool`): Optional. :obj:`True`, if the bot can change group the
             sticker set.
+        linked_chat_id (:obj:`int`): Optional. Unique identifier for the linked chat, i.e. the
+            discussion group identifier for a channel and vice versa; for supergroups and channel
+            chats. Returned only in :meth:`telegram.Bot.get_chat`.
+        location (:class:`telegram.ChatLocation`): Optional. For supergroups, the location to which
```
harshil21, line 65: That 'Optional' should be inside the brackets right? Same for bio.

Bibo-Joshi: Only for the arguments, not for the attributes. so in the change below, but thanks for the catch :)

PR #2188, opened 2020-11-07
Api 5.0 WP 7 (Live Locations)

Bibo-Joshi:


telegram/bot.py
```diff
@@ -1323,8 +1326,15 @@ def send_location(
             latitude (:obj:`float`, optional): Latitude of location.
             longitude (:obj:`float`, optional): Longitude of location.
             location (:class:`telegram.Location`, optional): The location to send.
+            horizontal_accuracy (:obj:`int`, optional): The radius of uncertainty for the location,
+                measured in meters; 0-1500
```
Poolitzer, line 1330:
```suggestion
                measured in meters; 0-1500.
```

telegram/bot.py
```diff
@@ -1405,6 +1424,13 @@ def edit_message_live_location(
             latitude (:obj:`float`, optional): Latitude of location.
             longitude (:obj:`float`, optional): Longitude of location.
             location (:class:`telegram.Location`, optional): The location to send.
+            horizontal_accuracy (:obj:`float`, optional): The radius of uncertainty for the
+                location, measured in meters; 0-1500
```
Poolitzer, line 1428:
```suggestion
                location, measured in meters; 0-1500.
```

telegram/ext/filters.py
```diff
@@ -901,6 +911,8 @@ def filter(self, message: Update) -> bool:
             :attr:`telegram.Message.group_chat_created`,
             :attr:`telegram.Message.supergroup_chat_created` or
             :attr:`telegram.Message.channel_chat_created`.
+        connected_website: Messages that contain
+            :attr:`telegram.Message.connected_website`.
```
Poolitzer, line 915: thats not introduced here, mistake or fix?

telegram/files/location.py
```diff
@@ -32,17 +32,50 @@ class Location(TelegramObject):
...
+        heading: int = None,
+        proximity_alert_radius: int = None,
+        **_kwargs: Any,
+    ):
         # Required
         self.longitude = float(longitude)
         self.latitude = float(latitude)
 
+        # Optionals
+        self.horizontal_accuracy = horizontal_accuracy
```
Poolitzer, line 74: why arent you doing a float(variable) if else here as well?

telegram/inline/inlinequeryresultlocation.py
```diff
@@ -93,3 +109,8 @@ def __init__(
         self.thumb_url = thumb_url
         self.thumb_width = thumb_width
         self.thumb_height = thumb_height
+        self.horizontal_accuracy = horizontal_accuracy
```
Poolitzer, line 112: same

Poolitzer (changes requested):

telegram/ext/filters.py
Bibo-Joshi: The filter already existed, just added it to the docstring

telegram/files/location.py
Bibo-Joshi: my though about `int()` was that it would convert a falsy passed `10.0` to `10`, but I'm not sure if even that is necessary … handling strings is not really necessary imo.

Poolitzer: I mean we are doing it in 70/71 as well with longitude and latitude

Bibo-Joshi: ah. yes. will edit

Bibo-Joshi:
IISC, all review points were fixed. merging.

PR #2190, opened 2020-11-07
Api 5.0 WP 10 (Dice)

Bibo-Joshi:
I don't completely understand how the value for the slot machine is computed from the symbols, so I didn't add a description. But I don't think that's strictly necessary … (https://t.me/pythontelegrambottalk/152143)

### DIFF
--- a/telegram/constants.py
+++ b/telegram/constants.py
@@ -83,6 +83,8 @@ Attributes:
     DICE_DICE (:obj:`str`): '🎲'
     DICE_DARTS (:obj:`str`): '🎯'
     DICE_BASKETBALL (:obj:`str`): '🏀'
+    DICE_FOOTBALL (:obj:`str`): '⚽'
+    DICE_SLOT_MACHINE (:obj:`str`): = '🎰'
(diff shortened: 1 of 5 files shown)


telegram/constants.py
```diff
@@ -83,6 +83,8 @@
     DICE_DICE (:obj:`str`): '🎲'
     DICE_DARTS (:obj:`str`): '🎯'
     DICE_BASKETBALL (:obj:`str`): '🏀'
+    DICE_FOOTBALL (:obj:`str`): '⚽'
+    DICE_SLOT_MACHINE (:obj:`str`): = '🎰'
```
Poolitzer, line 87: so the = is wrong I think

Poolitzer (commented):
one minor thing

telegram/constants.py
Bibo-Joshi, line 87:
```suggestion
    DICE_SLOT_MACHINE (:obj:`str`): '🎰'
```

Poolitzer (approved):
LGTM

PR #2193, opened 2020-11-07
Api 5.0 WP 9.5 (Google Place Venues)

Bibo-Joshi:
TBH I didn't try to send `InputVenueMessageContent` and `ILQRVenue` with both foursquare and google place details, but as that doesn't seem to work to `send_venue`, I assume it doesn't for those either …

### DIFF
--- a/telegram/inline/inlinequeryresultvenue.py
+++ b/telegram/inline/inlinequeryresultvenue.py
@@ -32,6 +32,10 @@ class InlineQueryResultVenue(InlineQueryResult):
     use :attr:`input_message_content` to send a message with the specified content instead of the
     venue.
 
+    Note:
+      Foursquare details and Google Pace details are mutually exclusive. However, this
+      behaviour is undocumented and might be changed by Telegram.
+
     Attributes:
         type (:obj:`str`): 'venue'.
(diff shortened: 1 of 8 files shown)


telegram/inline/inlinequeryresultvenue.py
```diff
@@ -41,8 +45,8 @@ class InlineQueryResultVenue(InlineQueryResult):
         address (:obj:`str`): Address of the venue.
         foursquare_id (:obj:`str`): Optional. Foursquare identifier of the venue if known.
         foursquare_type (:obj:`str`): Optional. Foursquare type of the venue, if known.
-            (For example, "arts_entertainment/default", "arts_entertainment/aquarium" or
-            "food/icecream".)
```
Poolitzer, line 45: hmm I like this actually. Why did you remove it, maintaining?

Poolitzer (commented):
just one question, changes look great

telegram/inline/inlinequeryresultvenue.py
Bibo-Joshi: Because as a rule of thumb, attribute docstrings should be shorter than the corresponding argument docstring. at least that's what I did up to now …

PR #2196, opened 2020-11-08
Api 5.0 WP 9.1

Poolitzer:
Adding copy message and poll question length constant

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -96,7 +96,7 @@ from telegram.constants import MAX_INLINE_QUERY_RESULTS
 from telegram.error import InvalidToken, TelegramError
 from telegram.utils.helpers import DEFAULT_NONE, DefaultValue, to_timestamp
 from telegram.utils.request import Request
-from telegram.utils.types import FileLike, JSONDict, SLT
+from telegram.utils.types import FileLike, JSONDict
 
 if TYPE_CHECKING:
     from telegram.ext import Defaults
@@ -4381,7 +4381,7 @@ class Bot(TelegramObject):
         message_id: Union[str, int],
         caption: str = None,
         parse_mode: str = None,
-        caption_entities: SLT[MessageEntity] = None,
+        caption_entities: Union[Tuple[MessageEntity, ...], List[MessageEntity]] = None,
         disable_notification: bool = False,
         reply_to_message_id: Union[int, str] = None,
         allow_sending_without_reply: bool = False,
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -69,6 +69,8 @@ from telegram import (
     Location,
     MaskPosition,
     Message,
+    MessageEntity,
+    MessageId,
     PassportElementError,
     PhotoSize,
     Poll,
@@ -94,7 +96,7 @@ from telegram.constants import MAX_INLINE_QUERY_RESULTS
 from telegram.error import InvalidToken, TelegramError
 from telegram.utils.helpers import DEFAULT_NONE, DefaultValue, to_timestamp
 from telegram.utils.request import Request
-from telegram.utils.types import FileLike, JSONDict
+from telegram.utils.types import FileLike, JSONDict, SLT
 
 if TYPE_CHECKING:
     from telegram.ext import Defaults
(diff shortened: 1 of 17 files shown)


telegram/poll.py
```diff
@@ -39,11 +39,11 @@ class PollOption(TelegramObject):
     considered equal, if their :attr:`text` and :attr:`voter_count` are equal.
 
     Attributes:
-        text (:obj:`str`): Option text, 1-100 characters.
+        text (:obj:`str`): Option text, 1-300 characters.
         voter_count (:obj:`int`): Number of users that voted for this option.
 
     Args:
-        text (:obj:`str`): Option text, 1-100 characters.
+        text (:obj:`str`): Option text, 1-300 characters.
```
harshil21, line 46: Option text is still 100 characters, as per bot api docs

telegram/bot.py
```diff
@@ -4371,6 +4373,89 @@ def set_my_commands(
...
 
+    @log
+    def copy_message(
+        self,
+        chat_id: Union[int, str],
+        from_chat_id: Union[str, int],
+        message_id: Union[str, int],
+        caption: str = None,
+        parse_mode: str = None,
+        caption_entities: SLT[MessageEntity] = None,
```
Bibo-Joshi, line 4384: that's a good idea 👍  we should do that for ILQR and IM*, too

telegram/message.py
```diff
@@ -924,6 +924,23 @@ def forward(self, chat_id: int, *args: Any, **kwargs: Any) -> 'Message':
             chat_id=chat_id, from_chat_id=self.chat_id, message_id=self.message_id, *args, **kwargs
         )
 
+    def copy(self, chat_id: int, *args: Any, **kwargs: Any) -> 'MessageId':
+        """Shortcut for::
+
+            bot.copy_message(chat_id=chat_id,
+                                from_chat_id=update.message.chat_id,
```
Bibo-Joshi, line 931: indention looks wrong

telegram/messageid.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
+# pylint: disable=R0903
```
Bibo-Joshi, line 2: nitpicking just b/c I went over those recently: please check, if we actually need this here :)

telegram/messageid.py
```diff
@@ -0,0 +1,39 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser Public License for more details.
+#
+# You should have received a copy of the GNU Lesser Public License
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+"""This module contains an object that represents a type of a Telegram Poll."""
```
Bibo-Joshi, line 20: I doubt it ;P

telegram/constants.py
```diff
@@ -223,7 +224,7 @@
 
 POLL_REGULAR: str = 'regular'
 POLL_QUIZ: str = 'quiz'
-
+POLL_QUESTION_LENGTH: int = 300
```
Bibo-Joshi, line 227: maybe rather `MAX_POLL_QUESTION_LENGTH` in accordance with the other `MAX_*` constants? Also, while we're at it, maybe also add `MAX_POLL_OPTION_LENGTH`?

telegram/poll.py
```diff
@@ -262,3 +262,5 @@ def parse_explanation_entities(self, types: List[str] = None) -> Dict[MessageEnt
     """:const:`telegram.constants.POLL_REGULAR`"""
     QUIZ: ClassVar[str] = constants.POLL_QUIZ
     """:const:`telegram.constants.POLL_QUIZ`"""
+    POLL_QUESTION_LENGTH: ClassVar[int] = constants.POLL_QUESTION_LENGTH
```
Bibo-Joshi, line 265: rather `MAX_QUESTION_LENGTH`?

Bibo-Joshi (changes requested):
Left some minor comments. Three more general ones:

* we need tests for `MessageID`
* `copy_message` should be tested for defaults, i.e. `parse_mode` and `allow_sending_without_reply` (maybe see my #2192 for how I tested that)
* If you're starting with shortcuts already, maybe it's a good idea to just do all of WP12 here

telegram/bot.py
Bibo-Joshi: On second thought, if we allow singles here, we need to handle that properly below. Maybe it's better to only allow `Union[Tuple[ME, ...], List[ME]]`

telegram/poll.py
Poolitzer: oh thats right I changed the wrong strings thanks!

telegram/bot.py
Poolitzer: oh right forgot about SLT also allowing singles

telegram/constants.py
Poolitzer: Oh. Yeah.

Poolitzer:
> test for default allow_sending_without_reply 

I mean thats hard to do rn because it isnt merged yet and I feel like cherry picking is over the top. Just wait for merging #2192?

tests/test_bot.py
```diff
@@ -1431,3 +1431,93 @@ def assertion(url, data, *args, **kwargs):
...
+            message_id=media_message.message_id,
+            caption="<b>Test</b>",
+            parse_mode=ParseMode.HTML,
+            reply_to_message_id=media_message.message_id,
+            reply_markup=keyboard,
+        )
+
+    @flaky(3, 1)
+    @pytest.mark.timeout(10)
+    def test_copy_message_without_reply(self, bot, chat_id, media_message):
```
Bibo-Joshi, line 1465: what exactly was the problem here to combine the two tests? shouldn't something like

```python
orig_post = bot.requst.post

def post(*args, **kwargs):
    ...
    return orig_post(*args, **kwargs)
```
do the trick? if it's indeed complicated for some reason, just leave it as is, though …

tests/test_bot.py
```diff
@@ -1431,3 +1431,93 @@ def assertion(url, data, *args, **kwargs):
...
+    @pytest.mark.parametrize(
+        'default_bot,custom',
+        [
+            ({'parse_mode': ParseMode.HTML}, None),
+            ({'parse_mode': False}, None),
+            ({'parse_mode': False}, True),
+        ],
+        indirect=['default_bot'],
+    )
+    def test_copy_message_with_default(self, default_bot, chat_id, media_message, custom):
```
Bibo-Joshi, line 1505: I don't really understand what the `custom` parameter does here … like what's the difference between the 2nd and the 3rd parameter combination?

tests/test_messageid.py
```diff
@@ -0,0 +1,53 @@
...
+        message_id_dict = message_id.to_dict()
+
+        assert isinstance(message_id_dict, dict)
+        assert message_id_dict['message_id'] == message_id.message_id
+
+    def test_equality(self):
+        a = MessageId(message_id=1)
+        b = MessageId(message_id=1)
+        c = MessageId(message_id=2)
+
```
Bibo-Joshi, line 48: please also test that a `MessageId` object is not equal to other PTB objects, e.g. try to compare to a `Message`

Bibo-Joshi (changes requested):
thanks for the updates! IMO we should wait for #2192 so that we can add the tests here and don't forget … Still, I left you a little work for the meantime :)

telegram/constants.py
```diff
@@ -119,6 +119,7 @@
 Attributes:
     POLL_REGULAR (:obj:`str`): 'regular'
     POLL_QUIZ (:obj:`str`): 'quiz'
+    POLL_QUESTION_LENGTH (:obj:`int`): 300
```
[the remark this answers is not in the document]
Poolitzer: oh damn documentation. I was really confused because I knew I added option length constant :D

telegram/poll.py
```diff
@@ -54,6 +54,9 @@ def __init__(self, text: str, voter_count: int, **_kwargs: Any):
 
         self._id_attrs = (self.text, self.voter_count)
 
+    MAX_POLL_OPTION_LENGTH: ClassVar[int] = constants.MAX_POLL_OPTION_LENGTH
```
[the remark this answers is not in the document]
Poolitzer: uhm, but here it is, being a constant, belonging to PollOption?

tests/test_bot.py
Poolitzer: The issue is here that it also tries to do the test in post for the temp send message request. I dont know man I hate tests. Im just gonna let it be split

tests/test_bot.py
Poolitzer: I copied that from somewhere else. Just gonna remove it then, didnt make much sense to me as well, but hey, I am just testing

tests/test_bot.py
Bibo-Joshi: ah okay. yeah, np

telegram/poll.py
```diff
@@ -54,6 +54,9 @@ def __init__(self, text: str, voter_count: int, **_kwargs: Any):
 
         self._id_attrs = (self.text, self.voter_count)
 
+    MAX_LENGTH: ClassVar[int] = constants.MAX_POLL_OPTION_LENGTH
+    """:const:`telegram.constants.POLL_QUESTION_LENGTH`"""
```
Bibo-Joshi, line 58:
```suggestion
    """:const:`telegram.constants.MAX_POLL_OPTION_LENGTH`"""
```

telegram/poll.py
```diff
@@ -262,3 +265,7 @@ def parse_explanation_entities(self, types: List[str] = None) -> Dict[MessageEnt
     """:const:`telegram.constants.POLL_REGULAR`"""
     QUIZ: ClassVar[str] = constants.POLL_QUIZ
     """:const:`telegram.constants.POLL_QUIZ`"""
+    MAX_QUESTION_LENGTH: ClassVar[int] = constants.MAX_POLL_QUESTION_LENGTH
+    """:const:`telegram.constants.POLL_QUESTION_LENGTH`"""
```
Bibo-Joshi, line 269:
```suggestion
    """:const:`telegram.constants.MAX_POLL_QUESTION_LENGTH`"""
```

telegram/poll.py
```diff
@@ -262,3 +265,7 @@ def parse_explanation_entities(self, types: List[str] = None) -> Dict[MessageEnt
...
+    MAX_OPTION_LENGTH: ClassVar[int] = constants.MAX_POLL_OPTION_LENGTH
+    """:const:`telegram.constants.POLL_QUESTION_LENGTH`"""
```
Bibo-Joshi, line 271:
```suggestion
    """:const:`telegram.constants.MAX_POLL_OPTION_LENGTH`"""
```

tests/test_bot.py
Bibo-Joshi: then please also remove the parameter alltogether to avoid future confusion ;)

Bibo-Joshi (changes requested):
some last nitpicks left ;)

tests/test_bot.py
Poolitzer: I dont see what you mean. The custom argument of the function? Its no in the code anymore

Bibo-Joshi: oh, was probably looking at the old diff. sorry for the confusion

PR #2151, opened 2020-10-16
Improve handling of custom objects in BP.insert/replace_bot

Bibo-Joshi:
if an object can't be copied, they're probably something special and don't contain a Bot instance. Therefore, those will now just be skipped.

Note: I noticed the latter while testing the changes on a personal project, where I have locks in some objects that have a custom pickle behaviour but not a custom copy behaivour

PR #2197, opened 2020-11-08
Update basepersistence.py

rojserbest:


Bibo-Joshi:
Thanks for the quick PR. We should add a unit test for this, though, ie. create a simple class that an error on comparison with a string check that this doesn't bother the persistence. Would you be up to adding that to `test_persistence.py`? You can integrate it into the existing tests for `insert/replace_bot`

rojserbest:
@Bibo-Joshi do you mean adding the fix to [`test/test_persistense.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/tests/test_persistence.py)? I wasn't able to do that because I wasn't able to find the `insert_bot` method.

Bibo-Joshi:
Not adding the fix to that file :D `tests/test_persistence.py` is a file containing unit tests for the persistence setup. What I mean is that the fix should be tested, probably by adjusting this [test](https://github.com/python-telegram-bot/python-telegram-bot/blob/d1438a9b238ce7d1c355eefeac329ff4b098ac19/tests/test_persistence.py#L440). See the contribution guide on how to run the tests locally.
If you're not familiar with pytest and don't want to put too much time into this, that's okay - especially the persistence testing is not always very easy. Just wanted to at least give you the opportunity ;)

Bibo-Joshi:
Just went ahead and added both fix and tests an #2151. Still, thanks for making the effort and finding the bug in the first place :)

PR #2202, opened 2020-11-10
Handle open(filename, 'r')

### LINKED ISSUE #2198: [BUG] Passing non-bytes file input leads to error
https://t.me/pythontelegrambotgroup/396541

TL;DR:

`send_document(open('text_file', 'rb'))` works but `send_document(open('text_file', 'r'))` raises is error.
This is, because we try to guess if the file is an image using `imghdr.what(None, stream)` in `InputFile.is_image`, which only works if `stream` is a bytes stream.
If I comment the `is_image` out, the file is sent without issue, so I guess we should just check if the input is bytes before calling `is_image`

Bibo-Joshi:
closes #2198

### DIFF
--- a/telegram/files/inputfile.py
+++ b/telegram/files/inputfile.py
@@ -20,6 +20,7 @@
 """This module contains an object that represents a Telegram InputFile."""
 
 import imghdr
+import logging
 import mimetypes
 import os
 from typing import IO, Optional, Tuple
@@ -84,11 +85,17 @@ class InputFile:
             :obj:`str`: The str mime-type of an image.
 
         """
-        image = imghdr.what(None, stream)
-        if image:
-            return 'image/%s' % image
-
-        raise TelegramError('Could not parse file content')
+        try:
+            image = imghdr.what(None, stream)
+            if image:
+                return 'image/%s' % image
+            raise TelegramError('Could not parse file content')
+        except Exception as exc:
+            logger = logging.getLogger(__name__)
+            logger.debug(
+                "Could not parse file content. Assuming that file is not an image.", exc_info=True
+            )
+            raise TelegramError('Could not parse file content') from exc
 
     @staticmethod
     def is_file(obj: object) -> bool:
--- a/telegram/files/inputfile.py
+++ b/telegram/files/inputfile.py
@@ -62,11 +62,11 @@ class InputFile:
         image_mime_type = self.is_image(self.input_file_content)
         if image_mime_type:
             self.mimetype = image_mime_type
+        elif self.filename:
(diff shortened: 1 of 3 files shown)


telegram/files/inputfile.py
```diff
@@ -59,9 +59,10 @@ def __init__(self, obj: IO, filename: str = None, attach: bool = None):
         elif hasattr(obj, 'name') and not isinstance(obj.name, int):
             self.filename = os.path.basename(obj.name)
 
-        try:
-            self.mimetype = self.is_image(self.input_file_content)
-        except TelegramError:
+        image_mime_type = self.is_image(self.input_file_content)
+        if image_mime_type:
+            self.mimetype = image_mime_type
+        else:
```
tsnoam, line 65: elif ...

telegram/files/inputfile.py
```diff
@@ -74,21 +75,27 @@ def field_tuple(self) -> Tuple[str, bytes, str]:
...
-    def is_image(stream: bytes) -> str:
+    def is_image(stream: bytes) -> Optional[str]:
         """Check if the content file is an image by analyzing its headers.
 
         Args:
             stream (:obj:`bytes`): A byte stream representing the content of a file.
 
         Returns:
-            :obj:`str`: The str mime-type of an image.
+            :obj:`str` | :obj:`None`: The mime-type of an image, if the input is an image, and
```
tsnoam, line 85:
```suggestion
            :obj:`str` | :obj:`None`: The mime-type of an image, if the input is an image, or
```

telegram/files/inputfile.py
```diff
@@ -74,21 +75,27 @@ def field_tuple(self) -> Tuple[str, bytes, str]:
...
-        image = imghdr.what(None, stream)
-        if image:
-            return 'image/%s' % image
-
-        raise TelegramError('Could not parse file content')
+        try:
+            image = imghdr.what(None, stream)
+            if image:
+                return f'image/{image}'
+            return None
```
tsnoam, line 93: `return None` is redundant. You can remove that line.

telegram/files/inputfile.py
```diff
@@ -74,21 +75,27 @@ def field_tuple(self) -> Tuple[str, bytes, str]:
...
+        except Exception:
+            logger.debug(
+                "Could not parse file content. Assuming that file is not an image.", exc_info=True
+            )
+            return None
```
tsnoam, line 98: `return None` is redundant. You can remove that line.

tsnoam (changes requested):
Some style nitpicking.
Fix or ignore at your discretion.

telegram/files/inputfile.py
Bibo-Joshi: mypy & pylint like that …

PR #2205, opened 2020-11-10
Api 5.0 WP 13 (New Filters)

Bibo-Joshi:
Some comments:

I ignore pylint complaining about some methods just calling `super`. I do that because I think it's important for all of those methods to have a proper docstring. IMO the usage of those filters is complicated enough to justify not just referring to the base class. figuring out what a base class for `Filters.{chat, via_bot, user, sender_chat}` does is not easy, if you're not familiar with the code.

On the fly I updated the example for `XOR` filter, as it didn't make any sense :D

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -38,7 +38,7 @@ from typing import (
     NoReturn,
 )
 
-from telegram import Chat, Message, MessageEntity, Update
+from telegram import Chat, Message, MessageEntity, Update, User
 
 __all__ = [
     'Filters',
@@ -69,7 +69,7 @@ class BaseFilter(ABC):
 
     Exclusive Or:
 
-        >>> (Filters.regex('To Be') ^ Filters.regex('Not To Be'))
+        >>> (Filters.regex('To Be') ^ Filters.regex('Not 2B'))
 
     Not:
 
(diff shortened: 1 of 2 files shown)


telegram/ext/filters.py
```diff
@@ -1654,158 +1610,164 @@ class chat(MessageFilter):
...
+        Examples:
+            * To filter for messages forwarded from a channel with ID ``-1234``, use
+              ``MessageHandler(Filters.sender_chat(-1234), callback_method)``.
+            * To filter for messages of anonymous admins in a super group with username
+              ``@anonymous``, use
+              ``MessageHandler(Filters.sender_chat(username='anonymous'), callback_method)``.
+            * To filter for messages forwarded from *any* channel, use
+              ``MessageHandler(Filters.sender_chat.channel, callback_method)``
+            * To filter for messages of anonymous admins in *any* super group, use
+              ``MessageHandler(Filters.sender_chat.super_group, callback_method)``
```
Poolitzer, line 1672:
```suggestion
            * To filter for messages forwarded from *any* channel, use
              ``MessageHandler(Filters.sender_chat.channel, callback_method)``.
            * To filter for messages of anonymous admins in *any* super group, use
              ``MessageHandler(Filters.sender_chat.super_group, callback_method)``.
```

telegram/ext/filters.py
```diff
@@ -1654,158 +1610,164 @@ class chat(MessageFilter):
...
+
+        Warning:
+            :attr:`chat_ids` will give a *copy* of the saved chat ids as :class:`frozenset`. This
```
Poolitzer, line 1675:
```suggestion
            :attr:`chat_ids` will retun a *copy* of the saved chat ids as :class:`frozenset`. This
```

telegram/ext/filters.py
```diff
@@ -1654,158 +1610,164 @@ class chat(MessageFilter):
...
+            that it is not causing race conditions, as this will complete replace the current set
+            of allowed chats.
+
+        Attributes:
+            chat_ids(set(:obj:`int`), optional): Which sender chat chat ID(s) to allow through.
+            usernames(set(:obj:`str`), optional): Which sender chat username(s) (without leading
+                '@') to allow through.
+            allow_empty(:obj:`bool`, optional): Whether updates should be processed, if no sender
+                chat is specified in :attr:`chat_ids` and :attr:`usernames`.
+            super_group: Messages whose sender chat is a super group.
```
Poolitzer, line 1688: thats wrong?

Poolitzer (changes requested):
nitpicking docs

telegram/ext/filters.py
Bibo-Joshi: what's wrong? o.0

Poolitzer: the super group attribute. It doesnt have one.
```suggestion
```

Poolitzer: and after some internal discussion I take this back.

Poolitzer (approved):
LGTM now

Issue #2208, opened 2020-11-12
[FEATURE] adding run_async in telegram.ext.Defaults


starry69:
Since most (if not all) people using PTB use asynchronous callbacks anyway, adding `run_async` parameter in `telegram.ext.Defaults` will be lot helpful for users, it'll also save them lot of time for adding `run_async=True` in each handlers, and ofcourse they can override the default `run_async` value for specific handler if they want to.. so it seems a win win in end!

Bibo-Joshi:
Hey. After short internal discussion, we think this is a great feature.

A note about the implementation: the `Dispatcher` gets a `Bot` as argument anyway, so we can just extract the default setting is `self.bot.defaults.run_async`. Maybe add a (private) property as shortcut that checks if the bot has defaults, or something like that.

@starry69, ofc you're welcome to PR yourself :) If you want to, please leave a note, so we don't do the work twice.

starry69:
Thanks for taking this into consideration! I'll PR soon..

starry69:
Ig i should close this now

PR #2210, opened 2020-11-13
Add Defaults.run_async support

starry69:
Adds `run_async` support in `telegram.ext.Defaults` as discussed in: https://github.com/python-telegram-bot/python-telegram-bot/issues/2208

### DIFF
--- a/telegram/ext/callbackqueryhandler.py
+++ b/telegram/ext/callbackqueryhandler.py
@@ -131,7 +131,7 @@ class CallbackQueryHandler(Handler):
         pass_groupdict: bool = False,
         pass_user_data: bool = False,
         pass_chat_data: bool = False,
-        run_async: DefaultValue = DEFAULT_FALSE,
+        run_async: Union[bool, DefaultValue] = DEFAULT_FALSE,
     ):
         super().__init__(
             callback,
--- a/telegram/ext/callbackqueryhandler.py
+++ b/telegram/ext/callbackqueryhandler.py
@@ -34,6 +34,7 @@ from typing import (
 
 from telegram import Update
 from telegram.utils.types import HandlerArg
+from telegram.utils.helpers import DefaultValue, DEFAULT_FALSE
 
 from .handler import Handler
 
@@ -130,7 +131,7 @@ class CallbackQueryHandler(Handler):
         pass_groupdict: bool = False,
         pass_user_data: bool = False,
         pass_chat_data: bool = False,
-        run_async: bool = False,
+        run_async: DefaultValue = DEFAULT_FALSE,
     ):
         super().__init__(
             callback,
--- a/telegram/ext/defaults.py
+++ b/telegram/ext/defaults.py
@@ -44,7 +44,8 @@ class Defaults:
         tzinfo (:obj:`tzinfo`): A timezone to be used for all date(time) objects appearing
             throughout PTB.
         run_async (:obj:`bool`): Optional. Default setting for the ``run_async`` parameter of
-            handlers and error handlers registered through :meth:`Dispatcher.add_handler`.
+            handlers and error handlers registered through :meth:`Dispatcher.add_handler` and
+            :meth:`Dispatcher.add_error_handler`.
 
     Parameters:
         parse_mode (:obj:`str`, optional): Send Markdown or HTML, if you want Telegram apps to show
@@ -64,8 +65,8 @@ class Defaults:
             somewhere, it will be assumed to be in ``tzinfo``. Must be a timezone provided by the
             ``pytz`` module. Defaults to UTC.
         run_async (:obj:`bool`, optional): Default setting for the ``run_async`` parameter of
-            handlers and error handlers registered through :meth:`Dispatcher.add_handler`.
-            Defaults to :obj:`False`.
+            handlers and error handlers registered through :meth:`Dispatcher.add_handler` and
+            :meth:`Dispatcher.add_error_handler`. Defaults to :obj:`False`.
     """
 
     def __init__(
--- a/telegram/ext/defaults.py
+++ b/telegram/ext/defaults.py
@@ -43,8 +43,8 @@ class Defaults:
             be ignored. Default: :obj:`True` in group chats and :obj:`False` in private chats.
         tzinfo (:obj:`tzinfo`): A timezone to be used for all date(time) objects appearing
             throughout PTB.
-        run_async (:obj:`bool`): Optional. If set to :obj:`True`, callbacks registerd in handlers
-            will asynchronously throughout PTB.
+        run_async (:obj:`bool`): Optional. Default setting for the ``run_async`` parameter of
+            handlers registered through :meth:`Dispatcher.add_handler`.
 
     Parameters:
         parse_mode (:obj:`str`, optional): Send Markdown or HTML, if you want Telegram apps to show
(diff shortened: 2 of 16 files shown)


telegram/ext/callbackqueryhandler.py
```diff
@@ -130,7 +131,7 @@ def __init__(
         pass_groupdict: bool = False,
         pass_user_data: bool = False,
         pass_chat_data: bool = False,
-        run_async: bool = False,
+        run_async: DefaultValue = DEFAULT_FALSE,
```
Bibo-Joshi, line 134:
```suggestion
        run_async: Union[bool, DefaultValue] = DEFAULT_FALSE,
```

same for the others

telegram/ext/defaults.py
```diff
@@ -61,6 +63,9 @@ class Defaults:
             appearing throughout PTB, i.e. if a timezone naive date(time) object is passed
             somewhere, it will be assumed to be in ``tzinfo``. Must be a timezone provided by the
             ``pytz`` module. Defaults to UTC.
+        run_async (:obj:`bool`, optional): If set to :obj:`True`, the callbacks registers in handlers will run
+            asynchronously, this parameter will ignore the default `run_async` parameter, whitch is :obj:`False`
+            for every handler.
```
Bibo-Joshi, line 68:
```suggestion
        run_async (:obj:`bool`, optional): Default setting for the ``run_async`` parameter of handlers registered through :meth:`Dispatcher.add_handler`. Defaults to :obj:`False`.
```

telegram/ext/defaults.py
```diff
@@ -43,6 +43,8 @@ class Defaults:
             be ignored. Default: :obj:`True` in group chats and :obj:`False` in private chats.
         tzinfo (:obj:`tzinfo`): A timezone to be used for all date(time) objects appearing
             throughout PTB.
+        run_async (:obj:`bool`): Optional. If set to :obj:`True`, callbacks registerd in handlers
+            will asynchronously throughout PTB.
```
Bibo-Joshi, line 47:
```suggestion
        run_async (:obj:`bool`): Optional. Default setting for the ``run_async`` parameter of handlers registered through :meth:`Dispatcher.add_handler`.
```

telegram/ext/handler.py
```diff
@@ -143,14 +144,19 @@ def handle_update(
                 the dispatcher.
 
         """
+        should_run_async = self.run_async
```
Bibo-Joshi, line 147: I think we can rename `should_run_async` to `run_async`. the `should` doesn't give any extra information imho

Bibo-Joshi (changes requested):
Thanks for the quick PR!
Besides some nitpicks that I left, please make sure that pre-commit & codecov are happy. please see the contribution guide on how to run pre-commit locally. `tests/test_dispatcher.py` would be, where there new tests should go and you can take inspiration from the existing tests :)
we can ignore `test_official` for now, as we didn't merge API 5.0 yet …

telegram/ext/handler.py
starry69: Thanks for reviewing.. I'll address requested changes soon..

starry69:
Will look into pre-commit and tests for `test/test_dispatcher.py` tomorrow...

starry69:
@Bibo-Joshi is everything alright now?

tests/test_dispatcher.py
```diff
@@ -174,6 +174,18 @@ def test_error_handler_that_raises_errors(self, dp):
 
         assert self.count == 1
 
+    def test_default_run_async(self, run_async_bot, dp):
```
Bibo-Joshi, line 177: I don't understand this test …

* according to the new fixture, `run_async_bot.defaults.run_async` is always `True`, so why the `if-else`?
* it doesn't even involve a handler, so it doesn't actually test the behaviour of `Handler.handle_update` recarding `run_async`. This also shows in the `codecov` report: the line marked in red is never executed in the tests

image

Bibo-Joshi (changes requested):
thanks for the update. apart from pre-commit not being happy, the test doesn't really make sense to me :/

tests/test_dispatcher.py
starry69: Oh i see, ig i should do more study around tests before making it, btw i noticed that `Dispatcher.add_error_handler` also has run_async parameter, i should implement default behaviour in error handler too right?

tests/test_dispatcher.py
```diff
@@ -174,6 +174,15 @@ def test_error_handler_that_raises_errors(self, dp):
 
         assert self.count == 1
 
+    def test_default_run_async(self, monkeypatch, dp):
```
Bibo-Joshi, line 177: mh, `dp.bot` doesn't have the default settings, iisc, which explains why the coverage is still failing. maybe set those manually here (but make sure to delete them at the end of the test)

tests/test_dispatcher.py
```diff
@@ -174,6 +174,15 @@ def test_error_handler_that_raises_errors(self, dp):
...
+        def test_callback(*args, **kwargs):
+            self.received = "running async"
+
+        dp.add_handler(MessageHandler(Filters.all, test_callback))
```
Bibo-Joshi, line 181: you need a different callback here. maybe just `lambda u, c: pass`. because the handler callback sets the flag, the test succeeds even though the default setting is not applied

Bibo-Joshi (changes requested):
Please also add a test for the error handler similar to the othe new test. in pycharm you can easily run the tests with coverage, btw, so you can check for coverage before pushing.
and please run `pre-commit` locally (see contribution guide) ;)

starry69:
Yeah sure, I'll add test for err handler aswell...

Bibo-Joshi:
> dunno why mypy is failing here, as it was passing in my local environment

mypy isn't failing, black and flake8 are ;)

starry69:
> > dunno why mypy is failing here, as it was passing in my local environment
> 
> mypy isn't failing, black and flake8 are ;)

Umm those were passing aswell ^_^
Screenshot_20201117-150426181-01

tests/test_dispatcher.py
```diff
@@ -174,6 +175,40 @@ def test_error_handler_that_raises_errors(self, dp):
 
         assert self.count == 1
 
+    def test_default_run_async_error_handler(self, dp, monkeypatch):
```
Bibo-Joshi, line 178: okay, the tests now test the proper thing :) let's improve them a bit more:

* please also test that passing `Defaults(run_async=False)` does not run stuff asynchorously (you can use `pytest.mark.parametrize` here, you can have a look at other tests that use that
* please do something like 

```python
dp.bot.defaults = Defaults(run_async=True)
try:
    # the actual test
finally:
    dp.bot.defaults = None
```

  to ensure that the defaults are removed at the end of the test, even if it fails

Bibo-Joshi (changes requested):
we're close :) left some nitpicks and also black still fails. you can run it manually with `black .` before commiting, don't know why pre-commit doesn't fail for you …

starry69:
Thanks for reviewing! I'll address suggested changes in evening..

starry69:
@Bibo-Joshi is everything alright now?

telegram/ext/defaults.py
```diff
@@ -43,6 +43,8 @@ class Defaults:
             be ignored. Default: :obj:`True` in group chats and :obj:`False` in private chats.
         tzinfo (:obj:`tzinfo`): A timezone to be used for all date(time) objects appearing
             throughout PTB.
+        run_async (:obj:`bool`): Optional. Default setting for the ``run_async`` parameter of
+            handlers and error handlers registered through :meth:`Dispatcher.add_handler`.
```
Bibo-Joshi, line 47: please link both `add_handler` and `add_error_handler` at the right places ;)

telegram/ext/defaults.py
```diff
@@ -61,6 +63,9 @@ class Defaults:
             appearing throughout PTB, i.e. if a timezone naive date(time) object is passed
             somewhere, it will be assumed to be in ``tzinfo``. Must be a timezone provided by the
             ``pytz`` module. Defaults to UTC.
+        run_async (:obj:`bool`, optional): Default setting for the ``run_async`` parameter of
+            handlers and error handlers registered through :meth:`Dispatcher.add_handler`.
```
Bibo-Joshi, line 67: same

Bibo-Joshi (changes requested):
just one documentation nitpick left :) btw, github already notifies me on updates ;)

Bibo-Joshi:
Thanks for your contribution and your patience ):

PR #2223, opened 2020-11-24
API 5.0 WP4.3: Handling of files in local mode

Bibo-Joshi:
- [x] add ability to pass `pathlib.Path` objects instead of strings

### DIFF
--- a/telegram/files/file.py
+++ b/telegram/files/file.py
@@ -133,7 +133,8 @@ class File(TelegramObject):
 
         if out:
             if local_file:
-                buf = open(url, "rb").read()
+                with open(url, 'rb') as file:
+                    buf = file.read()
             else:
                 buf = self.bot.request.retrieve(url)
                 if self._credentials:
@@ -152,14 +153,11 @@ class File(TelegramObject):
         else:
             filename = os.path.join(os.getcwd(), self.file_id)
 
-        if local_file:
-            buf = open(url, "rb").read()
-        else:
-            buf = self.bot.request.retrieve(url, timeout=timeout)
-            if self._credentials:
-                buf = decrypt(
-                    b64decode(self._credentials.secret), b64decode(self._credentials.hash), buf
-                )
+        buf = self.bot.request.retrieve(url, timeout=timeout)
+        if self._credentials:
+            buf = decrypt(
+                b64decode(self._credentials.secret), b64decode(self._credentials.hash), buf
+            )
         with open(filename, 'wb') as fobj:
             fobj.write(buf)
         return filename
@@ -187,7 +185,8 @@ class File(TelegramObject):
         if buf is None:
             buf = bytearray()
         if is_local_file(self.file_path):
-            buf.extend(open(self.file_path, "rb").read())
+            with open(self.file_path, "rb") as file:
+                buf.extend(file.read())
         else:
             buf.extend(self.bot.request.retrieve(self._get_encoded_url()))
         return buf
--- a/telegram/files/file.py
+++ b/telegram/files/file.py
@@ -18,6 +18,7 @@
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains an object that represents a Telegram File."""
 import os
+import shutil
 import urllib.parse as urllib_parse
 from base64 import b64decode
 from os.path import basename
@@ -144,6 +145,10 @@ class File(TelegramObject):
             out.write(buf)
             return out
 
+        if custom_path and local_file:
+            shutil.copyfile(self.file_path, custom_path)
+            return custom_path
+
         if custom_path:
             filename = custom_path
         elif local_file:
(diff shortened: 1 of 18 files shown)


Bibo-Joshi:
Theoretically this is ready for review.

* pytest fails seem unrelated
* don't know why coverage is "not affected". need to double check before merging

If I get to it before this is reviewed and merged, I'll try to add the `file://` parsing logic. Otherwise it get's its own pr …

Bibo-Joshi:
Update: @Poolitzer tested the local file get & download behaviour on local bot apis on both windows and ubuntu.

I just pushed a commit to also allow passing (relative) local paths to *send* files. Mock-tests succeed, will double check on ubuntu in a moment.

Bibo-Joshi:
Tested `send_document` and `send_media_group` and both work.

onyexty (approved):

telegram/files/file.py
```diff
@@ -119,30 +123,45 @@ def download(
...
 
         if out:
-            buf = self.bot.request.retrieve(url)
-            if self._credentials:
-                buf = decrypt(
-                    b64decode(self._credentials.secret), b64decode(self._credentials.hash), buf
-                )
+            if local_file:
+                with open(url, 'rb') as file:
+                    buf = file.read()
```
tsnoam, line 137: I think there are more efficient ways to copy a file than reading a file into python memory.
Maybe the shutil package can assist here.

telegram/utils/helpers.py
```diff
@@ -55,6 +56,79 @@ def get_signal_name(signum: int) -> str:
...
+
+    Args:
+        obj (:obj:`str`): The string to check.
+        absolute (:obj:`bool`): Optional. Whether to allow only absolute paths. Defaults to
+            :obj:`True`.
+    """
+    if obj is None:
+        return False
+
+    path = Path(obj) if isinstance(obj, str) else obj
```
tsnoam, line 71: I think bytes can also be a filename.

Maybe just:
```
path = Path(obj)
```

telegram/utils/helpers.py
```diff
@@ -55,6 +56,79 @@ def get_signal_name(signum: int) -> str:
...
+    try:
+        if path.exists() and path.is_file():
```
tsnoam, line 73: No need for path.exists ?

telegram/utils/helpers.py
```diff
@@ -55,6 +56,79 @@ def get_signal_name(signum: int) -> str:
...
+    Args:
+        obj (:obj:`str`): The string to check.
+        absolute (:obj:`bool`): Optional. Whether to allow only absolute paths. Defaults to
+            :obj:`True`.
+    """
+    if obj is None:
+        return False
+
+    path = Path(obj) if isinstance(obj, str) else obj
+    try:
```
tsnoam, line 72: Why do we need try except here?

telegram/utils/helpers.py
```diff
@@ -55,6 +56,79 @@ def get_signal_name(signum: int) -> str:
...
+    Returns:
+        :obj:`str` | :class:`telegram.InputFile` | :obj:`object`: The parsed input or the untouched
+            :attr:`file_input`, in case it's no valid file input.
+    """
+    # Importing on file-level yields cyclic Import Errors
+    from telegram import InputFile  # pylint: disable=C0415
+
+    if isinstance(file_input, str) and file_input.startswith('file://'):
+        return file_input
+    if isinstance(file_input, (str, Path)):
```
tsnoam, line 116: I think bytes may also be valid for a filename.

telegram/utils/helpers.py
```diff
@@ -55,6 +56,79 @@ def get_signal_name(signum: int) -> str:
     return _signames[signum]
 
 
+def is_local_file(obj: Optional[Union[str, Path]], absolute: bool = True) -> bool:
```
tsnoam, line 59: Why does it matter if the file is absolute or not?

telegram/utils/helpers.py
```diff
@@ -55,6 +56,79 @@ def get_signal_name(signum: int) -> str:
...
+            :attr:`file_input`, in case it's no valid file input.
+    """
+    # Importing on file-level yields cyclic Import Errors
+    from telegram import InputFile  # pylint: disable=C0415
+
+    if isinstance(file_input, str) and file_input.startswith('file://'):
+        return file_input
+    if isinstance(file_input, (str, Path)):
+        if is_local_file(file_input, absolute=True):
+            out = f'file://{file_input}'
```
tsnoam, line 118: I don't think we should care if the file is absolute or not.
I think that this will be good for both cases:
```python
if is_local_file(file_input):
    out = f'file://{Path(file_input).absolute()}'
```

telegram/files/file.py
Bibo-Joshi: You're right. I think `shutil.copyfile()` would be the one to use.

telegram/utils/helpers.py
Bibo-Joshi: jup, forgot that `Path(Path())` works …

telegram/utils/helpers.py
Bibo-Joshi: `Path('https://some-domain.org').is_file()` throws an error instead of returning `False` …

telegram/utils/helpers.py
Bibo-Joshi: Will check

Bibo-Joshi: nope, `Path` only accepts `str` or `os.PathLike` (>> `Path`)

Poolitzer (changes requested):
Small doc suggestions. Code looks good!

telegram/utils/helpers.py
Bibo-Joshi: nah, `path.is_file()` can still return `False` if the passed string looks like a valid path but the file doesn't exist

PR #2240, opened 2020-12-09
Improve signature of shortcuts

### LINKED ISSUE #2238: [ENHANCEMENT] Improve callback signature of shortcuts
We just pass kwargs and args to the underlying function call currently for maintenance easiness. This however means that type hinting/suggestions from IDEs don't live up to their full potential. One of ptb greatest feature are our shortcuts, sparing the developers a lot of time writing out user or chat ids etc. We should spend some time improving this part of our library in my opinion.

If we want to implement this feature, at least testing for it would be a requirement, so whenever we change the underlying functions or add new shortcuts, the test tells us that we have to change the type hints/docs/etc.

Bibo-Joshi:
First POC for #2238. It's probably good to discuss it first, before applying the changes to all the other shortcuts.

Some notes:
* As per my comment on #2238, I skipped documentation of the arguments
* manually passing all the arguments to the bot methods seemed tiresome to me and is also error prone (we would have to test passing aaaaall the args), so I came up with `get_shortcut_kwargs`, which can be used in combination with pythons built-in `locals()` method. It will probably have to get a argument `local_vars_to_ignore` for some of the shortcuts, but we'll see.
* put the signature test (including annotations) in `conftest.py` to reduce code duplication in the tests
* added `Message._new_quote` instead of editing `Message._quote` for now, so the test won't fail immediately …

when ready, closes #2238


Poolitzer (approved):
LGTM. Details in dev group

examples/rawapibot.py
```diff
@@ -52,7 +52,7 @@ def echo(bot: telegram.Bot) -> None:
 
         if update.message:  # your bot can receive updates without messages
             # Reply to the message
-            update.message.reply_text(update.message.text)
+            update.message.reply_text(update.message.text)  # type: ignore[arg-type]
```
Poolitzer, line 55: that looks kinda wrong here but I guess we can leave it in?

Poolitzer (commented):
okay, looks good now. are we doing it like this?

examples/rawapibot.py
Bibo-Joshi: Mypy just complains about `message.text` being `Optional[str]` instead of `str`. The only alternative would be to cast it to str, which might be even more irritation for newbies

Bibo-Joshi:
Here is a (hopefully complete) list of all TG classes with shortcuts. Put your name behind it, when you start working on it. that way we can keep track of who does what. Don't forget to update the tests ;P

- [ ] CallbackQuery
- [ ] Chat
- [ ] Message (@Bibo-Joshi )
- [ ] Poll (doesn't have one yet, but we should add one for `stop_poll`, while at it)
- [ ] User

Bibo-Joshi:
TimeOut-failures seem unrelated and I think I got all the shortcuts. Ready for review.

examples/rawapibot.py
blueset: Could this be better if we have add an assert befor this line? Just to assure mypy that there is something in `Message.text`.


```suggestion
            assert update.message.text
            update.message.reply_text(update.message.text)
```

Bibo-Joshi: I would be surporised if MyPy parsed that correctly. But even if it does, personally I find an assertion superfluous here, as `MessageHandler` already checked that, i.e. adding a run time relevant line only for a type checker seems strange to me.

blueset: `mypy` do honor type guards/checks within a function, as it helps narrow down `Union`s and `Optional`s (which is just a shortcut of `Union`s) of a variable in the flow. 

`mypy` only considers the type definition passed in to the function (i.e. `Message` for `update.message`) which is a general type definition for all possible messages. which is passed from the handler. In the dispatcher-handler pattern, if we want to tell `mypy` about the optional-ness outcome of each specific handler, we need to show that in the type definition, potentially thourgh a super long generic type definition that could definitely look much more “ugly” than a simple type guard in the processor function. 

Otherwise, mypy would take the `def my_processor(update: Update, ...):` literally, as if `my_processor` could be called by any code with any valid `Update`s.

<details><summary>Example</summary>

```python
from typing import Optional, List

class Message:
    text: Optional[str]


def process_message(m: Message):
    # With a type guard
    if m.text is not None:
        print(m.text.split())
        
    # Without a type guard
    print(m.text.split())
    

def mock_dispatcher(ms: List[Message]):
    for m in ms:
        if m.text is not None:
            process_message(m)


message0 = Message()
message0.text = None

message1 = Message()
message1.text = "Hello world."

mock_dispatcher([message0, message1])

```

Gives the following error in `mypy`:

```
main.py:13: error: Item "None" of "Optional[str]" has no attribute "split"
Found 1 error in 1 file (checked 1 source file)
```

Result ran with https://mypy-play.net/?mypy=latest&python=3.9
</details>

Furthermore, looking back at this specific piece code, the `echo` function is just looping through the `get_updates()` results without any handlers, so a check on the `None`ness of `Message.text` is needed anyway.

blueset: On a sidenote, the opening comment of this file seems to be pointing itself for a `telegram.ext` example. Should it be something else instead?

https://github.com/python-telegram-bot/python-telegram-bot/blob/80b34811ab7cfb633d703f88d2a06df6e71d3551/examples/rawapibot.py#L4-L9

Bibo-Joshi: > Furthermore, looking back at this specific piece code, the echo function is just looping through the get_updates() results without any handlers, so a check on the Noneness of Message.text is needed anyway.

Good catch, you're right!

> On a sidenote, the opening comment of this file seems to be pointing itself for a telegram.ext example. Should it be something else instead?

Ups, seems the search & replace went too far, when we moved `echobot2.py` -> `echobot.py` and the former `echobot.py` -> `rawapibot.py`

Poolitzer (approved):
LGTM

PR #2243, opened 2020-12-13
Improve typing & docs of handlers

### LINKED ISSUE #2242: [BUG] Incorrect type hinting for Handlers
### Steps to reproduce

This line:

https://github.com/python-telegram-bot/python-telegram-bot/blob/73b0e29a308da7ebb35328e5731bee8db0c423be/telegram/ext/commandhandler.py#L94

contradicts with this line:

https://github.com/python-telegram-bot/python-telegram-bot/blob/73b0e29a308da7ebb35328e5731bee8db0c423be/telegram/utils/types.py#L36

And I think that the `update` parameter can actually never be `str`, right?

### Expected behaviour

No mypy issue for

```py
dispatcher.add_handler(telegram.ext.CommandHandler('start', start_command_handler))
```

and

```py
def start_command_handler(update: telegram.Update, context: telegram.ext.CallbackContext) -> None:
```

### Actual behaviour

```
Argument 2 to "CommandHandler" has incompatible type "Callable[[Update, CallbackContext], None]"; expected "Callable[[Union[str, Update], CallbackContext], None]" 
```

### Configuration

**Operating System:**

macOS 10.15.7 (19H15)

``$ python -m telegram``

python-telegram-bot 13.1
certifi 2020.11.08
Python 3.9.0 (default, Nov 21 2020, 14:01:50)  [Clang 12.0.0 (clang-1200.0.32.27)]

### Logs

Bibo-Joshi:
Closes #2242

I made `Handler` be a generic class, this allows us to use the "update type" consistently in the annotation of the callback and the methods. Also updated the docs a bit.

@revolter maybe you can install this branch via `pip install git+https://github.com/python-telegram-bot/python-telegram-bot.git@fix-2242` and check if it acutally works for you?

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -35,7 +35,7 @@ from telegram.ext import (
     InlineQueryHandler,
 )
 from telegram.utils.promise import Promise
-from telegram.utils.types import ConversationDict, HandlerArg
+from telegram.utils.types import ConversationDict
 
 if TYPE_CHECKING:
     from telegram.ext import Dispatcher, Job
@@ -56,7 +56,7 @@ class _ConversationTimeoutContext:
         self.callback_context = callback_context
 
 
-class ConversationHandler(Handler):
+class ConversationHandler(Handler[Update]):
     """
     A handler to hold a conversation with a single user by managing four collections of other
     handlers.
@@ -391,13 +391,13 @@ class ConversationHandler(Handler):
 
         return tuple(key)
 
-    def check_update(self, update: HandlerArg) -> CheckUpdateType:  # pylint: disable=R0911
+    def check_update(self, update: Any) -> CheckUpdateType:  # pylint: disable=R0911
         """
         Determines whether an update should be handled by this conversationhandler, and if so in
         which state the conversation currently is.
 
         Args:
-            update (:class:`telegram.Update`): Incoming telegram update.
+            update (:class:`telegram.Update` | :obj:`object`): Incoming update.
 
         Returns:
             :obj:`bool`
@@ -487,7 +487,7 @@ class ConversationHandler(Handler):
 
     def handle_update(  # type: ignore[override]
         self,
-        update: HandlerArg,
+        update: Update,
         dispatcher: 'Dispatcher',
         check_result: CheckUpdateType,
         context: CallbackContext = None,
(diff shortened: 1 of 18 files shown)


revolter (approved):
❤️

telegram/ext/conversationhandler.py
```diff
@@ -487,7 +487,7 @@ def check_update(self, update: HandlerArg) -> CheckUpdateType:  # pylint: disabl
 
     def handle_update(  # type: ignore[override]
         self,
-        update: HandlerArg,
+        update: Update,
```
Poolitzer, line 490: https://github.com/python-telegram-bot/python-telegram-bot/blob/1ddffc2f6321218fd5ec03af6bf3187de565deb0/telegram/ext/conversationhandler.py#L500

why is this still only a telegram update?

telegram/ext/dispatcher.py
```diff
@@ -632,12 +631,12 @@ def remove_error_handler(self, callback: Callable[[Any, CallbackContext], None])
...
 
     def dispatch_error(
-        self, update: Optional[HandlerArg], error: Exception, promise: Promise = None
+        self, update: Optional[Any], error: Exception, promise: Promise = None
     ) -> None:
         """Dispatches an error.
 
         Args:
-            update (:obj:`str` | :class:`telegram.Update` | None): The update that caused the error
+            update (:obj:`Any` | :class:`telegram.Update` | None): The update that caused the error
```
Poolitzer, line 639:
```suggestion
            update (:obj:`Any` | :class:`telegram.Update` | None): The update that caused the error.
```

telegram/ext/shippingqueryhandler.py
```diff
@@ -82,11 +81,11 @@ class ShippingQueryHandler(Handler):
 
     """
 
-    def check_update(self, update: HandlerArg) -> bool:
+    def check_update(self, update: Any) -> bool:
         """Determines whether an update should be passed to this handlers :attr:`callback`.
 
         Args:
-            update (:class:`telegram.Update`): Incoming telegram update.
+            update (:class:`telegram.Update` | :obj:`object`): Incoming telegram update.
```
Poolitzer, line 88:
```suggestion
            update (:class:`telegram.Update` | :obj:`object`): Incoming update.
```

Poolitzer (changes requested):
left a comment here and there

telegram/ext/conversationhandler.py
Bibo-Joshi: `check_update` must accept `Any`, `handle_update` needs only accept what `check_update` allows, so `tg.Update` for all the TG-handlers

Poolitzer: ahhh I see because we only allow telegram updates in check update, right.

Poolitzer (approved):
LGTM

PR #2266, opened 2020-12-23
Try testing on macos

Bibo-Joshi:


Bibo-Joshi:
* timing seems to be instable on macOS. we skip `JobQueue` tests on windows, should do that here, too. Maybe even for the `ConversationHandler` timeout thingies
* For the last three tests, the game picture is wrong and the bots don't have rights to expor the chat invite link of the corresponding group. Need to check which bots that are and change that

Bibo-Joshi:
Test failures seem to be down to the usual stuff. Tried to make them a bit more stable by xfailing all `TimedOut` errors and making the sleeps in the ConversationHandler-timeout tests a bit more generous.

Poolitzer:
I mean its still failing but also on windows so

PR #2282, opened 2020-12-31
Add 1.5 constants

Bibo-Joshi:
Adds a constant for max text length in `answer_callback_query` and the corresponding shortcut in `CallbackQuery`. Also adds a shortcut in `InlineQuery` for the max number of results.

Change is pretty easy, but @Poolitzer maybe you still can have a glance? Just in case you can come up with bettern names for the new constant & the shortcuts.


Poolitzer (approved):
More constants

Bibo-Joshi:
Test fails unrelated. merging.

PR #2285, opened 2021-01-01
Adjust calling of update_persistence

Bibo-Joshi:
This does two things:

* Respect the rather new `Defaults.run_async` when deciding if `update_persistence` should be called or not. This was just overlooked by me in #2210
* call `update_persistence` only once per update and only if the update was handled by at least once sync handler. Having a update handled by multiple groups can slow down the bot significantly otherwise.

Also this actually documents how stuff is updated (and fixes the annotation of `process_update` on the fly)

### DIFF
--- a/telegram/ext/defaults.py
+++ b/telegram/ext/defaults.py
@@ -173,7 +173,7 @@ class Defaults:
         )
 
     @property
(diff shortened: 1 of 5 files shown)


Poolitzer (approved):
LGTM

PR #2291, opened 2021-01-04
Add new shortcuts to Chat

### LINKED ISSUE #2284: [FEATURE] Shortcuts for Chat
Add `Chat.{promote, restrict}_member`.
We could also think about adding `ChatMember.{kick, unban, promote, restrict}`, though the usecase is small, as a `ChatMember` is only returned by `get_chat_member`, so you'd have to do something like

```pyhon
update.effective_chat.get_member(user_id).kick()
```
which is longer than

```python
update.effective_chat.kick_member(user_id)
```

Bibo-Joshi:
closes #2284

on the fly replaces some `.format(…)` left in `User`

### DIFF
--- a/telegram/chat.py
+++ b/telegram/chat.py
@@ -195,6 +195,10 @@ class Chat(TelegramObject):
         :obj:`str`: Convenience property. If :attr:`first_name` is not :obj:`None` gives,
         :attr:`first_name` followed by (if available) :attr:`last_name`.
 
+        Note:
+            :attr:`full_name` will always be :obj:`None`, if the chat is a (super)group or
+            channel.
+
         .. versionadded:: 13.2
         """
         if not self.first_name:
--- a/telegram/chat.py
+++ b/telegram/chat.py
@@ -190,6 +190,20 @@ class Chat(TelegramObject):
         self._id_attrs = (self.id,)
 
     @property
+    def full_name(self) -> Optional[str]:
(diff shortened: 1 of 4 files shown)


telegram/chat.py
```diff
@@ -189,6 +189,20 @@ def __init__(
...
+    @property
+    def full_name(self) -> Optional[str]:
+        """
+        :obj:`str`: Convenience property. If :attr:`first_name` is not :obj:`None` gives,
+        :attr:`first_name` followed by (if available) :attr:`last_name`.
+
+        .. versionadded:: 13.2
+        """
+        if not self.first_name:
+            return None
```
Poolitzer, line 201: wait, so this will always be None for not user chats? I think the docstring should reflect this, I wouldn't know as a first time API user that first_name is user only. Im not sure why people would use this over .effective_user, since that syntax is way clearer, but okay

Poolitzer, line 201: wait, so this will always be None for not user chats? I think the docstring should reflect this, I wouldn't know as a first time API user that first_name is user only. Im not sure why people would use this over .effective_user, since that syntax is way clearer, but okay

Poolitzer (commented):
LGTM

telegram/chat.py
Poolitzer: yay

PR #2292, opened 2021-01-06
Add run_async parameter to ConversationHandler

### LINKED ISSUE #2283: [FEATURE] ConversationHandler with option to override all handlers' run async option
I have a long `ConversationHandler` in which I want all of its handlers to be able to run asynchronously. To achieve this, I have to set every handler in this `ConversationHandler` with `run_async=True`. I found this a bit repetitive and thought that there might be an easier way.

I think it would be nice if there is an option in `ConversationHandler` like `all_handlers_run_async: bool`. When it is set to `True`, then all the handlers in this `ConversationHandler` would be overridden with `run_async=True`. 

We can also give a bit more flexibility by adding options like `all_entry_points_run_async`, `all_states_run_async` and `all_fallbacks_run_async`, where it only overrides a subset of handlers with `run_async=True`

N/A

#### Additional context
N/A

Let me know what you think and Happy New Year!

zeshuaro:
- Add `run_async` parameter to `ConversationHandler`
- When `run_async` is set to `True`, overrides all handlers in this `ConversationHandler` with `run_async` equals to `True`
- Closes #2283

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -136,8 +136,13 @@ class ConversationHandler(Handler[Update]):
         map_to_parent (Dict[:obj:`object`, :obj:`object`], optional): A :obj:`dict` that can be
             used to instruct a nested conversationhandler to transition into a mapped state on
             its parent conversationhandler in place of a specified nested state.
-        run_async (:obj:`bool`, optional): Pass :obj:`True` to override the
-            :attr:`Handler.run_async` setting of all handlers.
+        run_async (:obj:`bool`, optional): Pass :obj:`True` to *override* the
+            :attr:`Handler.run_async` setting of all handlers (in :attr:`entry_points`,
+            :attr:`states` and :attr:`fallbacks`).
+
+            Note:
+                If set to :obj:`True`, you should not pass a handler instance, that needs to be
+                run synchronously in another context.
 
             .. versionadded:: 13.2
 
@@ -172,9 +177,8 @@ class ConversationHandler(Handler[Update]):
         map_to_parent (Dict[:obj:`object`, :obj:`object`]): Optional. A :obj:`dict` that can be
             used to instruct a nested conversationhandler to transition into a mapped state on
             its parent conversationhandler in place of a specified nested state.
-        run_async (:obj:`bool`): Pass :obj:`True` to override the
-            :attr:`Handler.run_async` setting of all handlers (in :attr:`entry_points`,
-            :attr:`states` and :attr:`fallbacks`).
+        run_async (:obj:`bool`): If :obj:`True`, will override the
+            :attr:`Handler.run_async` setting of all internal handlers on initialization.
 
             .. versionadded:: 13.2
 
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -138,7 +138,9 @@ class ConversationHandler(Handler[Update]):
             its parent conversationhandler in place of a specified nested state.
         run_async (:obj:`bool`, optional): Pass :obj:`True` to override the
             :attr:`Handler.run_async` setting of all handlers (in :attr:`entry_points`,
-            :attr:`states` and :attr:`fallbacks:`.
+            :attr:`states` and :attr:`fallbacks`.
+
+            .. versionadded:: 13.2
 
     Raises:
         ValueError
@@ -173,7 +175,9 @@ class ConversationHandler(Handler[Update]):
             its parent conversationhandler in place of a specified nested state.
         run_async (:obj:`bool`): Pass :obj:`True` to override the
             :attr:`Handler.run_async` setting of all handlers (in :attr:`entry_points`,
-            :attr:`states` and :attr:`fallbacks:`.
+            :attr:`states` and :attr:`fallbacks`.
+
+            .. versionadded:: 13.2
 
     """
 
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -99,6 +99,11 @@ class ConversationHandler(Handler[Update]):
         :attr:`END` to end the *parent* conversation from within the nested one. For an example on
         nested :class:`ConversationHandler` s, see our `examples`_.
 
+        Note that setting :attr:`run_async` to :obj:`True` will override the attribute
+        :attr:`run_async` in all the handlers used in this :class:`ConversationHandler` and set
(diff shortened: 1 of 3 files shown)


telegram/ext/conversationhandler.py
```diff
@@ -136,6 +141,9 @@ class ConversationHandler(Handler[Update]):
         map_to_parent (Dict[:obj:`object`, :obj:`object`], optional): A :obj:`dict` that can be
             used to instruct a nested conversationhandler to transition into a mapped state on
             its parent conversationhandler in place of a specified nested state.
+        run_async (:obj:`bool`, optional): If set to :obj:`True`, all the handlers in this
+            :class:`ConversationHandler` will be set with :attr:`run_async` equals to :obj:`True`.
```
Bibo-Joshi, line 145: The wording is somewhat hard to understand (same below). suggestion:

```suggestion
        run_async (:obj:`bool`, optional): Pass :obj:`True` to override the :attr:`Handler.run_async` setting of all handlers (in :attr:`entry_points`, :attr:`states` and :attr:`fallbacks:`.
```

I guess this formulation would also make above note unnecessary …

tests/test_conversationhandler.py
```diff
@@ -1429,3 +1429,35 @@ def test_callback(u, c):
...
+        )
+
+        all_handlers = conv_handler.entry_points + conv_handler.fallbacks
+        for state_handlers in conv_handler.states.values():
+            all_handlers += state_handlers
+
+        for handler in all_handlers:
+            assert handler.run_async
+
+    def test_conversation_handler_run_async_false(self, dp):
```
Bibo-Joshi, line 1448: maybe make the entry_point async, to check that that won't be overriden to `entry_point[0].run_async=False`

Bibo-Joshi (changes requested):
Thanks for your PR!
Looking at the changes, I was wondering if maybe it's a better idea to not override the setting of the handlers but only change it during `handle_update`, i.e. something along the lines

```python
setting = handler.rus_async
handler.run_async = True if self.run_async else handler.run_async
handler.handle_updater(update)
handler.run_async = setting
```

but I'm not 100% sure if that could still have side effects if the same handler object is used in two places … Currently `Dispatcher.process_update` is calling `handle_update()` one after the other, even if internally they might call `Dispatcher.run_async` but if that's changed at some point and the above approach will be hard to debug. and maybe the use case for having the same handler object in multiple places is small enough to just go with overriding.
@Poolitzer do you have an opinion on this?

zeshuaro:
How about deep copying the handlers and with `run_async` set to `True`? Would that be easier and also solves the problem of users passing the same handlers in different places?

tests/test_conversationhandler.py
zeshuaro: I've already have this in the test, see line 1450 for setting the `CommandHandler` with `run_async=True` and testing it in lines 1455-1456.

Bibo-Joshi:
> How about deep copying the handlers and with `run_async` set to `True`? Would that be easier and also solves the problem of users passing the same handlers in different places?

Mh, that's risky. What do we do if someone uses a custom handler with a `Lock()` in it? that can't be deepcopied.

zeshuaro:
> > How about deep copying the handlers and with `run_async` set to `True`? Would that be easier and also solves the problem of users passing the same handlers in different places?
> 
> Mh, that's risky. What do we do if someone uses a custom handler with a `Lock()` in it? that can't be deepcopied.

Right, didn't know we can define a custom handler but that makes sense.

Going back to what we have implemented in this PR. If a user set `run_async=True` in the `ConversationHandler`, then they will expect all handlers to be configured with `run_async=True`, i.e. this handler should work and run with `run_async=True`. And therefore can we assume if the same handler instance is used elsewhere, it will run with `run_async=True` as well?

telegram/ext/conversationhandler.py
zeshuaro: Updated with new commits

Bibo-Joshi:
> Going back to what we have implemented in this PR. If a user sets `run_async=True` in the `ConversationHandler`, then they will expect all handlers to be configured with `run_async=True`, i.e. these handlers should work and run with `run_async=True`. And therefore can we assume if the same handler instance is used elsewhere, it will run with `run_async=True` as well?

Yes, my point is that that's rather unexpected. With proper documentation I think I'm okay with it but that's why I asked for pools opinion.

BTW, before I forget: we started adding `.. versionadded:: 13.2` thingies, please add that to the new argument ;)

Poolitzer:
> @Poolitzer do you have an opinion on this?

I do not. what would be the upside to your suggested solution?

telegram/ext/conversationhandler.py
```diff
@@ -136,6 +136,11 @@ class ConversationHandler(Handler[Update]):
         map_to_parent (Dict[:obj:`object`, :obj:`object`], optional): A :obj:`dict` that can be
             used to instruct a nested conversationhandler to transition into a mapped state on
             its parent conversationhandler in place of a specified nested state.
+        run_async (:obj:`bool`, optional): Pass :obj:`True` to override the
+            :attr:`Handler.run_async` setting of all handlers (in :attr:`entry_points`,
+            :attr:`states` and :attr:`fallbacks`.
```
Poolitzer, line 141:
```suggestion
        run_async (:obj:`bool`, optional): Pass :obj:`True` to override the
            :attr:`Handler.run_async` setting of all handlers.
```

We try to keep the extensive documentation to the Parameters and have shorter ones which are sufficient in the attributes :)

telegram/ext/conversationhandler.py
```diff
@@ -168,6 +173,11 @@ class ConversationHandler(Handler[Update]):
         map_to_parent (Dict[:obj:`object`, :obj:`object`]): Optional. A :obj:`dict` that can be
             used to instruct a nested conversationhandler to transition into a mapped state on
             its parent conversationhandler in place of a specified nested state.
+        run_async (:obj:`bool`): Pass :obj:`True` to override the
+            :attr:`Handler.run_async` setting of all handlers (in :attr:`entry_points`,
+            :attr:`states` and :attr:`fallbacks`.
```
Poolitzer, line 178:
```suggestion
            :attr:`states` and :attr:`fallbacks`).
```

look at me ~go~ closing parentheses

Poolitzer (changes requested):
docs nit-picking

Bibo-Joshi:
I pushed a little bit more doc nitpicking … Thank you for your contribution @zeshuaro ! Merging.

PR #2301, opened 2021-01-07
Add pyupgrade to pre-commit

Bibo-Joshi:
Stumbled accros that project it thought it's neat. TBH, I'm not 100% we should add it to the pre-commit as it's somewhat young (2017) and doesn't seem well recognized in terms of stars, forks and dependent packages (https://github.com/asottile/pyupgrade).
Then again, dependent packages don't list pre-commit iirc, and it's apparently rather actively maintained by the pre-commit author himself.
Also I didn't find an option to make it *not* change files in pre-commit (it fails, but automatically modifies the files).
Anyway, even if we don't add it, at least the changes it made seem valid :)

### DIFF
--- a/telegram/ext/picklepersistence.py
+++ b/telegram/ext/picklepersistence.py
@@ -105,7 +105,7 @@ class PicklePersistence(BasePersistence):
                 # For backwards compatibility with files not containing bot data
                 self.bot_data = data.get('bot_data', {})
                 self.conversations = data['conversations']
-        except IOError:
+        except OSError:
             self.conversations = dict()
             self.user_data = defaultdict(dict)
             self.chat_data = defaultdict(dict)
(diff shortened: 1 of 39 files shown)


Poolitzer:
having a quick glance at the files, they look alright to me. I would vote in favour of implementing this

telegram/ext/picklepersistence.py
```diff
@@ -105,7 +105,7 @@ def load_singlefile(self) -> None:
                 # For backwards compatibility with files not containing bot data
                 self.bot_data = data.get('bot_data', {})
                 self.conversations = data['conversations']
-        except IOError:
+        except OSError:
```
Poolitzer, line 108: we sure this change is alright?

Poolitzer (approved):
alright, looks good, just one comment

telegram/ext/picklepersistence.py
Bibo-Joshi: yeah, looked it up, some exceptions were merged into OSError. just below [here](https://docs.python.org/3/library/exceptions.html#OSError.filename2)

Poolitzer: cool

Poolitzer (approved):

Bibo-Joshi:
I'm gonna go ahead and merge. If we dislike it at some point, we can kick it, but we don't have to bee too strict on dev dependencies I guess …

PR #2312, opened 2021-01-11
Fix: overly escape and offset error for markdown v2 and HTML symbols when nested

blueset:
Previously, PTB is producing incorrect results when formatting a message into the Markdown v2 format (`Message.(text|caption)_markdown_v2`). Specifically, when there is any of the need-to-escape characters <code>_*[]()~`>#+-=|{}.!</code> appear in a nested entities, the last few characters was repeated at the end of the range. 

For example

Message | What PTB gives | Expected result
-- | -- | --
`<b><hashtag>#boldhashtag</hashtag></b>` | `*#boldhashtag*g` | `*#boldhashtag*`
`<b><i>a{b+c}d</i></b>` | `*_a\{b\+c\}d_\\}d*` | `*_a\{b\+c\}d_*`

_\* `<hashtag />` is used here just to represent the hashtag entity, Telegram does not support such a tag._

The issue is suspected to be caused by the used of _escaped_ text sequence for nested entities, which is causing a discrepancy in the length of text due to the introduction of `\`.

This PR fixes the issue by using the original text (`orig_text`) as a basis for nested entity parsing. It also includes updated tests to address this issue.

### DIFF
--- a/telegram/message.py
+++ b/telegram/message.py
@@ -2171,11 +2171,12 @@ class Message(TelegramObject):
                 }
                 parsed_entities.extend(list(nested_entities.keys()))
 
+                orig_text = text
                 text = escape(text)
 
                 if nested_entities:
                     text = Message._parse_html(
-                        text, nested_entities, urled=urled, offset=entity.offset
+                        orig_text, nested_entities, urled=urled, offset=entity.offset
                     )
 
                 if entity.type == MessageEntity.TEXT_LINK:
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
Hi, thanks for the finding and the PR! On a closer look, the same problem happens with HTML formatting (replace `a{b+c}` with `a<b+c}`, where the `<` needs to be escaped in HTML). Would you mind applying the fix to `_parse_html` and adjusting the tests accordingly?

blueset:
@bibo-joshi, nice spot! I just added the same fix to the HTML parser as well. For the tests, I included the `>` symbol in the message text, which is the only one required to be escaped in both format.

Bibo-Joshi:
Perfect, thanks for the contribution :)

PR #2320, opened 2021-01-15
Make telegram.Bot comparable

### LINKED ISSUE #2319: [FEATURE] Make tg.Bot comparable
`tg.Bot` didn't get rich comparison in #1724, though there are use cases when comparison of `Bot` with `User` objects would be helpful. Most notable when checking if the bot is one of the new chat members reportet in a status update; with rich comparison that would be as easy as

```python
context.bot in update.effective_message.new_chat_members
```

Currently one can use `bot_instance.bot`, which holds the `User` instance for comparison, but it's undocumented and unintuitive. Proposal:

* Make `Bot.__eq__(self, other)` basically return `self.bot == other` and document the behaviour accordingly
* we need to make sure that `self.bot` is already set. For the `Bot.{id, username, …}` properties, that's currently done with the `@info` decorator. I think it would be cleaner to just convert `self.bot` into a property that calls `self.get_me()` if needed.

starry69:
Closes #2319

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -129,8 +129,8 @@ class Bot(TelegramObject):
     """This object represents a Telegram Bot.
 
     .. versionadded:: 13.2
-    Objects of this class are comparable in terms of equality. Two objects of this class are
-    considered equal, if their :attr:`bot` is equal.
+        Objects of this class are comparable in terms of equality. Two objects of this class are
+        considered equal, if their :attr:`bot` is equal.
 
     Note:
         Most bot methods have the argument ``api_kwargs`` which allows to pass arbitrary keywords
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -110,22 +110,6 @@ if TYPE_CHECKING:
 RT = TypeVar('RT')
 
 
-def info(func: Callable[..., RT]) -> Callable[..., RT]:
-    # pylint: disable=W0212
-    @functools.wraps(func)
-    def decorator(self: 'Bot', *args: Any, **kwargs: Any) -> RT:
-        if not self.bot:
-            self.get_me()
-
-        if self._commands is None:
-            self.get_my_commands()
-
-        result = func(self, *args, **kwargs)
-        return result
-
-    return decorator
-
-
 def log(
     func: Callable[..., RT], *args: Any, **kwargs: Any  # pylint: disable=W0613
 ) -> Callable[..., RT]:
@@ -217,7 +201,7 @@ class Bot(TelegramObject):
 
         self.base_url = str(base_url) + str(self.token)
         self.base_file_url = str(base_file_url) + str(self.token)
-        self.bot: Optional[User] = None
+        self._bot: Optional[User] = None
         self._commands: Optional[List[BotCommand]] = None
         self._request = request or Request()
         self.logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


telegram/bot.py
```diff
@@ -303,67 +287,68 @@ def _validate_token(token: str) -> str:
         return token
 
     @property  # type: ignore
-    @info
+    def bot(self) -> User:
+        """:class:`telegram.User`: Instance for the bot."""
```
Bibo-Joshi, line 291:
```suggestion
        """:class:`telegram.User`: User instance for the bot as returned by :meth:`get_me`."""
```

telegram/bot.py
```diff
@@ -303,67 +287,68 @@ def _validate_token(token: str) -> str:
...
+
+        if self._bot is None:
+            self.get_me()
```
Bibo-Joshi, line 294: this is probably why everthing's  failing ;)

```suggestion
            self._bot = self.get_me()
```

telegram/bot.py
```diff
@@ -303,67 +287,68 @@ def _validate_token(token: str) -> str:
...
         return self.bot.supports_inline_queries  # type: ignore
 
     @property  # type: ignore
-    @info
     def commands(self) -> List[BotCommand]:
         """List[:class:`BotCommand`]: Bot's commands."""
 
-        return self._commands or []
+        if self._commands is None:
+            self.get_my_commands()
```
Bibo-Joshi, line 350:
```suggestion
           self._commands = self.get_my_commands()
```

tests/test_bot.py
```diff
@@ -145,6 +145,12 @@ def test_get_me_and_properties(self, bot):
         assert f'https://t.me/{get_me_bot.username}' == bot.link
         assert commands == bot.commands
 
+    def test_comparison(self, bot):
```
Bibo-Joshi, line 148: I think this test should be elaborated. please have a look at the other TG object classes tests for inspiration ;)

Bibo-Joshi (changes requested):

telegram/bot.py
starry69: Oh i thought as `bot.get_me()` call sets attr anyway, so we can leave that...

Bibo-Joshi (changes requested):
Mh, actually you're right about `self.get_me()` already setting `self._bot`, but imo it's still more readible to assign the return value anyway.

Some more notes:

* Please document that bot is now comparable (See e.g. [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/be54cf4ecedfd52c1aa684eb03ad5ef9206b4b1a/telegram/chat.py#L62)) probably just do it als
  ```python
  .. versionadded:: 13.2
     Objects of this class …
  ```
* After a closer look the tests fail b/c some tests want to monkeypatch `bot.bot`, which is no longer possible. Ig just making them patch `bot._bot` instead should be fine
* Some other tests fail, b/c `Bot` is no longer hashable. `Bot.__hash__` should return `hash(self.bot)`, as that's what we're comparing with. see [here](https://docs.python.org/3.5/reference/datamodel.html#object.__hash__)
* For what the equality tests should look like please see [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/be54cf4ecedfd52c1aa684eb03ad5ef9206b4b1a/tests/test_user.py#L539)

telegram/bot.py
```diff
@@ -144,6 +128,10 @@ def decorator(self: 'Bot', *args: Any, **kwargs: Any) -> RT:  # pylint: disable=
 class Bot(TelegramObject):
     """This object represents a Telegram Bot.
 
+    .. versionadded:: 13.2
+    Objects of this class are comparable in terms of equality. Two objects of this class are
+    considered equal, if their :attr:`bot` is equal.
+
```
Bibo-Joshi, line 134: Can you build the docs locally and check if this is rendered correctly? IIRC the text below the directive needs to be indented for sphinx to see it as explanation.

telegram/bot.py
```diff
@@ -302,68 +290,69 @@ def _validate_token(token: str) -> str:
...
-        return self.bot.first_name  # type: ignore
+        return self.bot.first_name
 
-    @property  # type: ignore
-    @info
+    @property
     def last_name(self) -> str:
         """:obj:`str`: Optional. Bot's last name."""
 
         return self.bot.last_name  # type: ignore
```
Bibo-Joshi, line 317: Do we still need those `# type: ignore` comments? 🤔 up to now that fact that `self.bot` was not `None` was hidden in the `@info` decorator, so I guess now mypy can do without them

Bibo-Joshi (commented):
Thanks for the updates! Just two comments left :)

telegram/bot.py
starry69: IMG_20210116_234042_007
Ig it's looks fine 🤔

telegram/bot.py
starry69: IMG_20210116_234042_007
Ig it looks fine 🤔

telegram/bot.py
starry69: Umm.. i removed most `# type: ignore` comments but removing those 2 was causing mypy to fail in my local environment. So i left them as it was

telegram/bot.py
Bibo-Joshi: mh, it's supposed to look like this:

image

telegram/bot.py
Bibo-Joshi: okay, let's leave them for now

telegram/bot.py
starry69: Oops, okay

PR #2321, opened 2021-01-15
Use object instead of Any where possible

Bibo-Joshi:
Apparently I only now understood the difference between `object` and `Any` as annotiation. TL;DR:

* `var: object` tells the typchecker that literally any object must be handled correctly
* `var: Any` just won't typecheck usages of `var` at all

So this drops `Any` wherever possible in favor of `object`


Bibo-Joshi:
I'll just go ahead and merge this. as noam once put it "type hints are basically documentation" …

PR #2323, opened 2021-01-16
Drop nightly tests & update badges

Bibo-Joshi:
uses shield.io for the dowsload badge instead of custom solution. https://shield.io uses https://pypistats.org

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,8 +3,6 @@ on:
   pull_request:
     branches:
       - master
-  schedule:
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
For future reference: the custom solution was added in #692, a year before the inital commit to the [pypistats.org repo](https://github.com/crflynn/pypistats.org/commit/741e385e1e50860e25a7863b3f7b68baa0f0ebed). Also pypi doesn't directly provide any download statistics directly, that's why this PR introduces 3rd party services.

PR #2335, opened 2021-01-25
extend deeplinking example

### LINKED ISSUE #2318: Showcase answer_callback_query with deeplinking URL
Instead of using the deeplinking URL directly in text links or inline buttons, one can also use them as `answer_inline_query(url=deeplinking_url)`, as done e.g. by @spoilerobot . I think it would be a good idea to include this option in the `deeplinking.py` example somehow.

starry69:
Closes #2318

### DIFF
--- a/examples/deeplinking.py
+++ b/examples/deeplinking.py
@@ -22,21 +22,31 @@ bot.
 import logging
 
 from telegram import ParseMode, InlineKeyboardMarkup, InlineKeyboardButton, Update
-from telegram.ext import Updater, CommandHandler, Filters, CallbackContext
+from telegram.ext import (
+    Updater,
+    CommandHandler,
+    CallbackQueryHandler,
+    Filters,
+    CallbackContext,
+)
 
 # Enable logging
 from telegram.utils import helpers
 
 logging.basicConfig(
-    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s', level=logging.INFO
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s", level=logging.INFO
 )
 
 logger = logging.getLogger(__name__)
 
 # Define constants that will allow us to reuse the deep-linking parameters.
-CHECK_THIS_OUT = 'check-this-out'
-USING_ENTITIES = 'using-entities-here'
-SO_COOL = 'so-cool'
+CHECK_THIS_OUT = "check-this-out"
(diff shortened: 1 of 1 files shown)


starry69:
That JobQueue fail seems unrelated to me

examples/deeplinking.py
```diff
@@ -22,21 +22,31 @@
...
+    Filters,
+    CallbackContext,
+)
 
 # Enable logging
 from telegram.utils import helpers
 
 logging.basicConfig(
-    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s', level=logging.INFO
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s", level=logging.INFO
```
Bibo-Joshi, line 37: Why were the quotes changed? did you do that manually or was that black? Same below

Bibo-Joshi (commented):
Looks good :) just one question

examples/deeplinking.py
starry69: Yep, done by black.

examples/deeplinking.py
```diff
@@ -71,6 +81,23 @@ def deep_linked_level_2(update: Update, context: CallbackContext) -> None:
...
+        reply_markup=InlineKeyboardMarkup(
+            [[InlineKeyboardButton(text="Like this!", callback_data=KEYBOARD_CALLBACKDATA)]]
+        ),
+    )
+
+
+def deep_link_level_3_callback(update: Update, context: CallbackContext) -> None:
+    """Answers CallbackQuery with deeplinking url."""
+    bot = context.bot
+    url = helpers.create_deep_linked_url(bot.get_me().username, USING_KEYBOARD)
```
Bibo-Joshi, line 95: doesn't `Bot` have a convenience property for the username? 🤔

examples/deeplinking.py
Bibo-Joshi: mh, strange that it didn't change that before … but whatever.

Bibo-Joshi (commented):
Found one more thing :)

examples/deeplinking.py
starry69: You're right, i just followed how it was done for other deeplinking levels without thinking, will change 👍

Bibo-Joshi (approved):

Issue #2372, opened 2021-02-09
Handle cryptography


Bibo-Joshi:
seems like #2370 was too short-sighted as

* the cyclic imports apparently are back in 3.4.3, see pyca/cryptography#5794
* Building PTB via `setup.py` apparently fails as for some reason this tried to build cryptography from sources? At least [on RTD](https://readthedocs.org/projects/python-telegram-bot/builds/12968269/)
* There is currently a heavy discussion in progress in pyca/cryptography#5771 as apparently the usage of Rust code in v3.4 has broken CIs of a lot of people

We'll need to check out

* How this affects our building process
* How we can reasonably protect our users from install problems. Mind you, installation problems of 3rd party deps is not fully our problem, but if the majority of our users can't install it, then our package is worthless.

One could just pin cryptography to < 3.4 for starters and just wait a bit until the dust settles …

Bibo-Joshi:
PS: RTD might be easy to fix with the [V2 configuration fil](https://docs.readthedocs.io/en/stable/config-file/v2.html#packages)e, which allows to install via pip instead of `setup.py`. Still we'll have to check out, if our release process or building from source process is affected.

PR #2377, opened 2021-02-09
Pin cryptography to 3.3.2

Bibo-Joshi:
See #2372

I propose to leave #2372 open and give the cryptography time to figure out their plans for the future. As we need cryptography only for passports (and I've never met anyone acutally using those …) it's not a main security issue. Nevertheless I subscribed to the cryptography release channel to keep an eye on the changelog for security fixes.

If the situation is more clear in the future we can think about how we want to approach the issue.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 certifi
-cryptography!=3.4
+cryptography==3.3.2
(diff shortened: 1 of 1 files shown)


Bibo-Joshi:
Closing in favor of #2386

PR #2386, opened 2021-02-13
Make cryptography optional

### LINKED ISSUE #2372: Handle cryptography

Bibo-Joshi:
as discussed offline.

Adds requirements-extra called `passport`. Installing via `pip install ptb[passport]` seems more intuitive to me than installing as `pip install ptb[crypto]` and it's more flexible in terms of which optional dependencies are actually needed for passports.

superseeds #2377 
closes #2372

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,10 +39,11 @@ jobs:
         # 3. Test everything else
         # The second one is achieved by mocking the imports, see test_no_passport.py for details
         run: |
-          pytest -v --cov -k test_no_passport.py
+          pytest -v -s --cov -k test_no_passport.py
           no_passport_exit=$?
           export TEST_NO_PASSPORT='false'
-          pytest -v --cov --cov-append
+          echo "TEST_NO_PASSPORT now has the value ${TEST_NO_PASSPORT}"
+          pytest -v -s --cov --cov-append
           passport_exit=$?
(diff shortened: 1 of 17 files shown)


Bibo-Joshi:
Upgrading pytest required a small workaround, see pytest-dev/pytest#8343. Documented with comments. Let's see if that did the trick.

Bibo-Joshi:
Yeay, CI runs 🥳 Codecov is shy for some reason, but it looks good in the [dashboard](https://app.codecov.io/gh/python-telegram-bot/python-telegram-bot/compare/2386).

Issue #2392, opened 2021-02-18
[BUG] Shortcuts lost their *args and **kwargs, resulting messagequeue isgroup= and other broke


standin-fecha:
There is no complicated stuff, someone just removed all *args and **kwargs from shortcuts such as reply_text in message, so old methods used with messagequeue wrapper and isgroup parameter stopped working because they simply can't pass this parameteres to strictly defined function. It's especially strange in case that messagequeue itself wasn't removed and from direct send_message methods works fine.

Poolitzer:
Hey, thank you for your issue report.

We introduced strongly typed shortcuts in our latest Version, 13.2. We also got rid of the kwargs and args parameters there, because they inconveniently hid typing errors. If you want to pass kwargs manually, we introduced `api_kwargs` as an argument where you can pass the kwargs as a dict.

Sadly, you said we broke messagequeue in the process. Can I ask you to provide a minimal working example (see [here](https://telegra.ph/Minimal-Working-Example-for-PTB-07-18) for what that is), so it is easier for me to do the debugging?

Btw, probably unrelated note, but we are going to give MQ a massive overhaul once we have proper async support, see #2139

standin-fecha:
Thanks a lot for the answer, a bit ugly construction api_kwargs={'isgroup': True} saved beautiful formal isgroup=True.

If I'm not wrong, it wasn't directly mentioned in changelog, so I missed it when the update came out and found strange error about it only now, while updating another Bot.

messagequeue itself working as usual, I'm talking only about shorcuts, which lost their direct option to pass isgroup without any error highlights or something until exact executing.

Yep, I knew about coming MQ refactoring, but it's open since 12 October and still in work, so feel a bit hopeless to finally get it. :) Working Defaults are the most wanted feature in it as for code beauty lover.

Poolitzer:
> Thanks a lot for the answer

No problem :)

>  a bit ugly construction api_kwargs={'isgroup': True} saved beautiful formal isgroup=True.

Agreed.

> If I'm not wrong, it wasn't directly mentioned in changelog,

It was one of the main features. (Explicit Signatures for Shortcuts (#2240). see [release notes](https://github.com/python-telegram-bot/python-telegram-bot/releases/tag/v13.2))

> and found strange error about it

I guess we could've warned more about this, yeah. It didn't occur to me that MQ would use kwargs directly, never used it though.

> which lost their direct option to pass isgroup without any error highlights or something until exact executing.

[...]

PR #2393, opened 2021-02-18
Deprecate MQ

Bibo-Joshi:
As discussed in #2392. Note that I did not add tests for the warnings to be raised due to the broken car analogy. I tested manually, though …


Poolitzer (commented):
good enough for me, but we should get the ok from other maintainers

PR #2409, opened 2021-03-02
Stabilize Tests

Bibo-Joshi:
see the failures e.g. [here](https://github.com/python-telegram-bot/python-telegram-bot/pull/2404/checks?check_run_id=2014971523)

### DIFF
--- a/tests/test_photo.py
+++ b/tests/test_photo.py
@@ -63,7 +63,7 @@ class TestPhoto:
     width = 800
     height = 800
     caption = '<b>PhotoTest</b> - *Caption*'
-    photo_file_url = 'https://python-telegram-bot.org/static/testfiles/telegram_new.jpg'
+    photo_file_url = 'https://python-telegram-bot.org/static/testfiles/telegram.jpg'
     file_size = 29176
 
     def test_creation(self, thumb, photo):
@@ -191,9 +191,6 @@ class TestPhoto:
         message = bot.send_photo(
             chat_id, photo_file, caption=test_string, caption_entities=entities
         )
-        # message = bot.send_photo(
-        #     chat_id, photo_file, caption=test_string, caption_entities=entities
-        # )
 
         assert message.caption == test_string
         assert message.caption_entities == entities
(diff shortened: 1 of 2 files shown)


tests/test_photo.py
```diff
@@ -63,7 +63,7 @@ class TestPhoto:
     width = 800
     height = 800
     caption = '<b>PhotoTest</b> - *Caption*'
-    photo_file_url = 'https://python-telegram-bot.org/static/testfiles/telegram_new.jpg'
+    photo_file_url = 'https://python-telegram-bot.org/static/testfiles/telegram.jpg'
```
Poolitzer, line 66: why?

tests/test_photo.py
```diff
@@ -83,7 +83,8 @@ def test_creation(self, thumb, photo):
     def test_expected_values(self, photo, thumb):
         assert photo.width == self.width
         assert photo.height == self.height
-        assert photo.file_size == self.file_size
+        # For some reason something changed at some point, that's why we allow two values here …
+        assert photo.file_size in [self.file_size, 27662]
```
Poolitzer, line 87: wont that break the general comparison experience between two photo object? or rather: Why are there two values?

tests/test_photo.py
```diff
@@ -305,18 +303,16 @@ def test_send_url_jpg_file(self, bot, chat_id, thumb, photo):
         assert isinstance(message.photo[0].file_unique_id, str)
         assert message.photo[0].file_id != ''
         assert message.photo[0].file_unique_id != ''
-        assert message.photo[0].width == thumb.width
-        assert message.photo[0].height == thumb.height
-        assert message.photo[0].file_size == thumb.file_size
+        # We used to test for width, height and file_size, but TG apparently started to treat
+        # sending by URL and sending by upload differently and it's not really our job anyway ...
```
Poolitzer, line 307: oh so it compresses these now? good god...

Poolitzer (commented):
look at me asking questions.

changes itself look good though

tests/test_photo.py
Bibo-Joshi: because iisc that's the file the `photo` fixture uses. I don't know what exactly the difference is, but it was a first guess to fix the tests and at least didn't make them worse …

tests/test_photo.py
Bibo-Joshi: because in recent tests suits for some reason in *some* (not all) os-py version combinatitions the second file size is generated …

Poolitzer: consistent?

tests/test_photo.py
Poolitzer: welp, okay.

tests/test_photo.py
Bibo-Joshi: yes, I've only ever seen 27662 as resulting file size in the failing tests. But, well, similar as for the other tests, it's not really our job to check the size and file size of the resulting photo, IG, we only want to make sure that a photo __is__ sent. If/How it's compressed is TGs decision. So we could also just drop those tests, if you're more comfortable with that

Poolitzer: I think I would be, yeah. Better then having to add a random number and hope telegram doesn't change how they compress

Bibo-Joshi: All right. Only dropped the file size test for now and kept dimensions. The comment should be clear enough in case that fails in the future, too.

Poolitzer: sure

Poolitzer (approved):

PR #2417, opened 2021-03-05
Improve timeouts handling in conversation handlers

### LINKED ISSUE #2407: Check promise states in conversation timeout callback | timeouts & nested conversations
When using `ConversationHandler` with async handler callbacks and timeout, the timeout may trigger falsely. This is the case, when the callback returns `CH.END`, which currently the timeout callback doesn't know.

This can be solved by making `CH._trigger_timeout` check if the current conversations state as follows:

```python
if isinstance(state, Promise):
    if state.done and state.result() == CH.END:
        return
```

This will still fail, if the promise is not yet done, when the timeout triggers but

1. it's a best effort solution
2. if a promise takes longer to run than the timeout is, then there are probably some other issues in the users code

Also when looking at the code I noticed that we don't handle nested conversations in `CH._trigger_timeout`, so that's a good opportunity to fix that …

starry69:
Closes #2407

### DIFF
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -467,9 +467,11 @@ class ConversationHandler(Handler[Update]):
         conversation_key: Tuple[int, ...],
     ) -> None:
         if new_state != self.END:
-            self.timeout_jobs[conversation_key] = dispatcher.job_queue.run_once(
+            self.timeout_jobs[
+                conversation_key
+            ] = dispatcher.job_queue.run_once(  # type:ignore[union-attr]
                 self._trigger_timeout,  # type: ignore[arg-type]
-                self.conversation_timeout,
+                self.conversation_timeout,  # type: ignore[arg-type]
                 context=_ConversationTimeoutContext(conversation_key, update, dispatcher, context),
             )
 
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -667,9 +667,8 @@ class ConversationHandler(Handler[Update]):
 
         # Backward compatibility with bots that do not use CallbackContext
         if isinstance(context, CallbackContext):
-            ctxt = cast(
-                _ConversationTimeoutContext, context.job.context  # type: ignore[union-attr]
-            )
+            job = context.job
+            ctxt = cast(_ConversationTimeoutContext, job.context)  # type: ignore[union-attr]
         else:
             ctxt = cast(_ConversationTimeoutContext, job.context)
 
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -22,8 +22,9 @@
 import logging
 import warnings
 import functools
+import datetime
 from threading import Lock
-from typing import TYPE_CHECKING, Dict, List, NoReturn, Optional, Tuple, cast, ClassVar
+from typing import TYPE_CHECKING, Dict, List, NoReturn, Optional, Union, Tuple, cast, ClassVar
 
 from telegram import Update
 from telegram.ext import (
@@ -146,9 +147,6 @@ class ConversationHandler(Handler[Update]):
             :attr:`ConversationHandler.TIMEOUT`.
 
             Note:
-                * `conversation_timeout` with `run_async` may fail, if the
-                  :class:`telegram.ext.utils.promise.Promise` takes longer to finish than the
-                  timeout.
                 * Using `conversation_timeout` with nested conversations is currently not
                   supported. You can still try to use it, but it will likely behave differently
                   from what you expect.
@@ -226,7 +224,7 @@ class ConversationHandler(Handler[Update]):
         per_chat: bool = True,
         per_user: bool = True,
         per_message: bool = False,
-        conversation_timeout: int = None,
+        conversation_timeout: Union[float, datetime.timedelta] = None,
         name: str = None,
         persistent: bool = False,
         map_to_parent: Dict[object, object] = None,
@@ -373,7 +371,9 @@ class ConversationHandler(Handler[Update]):
         raise ValueError('You can not assign a new value to per_message after initialization.')
 
     @property
-    def conversation_timeout(self) -> Optional[int]:
+    def conversation_timeout(
+        self,
+    ) -> Optional[Union[float, datetime.timedelta]]:
         return self._conversation_timeout
 
     @conversation_timeout.setter
@@ -467,13 +467,19 @@ class ConversationHandler(Handler[Update]):
         conversation_key: Tuple[int, ...],
     ) -> None:
         if new_state != self.END:
-            self.timeout_jobs[
-                conversation_key
-            ] = dispatcher.job_queue.run_once(  # type:ignore[union-attr]
-                self._trigger_timeout,  # type: ignore[arg-type]
-                self.conversation_timeout,  # type: ignore[arg-type]
-                context=_ConversationTimeoutContext(conversation_key, update, dispatcher, context),
-            )
+            try:
+                # both job_queue & conversation_timeout are checked before calling _schedule_job
+                j_queue = dispatcher.job_queue
+                self.timeout_jobs[conversation_key] = j_queue.run_once(  # type: ignore[union-attr]
+                    self._trigger_timeout,  # type: ignore[arg-type]
+                    self.conversation_timeout,  # type: ignore[arg-type]
+                    context=_ConversationTimeoutContext(
+                        conversation_key, update, dispatcher, context
+                    ),
+                )
+            except Exception as exc:
+                self.logger.exception("Failed to add timeout job due to exception")
+                self.logger.exception("%s", exc)
 
     def check_update(self, update: object) -> CheckUpdateType:  # pylint: disable=R0911
         """
@@ -596,20 +602,25 @@ class ConversationHandler(Handler[Update]):
             new_state = exception.state
             raise_dp_handler_stop = True
         with self._timeout_jobs_lock:
-            if self.conversation_timeout and dispatcher.job_queue:
-                # Add the new timeout job
-                if isinstance(new_state, Promise):
-                    new_state.add_done_callback(
-                        functools.partial(
-                            self._schedule_job,
-                            dispatcher=dispatcher,
-                            update=update,
-                            context=context,
-                            conversation_key=conversation_key,
+            if self.conversation_timeout:
+                if dispatcher.job_queue is not None:
+                    # Add the new timeout job
+                    if isinstance(new_state, Promise):
+                        new_state.add_done_callback(
+                            functools.partial(
+                                self._schedule_job,
+                                dispatcher=dispatcher,
+                                update=update,
+                                context=context,
+                                conversation_key=conversation_key,
+                            )
                         )
-                    )
+                    else:
+                        self._schedule_job(
+                            new_state, dispatcher, update, context, conversation_key
+                        )
                 else:
-                    self._schedule_job(new_state, dispatcher, update, context, conversation_key)
+                    self.logger.warning("`conversation_timeout` can't work without JobQueue!")
 
         if isinstance(self.map_to_parent, dict) and new_state in self.map_to_parent:
             self.update_state(self.END, conversation_key)
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -617,7 +617,7 @@ class ConversationHandler(Handler[Update]):
                                 conversation_key=conversation_key,
                             )
                         )
-                    else:
+                    elif new_state != self.END:
                         self._schedule_job(
                             new_state, dispatcher, update, context, conversation_key
                         )
--- a/telegram/ext/conversationhandler.py
+++ b/telegram/ext/conversationhandler.py
@@ -147,9 +147,9 @@ class ConversationHandler(Handler[Update]):
             :attr:`ConversationHandler.TIMEOUT`.
 
             Note:
-                * Using `conversation_timeout` with nested conversations is currently not
-                  supported. You can still try to use it, but it will likely behave differently
-                  from what you expect.
+                 Using `conversation_timeout` with nested conversations is currently not
+                 supported. You can still try to use it, but it will likely behave differently
+                 from what you expect.
 
 
         name (:obj:`str`, optional): The name for this conversationhandler. Required for
@@ -471,14 +471,16 @@ class ConversationHandler(Handler[Update]):
                 # both job_queue & conversation_timeout are checked before calling _schedule_job
                 j_queue = dispatcher.job_queue
                 self.timeout_jobs[conversation_key] = j_queue.run_once(  # type: ignore[union-attr]
-                    self._trigger_timeout,  # type: ignore[arg-type]
+                    self._trigger_timeout,
                     self.conversation_timeout,  # type: ignore[arg-type]
                     context=_ConversationTimeoutContext(
                         conversation_key, update, dispatcher, context
                     ),
                 )
             except Exception as exc:
-                self.logger.exception("Failed to add timeout job due to exception")
+                self.logger.exception(
+                    "Failed to schedule timeout job due to the following exception:"
+                )
                 self.logger.exception("%s", exc)
 
     def check_update(self, update: object) -> CheckUpdateType:  # pylint: disable=R0911
@@ -620,7 +622,9 @@ class ConversationHandler(Handler[Update]):
                             new_state, dispatcher, update, context, conversation_key
                         )
                 else:
-                    self.logger.warning("`conversation_timeout` can't work without JobQueue!")
+                    self.logger.warning(
+                        "Ignoring `conversation_timeout` because the Dispatcher has no JobQueue."
+                    )
 
         if isinstance(self.map_to_parent, dict) and new_state in self.map_to_parent:
             self.update_state(self.END, conversation_key)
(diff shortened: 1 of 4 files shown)


Bibo-Joshi (changes requested):
Besides the comment I left, I'm still missing a few things:

* We should add a note that `timeout` with `run_async` may fail, if the promise takes longer than the timeout (as described in #2417)
* We need a test that makes sure the timeout handlers are not triggered, if the promise resolves to `CH.END`. Ideally, we'd also have a test that makes sure that stuff still works if the promise raises an exception
* Timeout of child conversations needs to be detected by parent conversations somehow. If we can't get that to work, the added warning needs to be reformulated and the limitation needs to be documented. picking up my idea from off-line to have `check_update` check if the child has timed out in the meantime: that's not stright forward, as the childs state is then `None` and also the child may be used in multiple parent conversations (having the same CH instance as child in multiple parents is surely somewhat crazy but we can't prevent that either …).

  TBH while typing this I feel like it may just not be worth the effort and we might just want to document the limitation for now and revisit only if someone complains …

PS: most of the test fails are due to #2409 …

tests/test_conversationhandler.py
```diff
@@ -753,6 +753,49 @@ def test_all_update_types(self, dp, bot, user1):
         assert not handler.check_update(Update(0, pre_checkout_query=pre_checkout_query))
         assert not handler.check_update(Update(0, shipping_query=shipping_query))
 
+    def test_promise_exception(self, dp, bot, user1, caplog):
+        def conv_entry(*a, **kw):
+            return 1
+
+        def raise_error(*a, **kw):
+            raise Exception("Oh no")
```
Bibo-Joshi, line 761: Please try to use some more descriptive strings ;)

tests/test_conversationhandler.py
```diff
@@ -1126,6 +1219,29 @@ def slowbrew(_bot, update):
...
+                            self.BREWING: [CommandHandler('startCoding', self.code)],
+                        },
+                        fallbacks=self.fallbacks,
+                    )
+                ]
+            },
+            fallbacks=self.fallbacks,
+            conversation_timeout=100,
+        )
+        assert len(recwarn) == 1
```
Bibo-Joshi, line 1239: This test doesn't test too much, if there is only 1 child conversaton in the first place ;)

telegram/ext/conversationhandler.py
```diff
@@ -143,6 +143,13 @@ class ConversationHandler(Handler[Update]):
             received update and the corresponding ``context`` will be handled by ALL the handler's
             who's :attr:`check_update` method returns :obj:`True` that are in the state
             :attr:`ConversationHandler.TIMEOUT`.
+
+            Note:
+                * `conversation_timeout` with `run_async` may fail, if the :obj:`Promise`
+                  takes longer to finish.
```
Bibo-Joshi, line 149: * :class:`telegram.ext.utils.promise.Promise`
* … longer to finish than than the timeout.

Bibo-Joshi (changes requested):
My third bullet point remains to be resolved. Do you have a preference on that?
Test failures: Seems like TG doesn't like our test bots anymore … I had problems with the fallback bots on local tests in the last time.

starry69:
> My third bullet point remains to be resolved. Do you have a preference on that?

Imo we should leave it for now as you said, and just add a note in documentation that: *using timeouts with nested conversation may cause troubles.* TBH I've never seen anyone complaining about timeouts in nested conversation triggering incorrectly, even when it was bugged, which tells that users rarely use timeouts in nested conversation, besides, now we also have a warning in place so ig we should be fine...

tests/test_conversationhandler.py
starry69: Oops you're right, gonna fix that

tests/test_conversationhandler.py
starry69: Okay 😅

telegram/ext/conversationhandler.py
starry69: 👍

Bibo-Joshi:
All right. I'd just to reformulate the note & warning in this case, to be more explicit. E.g. something like

> Using `conversation_timeout` with nested conversations is currently not supported. You can still try to use it, but it will likely behave differntly from what you expect.

for both

Bibo-Joshi (approved):
Thanks! I'll wait with merging, though, let's first see if we can get the tests to work again …

starry69:
> Thanks! I'll wait with merging, though, let's first see if we can get the tests to work again …

Yep sure 👍

telegram/ext/conversationhandler.py
```diff
@@ -423,6 +443,20 @@ def _get_key(self, update: Update) -> Tuple[int, ...]:
 
         return tuple(key)
 
+    def _resovle_promomise(self, state: Tuple) -> object:
```
Poolitzer, line 446: resovle promomise, hihi
```suggestion
    def _resolve_promise(self, state: Tuple) -> object:
```

telegram/ext/conversationhandler.py
```diff
@@ -455,21 +489,14 @@ def check_update(self, update: object) -> CheckUpdateType:  # pylint: disable=R0
...
-                    res = old_state
-                finally:
-                    if res is None and old_state is None:
-                        res = self.END
-                    self.update_state(res, key)
-                    with self._conversations_lock:
-                        state = self.conversations.get(key)
+            # check if promise is finished or not
+            if state[1].done.wait(0):
+                res = self._resovle_promomise(state)
```
Poolitzer, line 494:
```suggestion
                res = self._resolve_promise(state)
```

telegram/ext/conversationhandler.py
```diff
@@ -608,10 +635,21 @@ def _trigger_timeout(self, context: _ConversationTimeoutContext, job: 'Job' = No
...
+            and isinstance(state[1], Promise)
+            and self._resovle_promomise(state) == self.END
+        ):
+            return
+
         with self._timeout_jobs_lock:
             found_job = self.timeout_jobs[context.conversation_key]
             if found_job is not job:
-                # The timeout has been canceled in handle_update
+                # The timeout has been cancelled in handle_update
```
Poolitzer, line 652: so he can speak english ;P

Poolitzer (changes requested):
code changes look good to me

telegram/ext/conversationhandler.py
starry69: Oh shit!

Poolitzer (approved):
GJ!

Bibo-Joshi:
tests now look good, but coverage is not happy. @starry69 can you check why line 647 in CH is not covered?

starry69:
> tests now look good, but coverage is not happy. @starry69 can you check why line 647 in CH is not covered?

Fixed

telegram/ext/conversationhandler.py
```diff
@@ -143,6 +144,16 @@ class ConversationHandler(Handler[Update]):
             received update and the corresponding ``context`` will be handled by ALL the handler's
             who's :attr:`check_update` method returns :obj:`True` that are in the state
             :attr:`ConversationHandler.TIMEOUT`.
+
+            Note:
+                * `conversation_timeout` with `run_async` may fail, if the
```
Bibo-Joshi, line 149: THis is now no longer true :)

telegram/ext/conversationhandler.py
```diff
@@ -423,6 +444,37 @@ def _get_key(self, update: Update) -> Tuple[int, ...]:
...
+        new_state: object,
+        dispatcher: 'Dispatcher',
+        update: Update,
+        context: Optional[CallbackContext],
+        conversation_key: Tuple[int, ...],
+    ) -> None:
+        if new_state != self.END:
+            self.timeout_jobs[
+                conversation_key
+            ] = dispatcher.job_queue.run_once(  # type:ignore[union-attr]
```
Bibo-Joshi, line 472: just curiosity: Why are those type ignore comments necessary?

telegram/ext/conversationhandler.py
```diff
@@ -455,21 +507,14 @@ def check_update(self, update: object) -> CheckUpdateType:  # pylint: disable=R0
...
-                    self.logger.exception("Promise function raised exception")
-                    self.logger.exception("%s", exc)
-                    res = old_state
-                finally:
-                    if res is None and old_state is None:
-                        res = self.END
-                    self.update_state(res, key)
-                    with self._conversations_lock:
-                        state = self.conversations.get(key)
+            # check if promise is finished or not
```
Bibo-Joshi, line 510: IISC we don't strictly need to have that logic in its own method anymore, but I'm fine with leaving it that way.

telegram/ext/utils/promise.py
```diff
@@ -106,6 +109,18 @@ def result(self, timeout: float = None) -> Optional[RT]:
             raise self._exception  # pylint: disable=raising-bad-type
         return self._result
 
+    def add_done_callback(self, callback: Callable) -> None:
+        """
+        Callback to be run when :class:`telegram.ext.utils.promise.Promise` becomes done.
```
Bibo-Joshi, line 114: Please document that the callback will be given `self.result()` as only positional arg

telegram/ext/utils/promise.py
```diff
@@ -83,6 +84,8 @@ def run(self) -> None:
 
         finally:
             self.done.set()
+            if self._done_callback:
```
Bibo-Joshi, line 87: We should do some `try-except` here for completeness. IISC we can only reasonably `pass` on exceptions, but it's better to make sure that ony exceptions here don't break anything …

telegram/ext/conversationhandler.py
Bibo-Joshi, line 472: We should add some try-except here and in the except clause either call `dispatcher.dispatch_error` or a least log something to warning or debug level about the timeout job not being set. This is relevant when the `done_callback` is actually run only in `Promise.run`, where we have no means of directly dispatching the error.

This case should also be tested, probably my mocking `run_once` to raise an excption.

tests/test_conversationhandler.py
```diff
@@ -753,6 +753,49 @@ def test_all_update_types(self, dp, bot, user1):
         assert not handler.check_update(Update(0, pre_checkout_query=pre_checkout_query))
         assert not handler.check_update(Update(0, shipping_query=shipping_query))
 
+    def test_promise_exception(self, dp, bot, user1, caplog):
```
Bibo-Joshi, line 756:
```suggestion
    def test_promise_exception(self, dp, bot, user1, caplog):
        """Here we make sure that when a run_async handle raises an exception, the state isn't changed"""
```

tests/test_promise.py
```diff
@@ -63,3 +63,16 @@ def callback():
 
         with pytest.raises(TelegramError, match='Error'):
             promise.result()
+
+    def test_done_callback(self):
```
Bibo-Joshi, line 67: Please also test adding the callback before calling `run()` and also both cases with `done_callback` raising an exception.

Bibo-Joshi (changes requested):
Thanks for re-working :)
Overall looks good, I left some remarks

telegram/ext/conversationhandler.py
starry69: Ah right

telegram/ext/conversationhandler.py
starry69: Mypy was complaining about `Optional[JobQueue]` not having `run_once` attr and invalid type of `self.conversation_timeout`..
IMG_20210321_171842_329

telegram/ext/conversationhandler.py
starry69: Imo it looks more readable that way, `check_update` is already a very big method.

telegram/ext/conversationhandler.py
Bibo-Joshi: Ah, I see - The checks for `conversation_timeout` and `dispatcher.job_queue` happen outside of the function. Still, I think we can improve a bit:

* `if disptacher.job_queue is None` in `handle_update` (line 599), make a log on warning level that timeouts can't work without jobqueue
* Correct the annotation for the `conversation_timeout` argument of `__init__` to `Union[float, timedelta]` and for the property to `Optional[Union[float, timedelta]]`
* Add a comment that both `dispatcher.job_queue` and `conversation_timeout` are checked before calling `_schedule_timeout`, so future-us don't worry about it :)

telegram/ext/conversationhandler.py
```diff
@@ -467,13 +467,19 @@ def _schedule_job(
...
+                j_queue = dispatcher.job_queue
+                self.timeout_jobs[conversation_key] = j_queue.run_once(  # type: ignore[union-attr]
+                    self._trigger_timeout,  # type: ignore[arg-type]
+                    self.conversation_timeout,  # type: ignore[arg-type]
+                    context=_ConversationTimeoutContext(
+                        conversation_key, update, dispatcher, context
+                    ),
+                )
+            except Exception as exc:
+                self.logger.exception("Failed to add timeout job due to exception")
```
Bibo-Joshi, line 481:
```suggestion
                self.logger.exception("Failed to schedule timeout job due to the following exception:")
```

telegram/ext/conversationhandler.py
```diff
@@ -467,13 +467,19 @@ def _schedule_job(
...
-            ] = dispatcher.job_queue.run_once(  # type:ignore[union-attr]
-                self._trigger_timeout,  # type: ignore[arg-type]
-                self.conversation_timeout,  # type: ignore[arg-type]
-                context=_ConversationTimeoutContext(conversation_key, update, dispatcher, context),
-            )
+            try:
+                # both job_queue & conversation_timeout are checked before calling _schedule_job
+                j_queue = dispatcher.job_queue
+                self.timeout_jobs[conversation_key] = j_queue.run_once(  # type: ignore[union-attr]
+                    self._trigger_timeout,  # type: ignore[arg-type]
```
Bibo-Joshi, line 474: I think the signature of `_trigger_timeout` is acutally wrong, the argument should be of type `CallbackContext` and this comment can be removed

telegram/ext/conversationhandler.py
```diff
@@ -596,20 +602,25 @@ def handle_update(  # type: ignore[override]
...
+                            )
+                        )
+                    else:
+                        self._schedule_job(
+                            new_state, dispatcher, update, context, conversation_key
                         )
-                    )
                 else:
-                    self._schedule_job(new_state, dispatcher, update, context, conversation_key)
+                    self.logger.warning("`conversation_timeout` can't work without JobQueue!")
```
Bibo-Joshi, line 623:
```suggestion
                    self.logger.warning("Ignoring `conversation_timeout` because the Dispatcher has no JobQueue.")
```

Let's try to be precise both for the users and for future-us 😉

Bibo-Joshi (commented):
Had a look at the updates and left some minor comments. Codecov complains that the added `except` in `Promise` isn't hit …

telegram/ext/conversationhandler.py
starry69: It raises error
```error: "CallbackContext" has no attribute "conversation_key"  [attr-defined]```

Comment (Bibo-Joshi) [reply telegram/ext/conversationhandler.py:474]: mh, that's probably because of the unfortunate naming in here:

https://github.com/python-telegram-bot/python-telegram-bot/blob/ccedd3a87d5d7c4ea2a651eba4e9a8bcfba5faba/telegram/ext/conversationhandler.py#L608

Changing that e.g. `cntxt = … ` should probably do the trick. You can tell that in #1920 I didn't put too much effort in getting everything perfect on the first try xD

**Review by SwayamSodha — APPROVED** [review]
Brilent  Code it's wonderful.

File: telegram/ext/conversationhandler.py [hunk 17]
```diff
@@ -551,15 +604,27 @@ def handle_update(  # type: ignore[override]
...
+                        new_state.add_done_callback(
+                            functools.partial(
+                                self._schedule_job,
+                                dispatcher=dispatcher,
+                                update=update,
+                                context=context,
+                                conversation_key=conversation_key,
+                            )
+                        )
+                    else:
```
Bibo-Joshi, line 620:
```suggestion
                    elif new_state != self.END:
```

if we already know that, we don't have to schedule a job for that … less code executed :)

telegram/ext/conversationhandler.py
```diff
@@ -423,6 +444,45 @@ def _get_key(self, update: Update) -> Tuple[int, ...]:
 
         return tuple(key)
 
+    def _resolve_promise(self, state: Tuple) -> object:
```
Bibo-Joshi, line 447:
```suggestion
    def _resolve_promise(self, state: Tuple[object, Promise]) -> object:
```

telegram/ext/utils/promise.py
```diff
@@ -83,6 +84,11 @@ def run(self) -> None:
...
+                try:
+                    self._done_callback(self.result())
+                except Exception:
+                    pass
```
Bibo-Joshi, line 91: When re-reading this, Noams force-ghost appeared and scolted me :D let's put a log here, something like
```suggestion
                    logger.warning('`done_callback` of a Promise raised the following exception. The exception won't be handled by error handlers.')
                    logger.warning(<traceback>)
```
that should also make codacy happy

Bibo-Joshi (changes requested):
Thanks for your patience @starry69 ! Almost done, just some minor nit picking left :)

telegram/ext/utils/promise.py
starry69: Lol, yea makes sense

telegram/ext/conversationhandler.py
starry69: I did same at first but mypy started raising:
```python
telegram/ext/conversationhandler.py:520: error: Argument 1 to "_resolve_promise" of "ConversationHandler" has incompatible type "Tuple[Any, ...]"; expected "Tuple[object, Promise]"  [arg-type]
```
Even after having `isinstance(state, tuple) and isinstance(state[1], Promise)` check in `handle_update` method, since `_resolve_promise()` was private method i just added `Tuple` after that.

Bibo-Joshi: ah, because `self.conversations` is typed to be `Dict[Tuple[int, …], Optional[object]]` . Well, then let's just leave it this way. It's not important.

Bibo-Joshi (approved):
I'm happy now :) @Poolitzer do you want to review again?

Poolitzer (approved):
Good QoL change!

Issue #2347, opened 2021-01-30
Deprecation: v15


Bibo-Joshi:
Things we should drop in v15. The list is probably not complete. This includes everything deprecated in v13.x.

- [ ] The `@run_async` decorator
- [ ] `Filters.{private, group}`
- [ ] `utils.{promise, webhookhandler}`
- [ ] `clean` argument of `Updater.start_{polling, webhook}`
- [ ] Think about removing `Defaults` from `tg.Bot` completey and moving them to `tg.ext.ExtBot`. We may have to think about how to handle signatures with `DefaultVaule`, then, i.e. if `_insert_defaults` should stay in `tg.Bot` or should be moved to `tg.ext.ExtBot`.

Poolitzer:
Why did you add this to the V14 stage?

Bibo-Joshi:
because we decided to rip the band aid off … We also add the items to #2346 and delete this one, I was just too lazy for that

Bibo-Joshi:
Closing in favor of #2346

PR #2419, opened 2021-03-06
Improve set_webhook

### LINKED ISSUE #2416: Check if we can be smarter about start_webhook
More precisely, check if we can call `set_webhook` automatically in more cases. Things to consider:

1. As of API 5.0 we have the `drop_pending_updates` parameter, so we can handle the `clean` logic more elegantly
2. SSL-settings can be applicable at two different "positions": the URL where TG sends the updates to and the webhookserver, where we fetch them. those two are not necessarily the same (e.g. reverse proxy)
3. citing jannes: "I think we could use the cert to set the webhook if available, but only use it for SSL termination if there is also a key provided" (which needs to be double checked)

Bibo-Joshi:
When ready closes #2416

Already tested 

* polling
* nginx reverse proxy webhooks with letsencrypt (not passing `cert` to TG) & with self-signed certificate (passing `cert` to TG)

To be done:

- [x] test having the built-in Webserver handle the SSL stuff. IISC I can't test that on my lolcal setup, though …
- [x] double check if docs are updated enough
- [x] unit tests
- [x] add dreprecation of `clean` to #2347 

This also

* deprecated the `clean` argument in favor of `drop_pending_updates` as a) that conforms with the TG nomenclature, b) that name is more speaking
* adds the `ip_address` parameter of `set_webhook` to `start_webhook` (tbh I don't really understand what it's good for, do we need some special casing for that?)

When ready:

- [ ] Update the wiki page on webhooks according to the changes

### Checklist for PRs

- [x] Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing changes (I did that but leaving it as todo here so we double check the version number before merging)

### DIFF
--- a/telegram/ext/updater.py
+++ b/telegram/ext/updater.py
@@ -255,12 +255,14 @@ class Updater:
         Args:
             poll_interval (:obj:`float`, optional): Time to wait between polling updates from
                 Telegram in seconds. Default is 0.0.
-            timeout (:obj:`float`, optional): Passed to :attr:`telegram.Bot.get_updates`.
+            timeout (:obj:`float`, optional): Passed to :meth:`telegram.Bot.get_updates`.
             drop_pending_updates (:obj:`bool`, optional): Whether to clean any pending updates on
                 Telegram servers before actually starting to poll. Default is :obj:`False`.
+
+                .. versionadded :: 13.4
             clean (:obj:`bool`, optional): Alias for ``drop_pending_updates``.
 
-                .. deprecated:: v13.4
+                .. deprecated:: 13.4
                     Use ``drop_pending_updates`` instead.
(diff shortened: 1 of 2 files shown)


Poolitzer:
> test having the built-in Webserver handle the SSL stuff. IISC I can't test that on my lolcal setup, though …

I can, I did, I can confirm it worked.

Bibo-Joshi:
On the last run, Codecov didn't show results for some reason, but the PR looks okay in the [dashboard](https://app.codecov.io/gh/python-telegram-bot/python-telegram-bot/compare/2419/diff).

jh0ker (approved):
LGTM

PR #2427, opened 2021-03-12
Set logging on vendored lib instead of upstream

Bibo-Joshi:
IISC that setting was just not changed when we started vendoring urllib3. see [this thread](https://t.me/pythontelegrambotgroup/446393) for reference.


Bibo-Joshi:
> OTOH, if someone is using a library we don't vendor maybe it's not our place to change the log level for that library.

I agree with, especially as this PR was triggered by someone actually having problems with us changing the log-level for the non-vendor version

Tests fail due to API update and the usual timeouts.

Merging.

PR #2441, opened 2021-03-23
Fix & improve some examples

### LINKED ISSUE #2440: [BUG] Broken button '4' in inlinekeyboard2.py example
When the inline button '4' in the inlinekeyboard2.py example is clicked, it sends the same button as an option in the keyboard menu. Thus, when the user clicks it again, Telegram raises an error saying new message content and reply markup is same. This can confuse a user.

### Steps to reproduce
1. Run inlinekeyboard2.py, click `/start`.

2. Click button '1'.

3. Click button '4'.

4.  Click button '4' again.
5.  Error raised from Telegram.

Potential fix:
Change which button is offered in the keyboard menu.

harshil21:
- closes #2440 
- some typo fixes
- Handle empty inline queries without raising errors.
- remove unnecessary if condition in passportbot.py
- use `ReplyKeyboardRemove()` when conversation is over.
- use `ForceReply` in an example.

### DIFF
--- a/examples/persistentconversationbot.py
+++ b/examples/persistentconversationbot.py
@@ -17,7 +17,7 @@ bot.
 import logging
 from typing import Dict
 
-from telegram import ReplyKeyboardMarkup, Update, ReplyKeyboardRemove, ForceReply
+from telegram import ReplyKeyboardMarkup, Update, ReplyKeyboardRemove
 from telegram.ext import (
     Updater,
     CommandHandler,
@@ -47,7 +47,7 @@ markup = ReplyKeyboardMarkup(reply_keyboard, one_time_keyboard=True)
 
 
 def facts_to_str(user_data: Dict[str, str]) -> str:
-    facts = list()
+    facts = []
 
     for key, value in user_data.items():
         facts.append(f'{key} - {value}')
@@ -81,15 +81,14 @@ def regular_choice(update: Update, context: CallbackContext) -> int:
         )
     else:
         reply_text = f'Your {text}? Yes, I would love to hear about that!'
-    update.message.reply_text(reply_text, reply_markup=ForceReply(selective=True))
+    update.message.reply_text(reply_text)
 
     return TYPING_REPLY
 
 
 def custom_choice(update: Update, _: CallbackContext) -> int:
     update.message.reply_text(
-        'Alright, please send me the category first, for example "Most impressive skill"',
-        reply_markup=ForceReply(selective=True),  # force a reply from the user
+        'Alright, please send me the category first, for example "Most impressive skill"'
     )
 
     return TYPING_CHOICE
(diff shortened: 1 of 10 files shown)


examples/persistentconversationbot.py
```diff
@@ -77,18 +77,19 @@ def regular_choice(update: Update, context: CallbackContext) -> int:
     context.user_data['choice'] = text
     if context.user_data.get(text):
         reply_text = (
-            f'Your {text}, I already know the following about that: {context.user_data[text]}'
+            f'Your {text}? I already know the following about that: {context.user_data[text]}'
         )
     else:
         reply_text = f'Your {text}? Yes, I would love to hear about that!'
-    update.message.reply_text(reply_text)
+    update.message.reply_text(reply_text, reply_markup=ForceReply(selective=True))
```
Bibo-Joshi, line 84: Mh, if people ask for examples on how to use `ForceReply` I feel like it's not a good idea to have them run a conversationbot … Could you mind moving this to `echobot.py`? You can have the `start` message mention the user, which should work with `selective=True`

examples/persistentconversationbot.py
```diff
@@ -77,18 +77,19 @@ def regular_choice(update: Update, context: CallbackContext) -> int:
...
 
     return TYPING_REPLY
 
 
 def custom_choice(update: Update, _: CallbackContext) -> int:
     update.message.reply_text(
-        'Alright, please send me the category first, ' 'for example "Most impressive skill"'
+        'Alright, please send me the category first, for example "Most impressive skill"',
```
Bibo-Joshi, line 91: same

Bibo-Joshi (changes requested):
Thanks for the improvements! Left a comment. Also you mentioned:

> * Handle empty inline queries without raising errors.

but I don't see that? If you want to add that, maybe an idea would be to use `switch_pm_text` to display a help message similarly to what rulesbot does. That feature is also little known.

harshil21:
> but I don't see that? 

Oops, I guess I discarded those changes by mistake. It was `if query == ""; return`. The exception raised when it's empty is 
`Can't parse inline query result: input message content is not specified`.

> maybe an idea would be to use `switch_pm_text` to display a help message similarly to what rulesbot does. 

So should I also put in `switch_pm_{text, parameter}`, along with the fix above?

examples/persistentconversationbot.py
harshil21: Alright. I included this here cause it felt the like the most natural place to put it.

Bibo-Joshi:
> > but I don't see that?
> 
> Oops, I guess I discarded those changes by mistake. It was `if query == ""; return`. The exception raised when it's empty is
> `Can't parse inline query result: input message content is not specified`.

Jup, makes sense :)

> > maybe an idea would be to use `switch_pm_text` to display a help message similarly to what rulesbot does.
> 
> So should I also put in `switch_pm_{text, parameter}`, along with the fix above?

If you like to. Not important, just an idea ;)

PR #2446, opened 2021-03-25
New Filters.forwarded_from filter

### LINKED ISSUE #2411: [FEATURE] new Filters.forwarded_from filter
This filter should make usages of `forward_from` and `forward_from_chat` in order to filter out messages which get forwarded by a certain chat.

This should inherit from `_ChatUserBaseFilter`, `Filters.user` or `Filters.sender_chat` are example implementations of this.

You can either create a new Filter like `Filters.forwarded_by` or you can enhance `Filters.forwarded` (behaviour could be copied from `Filters.chat_type`). I would prefer the second option in order to reduce the overall numbers of filters, other maintainers?

harshil21:
Closes #2411

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -1735,12 +1735,20 @@ officedocument.wordprocessingml.document")``.
     class forwarded_from(_ChatUserBaseFilter):
         # pylint: disable=W0235
         """Filters messages to allow only those which are forwarded from the specified chat ID(s)
-        or username(s).
+        or username(s) using :attr:`telegram.Message.forwarded_from` and
+        :attr:`telegram.Message.forwarded_from_chat`.
 
         .. versionadded:: 13.5
 
         Examples:
             ``MessageHandler(Filters.forwarded_from(chat_id=1234), callback_method)``
+
+        Note:
+            When a user has disallowed adding a link to their account while forwarding their
+            messages, this filter will *not* work since both
+            :attr:`telegram.Message.forwarded_from` and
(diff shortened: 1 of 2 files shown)


Bibo-Joshi (commented):
Thanks for the PR!
Code looks good, but I'm unsure about what we filter: We have `forward_from` and `forward_from_chat` as attributes. Can we rule out that both are non-`None` at the same time? If not, which should we use? Even if we can rule it out, maybe it's just cleaner to provide both `Filters.forward_{from, from_chat}`. In any case, the docstring should explicitly mention, which attribute of the `message` is checked.

harshil21:
> Can we rule out that both are non-`None` at the same time?

Yeah, only one of `.forward_from` or `.forward_from_chat` is `None` for a forwarded message, except when the account is hidden by the user (privacy setting), then both those attributes are `None`. So the filter would fail in that case. We should mention that too in the docs.


> maybe it's just cleaner to provide both `Filters.forward_{from, from_chat}`.

They do have the same function, so as I user I would not like the hassle of switching between the two.

> In any case, the docstring should explicitly mention, which attribute of the `message` is checked.

Agreed.

Bibo-Joshi:
> > Can we rule out that both are non-`None` at the same time?
> 
> Yeah, only one of `.forward_from` or `.forward_from_chat` is `None` for a forwarded message, except when the account is hidden by the user (privacy setting), then both those attributes are `None`. So the filter would fail in that case. We should mention that too in the docs.

but that not documented, right? So behavior could change … Unlike, for sure.

PR #2450, opened 2021-03-30
chore: refactor code quality issues

akshgpt7:
## Description

Hey :wave:, I'm a Developer Outreach at [DeepSource](https://deepsource.io/) and ran DeepSource analysis on my fork of the repo. It found some interesting [code quality improvements](https://deepsource.io/gh/akshgpt7/python-telegram-bot/) to consider.

This PR fixes a few of the issues detected for you to assess if it is right for you.
Happy to provide the tweaks separately otherwise :)

## Summary of changes

- Use literal syntax instead of function calls to create data structures.
- Refactor unnecessary use of comprehension.
- Remove redundant `None` default.
- Remove assert statement from non-test files.
- Remove unused imports.
- Add `.deepsource.toml` file for continuous analysis on bug risks/performance/code-quality issues on new changes.


examples/echobot.py
```diff
@@ -34,7 +34,7 @@ def start(update: Update, _: CallbackContext) -> None:
     """Send a message when the command /start is issued."""
     user = update.effective_user
     update.message.reply_markdown_v2(
-        f'Hi {user.mention_markdown_v2()}\!',
+        f'Hi {user.mention_markdown_v2()}!',
```
Bibo-Joshi, line 37: the `!` needs to be escaped for MarkdownV2. Please instead make it a raw string. Apparently I missed that in the latest commit to master …

telegram/bot.py
```diff
@@ -3457,7 +3457,8 @@ def answer_shipping_query(  # pylint: disable=C0103
         data: JSONDict = {'shipping_query_id': shipping_query_id, 'ok': ok}
 
         if ok:
-            assert shipping_options
+            if not shipping_options:
+                raise AssertionError
```
Bibo-Joshi, line 3461: What is the benefit of this except for maybe using `assert` in non-test files being considered bad practice? It basically does the same thing, but uses more words.

Bibo-Joshi (changes requested):
Hey, thanks for the PR. Always appreciated.

Regarding DeepSource: We currently use Codacy for code quality analysis and IIRC we could even find more issues with that, if we'd tighten the settings. Why should we choose to switch to DeepSource rather than sticking with Codacy or switching to other services like e.g. DeepCode.ai? And I mean apart from you getting paid by them? 😜

telegram/bot.py
akshgpt7: Python has an option to compile the optimized bytecode and create the respective .pyo files by using the options `-O` and `-OO`.

During this, these basic optimizations are done:
- All the assert statements are removed
- All docstrings are removed (when `-OO` is selected)
- Value of the debug built-in variable is set to False

Now, if someone never intends to run the optimized bytecode, they can ignore this issue. But, it is still recommended not to have it in non-test files.
A better way for internal self-checks is to explicitly check and raise respective Error using an if statement.

Bibo-Joshi: Fair enough. Please leave a short comment for that, though, along the lines "We don't user assert directly, as that's dropped in optimized bytecode". As a side note, please try to not force-push as that makes it hard to keep track of the PR ;)

Bibo-Joshi:
I just merged #2451 to master, which fixes the failing tests, so please merge from master.

akshgpt7:
> Why should we choose to switch to DeepSource rather than sticking with Codacy or switching to other services like e.g. DeepCode.ai?

@Bibo-Joshi There are various added benefits that DeepSource offers:
- Automatic code formatting to comply with the code style (through Transformers).
- Option to Autofix detected issues through one click.
- DeepSource Python analyzer detects [520 issues](https://deepsource.io/directory/analyzer/python/) as of now (new support being continuously added).
- Less than 5% false positive reports in DeepSource.
- Option to report false positives directly.

These are of course some of the benefits, and just for you to assess if the tool is right for the project :)
If you have any questions, feel free to reach out or have a look at the [docs](https://deepsource.io/docs).

Bibo-Joshi:
Thanks for the info! We'll have to discuss it in the dev team, but I'd like to merge your PR first. Would you mind removing the deepsource config file?

akshgpt7:
No problem, I've removed the config file @Bibo-Joshi.
If you want, I can open an issue for the same, but it's fine if you want to discuss it internally too :)

Bibo-Joshi:
Thanks for the contribution @akshgpt7 ! We'll discuss switching in the dev team - your example config file is already a great starting point :)

PR #2464, opened 2021-04-13
Add warnings

starry69:


telegram/ext/dispatcher.py
```diff
@@ -162,6 +162,11 @@ def __init__(
                 stacklevel=3,
             )
 
+        if self.workers < 1:
+            self.logger.warning(
+                'You must need to have atleast one worker for `run_async` to work properly.'
```
Bibo-Joshi, line 167: Maybe rather something like "Asynchronous callbacks can not be processed without at least one worker thread"?

starry69: Yea that's better

telegram/ext/dispatcher.py
```diff
@@ -162,6 +162,11 @@ def __init__(
                 stacklevel=3,
             )
 
+        if self.workers < 1:
+            self.logger.warning(
```
Bibo-Joshi, line 166: On a second look, this cought my eye. In other places, we use `warnings.warn` and I have the impression, that we're not very consistent on when we use what. I think I'll just open an issue later today to discuss this in general for v14, but giving it a second thought here already might also not hurt …

starry69: I guess we should use `logger.warning` for runtime warnings like [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/9eba935b113524fea330f7fb17fcfd8db660a3c0/telegram/ext/conversationhandler.py#L481) and `warnings.warn` while initialising some class. Don't have any specific reason to so just a thought..

starry69: Or we could use `logger.warning` everywhere except those places where we need to use `TelegramDepriciatiobWarning`.

Bibo-Joshi: Did a bit of digging, see #2467. From that I'd say to use `warnings.warn` here …

Bibo-Joshi:
BTW, #2466 gave me the idea that we could do a warning when a `ConversationHandler` handler returns a state that is unknown, b/c in that case the conversation will break down. Maybe this could fit into this PR - we'd be making it a "add warings" pr :D

Here IG that I'd also go with `warnings.warn`, b/c the user can very well make sure that they only return known states. This could be added in `update_state`.

telegram/ext/dispatcher.py
starry69: Alright 👍

harshil21:
> do a warning when a `ConversationHandler` handler returns a state that is unknown, b/c in that case the conversation will break down. 

+1, will be easier to debug CH!

Bibo-Joshi:
Aaaand another warning we could add: if in CH `allow_reentry` is `True` and the CH has nested conversations, those should also have `allow_reentry=True`. See https://t.me/pythontelegrambotgroup/459818. This also deserves a comment in the docs. This should be double checked though, if that actually works

starry69:
> BTW, #2466 gave me the idea that we could do a warning when a `ConversationHandler` handler returns a state that is unknown, b/c in that case the conversation will break down.

When returning state that is unknown to conversation handler, the conversation gets stucked as long as user cancels it manually (via fallback) or until timeout gets triggered, shouldn't we also close the conversation after warning like: "Conversation handler returned unknown state, closing the current conversation..." Instead of just warning

Bibo-Joshi:
> When returning state that is unknown to conversation handler, the conversation gets stucked as long as user cancels it manually (via fallback) or until timeout gets triggered, shouldn't we also close the conversation after warning like: "Conversation handler returned unknown state, closing the current conversation..." Instead of just warning

Mh, I don't think that's a good idea. Firstly, there's a chance we might close falsely in some edge cases. Also I think that it's rather surprising behaviour for CH to close conversation on its own and makes debugging even harder. And finally it would be somewhat breaking…

starry69:
> > When returning state that is unknown to conversation handler, the conversation gets stucked as long as user cancels it manually (via fallback) or until timeout gets triggered, shouldn't we also close the conversation after warning like: "Conversation handler returned unknown state, closing the current conversation..." Instead of just warning
> 
> Mh, I don't think that's a good idea. Firstly, there's a chance we might close falsely in some edge cases. Also I think that it's rather surprising behaviour for CH to close conversation on its own and makes debugging even harder. And finally it would be somewhat breaking…

Oh right, makes sense.

telegram/ext/conversationhandler.py
```diff
@@ -592,6 +592,10 @@ def update_state(self, new_state: object, key: Tuple[int, ...]) -> None:
                     )
 
         elif new_state is not None:
+            if new_state not in self.states:
+                warnings.warn(
+                    "Handler returned state which is unknown to the ConversationHandler."
```
Bibo-Joshi, line 597: Just an idea: If the hanlder has a name, we could include that in the message. Makes debugging easier :) And maybe including the unknown state is useful, too …

telegram/ext/conversationhandler.py
```diff
@@ -594,7 +594,8 @@ def update_state(self, new_state: object, key: Tuple[int, ...]) -> None:
         elif new_state is not None:
             if new_state not in self.states:
                 warnings.warn(
-                    "Handler returned state which is unknown to the ConversationHandler."
+                    f"Handler returned state {new_state} which is unknown to the "
+                    f"ConversationHandler {self.name if self.name is not None else ''}."
```
Bibo-Joshi, line 598: Just one very minor nitpick left: when the CH has no name, there will be a trailing whitespace. We can do better, even in warnings :D 

After this, I promise I'll merge! 😉

Bibo-Joshi (approved):
Thanks for your patience!

Bibo-Joshi:
Wen can ignore deepsource until I'm done with #2454. Merging

PR #2480, opened 2021-04-21
Improve tests

harshil21:
Removes the pytest-timeout plugin, since we have have [patient_request_wrapper](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/tests/bots.py#L79), which already handles timeout errors. Also stops tests from failing unnecessarily!

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,6 +21,7 @@ import functools
 import inspect
 import os
 import re
+import threading
 from collections import defaultdict
 from queue import Queue
 from threading import Thread, Event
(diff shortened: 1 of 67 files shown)


Bibo-Joshi (commented):
Cool stuff! 🦾 

about flaky: tests that don't make a requests to TG shouldn't be flaky. the idea behind flaky is to make non-deterministic tests more stable. I can agree to the conversation timeout thingy, because timing is not always really deterministic but IISC the other two tests shouldn't be flaky and I don't recall them being unreliable for me …

tests/conftest.py
```diff
@@ -339,6 +340,46 @@ def timezone(tzinfo):
     return tzinfo
 
 
+timed_out = False
+
```
Bibo-Joshi, line 344: Can you leave a comment somewhere around here explaining why we do this "manually" instead of using `pytest-timeout`?

tests/conftest.py
```diff
@@ -339,6 +340,46 @@ def timezone(tzinfo):
...
+        timeout_setup(item, max_time)  # Set and start timer
+    yield  # This executes the test function and later the code after this is run.
+    if marker:
+        timeout_teardown(item)  # Cancel timer if timeout didn't exceed.
+        if timed_out:  # xfail test if exceeded!
+            pytest.xfail(reason=f"Timeout of {max_time}s exceeded.")
+
+
+def timeout_setup(item, max_time):
+    global timed_out
```
Bibo-Joshi, line 361: when I see `global` I'm always wondering if it can be done differently 😬

Issue #2478, opened 2021-04-20
"Edit on GitHub" links in docs are misleading


jhsoby:
### Steps to reproduce
1. Go to a page in the docs, e.g. this one: https://python-telegram-bot.readthedocs.io/en/stable/telegram.ext.filters.html?highlight=caption#telegram.ext.filters.Filters.caption_regex

2. Click "Edit on GitHub" at the top right.

3. Youˈll end up at https://github.com/python-telegram-bot/python-telegram-bot/blob/ccedd3a87d5d7c4ea2a651eba4e9a8bcfba5faba/docs/source/telegram.ext.filters.rst

### Expected behaviour
You will end up somewhere where you can edit the actual text (e.g. if you found a spelling mistake).

### Actual behaviour
You end up on a template page (?) that apparently fetches the actual page content from somewhere else (where exactly is not obvious).

### Configuration
**Operating System:** Irrelevant

**Version of Python, python-telegram-bot & dependencies:** Irrelevant

Poolitzer:
thanks for reporting, I will look into it, it is sadly not trivial

Bibo-Joshi:
Hi.

We use the software Sphinx to build the docs and it's common practice to write the docstrings directly into the code rather than into separate files. One of the main selling points of Sphinx is that it can automatically extract the docrstings, which is done in the template you saw: 

```
.. automodule:: telegram.ext.filters
```

this tells sphinx to document everything in the module `telegram.ext.filters` - which also tells you, where the actual documentation is ;)

Note that depending on the page, the file linked by RTD may well contain the actual info, such as for the start page https://python-telegram-bot.readthedocs.io/en/stable/ - because that is manually typed. Moreover, one page on RTD may contain docstrings from multiple python files, so linking directly to the source code is not exactly trivial.

A quick google search tells me that there is no easy flag to tell RTD to hide the "edit on github" button.

IMHO we should just leave it as is. If someone is determined to fix a typo, they are very welcome to report it in this bug tracker, if they can't figure out where it needs to be fixed :)

Poolitzer:
> Moreover, one page on RTD may contain docstrings from multiple python files

Do we have this currently in our project?

jhsoby:
I donˈt know how the RST format works, but I assume itˈs possible to add comments to it somehow. I would suggest adding comments to each of the RST files that explain that the documentation lives in the source file itself, to give a pointer for where people should look.

[...]

PR #2483, opened 2021-04-23
Add manual darkmode to docs

Bibo-Joshi:
So, I was procastinating and noticed that [pillows docs](https://pillow.readthedocs.io/en/stable/) have a dark mode that's activated when your browser does the `prefers-color-scheme` thingy. My eyes like dark mode, so I had a look at how they did that.

Apparently the RTD theme won't have a native dark mode for a while (readthedocs/sphinx_rtd_theme#224 is open since 2015, so ~as old as PTB). So they just took the https://darkreader.org Chromium-Extension and exported the CSS generated by that. Looks neat IMHO. Can be previewed at https://python-telegram-bot.readthedocs.io/en/docs-dark-mode/. If you don't find the dark-mode setting in your browser, here are some screenshots:

<details>
  <summary>Click to expand</summary>
  <p>

  image
  
  image
  
  image
  </p>
</details>

Alternative approaches would be to go with different sphinx templates like [Furo](https://pradyunsg.me/furo/reference/), but I actually like the RTD theme and this manual approach isn't maintenance-heavy.

Obviously this is anything but pressing, but still nice-to-have IMO. If you like it, we can merge, if not I won't be too sad 🙂


harshil21:
Definitely like the dark mode, since we're improving the docs anyway, how about these features/changes (maybe in another PR?):

- Make the list of classes in `index.rst` a dropdown one, so that the user is less overwhelmed at first look and it's easier to what's where. Perhaps [this setting](https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-collapse_navigation) can help?

- Looks like we can also [change where the 'edit on github' url](https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-github_url) points to (closing #2478). We can just link it to the corresponding file since most of our doc pages are from one file. Might have to do this for every file though.

- [Styling the external links](https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-style_external_links) might look nice. Dunno how it looks like tho.

Bibo-Joshi:
Nice ideas @harshil21 ! the external-links thingy was easy, pushing that directly to doc-fixes in a sec.

`edit on github`: IISC that would require either manually inserting `:github_url: https://github.com/ptb/ptb/blos/master/telegram/…` in each file or doing something hacky as proposed at https://github.com/readthedocs/sphinx_rtd_theme/issues/465#issuecomment-521272830, see https://github.com/peterjc/thapbi-pict/commit/f14b1df17ed3faf382a63095aa6e22f519e71957 . IMO the first is tedious & the second error-prone (and will probably break at some point). As I personally don't see a great value, I won't try to get that to work - but also won't stand in the way of it :) 

regarding the reference list: I've played around a bit, but didn't find a completey satisfactory solution. The best I could do was adding a `:hidden:` to the `toctree::` directives in `index.rst` which hides doesn't render them on the start page but still includes them in the TOC in the navigation bar. Don't know if that's what we want …

VChernomYA (approved):

VChernomYA:
[](https://github.com/telegramdesktop/tdesktop/pull/16185/files)

Poolitzer (approved):
LGTM

harshil21 (approved):

PR #2495, opened 2021-04-30
Doc Fixes

Bibo-Joshi:
Gathering doc fixes to be merged on the next release. RTD-Build is at https://python-telegram-bot.readthedocs.io/en/doc-fixes

* Release notes should mention #2510


Poolitzer:
im not sure if I like the comments but I went with it.

examples/chatmemberbot.py
```diff
@@ -113,7 +113,11 @@ def show_chats(update: Update, context: CallbackContext) -> None:
     update.effective_message.reply_text(text)
 
 
-def greet_chat_members(update: Update, _: CallbackContext) -> None:
+# Best practice would be to replace context with an underscore,
+# since context is an unused local variable.
+# This being an example and not having context present confusing beginners,
```
Bibo-Joshi, line 118: Because this is an example and the underscore might confuse Python beginners, ... ?

Also looking at the repetition, what would you think of putting this remark in the Readme instead?

Poolitzer: yeah, I like that.

examples/README.md
```diff
@@ -4,6 +4,8 @@ In this folder are small examples to show what a bot written with `python-telegr
 
 All examples are licensed under the [CC0 License](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/LICENSE.txt) and are therefore fully dedicated to the public domain. You can use them as the base for your own bots without worrying about copyrights.
 
+Do note that we ignore one pythonic convention. Best practice would dictate, in many handler callbacks function signatures, to replace context with an underscore, since context is an unused local variable in those callbacks. However, since these are examples and not having context present lead to confused beginners, we decided to have it present.
```
Bibo-Joshi, line 7:
```suggestion
Do note that we ignore one pythonic convention. Best practice would dictate, in many handler callbacks function signatures, to replace the argument `context` with an underscore, since `context` is an unused local variable in those callbacks. However, since these are examples and not having a name for that argument confuses beginners, we decided to have it present.
```

PR #2503, opened 2021-05-03
Fixing timerbot.py Closes #2502

### LINKED ISSUE #2502: [BUG] Example for timerbot.py runs only once (and not scheduled as per timer)
### Steps to reproduce
1. Run the file timerbot.py
2. Use command /set <seconds> to  set a timer for alarm

### Expected behaviour
"Beep" message should be sent repeatedly after every n seconds.

### Actual behaviour
"Beep" message is sent only once and job is removed from job queue automatically.

### Configuration
Windows/Linux

### Logs
2021-05-03 18:30:11,108 - apscheduler.scheduler - INFO - Scheduler started
2021-05-03 18:30:19,167 - apscheduler.scheduler - INFO - Added job "263784925" to job store "default"
2021-05-03 18:30:24,101 - apscheduler.executors.default - INFO - Running job "263784925 (trigger: date[2021-05-03 18:30:24 UTC], next run at: 2021-05-03 18:30:24 UTC)"
 (scheduled at 2021-05-03 18:30:24.100786+00:00)
2021-05-03 18:30:24,101 - apscheduler.scheduler - INFO - Removed job 6ac7d08abaec4c10a2c89bc720dbc264
2021-05-03 18:30:24,360 - apscheduler.executors.default - INFO - Job "263784925 (trigger: date[2021-05-03 18:30:24 UTC], next run at: 2021-05-03 18:30:24 UTC)" execute
d successfully

### Proposed Solution
Use context.job_queue.run_repeating instead of context.job_queue.run_once in set_timer method

Let me know if issue needs to be fixed, will raise a PR for the same.

fathomer:
Using context.job_queue.run_repeating instead on run_once as timer alarm job needs to be run repeatedly based on the timer.
Currently it only runs once and removes the job from queue automatically.

Check #2502 for more info on the issue.

### DIFF
--- a/examples/timerbot.py
+++ b/examples/timerbot.py
@@ -64,7 +64,7 @@ def set_timer(update: Update, context: CallbackContext) -> None:
             return
 
         job_removed = remove_job_if_exists(str(chat_id), context)
-        context.job_queue.run_once(alarm, due, context=chat_id, name=str(chat_id))
+        context.job_queue.run_repeating(alarm, due, context=chat_id, name=str(chat_id))
 
         text = 'Timer successfully set!'
         if job_removed:


Bibo-Joshi:
Hi. thanks for the PR, but the `timerbot.py` example is, well, an example. If you expect a timer to run repeatedly, you're welcome to implement that for yourself. However, I don't see the need to change the example 😉

fathomer:
> Hi. thanks for the PR, but the `timerbot.py` example is, well, an example. If you expect a timer to run repeatedly, you're welcome to implement that for yourself. However, I don't see the need to change the example 😉
Hi @Bibo-Joshi 
Understandable, but what even is the point of using a timer and setting a time interval, if it is going to run only once and never again and no where that time interval is getting used.
I myself was using this amazing library for the first time today, and taking this example as a reference, I spent a significant time in figuring out what I was doing wrong and why the message isn't being sent at a set interval as it should.
Anyways, it's ok if you still think this is not required , but still I strongly feel this is a bug and not a feature :sweat_smile:

Bibo-Joshi:
There are many valid use cases for a one-time alarm. Like boiling an egg. You don't want the alarm to go off every 4 minutes - one time is perfectly fine. Anyway, the example is just there to illustrate how `JobQueue` can be used in general, not to be a state of the art-solution for a specific use case.

fathomer:
> There are many valid use cases for a one-time alarm. Like boiling an egg. You don't want the alarm to go off every 4 minutes - one time is perfectly fine. Anyway, the example is just there to illustrate how `JobQueue` can be used in general, not to be a state of the art-solution for a specific use case.

Ah yes yes, I missed the whole point and only now I see it. :)
Thanks for the explanations and Sorry for the trouble, I was very wrong in the understanding of that example, my bad.

PR #2510, opened 2021-05-08
url on InlineKeyboardButton is optional

bastbnl:
It's a docstring update, marking `url` optional as it is on https://python-telegram-bot.readthedocs.io/en/stable/telegram.inlinekeyboardbutton.html describing the purpose of the url in more detail

I assumed statements concerning versionadded etc would be too much for this

### DIFF
--- a/telegram/inline/inlinekeyboardbutton.py
+++ b/telegram/inline/inlinekeyboardbutton.py
@@ -40,7 +40,7 @@ class InlineKeyboardButton(TelegramObject):
 
     Args:
         text (:obj:`str`): Label text on the button.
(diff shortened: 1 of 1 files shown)


Poolitzer:
hey, good catch. I moved the base from master to doc fixes, since we collect said doc fixes in there for the release. Sadly, there is a master commit which would be wrong in the doc-fixes branch. Would you want to fix this? Otherwise I can cherry pick and dismiss the PR.

Bibo-Joshi:
Just merged master into doc-fixes :) Thanks for the catch!

PR #2520, opened 2021-05-14
Increase cbqh coverage

Poolitzer:
### PR to increase CallbackQueryHandler coverage

this makes sure that one else statement explicitly returns None and adds a test for it.

### DIFF
--- a/telegram/ext/callbackqueryhandler.py
+++ b/telegram/ext/callbackqueryhandler.py
@@ -157,6 +157,7 @@ class CallbackQueryHandler(Handler[Update]):
             :obj:`bool`
 
         """
+        # testing a suspicion
         if isinstance(update, Update) and update.callback_query:
             if self.pattern:
                 if update.callback_query.data:
--- a/telegram/ext/callbackqueryhandler.py
+++ b/telegram/ext/callbackqueryhandler.py
@@ -157,9 +157,9 @@ class CallbackQueryHandler(Handler[Update]):
             :obj:`bool`
 
         """
-        # testing a suspicion
         if isinstance(update, Update) and update.callback_query:
             if self.pattern:
+                # testing a suspicion (lets do it directly here)
                 if update.callback_query.data:
                     match = re.match(self.pattern, update.callback_query.data)
                     if match:
--- a/telegram/ext/callbackqueryhandler.py
+++ b/telegram/ext/callbackqueryhandler.py
@@ -81,7 +81,8 @@ class CallbackQueryHandler(Handler[Update]):
             DEPRECATED: Please switch to context based callbacks.
         pattern (:obj:`str` | `Pattern`, optional): Regex pattern. If not :obj:`None`, ``re.match``
             is used on :attr:`telegram.CallbackQuery.data` to determine if an update should be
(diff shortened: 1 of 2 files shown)


telegram/ext/callbackqueryhandler.py
```diff
@@ -81,7 +81,8 @@ class CallbackQueryHandler(Handler[Update]):
             DEPRECATED: Please switch to context based callbacks.
         pattern (:obj:`str` | `Pattern`, optional): Regex pattern. If not :obj:`None`, ``re.match``
             is used on :attr:`telegram.CallbackQuery.data` to determine if an update should be
-            handled by this handler.
+            handled by this handler. If :attr:`telegram.CallbackQuery.data` is not present, the
+            CallbackQuery update will not be handled.
```
Bibo-Joshi, line 85:
```suggestion
            :class:`telegram.CallbackQuery` update will not be handled.
```
or just "the update" …

telegram/ext/callbackqueryhandler.py
```diff
@@ -163,6 +164,8 @@ def check_update(self, update: object) -> Optional[Union[bool, object]]:
                     match = re.match(self.pattern, update.callback_query.data)
                     if match:
                         return match
+                else:
+                    return None
```
Bibo-Joshi, line 168: This isn't necessary to up the coverage ;)

Poolitzer: and here I thought thats the whole point? I am confused now.

Bibo-Joshi: no, codecov only tells us to have a test for the if-clause giving "true" and a test for the if-clause giving "false". It doesn't tell us that we need to have an else-clause 😉

Poolitzer: ohhhhh. Thats what its doing. And it knows where the line is tested in pytest. Damn thats smart coding

Bibo-Joshi: > And it knows where the line is tested in pytest

No, it doesn't. Running pytest with the right arguments (and additional dependencies) just generates a report about which lines are covered and codecov can read that

Bibo-Joshi: Dependencies: 

https://github.com/python-telegram-bot/python-telegram-bot/blob/9737b1d3c78bdb8d79d726e27a2b37f840fe3f55/.github/workflows/test.yml#L31

arguments:

https://github.com/python-telegram-bot/python-telegram-bot/blob/9737b1d3c78bdb8d79d726e27a2b37f840fe3f55/.github/workflows/test.yml#L49

uploading report:

https://github.com/python-telegram-bot/python-telegram-bot/blob/9737b1d3c78bdb8d79d726e27a2b37f840fe3f55/.github/workflows/test.yml#L62

see e.g. https://github.com/python-telegram-bot/python-telegram-bot/runs/1937444759#step:7:38

telegram/ext/callbackqueryhandler.py
Poolitzer: nah I like the link

PR #2522, opened 2021-05-14
Stabilize some tests

harshil21:
Tries to fix tests like: 

- [ ]  [test_clean_depr_warn_polling](https://github.com/python-telegram-bot/python-telegram-bot/runs/2493716321?check_suite_focus=true), 
- [ ]  [test_clean_depr_warn_webhook](https://github.com/python-telegram-bot/python-telegram-bot/runs/2582381833?check_suite_focus=true), 
First two will be fixed when the tests actually fail and the error is printed to console
- [x]  [Stickerset_not_modified](https://github.com/python-telegram-bot/python-telegram-bot/runs/2518489659?check_suite_focus=true), 
- [x]  [test_idle](https://github.com/python-telegram-bot/python-telegram-bot/runs/2533442877?check_suite_focus=true), 
- [x]  [Botscore_not_modified](https://github.com/python-telegram-bot/python-telegram-bot/runs/2582219566?check_suite_focus=true)

Checks will be run many times to expose the failing ones

### DIFF
--- a/tests/test_bot.py
+++ b/tests/test_bot.py
@@ -94,17 +94,17 @@ def inline_results():
     return inline_results_callback()
 
 
-@pytest.fixture(scope='class')
-def game_bot():
-    return Bot(FALLBACKS[0]["token"])
-
-
-cond = False
+BASE_GAME_SCORE = 60  # Base game score for game tests
+cond = False  # This condition is only relevant for github actions game tests.
 
 if GITHUB_ACTION:
     cond = True
 
-xfail = pytest.mark.xfail(cond, reason='Can fail due to race conditions')
+xfail = pytest.mark.xfail(
+    cond,
+    reason='Can fail due to race conditions when multiple test suites '
+    'with the same bot token are run at the same time',
+)
 
 
 class TestBot:
@@ -1283,34 +1283,34 @@ class TestBot:
                 )
 
     @xfail
-    def test_set_game_score_1(self, game_bot, chat_id):
+    def test_set_game_score_1(self, bot, chat_id):
         # NOTE: numbering of methods assures proper order between test_set_game_scoreX methods
         # First, test setting a score.
         game_short_name = 'test_game'
-        game = game_bot.send_game(chat_id, game_short_name)
+        game = bot.send_game(chat_id, game_short_name)
 
-        message = game_bot.set_game_score(
+        message = bot.set_game_score(
             user_id=chat_id,
-            score=60,
+            score=BASE_GAME_SCORE,  # Score value is relevant for other set_game_score_* tests!
             chat_id=game.chat_id,
             message_id=game.message_id,
         )
 
-        # We don't test for animation file id, since it actually changes when the score is updated
         assert message.game.description == game.game.description
         assert message.game.photo[0].file_size == game.game.photo[0].file_size
+        assert message.game.animation.file_unique_id == game.game.animation.file_unique_id
         assert message.game.text != game.game.text
 
     @xfail
-    def test_set_game_score_2(self, game_bot, chat_id):
+    def test_set_game_score_2(self, bot, chat_id):
         # NOTE: numbering of methods assures proper order between test_set_game_scoreX methods
         # Test setting a score higher than previous
         game_short_name = 'test_game'
-        game = game_bot.send_game(chat_id, game_short_name)
+        game = bot.send_game(chat_id, game_short_name)
 
-        score = 61
+        score = BASE_GAME_SCORE + 1
 
-        message = game_bot.set_game_score(
+        message = bot.set_game_score(
             user_id=chat_id,
             score=score,
             chat_id=game.chat_id,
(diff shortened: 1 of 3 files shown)


tests/test_bot.py
```diff
@@ -1332,27 +1336,32 @@ def test_set_game_score_3(self, bot, chat_id):
...
+            # For some reason the returned message doesn't contain the updated score. need to fetch
+            # the game again...
+            game2 = bot.send_game(chat_id, game_short_name)
+            assert str(score) in game2.game.text
 
-        # For some reason the returned message does not contain the updated score. need to fetch
-        # the game again...
-        game2 = bot.send_game(chat_id, game_short_name)
-        assert str(score) in game2.game.text
+        expect_bad_request(func, 'Bot_score_not_modified', 'This test can be flaky')
```
Bibo-Joshi, line 1364: I guess the more important reason here is that this also depends on the previous tests to run properly? (same above)

tests/test_updater.py
```diff
@@ -478,6 +482,9 @@ def test_idle(self, updater, caplog):
         with caplog.at_level(logging.INFO):
             updater.idle()
 
+        if caplog.records[-1].getMessage().startswith('Error while getting Updates: Conflict'):
```
Bibo-Joshi, line 485: wait, what? Will need to double check, but I IIRC we have 1 bot token per test suite, so this should never happen?!

Bibo-Joshi: Yes, we have 3.6-3.9 on 3 OSes = 4×3 = 12 suits and 12 test bots … But this can happen when running tests for different PRs in parallel, probably? We should at least add a comment for that, IG

tests/test_bot.py
harshil21: Yea but I ran the tests individually and they passed. So not sure what that was about.

tests/test_updater.py
harshil21: > But this can happen when running tests for different PRs in parallel, probably? 

It happens roughly once a week iirc, so yeah

> We should at least add a comment for that, IG

ok will include in next commit once I find out why the first two tests fail at times

tests/test_bot.py
Bibo-Joshi: mhkay. Still, "This test can be flaky" seems a bit strange as informations given that the test already has the `@flaky` decorator 😉 Would making it more flaky (e.g. `@flaky(5,1)` - 1 out of 5 runs) help?

harshil21: not sure, but I'll try by only keeping the flaky and testing...

harshil21:
There were many reasons why the game tests were failing -

- Not using one single bot. This caused problems since different bots were keeping track of different scores across different tests...
- `animation_file_id` actually changes every time the game score is edited, but the tests were checking if they were the same...
- `@flaky` on game tests were useless or actually made it more unstable as the same scores were being used, raising the error again.

Despite fixing this, there is still a chance that this can fail due to race conditions (many jobs running at once), so an `xfail` was added.

Onto the next test...

tests/test_bot.py
```diff
@@ -98,6 +94,19 @@ def inline_results():
     return inline_results_callback()
 
 
+@pytest.fixture(scope='class')
+def game_bot():
+    return Bot(FALLBACKS[0]["token"])
+
+
+cond = False
```
Bibo-Joshi, line 102: Please leave a comment here explaining what this is good for

tests/test_bot.py
```diff
@@ -1273,37 +1282,35 @@ def test_send_game_default_allow_sending_without_reply(self, default_bot, chat_i
...
-            assert message.game.text != game.game.text
+        message = game_bot.set_game_score(
+            user_id=chat_id,
+            score=60,
+            chat_id=game.chat_id,
+            message_id=game.message_id,
+        )
 
-        expect_bad_request(func, 'Bot_score_not_modified', 'This test is a diva for some reason.')
+        # We don't test for animation file id, since it actually changes when the score is updated
```
Bibo-Joshi, line 1299: can't we just test with `file_unique_id`?

tests/test_bot.py
```diff
@@ -98,6 +94,19 @@ def inline_results():
...
+def game_bot():
+    return Bot(FALLBACKS[0]["token"])
+
+
+cond = False
+
+if GITHUB_ACTION:
+    cond = True
+
+xfail = pytest.mark.xfail(cond, reason='Can fail due to race conditions')
```
Bibo-Joshi, line 107:
```suggestion
xfail = pytest.mark.xfail(cond, reason='Can fail due to race conditions when multiple test suits with the same bot token run at the same time')
```

tests/test_bot.py
```diff
@@ -1273,37 +1282,35 @@ def test_send_game_default_allow_sending_without_reply(self, default_bot, chat_i
...
-                message_id=game.message_id,
-            )
-
-            assert message.game.description == game.game.description
-            assert message.game.animation.file_id == game.game.animation.file_id
-            assert message.game.photo[0].file_size == game.game.photo[0].file_size
-            assert message.game.text != game.game.text
+        message = game_bot.set_game_score(
+            user_id=chat_id,
+            score=60,
```
Bibo-Joshi, line 1294: Can we keep tha `CONSTANT + offset` pattern for score? Or leave a comment `# score value is relevant for the other set_game_score tests!`?

tests/test_bot.py
```diff
@@ -98,6 +94,19 @@ def inline_results():
     return inline_results_callback()
 
 
+@pytest.fixture(scope='class')
+def game_bot():
+    return Bot(FALLBACKS[0]["token"])
```
Bibo-Joshi, line 99: I don't understand why this is necessary … The `bot` fixture is session scoped, so the `test_s_g_s_*` tests all run with the same bot token anyway

https://github.com/python-telegram-bot/python-telegram-bot/blob/9737b1d3c78bdb8d79d726e27a2b37f840fe3f55/tests/conftest.py#L84-L86

Bibo-Joshi (commented):
I admire your patience with the tests 👍

tests/test_bot.py
harshil21: You're right, I thought the scope meant something else. I was testing each function individually and that's why I noticed that 🤦🏼

tests/test_bot.py
harshil21: Alright

tests/test_bot.py
harshil21: yea sure, will add

tests/test_bot.py
Bibo-Joshi: ah, that makes sense :D np

Poolitzer:
changes look good to me

PR #2523, opened 2021-05-15
Ignore classes on replace/insert_bot

### LINKED ISSUE #2513: [BUG] Bot.insert/replace_bot with types
Yes, here we go again 😬 
If your data contains a class, you get `AttributeError: attribute '__dict__' of 'type' objects is not writable`.
I guess we can just skip types in the insert/replace logic. If a class holds a reference to a Bot it must be a class variable and that should be reconstructed at runtime. But this logic mostly comes from my understanding of how pickle persists classes, so this needs some double checking, if there are implications for other cases.

Bibo-Joshi:
Closes #2513

### Checklist for PRs

- [x] `.. versionadded:: version` not necessary, this is more a bug fix than a change …

### DIFF
--- a/telegram/ext/basepersistence.py
+++ b/telegram/ext/basepersistence.py
@@ -137,7 +137,8 @@ class BasePersistence(ABC):
         Replaces all instances of :class:`telegram.Bot` that occur within the passed object with
         :attr:`REPLACED_BOT`. Currently, this handles objects of type ``list``, ``tuple``, ``set``,
         ``frozenset``, ``dict``, ``defaultdict`` and objects that have a ``__dict__`` or
-        ``__slot__`` attribute, excluding objects that can't be copied with `copy.copy`.
+        ``__slot__`` attribute, excluding objects that can't be copied with `copy.copy` and
+        classes.
 
         Args:
             obj (:obj:`object`): The object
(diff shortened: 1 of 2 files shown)


telegram/ext/basepersistence.py
```diff
@@ -215,7 +224,8 @@ def insert_bot(self, obj: object) -> object:
         Replaces all instances of :attr:`REPLACED_BOT` that occur within the passed object with
         :attr:`bot`. Currently, this handles objects of type ``list``, ``tuple``, ``set``,
         ``frozenset``, ``dict``, ``defaultdict`` and objects that have a ``__dict__`` or
-        ``__slot__`` attribute, excluding objects that can't be copied with `copy.copy`.
+        ``__slot__`` attribute, excluding objects that can't be copied with `copy.copy` and
+        classes.
```
harshil21, line 228: same

telegram/ext/basepersistence.py
```diff
@@ -248,6 +258,14 @@ def _insert_bot(self, obj: object, memo: Dict[int, object]) -> object:  # pylint
             new_immutable = obj.__class__(self._insert_bot(item, memo) for item in obj)
             memo[obj_id] = new_immutable
             return new_immutable
+        if isinstance(obj, type):
+            # classes usually do have a __dict__, but it's not writable
+            warnings.warn(
+                'BasePersistence.insert_bot does not handle classes. See '
+                'the docs of BasePersistence.replace_bot for more information.',
```
harshil21, line 265: insert_bot

harshil21 (changes requested):

Bibo-Joshi:
Test failure apparently due to GH Actions currently having problems, see https://www.githubstatus.com
When that's resolved & we can rerun, I guess I'll merge. Thanks for the review @harshil21 !

harshil21 (approved):

PR #2528, opened 2021-05-19
Attachment filter

Poolitzer:
This is a very simple filter. It filters messages where effective_attachment is set to something.

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -2082,9 +2082,9 @@ officedocument.wordprocessingml.document")``.
             return bool(message.effective_attachment)
 
     attachment = _Attachment()
-    """Messages where :meth:`telegram.Message.effective_attachment` is not None.
+    """Messages that contain :meth:`telegram.Message.effective_attachment`.
 
-        See the documentation of that attribute when that is the case.
+
         .. versionadded:: 13.6"""
 
     class _UpdateType(UpdateFilter):
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -2084,7 +2084,8 @@ officedocument.wordprocessingml.document")``.
     attachment = _Attachment()
     """Messages where :meth:`telegram.Message.effective_attachment` is not None.
 
-        See the documentation of that attribute when that is the case."""
+        See the documentation of that attribute when that is the case.
+        .. versionadded:: 13.6"""
 
     class _UpdateType(UpdateFilter):
         name = 'Filters.update'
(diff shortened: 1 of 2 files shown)


telegram/ext/filters.py
```diff
@@ -2075,6 +2075,18 @@ def filter(self, message: Message) -> bool:
...
             )
 
+    class _Attachment(MessageFilter):
+        name = 'Filters.attachment'
+
+        def filter(self, message: Message) -> bool:
+            return bool(message.effective_attachment)
+
+    attachment = _Attachment()
+    """Messages where :meth:`telegram.Message.effective_attachment` is not None.
```
Bibo-Joshi, line 2085:
```suggestion
    """Messages that contain :meth:`telegram.Message.effective_attachment`.
```
? This matches the other docstrings. or at least ``:obj:`None` `` 😉

telegram/ext/filters.py
```diff
@@ -2075,6 +2075,18 @@ def filter(self, message: Message) -> bool:
...
+
+        See the documentation of that attribute when that is the case.
+        .. versionadded:: 13.6"""
```
Bibo-Joshi, line 2088:
```suggestion
        See the documentation of that attribute for more details?
        
        .. versionadded:: 13.6
        """
```
IMHO we can just skip the first sentence. We don't have something like that for the other filters either
IIRC the new line before ..versionadded is necessary for it to render properly …

tests/test_filters.py
```diff
@@ -2173,3 +2173,18 @@ def test_filters_via_bot_repr(self):
 
         with pytest.raises(RuntimeError, match='Cannot set name'):
             f.name = 'foo'
+
+    def test_filters_attachment(self, update):
+        assert not Filters.attachment(update)
+        # we need to define a new Update (or rather, message class) here because
+        # effective_attachment is only evaluated once per class, and the filter relies on that
```
Bibo-Joshi, line 2180:
```suggestion
        # effective_attachment is only evaluated once per instance, and the filter relies on that
```

Bibo-Joshi (commented):
just docstring nit picking :)

telegram/ext/filters.py
Bibo-Joshi: > > IMHO we can just skip the first sentence
> 
> Sure. I just wanted to point out that that is the case, since I feel like thats less obvious then other cases

your call

Poolitzer:
```suggestion

        .. versionadded:: 13.6"""
```

PR #2541, opened 2021-05-30
Fix slot tests

Bibo-Joshi:
* Makes `test_class_has_slots_and_dict` use the correct relative path for our test setup, so it actually tests something :D 
* Handles windows
* handles the `__dict__` check for classes where `__dict__` is in some parents `__slots__`, but we can't see that through MRO b/c of python magic (most notably for ABCs)

This caused some test failures in #1844 #2622 and I figured it's cleaner to just handles this in master directy …


harshil21 (approved):
Oops my bad, looks fine.

Issue #2542, opened 2021-06-04
[BUG] promise mistreats pooled_function exceptions as done_callback exceptions


angadsingh:
### Steps to reproduce
1. At the lowest level, create a `telegram.ext.utils.Promise`, throw an exception from the `pooled_function` on which the promise was built.
2. Use `DelayQueue` using `@mq.queuedmessage` decorator on a subclassed Bot class. any exception like TimeOut or flood control within a `send_message` for example ends up with this problem
3. Use MessageQueue - I know this is deprecated, but that isn't the problematic code here. Promise itself isn't correctly handling exceptions.

If you get any telegram exception, it just gets eaten up by Promise's run method in this specific area:
```python
            if self._done_callback:
                try:
                    self._done_callback(self.result())
                except Exception as exc:
                    logger.warning(
                        "`done_callback` of a Promise raised the following exception."
                        " The exception won't be handled by error handlers."
                    )
                    logger.warning("Full traceback:", exc_info=exc)
```

Here's the bug: even if there's an exception in `self.result()` it will treat it as an exception from `self._done_callback()` which isn't correct.

The funny thing is that code speaks out the problem on its own:
```python
      try:
            self._result = self.pooled_function(*self.args, **self.kwargs)

        except Exception as exc:
            self._exception = exc
```
if there's an exception in `pooled_function`, it is stored in `self._exception`

and then it is raised from `self.result()`:
```python
    def result(self, timeout: float = None) -> Optional[RT]:
        self.done.wait(timeout=timeout)
        if self._exception is not None:
            raise self._exception  # pylint: disable=raising-bad-type
        return self._result
```
why? this way all exception from `pooled_function` are forcibly going to be taken as exceptions of `done_callback`, eaten up by the promise code and never raised.

### Expected behaviour
`Promise` should ideally have another method called `add_exception_callback` or similar just like `add_done_callback`, and pass any exception from `pooled_function` to that callback

### Actual behaviour
Promise mistreats `pooled_function` exceptions as `done_callback` exceptions

### Configuration
**Operating System:**
linux

**Version of Python, python-telegram-bot & dependencies:**
```
Bot API 5.2
certifi 2020.12.05
Python 3.8.5 (default, Jul 21 2020, 10:42:08)  [Clang 11.0.0 (clang-1100.0.33.17)]
```
### Logs
Insert logs here (if necessary)
```
[...]

PR #2543, opened 2021-06-04
proposed fix for #2542: proper treatment of pooled_function exception…

angadsingh:
send the pooled_function exception back through done_callback instead of eating it up, mistreating it as a done_callback exception

### DIFF
--- a/telegram/ext/utils/promise.py
+++ b/telegram/ext/utils/promise.py
@@ -92,23 +92,23 @@ class Promise:
 
     def run(self) -> None:
         """Calls the :attr:`pooled_function` callable."""
+
         try:
             self._result = self.pooled_function(*self.args, **self.kwargs)
-
         except Exception as exc:
             self._exception = exc
-
         finally:
             self.done.set()
             if self._done_callback:
                 try:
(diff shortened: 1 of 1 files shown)


Bibo-Joshi:
hey Glad to see that you are eager to fix this. However, your change is immediately breaking (especially for `ConversationHandler`), because you pass an unexpected argument to the `done_callbacks`.

> but for now this fix is the only way for letting users of the library retry on timeouts, other network exceptions or flood limit exceptions from telegram's API

You can also just add proper error handling directly in your `pooled_function`. `Promise` is used two places:

* MessageQueue - deprecated anyway, so no use to add new features
* run_async, including ConversationHandler - here callbacks get error handling via dispatchers error handlers

Moreover, `done_callback` is only used in ConversationHandler and we don't need additional error handling there.

If you want to, you are very welcome to update your PR to the approach detailed in https://github.com/python-telegram-bot/python-telegram-bot/issues/2542#issuecomment-854629317 :)

Bibo-Joshi:
superseded by #2544. Nevertheless, thanks for taking the time to PR :)

PR #2547, opened 2021-06-06
Include max_connections in start_webhook

GauthamramRavichandran:
It was earlier possible to set max_connections using the set_webhook method. With the 

1. Added `max_connections` args to `start_webhook` method
2. Updated the doc_string
3. Tested, it was able to set `max_connections` to user set value

### DIFF
--- a/telegram/ext/updater.py
+++ b/telegram/ext/updater.py
@@ -412,10 +412,10 @@ class Updater:
                    Since version 13.6, ``tornade>=6.1`` is required, which resolves the former
                    issue.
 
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
Hi. Thanks for the PR! I took the liberty to update the docs & add a test case so that we can include this in the upcoming release :)

GauthamramRavichandran:
Thanks!

Bibo-Joshi:
Test failures are due to codecov not being able to upload the coverage reports for some reason … I'll merge anyway, as coverage should be fine :)

Issue #2346, opened 2021-01-30
Deprecation: v14


Bibo-Joshi:
Things we should drop in v14. The list is probably not complete. This includes everything deprecated in v12.x and v13.x.
---
- [x] DRop support for py3.6 - reaches EOL in december anyway … #2619 & #2636

----

- [x] Don't know why we skipped that in #1905, but as BasePersistence is an interface, it makes sense for all methods to be abcs. As flush currently doesn't raise NotImplementedError, this is strictly speaking breaking, so putting this on the v14 milestone. #2624
- [x] non-context based callbacks (i.e. all the `pass_*` arguments) #2617
- [x] `allow_edited` argument of CommandHandler #2617
- [x] non-filter arguments of MessageHandler #2617
- [x] `RegexHandler` #2617
- [x] `bot` argument of `JobQueue` #2597
- [x] `use_context` argument of `Dispatcher` and `Updater` (this is discussable. we could keep that until v15 as convenience for everyone who already switched to context based callbacks - then again, for adjusting your code to v14, this will be the easiest part …)  #2617
- [x] `MessageQueue` & `DelayQueue` as a whole (we might want to add some of the logic back in #2139 after the asyncio switch, but maybe it's cleaner to just remove everything in case #2139 will make it only to v14.1 or so)  #2644
- [x] drop `__dict__` from `__slots__` (users won't be able to set custom attributes on PTB objects) #2619
- [x] make `BP.refresh_*_data` abstract methods #2624
- [x] make `Persistence.store_callback_data` default to `True` and make `BP.get/update_callback_data` abstract methods. #2624
- [x] Move `TelegramDecryptionError` to `error.py` (https://github.com/python-telegram-bot/python-telegram-bot/pull/2345/commits/851f0b7d2e50eb0791600e946a96b8062b8ba6d1, it was reverted since it was breaking) #2621 
---
Originally from #2347:
- [x] The `@run_async` decorator #2644
- [x] `Filters.{private, group}` #2644
- [x] `utils.{promise, webhookhandler}` #2644
- [x] `clean` argument of `Updater.start_{polling, webhook}` #2644
- [x] Think about removing `Defaults` from `tg.Bot` completey and moving them to `tg.ext.ExtBot`. We may have to think about how to handle signatures with `DefaultVaule`, then, i.e. if `_insert_defaults` should stay in `tg.Bot` or should be moved to `tg.ext.ExtBot`. #2648
---
[...]

PR #2550, opened 2021-06-06
Remove deprecated arg from deeplinking.py

harshil21:
for some reason `pass_args=True` was still being passed to a `CommandHandler`


Bibo-Joshi:
Maybe we can do this together with #2346?

Poolitzer:
I dont see the need, ou contraire. This example throws a warning, we should merge this asap to avoid bad copde examples.

PR #2553, opened 2021-06-07
Don't throw warning when passing defaults to ExtBot

Bibo-Joshi:
See https://t.me/pythontelegrambotgroup/483120

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -192,11 +192,7 @@ class Bot(TelegramObject):
         # Gather default
         self.defaults = defaults
 
-        # We need to be a bit tricky here because
-        # 1. __init__ is called by the subclasses where Defaults is not deprecated
-        # 2. this must work in ptb-raw as well, were we can't import ExtBot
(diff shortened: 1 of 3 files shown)


Poolitzer:
Two things I don't understand.

I hope I understand this correctly though: We don't want to put the defaults logic into ExtBot yet because that would be breaking, but we intend to do this in the future.

If we don't want to remove the logic from Bot just yet, why are you still passing it in init? You could just set it manually after creating the Bot instance. This way you catch people doing it in their Bot() init, while still be able to do it yourself. I mean your solution works, but it just seems a tad complicated.

Secondly, I would actually release a fix release for this. Maybe wait a week (or some more) to see if more bugs show themselves, but this is a warning which shows up when you do everything correctly. I know its just a warning, but people tend to treat warnings as errors (which I personally support), and I bet we will get a lot of people asking the same question as soon as they update, which we can avoid and should imo.

Poolitzer (approved):
Looks good to me!

PR #2561, opened 2021-06-12
BP.insert/replace_bot: Handle objects with __dict__ in __slots__

Bibo-Joshi:
Never ending story 😬  Anyway, good that we didn't make a bug fix release yet


Poolitzer:
what error is that fixing?

Poolitzer (approved):
Code looks good

PR #2570, opened 2021-06-24
Pass job to error handler where possible

### LINKED ISSUE #2568: [FEATURE] Pass jobs to error handlers
When a job callback raises an error, there's currently no way to link the exception to the job in an error handler

* Add a paremeter `job` to `CallbackContext.from_error`
* Adjust the calls in `JobQueue` accordingly

#### Additional context
Reported in the usergroup, discussion strated here: https://t.me/pythontelegrambotgroup/489719

Bibo-Joshi:
Closes #2568 - at least partly.

For jobs that run only once, APS already drops the job from memory before reporting any exception so that we can't get it back by its ID. One could maybe try to work around that somehow by reworking some of the internals of `JobQueue` (bascially tell APS to run `job.run(dispatcher)` instead of running the `callback` directly, but I'm not sure if that's worth it.

That would also make it even more complicated for custom persistence solutions like the recently introduced `ptbcontrib/ptb_sqlalchemy_jobstore` (even though this shouldn't be the main reason to to do it)
The main use case probably is to stop repeating jobs that cause exceptions, anyway.

Because this adds a new parameter to `CallbackContext.from_error`, this is maybe kind of breaking if people subclassed for `ContextTypes` and overrode that. Anyway, not that urgent so I'll be putting it on the v14 milestone for now.

**edit:** We could just make `tg.ext.Job` callable (where `job(dispatcher)` just does `job.run(dispatcher)`). This would probably solve all of the issues mentioned above 🤔 

### Checklist for PRs

- [ ] Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings  - did that but should be updated before merging

### DIFF
--- a/telegram/ext/callbackcontext.py
+++ b/telegram/ext/callbackcontext.py
@@ -88,7 +88,8 @@ class CallbackContext(Generic[UD, CD, BD]):
             that raised the error. Only present when the raising function was run asynchronously
             using :meth:`telegram.ext.Dispatcher.run_async`.
         job (:class:`telegram.ext.Job`): Optional. The job which originated this callback.
-            Only present when passed to the callback of :class:`telegram.ext.Job`.
+            Only present when passed to the callback of :class:`telegram.ext.Job` or in error
+            handlers if the error is caused by a `repeating` job.
 
     """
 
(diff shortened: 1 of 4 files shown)


Bibo-Joshi:
closing in favor of #2692

PR #2572, opened 2021-06-25
API 5.3

Bibo-Joshi:
* `BotCommandScope*` is just for input, never output (at least for now), so we don't really need a proper `de_json`. But I realized that only after I had implemented it … If you think it's superfluous, we can remove
* Leaving the checklists empty for now so we can go over them once in the end.

### Checklist for PRs

* New classes:
    - [x] `__init__` accepts `**_kwargs`
    
* Added new shortcuts:
    - [x] In `Chat` & `User` for all methods that accept `chat/user_id` -> We need to deprecate `Chat.get_members_count` in favor of a new `Chat.get_member_count` and the same for `kick_chat_member` -> `ban_chat_member`
    - ~[ ] In `Message` for all methods that accept `chat_id` and `message_id`~
    - ~[ ] For new `Message` shortcuts: Added `quote` argument if methods accepts `reply_to_message_id`~
    - ~[ ] In `CallbackQuery` for all methods that accept either `chat_id` and `message_id` or `inline_message_id`~
    
* If relevant:

    - ~[ ] Added new handlers for new update types~
    - ~[ ] Added new filters for new message (sub)types~
    - [x] Updated the Bot API version number in all places: `README.rst` and `README_RAW.rst` (including the badge), as well as `telegram.constants.BOT_API_VERSION`
    - ~[ ] Added logic for arbitrary callback data in `tg.ext.Bot` for new methods that either accept a `reply_markup` in some form or have a return type that is/contains `telegram.Message`~

### DIFF
--- a/docs/source/telegram.chatmemberadministrator.rst
+++ b/docs/source/telegram.chatmemberadministrator.rst
@@ -1,7 +1,7 @@
 :github_url: https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/chatmember.py
 
 telegram.ChatMemberAdministrator
-===================
+================================
 
 .. autoclass:: telegram.ChatMemberAdministrator
     :members:
--- a/docs/source/telegram.chatmemberadministrator.rst
+++ b/docs/source/telegram.chatmemberadministrator.rst
@@ -1,0 +1,8 @@
+:github_url: https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/chatmember.py
+
+telegram.ChatMemberAdministrator
+===================
+
+.. autoclass:: telegram.ChatMemberAdministrator
+    :members:
+    :show-inheritance:
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -2316,6 +2316,62 @@ class Bot(TelegramObject):
         supergroups and channels, the user will not be able to return to the group on their own
         using invite links, etc., unless unbanned first. The bot must be an administrator in the
         chat for this to work and must have the appropriate admin rights.
+        Args:
+            chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target group or username
+                of the target supergroup or channel (in the format ``@channelusername``).
+            user_id (:obj:`int`): Unique identifier of the target user.
+            timeout (:obj:`int` | :obj:`float`, optional): If this value is specified, use it as
+                the read timeout from the server (instead of the one specified during creation of
+                the connection pool).
+            until_date (:obj:`int` | :obj:`datetime.datetime`, optional): Date when the user will
+                be unbanned, unix time. If user is banned for more than 366 days or less than 30
+                seconds from the current time they are considered to be banned forever. Applied
+                for supergroups and channels only.
+                For timezone naive :obj:`datetime.datetime` objects, the default timezone of the
+                bot will be used.
+            revoke_messages (:obj:`bool`, optional): Pass :obj:`True` to delete all messages from
+                the chat for the user that is being removed. If :obj:`False`, the user will be able
+                to see messages in the group that were sent before the user was removed.
+                Always :obj:`True` for supergroups and channels.
+                .. versionadded:: 13.4
+            api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
+                Telegram API.
+        Returns:
+            :obj:`bool`: On success, :obj:`True` is returned.
+        Raises:
+            :class:`telegram.error.TelegramError`
+        """
+        warnings.warn(
+            '`bot.kick_chat_member` is deprecated. Use `bot.ban_chat_member` instead.',
+            TelegramDeprecationWarning,
+            stacklevel=2,
+        )
+        return self.ban_chat_member(
+            chat_id=chat_id,
+            user_id=user_id,
+            timeout=timeout,
+            until_date=until_date,
+            api_kwargs=api_kwargs,
+            revoke_messages=revoke_messages,
+        )
+
+    @log
+    def ban_chat_member(
+        self,
+        chat_id: Union[str, int],
+        user_id: Union[str, int],
+        timeout: ODVInput[float] = DEFAULT_NONE,
+        until_date: Union[int, datetime] = None,
+        api_kwargs: JSONDict = None,
+        revoke_messages: bool = None,
+    ) -> bool:
+        """
+        Use this method to ban a user from a group, supergroup or a channel. In the case of
+        supergroups and channels, the user will not be able to return to the group on their own
+        using invite links, etc., unless unbanned first. The bot must be an administrator in the
+        chat for this to work and must have the appropriate admin rights.
+
+         .. versionadded:: 13.7
 
         Args:
             chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target group or username
@@ -2358,7 +2414,7 @@ class Bot(TelegramObject):
         if revoke_messages is not None:
             data['revoke_messages'] = revoke_messages
 
-        result = self._post('kickChatMember', data, timeout=timeout, api_kwargs=api_kwargs)
+        result = self._post('banChatMember', data, timeout=timeout, api_kwargs=api_kwargs)
 
         return result  # type: ignore[return-value]
 
@@ -3072,6 +3128,38 @@ class Bot(TelegramObject):
                 the connection pool).
             api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
                 Telegram API.
+        Returns:
+            :obj:`int`: Number of members in the chat.
+        Raises:
+            :class:`telegram.error.TelegramError`
+        """
+        warnings.warn(
+            '`bot.get_chat_members_count` is depreciated. '
+            'Use `bot.get_chat_member_count` instead.',
+            TelegramDeprecationWarning,
+            stacklevel=2,
+        )
+        return self.get_chat_member_count(chat_id=chat_id, timeout=timeout, api_kwargs=api_kwargs)
+
+    @log
+    def get_chat_member_count(
+        self,
+        chat_id: Union[str, int],
+        timeout: ODVInput[float] = DEFAULT_NONE,
+        api_kwargs: JSONDict = None,
+    ) -> int:
+        """Use this method to get the number of members in a chat.
+
+         .. versionadded:: 13.7
+
+        Args:
+            chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target chat or username
+                of the target supergroup or channel (in the format ``@channelusername``).
+            timeout (:obj:`int` | :obj:`float`, optional): If this value is specified, use it as
+                the read timeout from the server (instead of the one specified during creation of
+                the connection pool).
+            api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
+                Telegram API.
 
         Returns:
             :obj:`int`: Number of members in the chat.
@@ -3082,7 +3170,7 @@ class Bot(TelegramObject):
         """
         data: JSONDict = {'chat_id': chat_id}
 
-        result = self._post('getChatMembersCount', data, timeout=timeout, api_kwargs=api_kwargs)
+        result = self._post('getChatMemberCount', data, timeout=timeout, api_kwargs=api_kwargs)
 
         return result  # type: ignore[return-value]
 
@@ -5210,6 +5298,8 @@ class Bot(TelegramObject):
     """Alias for :meth:`get_user_profile_photos`"""
     getFile = get_file
     """Alias for :meth:`get_file`"""
+    banChatMember = ban_chat_member
+    """Alias for :meth:`ban_chat_member`"""
     kickChatMember = kick_chat_member
     """Alias for :meth:`kick_chat_member`"""
     unbanChatMember = unban_chat_member
@@ -5242,6 +5332,8 @@ class Bot(TelegramObject):
     """Alias for :meth:`set_chat_sticker_set`"""
     deleteChatStickerSet = delete_chat_sticker_set
     """Alias for :meth:`delete_chat_sticker_set`"""
+    getChatMemberCount = get_chat_member_count
+    """Alias for :meth:`get_chat_member_count`"""
     getChatMembersCount = get_chat_members_count
     """Alias for :meth:`get_chat_members_count`"""
     getWebhookInfo = get_webhook_info
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -57,6 +57,7 @@ from telegram import (
     Animation,
     Audio,
     BotCommand,
+    BotCommandScope,
     Chat,
     ChatMember,
     ChatPermissions,
@@ -2316,6 +2317,7 @@ class Bot(TelegramObject):
         supergroups and channels, the user will not be able to return to the group on their own
         using invite links, etc., unless unbanned first. The bot must be an administrator in the
         chat for this to work and must have the appropriate admin rights.
+
         Args:
             chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target group or username
                 of the target supergroup or channel (in the format ``@channelusername``).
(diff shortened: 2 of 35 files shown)


telegram/botcommandscope.py
```diff
@@ -0,0 +1,261 @@
...
+    CHAT_MEMBER = constants.BOT_COMMAND_SCOPE_CHAT_MEMBER
+    """:const:`telegram.constants.BOT_COMMAND_SCOPE_CHAT_MEMBER`"""
+
+    def __init__(self, type: str, **_kwargs: Any):
+        self.type = type
+        self._id_attrs = (self.type,)
+
+    @classmethod
+    def de_json(cls, data: Optional[JSONDict], bot: 'Bot') -> Optional['BotCommandScope']:
+        """Converts JSON data to a the appropriate :class:`BotCommandScope` object, i.e. takes
```
harshil21, line 83:
```suggestion
        """Converts JSON data to the appropriate :class:`BotCommandScope` object, i.e. takes
```

harshil21 (commented):
scopes look good - one tiny doc fix.

telegram/bot.py
```diff
@@ -2315,8 +2315,64 @@ def kick_chat_member(
...
+                the chat for the user that is being removed. If :obj:`False`, the user will be able
+                to see messages in the group that were sent before the user was removed.
+                Always :obj:`True` for supergroups and channels.
+                .. versionadded:: 13.4
+            api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
+                Telegram API.
+        Returns:
+            :obj:`bool`: On success, :obj:`True` is returned.
+        Raises:
+            :class:`telegram.error.TelegramError`
```
harshil21: can we remove this doc and just link to `ban_chat_member` docs? Also need to add the deprecated tag here

telegram/bot.py
```diff
@@ -3064,6 +3125,38 @@ def get_chat_members_count(
     ) -> int:
         """Use this method to get the number of members in a chat.
 
+        Args:
```
Bibo-Joshi, line 3128: here we can also just link to the new method

telegram/forcereply.py
```diff
@@ -31,31 +31,48 @@ class ForceReply(ReplyMarkup):
     to sacrifice privacy mode.
 
     Objects of this class are comparable in terms of equality. Two objects of this class are
-    considered equal, if their :attr:`selective` is equal.
+    considered equal, if their :attr:`selective`, and :attr:`input_field_placeholder` are equal.
```
Bibo-Joshi, line 34: strictly speaking, this is breaking and for most objects, we only used the optionals for equality comparison, if it's the "exactly one of the optionals" case … 🤔

telegram/forcereply.py
```diff
@@ -31,31 +31,48 @@ class ForceReply(ReplyMarkup):
...
 
     Args:
         selective (:obj:`bool`, optional): Use this parameter if you want to force reply from
             specific users only. Targets:
 
-            1) Users that are @mentioned in the text of the Message object.
-            2) If the bot's message is a reply (has reply_to_message_id), sender of the
-               original message.
+            1) Users that are @mentioned in the :attr:`~telegram.Message.text` of the
```
Bibo-Joshi, line 40:
```suggestion
            1) Users that are @mentioned in the :attr:`telegram.Message.text` of the
```

telegram/replykeyboardmarkup.py
```diff
@@ -48,12 +48,18 @@ class ReplyKeyboardMarkup(ReplyMarkup):
         selective (:obj:`bool`, optional): Use this parameter if you want to show the keyboard to
             specific users only. Targets:
 
-            1) Users that are @mentioned in the text of the Message object.
-            2) If the bot's message is a reply (has ``reply_to_message_id``), sender of the
-               original message.
+            1) Users that are @mentioned in the :attr:`~telegram.Message.text` of the
```
Bibo-Joshi, line 51:
```suggestion
            1) Users that are @mentioned in the :attr:`telegram.Message.text` of the
```

telegram/replykeyboardmarkup.py
```diff
@@ -48,12 +48,18 @@ class ReplyKeyboardMarkup(ReplyMarkup):
...
+               :class:`telegram.Message` object.
+            2) If the bot's message is a reply (has reply_to_message_id), sender of the original
```
Bibo-Joshi, line 53: did you remove the monospace-formatting on purpose?

telegram/forcereply.py
harshil21: that's actually intended - it adds the link to the doc and shows as '[text](https://python-telegram-bot.readthedocs.io/en/stable/telegram.message.html#telegram.Message.text)' instead of '[telegram.Message.text](https://python-telegram-bot.readthedocs.io/en/stable/telegram.message.html#telegram.Message.text)', so the resulting sentence makes little more sense

telegram/forcereply.py
harshil21: ooo will revert

telegram/replykeyboardmarkup.py
harshil21: nope, wanted to link it but then realized that we can't link arguments of methods. adding it back...

telegram/bot.py
```diff
@@ -3064,6 +3125,38 @@ def get_chat_members_count(
...
+                the connection pool).
+            api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
+                Telegram API.
+        Returns:
+            :obj:`int`: Number of members in the chat.
+        Raises:
+            :class:`telegram.error.TelegramError`
+        """
+        warnings.warn(
+            '`bot.get_chat_members_count` is depreciated. '
```
harshil21, line 3142:
```suggestion
            '`bot.get_chat_members_count` is deprecated. '
```
Also apply in `test_bot.py`

telegram/forcereply.py
Bibo-Joshi: ah, good to know, thought it was a typo. never mind, then :)

telegram/bot.py
```diff
@@ -5096,9 +5109,11 @@ def get_my_commands(
 
         result = self._post('getMyCommands', data, timeout=timeout, api_kwargs=api_kwargs)
 
-        self._commands = BotCommand.de_list(result, self)  # type: ignore[assignment,arg-type]
+        if (scope is None or scope.type == 'default') and language_code is None:
```
Bibo-Joshi, line 5112: I added constants for the types, maybe use those :)

telegram/bot.py
```diff
@@ -5152,9 +5167,10 @@ def set_my_commands(
 
         result = self._post('setMyCommands', data, timeout=timeout, api_kwargs=api_kwargs)
 
-        # Set commands. No need to check for outcome.
+        # Set commands only for default scope. No need to check for outcome.
         # If request failed, we won't come this far
-        self._commands = cmds
+        if (scope is None or scope.type == 'default') and language_code is None:
```
Bibo-Joshi, line 5172: same

telegram/bot.py
starry69: Yea, make sense

tests/test_chatmember.py
```diff
@@ -16,112 +16,114 @@
...
+        status = chatmember_class_and_status[1]
+
+        assert cls.de_json({}, bot) is None
+
+        json_dict = {'status': status, 'user': user.to_dict()}
+        chatmember_type = ChatMember.de_json(json_dict, bot)
+
+        assert isinstance(chatmember_type, ChatMember)
+        assert isinstance(chatmember_type, cls)
+        assert chatmember_type.status == status
```
Bibo-Joshi, line 90: I think it would be nice to test at least some of the attributes that are in the subclasses …

telegram/chatmember.py
```diff
@@ -29,7 +29,15 @@
...
-    """This object contains information about one member of a chat.
+    """Base class for Telegram ChatMember Objects..
+    Currently, the following 6 types of chat members are supported:
+
+    * :class:`telegram.ChatMemberOwner`
+    * :class:`telegram.ChatMemberAdministrator`
+    * :class:`telegram.ChatMemberMember`
+    * :class:`telegram.ChatMemberRestricted`
+    * :class:`telegram.ChatMemberLeft`
+    * :class:`telegram.ChatMemberBanned`
```
Bibo-Joshi, line 40: did we want to deprecated the arguments/attributes that are not common to all the subclasses?

telegram/chatmember.py
```diff
@@ -251,3 +268,341 @@ def to_dict(self) -> JSONDict:
...
+            this user.
+        is_anonymous (:obj:`bool`): Optional. :obj:`True`, if the user's
+            presence in the chat is hidden.
+    """
+
+    __slots__ = ()
+
+    def __init__(
+        self,
+        status: str,
```
Bibo-Joshi, line 302: status *must* be `'creator'` here, so I'd vote to skip this argument and just pass the right value to super (like we do for e.g. `InlineQueryResult*`). Same for the other classes.

telegram/chatmember.py
```diff
@@ -251,3 +268,341 @@ def to_dict(self) -> JSONDict:
...
+class ChatMemberOwner(ChatMember):
+    """
+    Represents a chat member that owns the chat
+    and has all administrator privileges.
+
+    .. versionadded:: 13.7
+
+    Args:
+        status (:obj:`str`): The member's status in the chat,
+            always “creator”.
```
Bibo-Joshi, line 282: maybe link the constant here?

telegram/chatmember.py
```diff
@@ -242,7 +250,16 @@ def de_json(cls, data: Optional[JSONDict], bot: 'Bot') -> Optional['ChatMember']
         data['user'] = User.de_json(data.get('user'), bot)
         data['until_date'] = from_timestamp(data.get('until_date', None))
 
-        return cls(**data)
+        _class_mapping: Dict[str, Type['ChatMember']] = {
```
Bibo-Joshi, line 253: I know that this setup comes from me, but at second thought it needs fine tuning: Because the subclasses inherit this behavior, currently `ChatMemberOwner.de_json` does not always return a `ChatMemberOwner` instance, which is probably unexpected. So we could either override in the subclasses or just insert an `if cls is ChatMember:` here 🤔 same for `BotCommandScope` …

The docstring could also be updated like for BotCommandScope as we're doing a bit more magic here :)

telegram/chatmember.py
```diff
@@ -29,7 +29,15 @@
 
 
 class ChatMember(TelegramObject):
-    """This object contains information about one member of a chat.
+    """Base class for Telegram ChatMember Objects..
```
harshil21, line 32:
```suggestion
    """Base class for Telegram ChatMember Objects.
```
Had to manually type this out since github removed suggestions?

tests/test_chatmember.py
```diff
@@ -16,112 +16,114 @@
...
-        d = ChatMember(User(2, '', False), ChatMember.ADMINISTRATOR)
-        d2 = ChatMember(User(1, '', False), ChatMember.CREATOR)
+    def test_slot_behaviour(self, chatmember_types, mro_slots):
+        for attr in chatmember_types.__slots__:
+            assert getattr(chatmember_types, attr, 'err') != 'err', f"got extra slot '{attr}'"
+        assert len(mro_slots(chatmember_types)) == len(
+            set(mro_slots(chatmember_types))
+        ), "duplicate slot"
+        with pytest.raises(AttributeError):
+            chatmember_types.custom
```
harshil21, line 77: imo we should still test if setting an attribute defined by the class doesn't raise a warning, i.e, the last 2 deleted lines from this test 

``` python
chatmember_types.custom, chatmember_types.status = 'warning!', self.status
assert len(recwarn) == 1 and 'custom' in str(recwarn[0].message), recwarn.list
```

telegram/chatmember.py
```diff
@@ -251,3 +268,341 @@ def to_dict(self) -> JSONDict:
...
+    and has all administrator privileges.
+
+    .. versionadded:: 13.7
+
+    Args:
+        status (:obj:`str`): The member's status in the chat,
+            always “creator”.
+        user (:class:`telegram.User`): Information about the user.
+        custom_title (:obj:`str`, optional): Custom title for this user.
+        is_anonymous (:obj:`bool`, optional):  :obj:`True`, if the
```
harshil21, line 285:
``` suggestion
        is_anonymous (:obj:`bool`, optional): :obj:`True`, if the
```

docs/source/telegram.chatmemberadministrator.rst
```diff
@@ -0,0 +1,8 @@
+:github_url: https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/chatmember.py
+
+telegram.ChatMemberAdministrator
+===================
```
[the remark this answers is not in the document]
starry69: Oh, right

telegram/chatmember.py
starry69: Only `status` and `user` are common in all subclasses, so you mean depreciate all other attributes from base class? If yes then how do we depreciate ig only way is by overriding `__getattr__()` and checking if user is trying to acess some other attribute than (`user`, `status`) then show the warning that these attributes are deprecated, and will be removed in future 🤔

telegram/chatmember.py
starry69: I would prefer second approach, as it is much cleaner to have just one `if` check in base class than overriding same method in all subclasses..

tests/test_chatmember.py
starry69: Okay.

telegram/chatmember.py
Bibo-Joshi: Mh, I guess I'd also be fine with just stating it in the docs …

starry69: And if we override `__getattr__()` doesn't it'll apply to it's subclasses too..

starry69: > Mh, I guess I'd also be fine with just stating it in the docs …

Hmmm okay, that would be better imo

telegram/chatmember.py
Bibo-Joshi: > Had to manually type this out since github removed suggestions?

works fine for me …

harshil21: ah turns out the suggestion button doesn't show up if you view changes from one commit. Which makes sense, but should show up for the latest commit ig

telegram/chatmember.py
```diff
@@ -48,102 +48,208 @@ class ChatMember(TelegramObject):
...
+
+            .. deprecated:: 13.7
+
         can_manage_chat (:obj:`bool`, optional): Administrators only. :obj:`True`, if the
             administrator can access the chat event log, chat statistics, message statistics in
             channels, see channel members, see anonymous administrators in supergroups and ignore
             slow mode. Implied by any other administrator privilege.
 
-            .. versionadded:: 13.4
+            .. deprecated:: 13.7
```
Bibo-Joshi, line 74: I think we can just keep both directives :)

tests/test_chatmember.py
```diff
@@ -17,50 +17,92 @@
...
+            set(mro_slots(chatmember_types))
+        ), "duplicate slot"
+        chatmember_types.custom, chatmember_types.status = 'warning!', chatmember_types.status
         assert len(recwarn) == 1 and 'custom' in str(recwarn[0].message), recwarn.list
+        with pytest.raises(AttributeError):
+            chatmember_types.custom2
 
-    def test_de_json_required_args(self, bot, user):
-        json_dict = {'user': user.to_dict(), 'status': self.status}
+    def test_de_json_required_args(self, bot, chatmember_class_and_status, user):
```
Bibo-Joshi, line 83: note to self: add a test making sure that `ChatMemberSubclass.de_json` only ever returns a `ChatMemberSubclass` object even when passed kwargs that suit a different subclass. same for botcommandscope

Bibo-Joshi:
Note about slots: I had remove the test lines that @harshil21 pointed out, b/c we are adding new classes and I thought that we don't need to have `__dict__` for them. But that's already added in the `TelegramObject` base class and it's more trouble to try & remove that for the new classes than to just have the deprecation warning raised :)

telegram/chatmember.py
```diff
@@ -29,113 +29,239 @@
...
+    * :class:`telegram.ChatMemberMember`
+    * :class:`telegram.ChatMemberRestricted`
+    * :class:`telegram.ChatMemberLeft`
+    * :class:`telegram.ChatMemberBanned`
 
     Objects of this class are comparable in terms of equality. Two objects of this class are
     considered equal, if their :attr:`user` and :attr:`status` are equal.
 
+    Note:
+        As of Bot API 5.3, :class:`ChatMember` is nothing bot the base class for the subclasses
```
harshil21, line 46:
```suggestion
        As of Bot API 5.3, :class:`ChatMember` is nothing but the base class for the subclasses
```
but bot

PR #2574, opened 2021-06-26
add stacklevel parameter to `set_new_attribute_deprecated`

harshil21:
Adds the `stacklevel` parameter and details the warning message to `set_new_attribute_deprecated` so users can pinpoint where the warning originates from.

In response to https://t.me/pythontelegrambotgroup/491855


Bibo-Joshi:
Ran a small test and noticed that it says e.g. "objects such as 'User'" instead of "objects such as 'telegram.User'" and for filters it will report name of the private class we use. I can live with both given that this warning will only be removed eventually and the latter is quite the edge case imo. Just wanted to note this before we go for merging … :)

Bibo-Joshi (approved):

Issue #2556, opened 2021-06-09
[DISCUSSION] Handling of inits


Bibo-Joshi:
# Motivation

`BasePersistence` and `Updater` currently accept a lot of parameters, which can become somewhat cluttered. For `BP`, this could get worse after #2524.

For `Updater`, we mostly allow all those arguments as convenience shortcuts so that users don't have to pass that stuff to `Dispatcher` and `Bot` on their own. However, I have the impression that it also leads to users not understanding the difference between the three classes `Updater`, `Dispatcher` and `Bot`. Ofc, to just use them, you don't have to understand them in detail, but it also quickly leads to questions like "how do I send a message without getting an update"? because people don't know/remember that you can instantiate a `Bot` without an `Updater`.

I'd like to think about refactoring how we handle those.

# `BasePersistence`

In this case it's probably rather easy: Instead of passing `store_bot_data`, `store_user_data` etc, we can just pass a tuple or - to be more user friendly - a namedtuple:

```
class PersistenceInput(typing.NamedTuple):
    bot_data: bool = True
    chat_data: bool = True
    ...
```

This can even be reused for new parameters that #2524 proposes and *could* be done non-breaking by allowing ways as mutually exclusive variants in a deprecation phase.

# `Updater`, `Dispatcher`, `Bot`

Ultimately, I'd like to have a pattern that is convenient for simple use cases but still separaters the three classes. 

## First idea

Pretty simple & straight forward: instead of passing all the args one by one to `Updater`, pass `bot_kwargs` & `dispatcher_kwargs` or namedtuples like above.

➕  Could be done non-breaking.
➖ even if you only want to pass the token, you'd have to pass `bot_params=BotParams(token=token)` or so

## Second idea

1. Make all of `Updater`, `Dispatcher` and `Bot` only accept what they actually need. So no changes for `Dispatcher` and `Bot`, bot `Updater` would change to `def __init__(self, dispatcher):`
2. Make `update_queue` optional for `Dispatcher` and have it be build on init as `self.update_queue = update_queue or Queue()`
3. Make `job_queue` such that by default a new `JobQueue` will be created, i.e. it will be opt-out instead of opt-in. For people who don't want to use threading & the `JobQueue`, this shouldn't be too bad, b/c the JQ only starts when we call `Dispatcher.start`. If they still want it to be `None`, they can pass `None`.
[...]

Issue #2524, opened 2021-05-15
[FEATURE/DISCUSSION]: Allow fine tuning of persistence setup


Bibo-Joshi:
There are two places in the current persistence setup where I think we can offer more customization for advanced users. All of this are just first ideas and up for discussion. If the dev team doesn't like the (use case value)/(maintenance workload) ratio, I'd be fine with not doing this.

1. Edit: We're dropping `insert/replace_data` see #2882 . 

    Original comment: 
We currently pass all data that goes through the persistence setup through `replace/insert_bot`. While this is the safest way to make sure no `Bot` instance gets serialized, it can also introduce some overhead which could get noticable when you have large, nested data structures in your `chat/bot/user_data` or are handling many updates. Adanced users will know where in their data they have bot instances that should get replaced. So we could offer settings like `replace_bots_in_chat/bot/user_data` (defaulting to True) and only call `replace/insert_data` when necessary.

     One should note that we'd have to re-indroduce deepcopy into the built-in persistence classes (depending on the `replace_bots_in_*` settings, that is). See https://github.com/python-telegram-bot/python-telegram-bot/pull/2262#discussion_r632762276

2. #2285 already made an effort to reduce the calls of `Dispatcher.update_persistence`. Still, updating after every update (or even handler call in case you're using `run_async`) might be unecessary, especially when you're getting updates frequently. Similarly to `PicklePersistence.on_flush` which allows users to control how often `PP` writes to file, we would add a flag telling `Dispatcher` to never automatically call `Dispatcher.update_persistence` at runtime (only on shutdown). Then users can call `Dispatcher.update_persistence` manually to their liking.

Poolitzer:
I like both of these suggestions. Since we redesign a lot for 14 anyway, and the resulting maintenance after doing it once shouldn't be too bad,

Paddy613:
****

Bibo-Joshi:
Got started on the first part of this in the https://github.com/python-telegram-bot/python-telegram-bot/tree/customize-replace-insert-bot branch. But we're currently re-evaluating if we want to proceed this.

second part may be independent.

Bibo-Joshi:
closed by #2893

PR #2604, opened 2021-07-24
Refactor init of Persistence

Bibo-Joshi:
First part of #2556

### Checklist for PRs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,9 +3,11 @@ on:
   pull_request:
     branches:
       - master
+      - v14
   push:
     branches: 
       - master
+      - v14
 
 jobs:
   pytest:
--- a/telegram/ext/basepersistence.py
+++ b/telegram/ext/basepersistence.py
@@ -21,7 +21,7 @@ import warnings
 from sys import version_info as py_ver
 from abc import ABC, abstractmethod
 from copy import copy
-from typing import Dict, Optional, Tuple, cast, ClassVar, Generic, DefaultDict
+from typing import Dict, Optional, Tuple, cast, ClassVar, Generic, DefaultDict, NamedTuple
 
 from telegram.utils.deprecate import set_new_attribute_deprecated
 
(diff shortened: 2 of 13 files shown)


.github/workflows/test.yml
```diff
@@ -3,9 +3,11 @@ on:
   pull_request:
     branches:
       - master
+      - v14
```
Poolitzer, line 6: this file (or rather the commit 3f2645f) shouldnt be in this branch right? Its in V14 already after all

telegram/ext/basepersistence.py
```diff
@@ -31,6 +31,33 @@
 from telegram.ext.utils.types import UD, CD, BD, ConversationDict, CDCData
 
 
+class PersistenceInput(NamedTuple):
+    """Convenience wrapper to group boolean input for :class:`BasePersistence`.
+
+    Args:
+        bot_data (:obj:`bool`, optional): Whether the setting should be applied for ``bot_data``.
```
Poolitzer, line 38: I dont think I like the sentence "Whether the setting". Why not call the thing by its name? "Whether this data (or data class?) will be saved".

Poolitzer (changes requested):
Good changes, only doc string suggestions

telegram/ext/basepersistence.py
Bibo-Joshi: Because this class can be used for other settings, too. Specifically for the first part of #2524 …

Poolitzer: I still dont get why you have to give the arguments such unspecific strings instead of specific ones? Other settings can get their own specific strings after all?

Bibo-Joshi: What I mean is with #2524, BPs init would be something like 

```python
def __init__(store_data: PersistenceInput = None, replace_bot: PersistenceInput = None):
```

where `replace_bot.bot_data` would tell you if `Bot` instances in `bot_data` should be replaced by `BP` before serialization or not. Now a doctsring `"Whether this data (or data class?) will be saved"` wouldn't make sense for `PersistenceInput.bot_data`.

Poolitzer: Alright, now I understand what you mean. I dont like it, but it's okay.

harshil21 (approved):
looks very good. Just one doubt.

telegram/ext/basepersistence.py
```diff
@@ -68,46 +95,28 @@ class BasePersistence(Generic[UD, CD, BD], ABC):
...
-            persistence class.
-        store_chat_data (:obj:`bool`): Optional. Whether chat_data should be saved by this
-            persistence class.
-        store_bot_data (:obj:`bool`): Optional. Whether bot_data should be saved by this
-            persistence class.
-        store_callback_data (:obj:`bool`): Optional. Whether callback_data should be saved by this
-            persistence class.
-
-            .. versionadded:: 13.6
+        store_data (:class:`PersistenceInput`): Specifies which kinds of data will be saved by this
```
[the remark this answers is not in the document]
Bibo-Joshi: yeah, that's basically why I dropped it - if you don't pass `store_data` on init, we'll just create an instance on our own …

PR #2619, opened 2021-08-08
Remove `__dict__` from `__slots__`

harshil21:
Notes about the changes (for further discussion + ease of review):

- `BasePersistence` needs `__dict__` (because of the replacement of methods we do in `__new__`), `CallbackContext` by design has `__dict__`, should any other classes have it too?
- `base.py` has `_id_attrs` as class variable. Should we consider setting this attribute in  `__new__`? If done, we can reduce some code duplication and fix the pycharm warning. Edit: Done.
- drops py3.6 related slots code, since we're dropping py3.6 support anyway.
- removed `_id_keys` in `document.py`, apparently its never been used.

Notes about tests:
- removed missing slot + warning about custom attribute tests - since python/pre-commit will now raise errors for both cases.
- `test_slots.py` now checks that each class _doesn't_ have a `__dict__`.
- subclassed `Dispatcher` in tests so we can add custom attributes, but a test failed `test_dispatcher.py/test_multiple_async_decorator` - didn't raise the error, so dispatcher has `__dict__`.
- fixed a test (`tests/test_inlinequeryhandler.py/test_chat_types`)
- modified `test_updater/test_webhook`, `test_conversationhandler.py/test_schedule_job_exception` to use `DictBot` and `DictJobQueue` to monkeypatch on it.

Also I try to group similar changes into one commit, so review by commit than viewing everything at once


harshil21:
py 3.6 related tests fail - as expected.

tests/conftest.py
```diff
@@ -89,14 +91,21 @@ def bot_info():
     return get_bot()
 
 
+class DictRequest(Request):  # Used for monkey patching attributes
```
Bibo-Joshi, line 94: Maybe just add a comment in front of all 3 classes explaning that they're here for monkey patching due to slots and `__dict__`

tests/conftest.py
```diff
@@ -149,6 +158,8 @@ def provider_token(bot_info):
 def create_dp(bot):
     # Dispatcher is heavy to init (due to many threads and such) so we have a single session
     # scoped one here, but before each test, reset it (dp fixture below)
+    # class DictDispatcher(Dispatcher):
+    #     pass
```
Bibo-Joshi, line 162: please remove

Bibo-Joshi (changes requested):
> `base.py` has `_id_attrs `as class variable. Should we consider setting this attribute in `__new__`? If done, it will also print the correct info when (`some_tg_object_class_instance.__class__._id_attrs`) is printed.

I don't understand the second part. Isn't `some_tg_object_class_instance.__class__._id_attrs` a class attribute as it's now?
Anyway, I agree that it would be cleaner to have `_id_attr` as instance attribute. We can either go with `__new__` or give `TelegramObject` an `__init__` and call it in all the subclasses. I feel like the second one would be somewhat cleaner, as AFAIK `__new__` isn't really there for initialization. In fact with `__new__`, we should double check if it makes any trouble with pickle 🤔 And calling `super().__init__()` is mostly a one time effort to add … OTOH I see that we already did this for filters … What do you think?


apart from this, I only have a few minor remarks (see below). Obviously I didn't check all the tests :D

harshil21:
> I don't understand the second part. Isn't `some_tg_object_class_instance.__class__._id_attrs` a class attribute as it's now?

Okay nvm that, I wanted to dig deeper into that cause pycharm now apparently flags that as a warning (saying  `_id_attrs` is read-only and that it has no attribute `_id_attrs` which doesn't make sense)

> OTOH I see that we already did this for filters …

No? In fact `BaseFilter` uses `__new__` to assign instance attributes, like what we can do here.

> What do you think?

I guess we don't really gain or lose anything out from doing this so I think its probably better to leave it as it is

telegram/ext/basepersistence.py
```diff
@@ -160,7 +147,7 @@ def update_callback_data_replace_bot(data: CDCData) -> None:
             obj_data, queue = data
             return update_callback_data((instance.replace_bot(obj_data), queue))
 
-        # We want to ignore TGDeprecation warnings so we use obj.__setattr__. Adds to __dict__
+        # Adds to __dict__
```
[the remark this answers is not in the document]
harshil21: yep, will change

tests/conftest.py
harshil21: I left that cause 
> subclassed Dispatcher in tests so we can add custom attributes, but a test failed `test_dispatcher.py/test_multiple_async_decorator` - didn't raise the error, so dispatcher has `__dict__` (If I can get some help on that test, then Dispatcher doesn't need a `__dict__`)

Bibo-Joshi: but these two lines are commented now, so they don't have any effect?

harshil21: Yea I'm hoping to get rid of `__dict__` in dispatcher soon, so after I can just uncomment that code

Bibo-Joshi:
> > OTOH I see that we already did this for filters …
> 
> No? In fact `BaseFilter` uses `__new__` to assign instance attributes, like what we can do here.

yes, filters uses `__new__`, that's what I meant :D
 
> > What do you think?
> 
> I guess we don't really gain or lose anything out from doing this so I think its probably better to leave it as it is

with `__new__` we can drop `_id_attrs` from the slots of all the subclasses (by instead just declaring that slot for `TelegramObject`) and we make it a proper instance attribute, which has at least some code quality value IMO …

harshil21:
Okay, moved the `_id_attrs` to `__new__`. Apparently type hints can't be applied on instances in `__new__` so had to type hint `_id_attrs` in other classes (wherever mypy complained).

telegram/base.py
```diff
@@ -37,22 +36,21 @@
...
-
     # Adding slots reduces memory usage & allows for faster attribute access.
     # Only instance variables should be added to __slots__.
-    # We add __dict__ here for backward compatibility & also to avoid repetition for subclasses.
-    __slots__ = ('__dict__',)
+    __slots__ = ('_id_attrs',)
+
+    def __new__(cls, *args: object, **kwargs: object) -> 'TelegramObject':  # pylint: disable=W0613
+        instance = super().__new__(cls)
+        instance._id_attrs = ()  # cannot add type hint since it is not read
```
Bibo-Joshi, line 45: maybe rather "since mypy doesn't support __new__ yet"? I think it would also be good to add a comment on why we do this here in the first place (some for filters)

harshil21: >  "since mypy doesn't support __new__ yet"?

Actually it turns out it's a python thing

> it would also be good to add a comment on why we do this here in the first place (some for filters)

👌

.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
```
Bibo-Joshi, line 18: we can also change two code-quality settings now:

* for pyupgrade to --py37-plus in the pre-commit yaml
* for black the target version to `['py37']` in pyproject.toml

telegram/base.py
```diff
@@ -36,13 +36,20 @@
 class TelegramObject:
     """Base class for most Telegram objects."""
 
+    # type hints in __new__ are not read by mypy (https://github.com/python/mypy/issues/1021). As a
+    # workaround we can type hint instance variables in __new__ using a syntax defined in PEP 526 -
+    # https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations
+    if TYPE_CHECKING:
+        _id_attrs: Tuple[object, ...]
```
Bibo-Joshi, line 43: Cool finding! 🦾

telegram/base.py
```diff
@@ -36,13 +36,20 @@
...
     # Adding slots reduces memory usage & allows for faster attribute access.
     # Only instance variables should be added to __slots__.
     __slots__ = ('_id_attrs',)
 
     def __new__(cls, *args: object, **kwargs: object) -> 'TelegramObject':  # pylint: disable=W0613
+        # We add _id_attrs in __new__ instead of __init__ since we want to add this to the slots
+        # w/o calling __init__ in all of the subclasses. This is what we also do in BaseFilter.
```
Bibo-Joshi, line 50: having that explanation here won't help you very much while reading the source of `BaseFilter`, though 😄 😋 please add the note the as well - in fact the workaround for type checking may also be applicable?

harshil21: > in fact the workaround for type checking may also be applicable?

For `BaseFilter`? Yeah but we don't have any issues with type hints for it? Is it needed?

Bibo-Joshi: Don't know … was just a thought since we have pretty much the same situation there …

Bibo-Joshi: Just for completeness: I think the difference is that for filters we have the `name` setter + getter, so the child classes don't try to access `_name` directly

PR #2620, opened 2021-08-09
make `BP.flush()` abstract

### LINKED ISSUE #2231: Make BP.flush an abstract method
Don't know why we skipped that in #1905, but as `BasePersistence` is a interface, it makes sense for all methods to be abcs. As `flush` currently doesn't raise `NotImplementedError`, this is strictly speaking breaking, so putting this on the v14 milestone.

harshil21:
Closes #2231


harshil21:
closing in favor of #2624

Issue #2468, opened 2021-04-13
Clear up imports policy


Bibo-Joshi:
We're currently not very consistent regarding what we make available directly via `telegram.__init__` and what we don't. Some examples:

* from `telegram.error` only `TelegramError` is available via `telegram.__init__`, but no of the other classes
* from `telegram.constants` only some few constants are available.
* `TelegramDeprecationWarning` is not available (see also #2467)
* the helper methods from `utils.helper` that are actually there to be used by users and not exclusively by PTB itself (e.g. `escape_markdown`, `mention_user_*`) are not available

IMO we should try to make a clear policy for this. IMHO it should be clear that all API-classes are directly available through `telgeram`. All the things mentioned above are from … lets call them "auxiliary modules". For these I basically see two options:

1. Make all of that available directly through `telegram`
2. Make none of it available through `telegram`. In this case I'd vote to at least move the above mentioned helper methods directly to `telegram.utils` and spare the user the `.helpers`, especially as users are in general not to be concerned with the other contents of `utils`.

Case 2 would be breaking & in that case we should wait until v14.
In any case I would aim to document the import policy somewhere … probably just at the index of the docs.
I'm slightly in favor of 2. What do the other @python-telegram-bot/developers think?

Poolitzer:
I don't like 1, because that would mean loosing all the tidiness we have right now, or at least the attempt for this. I would really like 2 though, we should make a new class for the helpers imo, and provide there all the methods we intend for users from the helpers :D

Bibo-Joshi:
With Noam also in favor of 2, I'll add the 💥 tag and postpone to v14

Poolitzer:
Random thought, maybe we can do like `from python-telegram-bot import ***`? This has two tremendous upsides: No namespace issues anymore with other libraries/user files. And three clear subcategories: `python-telegram-bot.telegram` for all pure API stuff, `python-telegram-bot.ext` for all ext stuff we did, and `python-telegram-bot.utils` for utils we provide.

Bibo-Joshi:
Interesting ideas! In fact "project name = name of the only package provided by the project" is strongly encouraged by [PEP 423](https://www.python.org/dev/peps/pep-0423/#id87).
However, hyphens in package names give syntax error (you'd have to do smth like `__import__('python-telgeram-bot')` and while underscores would be possible, [PEP 8](https://www.python.org/dev/peps/pep-0008/#package-and-module-names) discourages using them.

[...]

PR #2621, opened 2021-08-09
move telegramdecryptionerror to error.py

harshil21:
The import policy (#2468) isn't decided for errors. So keeping it as it is (defined in `__all__`)

### DIFF
--- a/telegram/error.py
+++ b/telegram/error.py
@@ -18,7 +18,7 @@
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 # pylint: disable=C0115
 """This module contains an object that represents Telegram errors."""
-from typing import Tuple
+from typing import Tuple, Union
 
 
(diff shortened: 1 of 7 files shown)


telegram/error.py
```diff
@@ -149,3 +149,16 @@ class Conflict(TelegramError):
 
     def __reduce__(self) -> Tuple[type, Tuple[str]]:
         return self.__class__, (self.message,)
+
+
+class TelegramDecryptionError(TelegramError):
```
Bibo-Joshi, line 154: While at it, We might want to rename the error - none of the others start with `Telegram`. Suggestions: Just `DecryptionError` or `PassportDecryptionError`

harshil21: PassportDecryptionError sounds more PTB specific, so I'll change it to that.

Poolitzer (approved):
Code changes look good to me!

PR #2628, opened 2021-08-11
Feat: Custom pytest marker

### LINKED ISSUE #2613: [FEATURE] Add custom pytest mark to conv file
I would like to mark new tests I develope with a custom mark, so I can select them with the pytest command. This results in the following error however: `pytest.PytestUnknownMarkWarning: Unknown pytest.mark.XXX- is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/mark.html`

I would like to have a custom mark added to the conf file so I can use it.

I can alternatively use -k to select the tests I want but that is more cumbersome in my opinion.

Poolitzer:
This PR adds a marker so one can run tests with it.

Closes #2613.

### DIFF
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -105,6 +105,8 @@ Here's how to make a one-off code change.
 
    - Please ensure that the code you write is well-tested.
 
+        - In addition to that, we provide the `dev` marker for pytest. If you write one or multiple tests and only want to run those, you can decorate them via `@pytest.mark.dev` and then run it with minimal overhead with `pytest ./path/to/test_file.py -m dev`.
+
    - Don’t break backward compatibility.
 
    - Add yourself to the AUTHORS.rst_ file in an alphabetical fashion.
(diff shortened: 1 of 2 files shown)


.github/CONTRIBUTING.rst
```diff
@@ -105,6 +105,8 @@ Here's how to make a one-off code change.
 
    - Please ensure that the code you write is well-tested.
 
+        - In addition to that, we provide the `dev` marker for pytest. If you write one or multiple tests and only want to run those, you can decorate them via `@pytest.mark.dev` and then run it with minimal overhead with `pytest ./path/to/test_file.py -m dev`.
```
harshil21, line 108: > `pytest ./path/to/test_file.py -m dev`

We don't need to specify the path, just the marker itself is fine

Poolitzer: yes, thats why I wrote minimal overhead. Specifying the file will decrease the amount of unnecessary tests pytest collects, which can be a lot in this project.

PR #2629, opened 2021-08-11
Fix readme: Installation with extra [ujson] should be [json].

Behoston:
Just a readme fix.

```
WARNING: python-telegram-bot 13.7 does not provide the extra 'ujson'
```

https://github.com/python-telegram-bot/python-telegram-bot/blob/master/setup.py#L85

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -144,7 +144,7 @@ Optional Dependencies
 PTB can be installed with optional dependencies:
 
 * ``pip install python-telegram-bot[passport]`` installs the `cryptography <https://cryptography.io>`_ library. Use this, if you want to use Telegram Passport related functionality.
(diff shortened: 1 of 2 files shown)


harshil21:
Thanks for the spot, I think it makes more sense to change it to `ujson` in the `setup.py`

Behoston:
agree, `[ujson]` is much clearer, but you need to extend existing setup to support both to backward compatibility. I'm not sure if extend will be ok, so I just fix readme to match code.

Bibo-Joshi:
that's good thinking @Behoston ! however, we're by now working towards v14, which will break a lot of things anyway. So that wouldn't necessarily be a no-go. Anyway, while I somewhat like the idea to rename `json` to `ujson`, the former makes it easier for us to switch to another json-library if we see fit at some point. The other optional requirements `socks` and `passport` are also more a description of what the requirements are for, not which lib would be installed.  So I think I'm slightly more in favor of adjusting the readme as you currently did.

Bibo-Joshi:
Thanks for the contribution!

PR #2646, opened 2021-08-31
Builder pattern for ext

### LINKED ISSUE #2556: [DISCUSSION] Handling of inits

Bibo-Joshi:
Closes #2556 

Note that I did not add an `ExtBotBuilder` class. Currently only the `request` argument of `ExtBot` would profit from with. Then again, the init of `ExtBot` may become more complicated in the future if we add more features. If we want to add a corresponding class, we need to think about if we still want to allow for manual instantiation of `ExtBot` or not …

### Checklist for PRs


telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+        self: BuilderType, updater_class: Type[Updater], kwargs: Dict[str, object] = None
+    ) -> BuilderType:
+        self._updater_class = updater_class
+        self._updater_kwargs = kwargs or {}
+        self._updater_class_was_set = True
+        return self
+
+    @check_if_already_set
+    def _set_token(self: BuilderType, token: str) -> BuilderType:
+        if self._bot_was_set:
```
tsnoam, line 358: Should we assert that `self._request` wasn't set as well?

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+            raise self._exception_builder('token', 'bot instance')
+        if self._dispatcher_check:
+            raise self._exception_builder('token', 'Dispatcher instance')
+        self._token = token
+        self._token_was_set = True
+        return self
+
+    @check_if_already_set
+    def _set_base_url(self: BuilderType, base_url: str) -> BuilderType:
+        if self._bot_was_set:
```
tsnoam, line 368: Should we assert that `self._request` wasn't set as well?

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+            raise self._exception_builder('base_url', 'bot instance')
+        if self._dispatcher_check:
+            raise self._exception_builder('base_url', 'Dispatcher instance')
+        self._base_url = base_url
+        self._base_url_was_set = True
+        return self
+
+    @check_if_already_set
+    def _set_base_file_url(self: BuilderType, base_file_url: str) -> BuilderType:
+        if self._bot_was_set:
```
tsnoam, line 378: Should we assert that `self._request` wasn't set as well?

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+            raise self._exception_builder('request', 'bot instance')
+        if self._dispatcher_check:
+            raise self._exception_builder('request', 'Dispatcher instance')
+        self._request = request
+        self._request_was_set = True
+        return self
+
+    @check_if_already_set
+    def _set_private_key(self: BuilderType, private_key: bytes) -> BuilderType:
+        if self._bot_was_set:
```
tsnoam, line 412: Should we assert that `self._request` wasn't set as well?

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+            raise self._exception_builder('private_key', 'bot instance')
+        if self._dispatcher_check:
+            raise self._exception_builder('private_key', 'Dispatcher instance')
+        self._private_key = private_key
+        self._private_key_was_set = True
+        return self
+
+    @check_if_already_set
+    def _set_private_key_password(self: BuilderType, private_key_password: bytes) -> BuilderType:
+        if self._bot_was_set:
```
tsnoam, line 422: Should we assert that `self._request` wasn't set as well?

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+    @check_if_already_set
+    def _set_update_queue(self: BuilderType, update_queue: Queue) -> BuilderType:
+        if self._dispatcher_check:
+            raise self._exception_builder('update_queue', 'Dispatcher instance')
+        self._update_queue = update_queue
+        self._update_queue_was_set = True
+        return self
+
+    @check_if_already_set
+    def _set_workers(self: BuilderType, workers: int) -> BuilderType:
```
tsnoam, line 478: In Rust there is `NonZeroU16`. I wonder if there is some equivalent in python (instead of `int`).

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+
+            dispatcher = DispatcherBuilder().token('TOKEN').build()
+
+    Please see the description of the individual methods for information on which arguments can be
+    set and what the defaults are when not called. When no default is mentioned, the argument will
+    not be used by default.
+
+    Note:
+        * Each method can be called at most once, e.g. you can't override arguments that were
+          already set.
```
tsnoam, line 578: Reminder: if we decide to remove this restriction, remove this note as well.

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+
+        .. seealso:: :meth:`request`
+
+        Args:
+            request_kwargs (Dict[:obj:`str`, :obj:`object`]): The keyword arguments.
+
+        Returns:
+            :class:`DispatcherBuilder`: The same builder with the updated argument.
+        """
+        return self._set_request_kwargs(request_kwargs)
```
tsnoam, line 685: codecov says that this line isn't covered.

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+
+        Args:
+            request (:class:`telegram.utils.Request`): The request object.
+
+        Returns:
+            :class:`DispatcherBuilder`: The same builder with the updated argument.
+        """
+        return self._set_request(request)
+
+    def private_key(self: BuilderType, private_key: bytes) -> BuilderType:
```
tsnoam, line 701: I wonder if `private_key` and `private_key_password` should be merged into a single setter.
After all, these two arguments will always be used together, they are checked together, etc. etc. (we might also want to fix the internal implementation as well).

In a sense, this is similar to how we allow setting a custom Dispatcher class, and taking kwargs as well.

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+
+            dispatcher = UpdaterBuilder().token('TOKEN').build()
+
+    Please see the description of the individual methods for information on which arguments can be
+    set and what the defaults are when not called. When no default is mentioned, the argument will
+    not be used by default.
+
+    Note:
+        * Each method can be called at most once, e.g. you can't override arguments that were
+          already set.
```
tsnoam, line 927: Reminder: if we decide to remove this restriction, remove this note as well.

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+        Args:
+            context_types (:class:`telegram.ext.ContextTypes`, optional): The context types.
+
+        Returns:
+            :class:`DispatcherBuilder`: The same builder with the updated argument.
+        """
+        return self._set_context_types(context_types)  # type: ignore[return-value]
+
+
+class UpdaterBuilder(_BaseBuilder[ODT, BT, CCT, UD, CD, BD, JQ, PT]):
```
tsnoam, line 908: Maybe inherit from `DispatcherBuilder`?

After all, all the setters of DispatcherBuilder are used by UpdaterBuilder

tests/conftest.py
```diff
@@ -106,7 +106,9 @@ class DictBot(Bot):
 
 @pytest.fixture(scope='session')
 def bot(bot_info):
```
tsnoam, line 108: Maybe use the builder here as well?

tsnoam (changes requested):

JosXa:
Why is it `DefaultContextType`, not just `DefaultContext`?

examples/conversationbot.py
```diff
@@ -18,25 +18,25 @@
...
-    Updater,
     CommandHandler,
     MessageHandler,
     Filters,
     ConversationHandler,
-    CallbackContext,
+    UpdaterBuilder,
 )
 
+from telegram.ext.utils.types import DefaultContextType
```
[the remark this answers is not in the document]
Bibo-Joshi: Will change

examples/passportbot.py
```diff
@@ -102,7 +104,7 @@ def main() -> None:
     """Start the bot."""
     # Create the Updater and pass it your token and private key
     with open('private.key', 'rb') as private_key:
-        updater = Updater("TOKEN", private_key=private_key.read())
+        updater = UpdaterBuilder().token("TOKEN").private_key(private_key.read()).build()
```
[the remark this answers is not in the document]
Bibo-Joshi: how much is `pathlib` considered "best practice" or "standard" by now? Question is: Does this acutally make the code more readible or does it just confuse newbies?

examples/paymentbot.py
```diff
@@ -130,7 +130,7 @@ def successful_payment_callback(update: Update, context: CallbackContext) -> Non
 def main() -> None:
     """Run the bot."""
     # Create the Updater and pass it your bot's token.
-    updater = Updater("TOKEN")
+    updater = UpdaterBuilder().token("TOKEN").build()
```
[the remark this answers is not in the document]
Bibo-Joshi: done

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+        DefCCT,
+        Dict,
+        Dict,
+        Dict,
+        JobQueue,
+        None,
+    ]
+
+
+def check_if_already_set(func: CT) -> CT:
```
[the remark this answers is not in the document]
Bibo-Joshi: Overriding just doesn't make any sense at all IMO, so from my pov it made sense to include this check in the flow that makes sure that the input can be reasonably be handled. But it would save same LOC …

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+            user_signal_handler=self._user_signal_handler,
+            exception_event=self._dispatcher.exception_event
+            if self._dispatcher
+            else self._exception_event,
+            builder_flag=True,
+            **self._updater_kwargs,
+        )
+
+    @staticmethod
+    def _exception_builder(arg_1: str, arg_2: str = None) -> RuntimeError:
```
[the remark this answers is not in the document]
Bibo-Joshi: 1. yeah, maybe. I guess I wanted to keep the number of helpers low :D 
2. true. though you can search for uses of `_exception_builder` ;)
3. fair enough
4. yes

The suggestion seounds reasonable. Though the first case would be irrelevant anyway if we are to remove `check_if_already_set()`

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+        '_updater_kwargs',
+    )
+
+    def __init__(self: 'InitBaseBuilder'):
+        # Instead of the *_was_set variables, we could work with e.g. _token = DEFAULT_NONE.
+        # However, this would make type hinting a *lot* more involved and reasonable type hinting
+        # accuracy is valuable for the builder classes.
+
+        self._token: str = ''
+        self._token_was_set = False
```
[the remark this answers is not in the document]
Bibo-Joshi: That's what I originally tried to do (note that we already fo `DEFAULT_NONE` for such cases ;P). The reason why I did not go through with this is:

https://github.com/python-telegram-bot/python-telegram-bot/blob/663b81c2cd78bb849c3e232c2bb9a62a84ed2482/telegram/ext/builders.py#L195-L197

But again, if we drop `check_if_already_set`, then we need neither `DEFAULT_NONE` nor `_*_was_set`

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+            workers=self._workers,
+            exception_event=self._exception_event,
+            job_queue=job_queue,
+            persistence=self._persistence,
+            context_types=self._context_types,
+            builder_flag=True,
+            **self._dispatcher_kwargs,
+        )
+
+        if isinstance(job_queue, JobQueue):
```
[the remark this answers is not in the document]
Bibo-Joshi: no, this makes sure that `job_queue is not None`. Don't know why I wrote it that way - will change

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+            job_queue.set_dispatcher(dispatcher)
```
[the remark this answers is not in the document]
Bibo-Joshi: mh, this was introduced in the original context PR #1100 and so far we didn't have any problems with it IISC. But I can check and see if converting it into a weakref causes any problems in the tests :)

telegram/ext/builders.py
```diff
@@ -0,0 +1,1338 @@
...
+
+    @staticmethod
+    def _exception_builder(arg_1: str, arg_2: str = None) -> RuntimeError:
+        if not arg_2:
+            return RuntimeError(f'The parameter `{arg_1}` was already set.')
+        return RuntimeError(f'The parameter `{arg_1}` can only be set, if the no {arg_2} was set.')
+
+    @property
+    def _dispatcher_check(self) -> bool:
+        return self._dispatcher_was_set and self._dispatcher is not None
```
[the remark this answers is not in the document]
Bibo-Joshi: see my reply above :)

telegram/ext/builders.py
Bibo-Joshi: not really. both `token` and `request` can bes passed to `Bot`, so if `_request` was set you can still set a token …

telegram/ext/builders.py
Bibo-Joshi: https://github.com/python-telegram-bot/python-telegram-bot/pull/2646#discussion_r706642955

telegram/ext/builders.py
Bibo-Joshi: https://github.com/python-telegram-bot/python-telegram-bot/pull/2646#discussion_r706642955

telegram/ext/builders.py
Bibo-Joshi: https://github.com/python-telegram-bot/python-telegram-bot/pull/2646#discussion_r706642955

telegram/ext/builders.py
Bibo-Joshi: https://github.com/python-telegram-bot/python-telegram-bot/pull/2646#discussion_r706642955

telegram/ext/builders.py
Bibo-Joshi: not AFAIK … the closet you can maybe get is `Literal[1,2,3,4]` which will allow only those numbers and iirc `Literal` is also new in py3.9

telegram/ext/builders.py
Bibo-Joshi: that makes sense 👍

tests/conftest.py
Bibo-Joshi: `Bot` and `ExtBot` currently don't have builders. See my first bullet point in the PR description.

if we want to add a builder for `tg.Bot`, this can only be optional, because one needs to be able to instantiate `tg.Bot` without `tg.ext`

Bibo-Joshi:
> Why is it `DefaultContextType`, not just `DefaultContext`?

@JosXa I chose the naming because it's only a type alias and not an actual class. I feared that `DefaultContext` could be easily confused with `CallbackContext` …

telegram/ext/builders.py
Bibo-Joshi: We'd still have to change the docstrings & type hints and the latter would require a lot of `# type: ignores` …

examples/passportbot.py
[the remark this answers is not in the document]
JosXa: `priv_key_file = Path(__file__).parent / 'private.key'` would make it independent of CWD 🤔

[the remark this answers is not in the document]
Bibo-Joshi: Okay, working on python-telegram-bot/ptbcontrib#29 pretty much removed my last doubts about `pathlib` 😄 But I guess I'd suggest to rather make a standalone PR that introduces pathlib whereever possible …

telegram/ext/builders.py
```diff
@@ -0,0 +1,1196 @@
...
+        self._persistence: ODVInput['BasePersistence'] = DEFAULT_NONE
+        self._context_types: DVInput[ContextTypes] = DefaultValue(ContextTypes())
+        self._dispatcher: ODVInput['Dispatcher'] = DEFAULT_NONE
+        self._user_signal_handler: Optional[Callable[[int, object], Any]] = None
+        self._dispatcher_class: DVInput[Type[Dispatcher]] = DefaultValue(Dispatcher)
+        self._dispatcher_kwargs: Dict[str, object] = {}
+        self._updater_class: Type[Updater] = Updater
+        self._updater_kwargs: Dict[str, object] = {}
+
+    def _get_connection_pool_size(self) -> int:
```
tsnoam, line 184: This might be my personal taste, but I prefer such a function to be a staticmethod and have it require as an input the workers count.
I see the following benefits:
1. It allows more flexibility for the consumer, passing a different worker count, depending on variables of its own control.
2. This approach makes it easier adding a unitest which tests the logic of this function, and this function alone (and maybe we should add  such a unitest)

i.e.
```python
def _get_connection_pool_size(workers: int) -> int:
    return workers + 4
```

telegram/ext/builders.py
```diff
@@ -0,0 +1,1196 @@
...
+            token=self._token,
+            base_url=DefaultValue.get_value(self._base_url),
+            base_file_url=DefaultValue.get_value(self._base_file_url),
+            private_key=DefaultValue.get_value(self._private_key),
+            private_key_password=DefaultValue.get_value(self._private_key_password),
+            defaults=DefaultValue.get_value(self._defaults),
+            arbitrary_callback_data=DefaultValue.get_value(self._arbitrary_callback_data),
+            request=self._request  # type: ignore[arg-type]
+            if self._request is not DEFAULT_NONE
+            else Request(**request_kwargs),  # pylint: disable=E1134
```
tsnoam, line 215: just a style comment: if you can't get the required value in one liner, better calculate it ahead. makes reading easier on the eye (especially that black has some weird preference not to indent the code here).  
Another reason would be because you only need to deduce what request_kwargs are iff you need to construct a request object.

i.e.
```python
request = self._request
if request is DEFAULT_NONE:
    request_kwargs = DefaultValue.get_value(self._request_kwargs)
    if "con_pool_size" not in request_kwargs:
        request_kwargs["con_pool_size"] = self._get_connection_pool_size()
    request = Request(**request_kwargs)

return ExtBot(
    .
    .
    request=request,
)

File: telegram/ext/builders.py [hunk 24]
```diff
@@ -0,0 +1,1196 @@
...
+            )
+
+        return self._updater_class(
+            dispatcher=self._dispatcher,
+            bot=self._dispatcher.bot if self._dispatcher else (self._bot or self._build_ext_bot()),
+            update_queue=self._update_queue,
+            user_signal_handler=self._user_signal_handler,
+            exception_event=self._dispatcher.exception_event
+            if self._dispatcher
+            else DefaultValue.get_value(self._exception_event),
```
tsnoam, line 272: Same comment about style as in the `_build_ext_bot()` function

telegram/ext/dispatcher.py
```diff
@@ -154,46 +160,21 @@ class Dispatcher(Generic[CCT, UD, CD, BD]):
...
-        self.job_queue = job_queue
-        self.workers = workers
-        self.context_types = cast(ContextTypes[CCT, UD, CD, BD], context_types or ContextTypes())
+    def __init__(self, **kwargs: object):
+        if not kwargs.pop('builder_flag', False):
+            warnings.warn(
+                '`Dispatcher` instances should be built via the `DispatcherBuilder`.',
+                UserWarning,
+                stacklevel=2,
+            )
```
tsnoam, line 169: 1. I really prefer that the function signature would not include kwargs, but rather the keyword args we expect.
2. If we want to force keyword args (as we've discussed in the developers group) we can do:
```python
def __init__(self, **, bot, update_queue, ...):
```
3. If we want to issue a warning, we can use `inspect` to check the program stack, and make sure we were called via `UpdaterBuilder`.

telegram/ext/updater.py
```diff
@@ -64,258 +65,88 @@ class Updater(Generic[CCT, UD, CD, BD]):
...
-
-        if defaults and bot:
-            warnings.warn(
-                'Passing defaults to an Updater has no effect when a Bot is passed '
-                'as well. Pass them to the Bot instead.',
-                TelegramDeprecationWarning,
-                stacklevel=2,
-            )
-        if arbitrary_callback_data is not DEFAULT_FALSE and bot:
+    def __init__(self, **kwargs: Any):
```
tsnoam, line 111: Same comment as for `Dispatcher.__init__()`

examples/passportbot.py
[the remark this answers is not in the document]
tsnoam: agreed. please open an issue so it won't be forgotten.

tsnoam (changes requested):

examples/passportbot.py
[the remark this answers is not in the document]
Bibo-Joshi: #2665

telegram/ext/dispatcher.py
Bibo-Joshi: My intention was to hide the signature as much as possible from the user. With a proper signature, IDEs will still suggest the required parameters for initializing `Updater/Dispatcher` manually …

And when doing that, I thought that I'd have to include `builder_flag` in the signature if I made the signature explicit. (You suggesting to you `inspect` magic is also new for me ;P)

But I'm not too passionate about it …

telegram/ext/updater.py
```diff
@@ -748,21 +593,18 @@ def bootstrap_onerr_cb(exc):
...
                 self.running = False
 
                 self._stop_httpd()
                 self._stop_dispatcher()
                 self._join_threads()
 
-                # Stop the Request instance only if it was created by the Updater
-                if self._request:
-                    self._request.stop()
-
```
Bibo-Joshi, line 765: Need to call this elsewhere …

telegram/ext/dispatcher.py
tsnoam: 1. **Kwargs, when you know all the possibile keyword arguments, is evil.
Hard to understand the code, leaves much more space for runtime errors and hard to work with an ide (basically it all goes together)
2. Why are we blocking manual construction of the Updater class? Beacsue we decided that this approach is obsolete. That's fine. It's good to block/warn on incorrect usage.
3. There are two approaches here: adding a "magic" argument which validates the code was called by the Builder class (and it bloats the amount of arguments) or, taking advantage of python's (black magic) reflection capabilities. Since this is something that happens only once at the start of the application (in contrary to constant usage during message processing), I don't mind using some of this black magic. So I'm just a bit in favor of it. Anyway, both approaches are valid in my eyes.
4. But back to article 1 above, I prefer a normal function signature. And as noted in my previous comment, we can still force using keyword arguments (as demonstrated in article 2 of that comment).

PR #2649, opened 2021-09-03
Move error_handler error handing into Dispatcher.dispatch_error

### LINKED ISSUE #2601: Refactor handling of exceptions raised by error handlers
Currently, we do something like

```python
try:
    dispatcher.dispatch_error(exception)
except:
    logger.execption(…)
```

everywhere, where we call `dispatcher.dispatch_error`, leading to some code repetition. I propose to instead move that logic directly into `Dispatcher.dispatch_error`. The only downside of this would be that currently log message in line 4 above can contain some explicit hint on where the exception happened, but one can also just read the traceback …

Edit: This would also cover cases where some `CustomContext.from_*` raises an exception, see the discussion at https://t.me/c/1494805131/17428

Bibo-Joshi:
Closes #2601

### Checklist for PRs


Bibo-Joshi:
closing in favor of #2660

PR #2657, opened 2021-09-14
Check that {Inline, Reply}Keyboards are of the correct type

harshil21:
If users don't input the keyboard in the correct format, i.e a list of lists, then the resulting error is hard to debug and its not obvious what went wrong. 

inspired from #2652

### DIFF
--- a/telegram/inline/inlinekeyboardmarkup.py
+++ b/telegram/inline/inlinekeyboardmarkup.py
@@ -53,6 +53,12 @@ class InlineKeyboardMarkup(ReplyMarkup):
 
         self._id_attrs = (self.inline_keyboard,)
 
+        if not self._check_keyboard_type(inline_keyboard):
+            raise ValueError(
+                "The parameter `inline_keyboard` should be a list of "
+                "list of InlineKeyboardButtons"
+            )
+
     def to_dict(self) -> JSONDict:
         """See :meth:`telegram.TelegramObject.to_dict`."""
         data = super().to_dict()
--- a/telegram/inline/inlinekeyboardmarkup.py
+++ b/telegram/inline/inlinekeyboardmarkup.py
@@ -48,16 +48,15 @@ class InlineKeyboardMarkup(ReplyMarkup):
     __slots__ = ('inline_keyboard',)
 
     def __init__(self, inline_keyboard: List[List[InlineKeyboardButton]], **_kwargs: Any):
-        # Required
-        self.inline_keyboard = inline_keyboard
-
-        self._id_attrs = (self.inline_keyboard,)
-
         if not self._check_keyboard_type(inline_keyboard):
             raise ValueError(
                 "The parameter `inline_keyboard` should be a list of "
                 "list of InlineKeyboardButtons"
             )
+        # Required
+        self.inline_keyboard = inline_keyboard
+
+        self._id_attrs = (self.inline_keyboard,)
 
     def to_dict(self) -> JSONDict:
         """See :meth:`telegram.TelegramObject.to_dict`."""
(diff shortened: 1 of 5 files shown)


telegram/replymarkup.py
```diff
@@ -31,3 +35,13 @@ class ReplyMarkup(TelegramObject):
     """
 
     __slots__ = ()
+
+    @staticmethod
+    def _check_keyboard_type(
+        keyboard: Sequence[Sequence[Union['KeyboardButton', 'str', 'InlineKeyboardButton']]]
+    ) -> bool:
```
Bibo-Joshi, line 42: shouldn't this rather just accept `object`? E.g. a confused user might just pass an `InlineKeyboardButton`, not even a simple list

telegram/inline/inlinekeyboardmarkup.py
```diff
@@ -53,6 +53,12 @@ def __init__(self, inline_keyboard: List[List[InlineKeyboardButton]], **_kwargs:
 
         self._id_attrs = (self.inline_keyboard,)
 
+        if not self._check_keyboard_type(inline_keyboard):
+            raise ValueError(
+                "The parameter `inline_keyboard` should be a list of "
+                "list of InlineKeyboardButtons"
+            )
```
Bibo-Joshi, line 60: Do we want to throw an error or just a warning? 🤔

Bibo-Joshi (commented):
👍 

can you call `check_keyboard_type` either at the end of `__init__` for both or at the start of `__init__` for both?

telegram/inline/inlinekeyboardmarkup.py
harshil21: if we warn then its going to raise an error later anyway, might as well do it earlier with a proper message

telegram/replymarkup.py
harshil21: I guess so, I just 'combined' the type hints of both the keyboards since its intended to be used only for them

harshil21: Okay nvm, mypy gives the error: 
`telegram/replymarkup.py:42: error: "object" has no attribute "__iter__"; maybe "__str__" or "__dir__"? (not iterable)  [attr-defined]`
 if you change it to `object`

Bibo-Joshi: > Also it's not really meant to be used by users since its private.

true. by "pass" I meant pass to `Inline/ReplyKeyboardMarkup`. My point is that if the helper function only accepts the correct type, then it doesn't really test much ;)

Bibo-Joshi: > Okay nvm, mypy gives the error:
> `telegram/replymarkup.py:42: error: "object" has no attribute "__iter__"; maybe "__str__" or "__dir__"? (not iterable) [attr-defined]`
> if you change it to `object`

that's exactly why it should be `object` :)

telegram/inline/inlinekeyboardmarkup.py
Bibo-Joshi: mh, yeah … makes sense :)

telegram/replymarkup.py
harshil21: oh I see, i'll just add a ignore comment then

telegram/replymarkup.py
```diff
@@ -31,3 +30,11 @@ class ReplyMarkup(TelegramObject):
     """
 
     __slots__ = ()
+
+    @staticmethod
+    def _check_keyboard_type(keyboard: object) -> bool:
+        """Checks if the keyboard provided is of the correct type - A list of lists."""
+        for row in keyboard:  # type: ignore[attr-defined]
```
Bibo-Joshi, line 37: this will throw an exception if `keyboard` is not iterable. please add a check that `keyboard` is a list and also add a corresponding test case :)

PR #2658, opened 2021-09-14
Improve performance of pylint

harshil21:
`pylint` used to run super quick, but a few months back it started to take 2+ mins on my machine. Apparently you can set the number of processors to speed it up. Setting it to 0 will auto detect no. of processors it can use. Much fast now! 😄

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,8 @@ repos:
         files: ^(telegram|examples)/.*\.py$
         args:
           - --rcfile=setup.cfg
-          - --jobs=0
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -20,6 +20,7 @@ repos:
         files: ^(telegram|examples)/.*\.py$
         args:
           - --rcfile=setup.cfg
+          - --jobs=0
```
Bibo-Joshi, line 23: Would you mind adding a comment explaining what this does and/or linking to the pylint docs?

PR #2662, opened 2021-09-16
Warnings overhaul

### LINKED ISSUE #2467: Overhaul of warnings
While reviewing #2464 I noticed that we're not very consistend on whether we use `logger.warning` or `warnings.warn` to issue warnings to users and I think we should try to overhaul that in v14. Here are a few things that came to my mind and that I'd like to discuss & tackle:

* We should add a subclass of `UserWarning` named `TelegramUserWarning` or similar. That way users can filter for those explicitly, if they want to.
* `TelegramDeprecationWarning` should then become a subclass of that new class
* Both should be properly included in the documentation. They exist so that users can use them in filtering, so users should be above to find them in the docs. To this end, they should probably also be easier importable, i.e. via `telegram.utils.TGWarning` or `telegram.deprecation.TGWarning`, maybe even `telegram.TGWarning` (see also #2468)
* Regarding when to use what: Citing the [python docs on logging](https://docs.python.org/3/howto/logging.html#when-to-use-logging):

  > **Issue a warning regarding a particular runtime event:**
  > warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
  >
  > logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

  The way I understand this, this would mean in particular that we should use `warnings.warn()` whenever

  * the user passes some non-reasonable value on initialization on classes that are usually set up manually (e.g. `Updater` and `Handlers`, in contrast to the TG types)
  * the way the users tries to use stuff does not correspond to how they set up the bot (thinking of e.g. [~this~](https://github.com/python-telegram-bot/python-telegram-bot/blob/conv-timeout/telegram/ext/conversationhandler.py#L625) [this](https://github.com/python-telegram-bot/python-telegram-bot/blob/bcec6f03cb1797d31039926f0c8c0f2306dcdf5a/telegram/ext/conversationhandler.py#L648-L651))
  * the users misuses something, the example that I have in mind is equality comparison for objects that have no `_id_attrs`

* We should make a wiki page on logging and warnings. We can then delete the logging section from the readme. the wiki page should also contain some basic info about how to catch warnings in the logger and how to filter warnings/logging entries made by PTB.
* **edit:** for warnings, it would be good to double check that the `stacklevel` is set appropriately 

As changing warnings <-> logs is somewhat breaking at least for people who filter those things, I'm putting this on the v14 milestone.

harshil21:
Closes #2467 

Notes:

- Made `Telegram{User, Deprecation}Warning` public and is accessible by `telegram.utils.deprecate`. We can change it later depending on #2468. Edit: Review now made it accessible via `telegram.utils.warnings`

- some warnings don't have stacklevels because they take place in another thread so its not possible to get the caller

- update some warning messages to be more descriptive.


telegram/ext/basepersistence.py
```diff
@@ -334,9 +335,9 @@ def _insert_bot(self, obj: object, memo: Dict[int, object]) -> object:  # pylint
         if isinstance(obj, type):
             # classes usually do have a __dict__, but it's not writable
             warnings.warn(
-                'BasePersistence.insert_bot does not handle classes. See '
-                'the docs of BasePersistence.insert_bot for more information.',
-                RuntimeWarning,
+                f'BasePersistence.insert_bot does not handle classes such as {obj.__name__!r}. '
+                'See the docs of BasePersistence.insert_bot for more information.',
+                category=TelegramUserWarning,
```
Bibo-Joshi, line 340: Maybe it's worth adding another subclass `TelegramRuntimeWarning`

Bibo-Joshi (commented):
Looks good so far! Comments:

* Even if import policy is changed in #2468, we should probably rename `utils.deprecate` to `utils.warnings` already since that file is now no longer exclusively for deprecation
* Just a thought: Maybe it's worth adding a function `utils.warnings.warn`, which acts as a shortcut for `warnings.warn` with a default `category=TelegramUserWarning`?
* Another thought: Maybe `PTB*Warning` is better than `TG*Warning` b/c the warnings are issued by PTB and not by TG?

telegram/utils/warnings.py
```diff
@@ -29,9 +31,22 @@ class TelegramUserWarning(UserWarning):
...
 # DeprecationWarning. We also subclass from TGUserWarning so users can easily 'switch off' warnings
-class TelegramDeprecationWarning(TelegramUserWarning, DeprecationWarning):
+class PTBDeprecationWarning(PTBUserWarning, DeprecationWarning):
     """Custom warning class for deprecations in this library."""
 
     __slots__ = ()
+
+
+def warn(message: str, category: Type[Warning] = PTBUserWarning, stacklevel: int = 1) -> None:
+    """Helper function used as a shortcut for warning with default values."""
```
Bibo-Joshi: Please add a note along the lines "The arguments cat and sl will be passed to warnigs.warn"

telegram/utils/warnings.py
```diff
@@ -29,9 +31,22 @@ class TelegramUserWarning(UserWarning):
...
+    warnings.warn(message, category=category, stacklevel=stacklevel)
```
Bibo-Joshi, line 63: maybe

```suggestion
    warnings.warn(message, category=category, stacklevel=stacklevel+1)
```
? Then we wouldn't have to think about that wan calling this function 🤔

telegram/utils/warnings.py
```diff
@@ -29,9 +31,22 @@ class TelegramUserWarning(UserWarning):
     __slots__ = ()
 
 
+class PTBRuntimeWarning(PTBUserWarning, RuntimeWarning):
+    """
+    Custom runtime warning class used for warnings in this library.
```
Bibo-Joshi, line 36:
```suggestion
    """Custom runtime warning class used for warnings in this library.
```

telegram/utils/warnings.py
harshil21: I guess.. I'll also specify that in the docs

telegram/utils/warnings.py
harshil21: added versionadded directive below it so doesn't matter

Issue #2633, opened 2021-08-18
Make documentation more accessible


Bibo-Joshi:
Hello world.

We receive regularly queries in our help group from people unable to read the support documents we provide. We believe this is mainly because we don't do a good enough job presenting them. This is where you come in: If you are new to the project and think "hey, this can be improved", and its not on the list yet, tell us about this.

One maintainer has compromised the thoughts **we** as maintainers made about this topic, but we value foreign input very much. Some discussion about this happened at https://t.me/pythontelegrambotgroup/504413 + answers.

**edit:** The list was extended with the ideas from the comments below and converted into a checklist so that it's easier to keep track of what's already been done.

The mentioned ideas:

- [x] Cross referencing. a LOT: -> #3110
    - [ ] within RTD: JobQueue -> CallbackContext.job_queue, Dispatcher.job_queue, Updater.job_queue. Bot.edit_message_text -> Message.edit_text. Similar for all other stuff that makes sense - bascically all "where is this used" references. sphinx has the nice `.. seealso::` directive that we can use for this  #2855
    - [ ] examples + RTD -> wiki: JobQueue -> wiki/jobqueue, Bot.arbitrary_callback_data -> wiki/arbitrary_callback_data. examples/timerbot -> jobqueue. etc.
    - [ ] RT + wiki -> examples. JobQueue -> timerbot.py, wiki/jobqueue -> timerbot.py etc.

   These should help a great deal when it comes to "where do I use this"? or "how is this used"?
- [ ] Maybe: On RTD, for the *really* newbies a "how to read sphinx docs", e.g. explain that "Parameters: token (str) – Bot’s unique authentication." means that the keyword argument ist named "token" and expects input of type str
- [ ] We have several notes and docstrings that are repeated very often across the docs, which makes it hard to maintain them. An idea is to use the `.. include::` directive to unify those. Unfortunately this currently doesn't work with the `napoleon` docstring style that we use - see sphinx-doc/sphinx#9939. If/when this can be tackled, this should include the following:
[...]

PR #2678, opened 2021-09-28
Update documentation of the ForceReply class

schvv31n:
### Checklist for PRs

- [X] Added myself alphabetically to `AUTHORS.rst` (optional)


Bibo-Joshi:
Hi. Thanks for the PR! However, this note is a bit overly specific IMHO. the [official Telegram docs](https://core.telegram.org/bots/api#updating-messages) already state that only messages without reply markup or with an inline keyboard can be edited.

One can surely argue that this note should be repeated in PTBs docs somewhere and I've added it to #2633 (see https://github.com/python-telegram-bot/python-telegram-bot/issues/2633#issuecomment-929348652). If you're interested, you're very welcome to tackle this in a new PR :)

PR #2686, opened 2021-10-01
fix @log annotation for pycharm and vscode

### LINKED ISSUE #2677: @log obscured wrapped function's signature in VS Code
### Expected behaviour
Hover on referenced function, its actual signature pops up

### Actual behaviour
Instead, decorator's signature was prompted, as in screenshot above

### Configuration
**Operating System:** macOS 11.4
**Version of Python, python-telegram-bot & dependencies:** Python 3.6.9 (pyenv), PTB 13.6

---

I first reported this to pylance (https://github.com/microsoft/pylance-release/issues/1470), and was pointed out that `@log()`'s annotation is to blame (https://github.com/microsoft/pylance-release/issues/1470#issuecomment-865288128):

`...` in `func: Callable[..., RT]` throws away parameter definition, and should use `ParamSpec` instead.

https://github.com/python-telegram-bot/python-telegram-bot/blob/105f1ccdb5283f19c97c76a8234f8e61df88ce60/telegram/bot.py#L121-L123

However, `ParamSpec` is pretty new (landed in Python 3.10) thus requires `typing_extension` package with older Python.

harshil21:
Closes #2677. 
Also adds a test for checking `@log`

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -2026,7 +2026,6 @@ class Bot(TelegramObject):
 
         return effective_results, next_offset
 
-    @_log
     @no_type_check  # mypy doesn't play too well with hasattr
     def _insert_defaults_for_ilq_results(  # pylint: disable=R0201
         self, res: 'InlineQueryResult'
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -35,6 +35,7 @@ from typing import (
     Dict,
     cast,
     Sequence,
+    Any,
 )
 
 try:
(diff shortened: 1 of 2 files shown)


Bibo-Joshi (commented):
do we have a test that actually checks that `@log` works?

harshil21:
I can't find one. I'll add it.

telegram/bot.py
```diff
@@ -222,6 +207,23 @@ def __init__(
                 private_key, password=private_key_password, backend=default_backend()
             )
 
+    # TODO: After https://youtrack.jetbrains.com/issue/PY-50952 is fixed, we can revisit this and
+    # consider adding Paramspec from typing_extensions to properly fix this. Currently a workaround
+    def log(  # type: ignore[no-untyped-def] # skipcq: PY-D0003
```
Bibo-Joshi, line 212: in light of our efforts to make clear what is public and what is not: let's make this a private method :)

telegram/bot.py
```diff
@@ -222,6 +207,23 @@ def __init__(
...
+        func: Any, *args, **kwargs  # pylint: disable=W0613
```
Bibo-Joshi, line 213: I think we can actually drop `*args, **kwargs`, right? I don't recall why they are here, but IISC we would only need them if we did things like `@log(*args)`
and we wanted to type out pylint error descriptions …

telegram/bot.py
harshil21: I thought of doing that, but then I saw that that didn't show up in the docs, so didn't change it. Will change anyway.

telegram/bot.py
harshil21: yep, even I didn't know why they are here when we don't use them

telegram/bot.py
Bibo-Joshi: IIRC that's just because it doesn't have a docstring. auto-complete will still show it …

PR #2687, opened 2021-10-01
Private modules

### LINKED ISSUE #2680: Further clear up imports policy: Private modules
Currently all the modules like `telegram.{message, chat, user, …}` look like they are public. However, we don't consider them to be part of the public API and they should not be treated as such by the user. The only thing that we guarantee to work is `from telegram import Message, Chat, User, …` and *not* `from telegram.message import Message`.

Rename all the private modules with a leading underscore, i.e. all files that contain classes that are imported in the `__init__`.
This is in accordance with [PEP8](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces): 

> Even with `__all__` set appropriately, internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore.

this should also be done for the `telegram.ext` package except for the module `telegram.ext.filters`. The packages `telegram(.ext).utils` should also be renamed to `telegram(.ext)._utils` and dropped from the documentation - unless this leads to pre-commit complaining about importing private packages. If it does, we'll have to re-evaluate.

#### Additional context
* This can be seen as extension of #2468 
* A PR for #2573 will likely introduce a new module. Depending on which issue get's resolved by a PR first, the other one will have to adapt to the changes.

kennethcheo:
Closes #2680 

I also standardized some of the module imports, except in cases of circular imports.

### Changes
- [x] Rename private modules in `telegram` package

### Todo
- [ ] Rename private modules in `telegram.ext` package
- [ ] Rename `telegram(.ext).utils` to `telegram(.ext)._utils`


telegram/ext/extbot.py
```diff
@@ -21,8 +21,9 @@
 from copy import copy
 from typing import Union, cast, List, Callable, Optional, Tuple, TypeVar, TYPE_CHECKING, Sequence
 
-import telegram.bot
+# import telegram.bot
```
Bibo-Joshi, line 24: please remove this line

telegram/__init__.py
```diff
@@ -34,103 +34,103 @@
...
-from .files.file import File
-from .parsemode import ParseMode
-from .messageentity import MessageEntity
-from .messageid import MessageId
-from .games.game import Game
-from .poll import Poll, PollOption, PollAnswer
-from .voicechat import (
+from ._chatmemberupdated import ChatMemberUpdated
+from ._chatpermissions import ChatPermissions
+from .files._photosize import PhotoSize
```
Bibo-Joshi, line 39: the subpackages should also be private, e.g. `telgram._{files, games, inline, passport, payment}`. Let's skip `vendor` for now, because that contains a submodule. I don't know how well that plays with renaming and also `vendor` will likely be gone after the switch to asyncio, anyway …

telegram/ext/_dispatcher.py
```diff
@@ -42,19 +42,16 @@
 
 from telegram import Update
 from telegram.error import TelegramError
-from telegram.ext import BasePersistence, ContextTypes
-from telegram.ext.handler import Handler
-import telegram.ext.extbot
-from telegram.ext.callbackdatacache import CallbackDataCache
+from telegram.ext import BasePersistence, ContextTypes, ExtBot, Handler
+from telegram.ext._callbackdatacache import CallbackDataCache
```
Bibo-Joshi, line 46: maybe replace `from tg.ext._*` with `from ._*`? 🤔

Bibo-Joshi (changes requested):
The DeepSource fail is unrelated - should work on the next commit. Please make sure to merge v14 into your branch, since there are currently merge conflicts :)

kennethcheo:
All private modules seem to be renamed with leading underscores, unless I missed any. I wasn't sure if the modules within `tg(.ext)._utils` should be private as well so I left them for now.

Apologies for the huge number of changes. I tried to make each PR as atomic as possible.

To note:
- I have not made any changes to the documentation or removed `tg.(.ext).utils` from the documentation.
- I tried to merge v14 into the branch but there were changes that I did not make (from the rebased commits?), so I left that for now as well.

Do let me know if there are any errors and I would be happy to fix them :)

telegram/ext/_dispatcher.py
```diff
@@ -42,63 +40,28 @@
...
-from telegram.ext.utils.promise import Promise
-from telegram.utils.helpers import DefaultValue, DEFAULT_FALSE
-from telegram.ext.utils.types import CCT, UD, CD, BD
+from telegram import Update
+from telegram.error import TelegramError
+from telegram.ext import BasePersistence, ContextTypes, ExtBot, Handler
+from telegram._utils.defaultvalue import DefaultValue, DEFAULT_FALSE
+from telegram._utils.warnings import warn
+from telegram.ext._utils.promise import Promise
+from telegram.ext._utils.types import CCT, UD, CD, BD
```
Bibo-Joshi, line 49: https://github.com/python-telegram-bot/python-telegram-bot/pull/2687#discussion_r720688636

Bibo-Joshi (changes requested):
Hi, thanks for the updates. Don't worry about the diff, that was to be expected :D  Some comments:

* merge conflicts: yeah, I had to rebase v14, that's were they come from - also the reason for a number of changes that are currently previewed … Sorry about that. I suggest that you address my commenst below and then rebase your branch onto v14 (I can also help with that). Then we can have a final look on a clean diff :)
* please do remove `(.ext).utils` from the docs. searching for `utils` in `docs/souce` and removing the corresponding mentions & files should be enough. The contribution guide also has details on building the docs locally for double checking (you can ignore the warnings about the filters module that you will see)
* `telegram.version.py` can also get an underscore - make sure to update `setup.py` line 52 accordingly
* additional underscores within `(.ext)._utils` are not needed 👍

kennethcheo:
All changes should be done and ready for review :)

I'm not too confident with rebasing commits so I will leave that up to someone more experienced.

Bibo-Joshi:
I did the rebasing & changed some more inputs, which tails a slots test - will have a closer look later :)

Bibo-Joshi:
I tried a bit to get `test_slots` to work, but didn't get too far … The problem is that we now have relative imports (like `from ..._utils.types import JSONDict`), which fail with `ValueError: attempted relative import beyond top-level package` during the programmatic module imports in `test_slots`.

@harshil21 you wrote that module those tests and are more familiar with `importlib`. Could you have a look at this and check if we can adapt the dynamic imports?

telegram/_utils/datetime.py
```diff
@@ -20,7 +20,7 @@
 
 .. versionchanged:: 14.0
    Previously, the contents of this module were available through the (no longer existing)
-   module ``telegram.utils.helpers``.
+   module ``telegram._utils.helpers``.
```
harshil21, line 23: technically speaking this shouldn't have been changed, but its fine imo

telegram/ext/_basepersistence.py
```diff
@@ -398,30 +398,30 @@ def _insert_bot(self, obj: object, memo: Dict[int, object]) -> object:  # pylint
     def get_user_data(self) -> DefaultDict[int, UD]:
         """Will be called by :class:`telegram.ext.Dispatcher` upon creation with a
         persistence object. It should return the ``user_data`` if stored, or an empty
-        :obj:`defaultdict(telegram.ext.utils.types.UD)` with integer keys.
+        :obj:`defaultdict(telegram.ext._utils.types.UD)` with integer keys.
```
harshil21, line 401: see this is what I was talking about, user will not be able to know what type is `UD`!

harshil21 (commented):
good changes, but just one objection

docs/source/telegram.ext.rst
```diff
@@ -53,11 +53,3 @@ Arbitrary Callback Data
...
     telegram.ext.callbackdatacache
     telegram.ext.invalidcallbackdata
-
-utils
------
-
-.. toctree::
-
-    telegram.ext.utils.promise
-    telegram.ext.utils.types
```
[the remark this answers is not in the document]
Bibo-Joshi: What benefit do you see in including modules in the documentation that are explicitly marked as private? As long as they were not private, it made sense and I added the warnings more or less so that we can say "I told you so" in case someone uses them and we break it. But with the utility modules being private now, I don't see a reason to include them in public docs.

The point about type aliases is valid though - thanks for catching this! Some solutions that I could image are:

1. rework documentation without mentioning the type aliases modules
2. make the `types` modules public, i.e. move `telgram(.ext)._utils.types` to `telegram(.ext).types` and remove the warning - then changes have to be considered breaking
3. Midway between the current state of v14 and 2.: Move the modules to `telegram(.ext).types` for convenience of the user, but keep the warning.
4. Another possible midway: The type aliases `CDCData` and `ConversationDict` might actually be of some use for advanced users that build their own persistence classes. We could add them as class variables to `CallbackDataCache` and `ConversationHandler` respectively, similarly to what I've done with `CallbackContext.DEFAULT_TYPE` in #2646. The rest is of no real interest for users IMO, so we could hide them in private modules and adapt the parts of the docs that mention them.

What do you think?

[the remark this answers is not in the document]
harshil21: > What benefit do you see in including modules in the documentation that are explicitly marked as private?

I was thinking that this could also lower the entry bar for new contributors since 'everything' is documented for them. But maybe it isn't that important since they can dig in the source code itself. Might also reduce new users from being overwhelmed by the amount of documentation.

Also I'd like to propose a 5th point:
- Keep it the way it is now. Meaning make `telegram(.ext).utils` a public module, but keep certain files which aren't meant to be used by users at all private and hidden from documentation. Like how `webhookhandler.py` is not included in the docs of  `telegram.ext.utils` at all. The directory structure would be like:

```
telegram.ext.utils:
    telegram.ext.utils.promise.Promise
    telegram.ext.utils.types Module

telegram.utils:
    telegram.utils.defaultvalue Module
    telegram.utils.types Module
```
The remaining files would be in `utils` but would be private and not in documentation (since they aren't being referenced anywhere else). And we keep the warning in all of them.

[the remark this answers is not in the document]
Bibo-Joshi: @kennethcheo 

Harshil & 1 had a chat offline and we came to the conclusion that all contents of `(ext)._utils` should stay private. This means that some parts of the documentation need to be updated, more precisely the docs shouldn't mention any contents of `(ext)._utils`. I'm aware that this wasn't part of the description in #2680. Still, if you would like to try & make the corresponding changes, you're very welcome to do so. If you don't want to, that's ofc fine as well.
I would be glad if you could give a short response here 🙂

kennethcheo:
Hi @Bibo-Joshi, I would be happy to update the documentation with the discussed changes. However, I am a little confused by what you mean by removing all mentions of `(ext)._utils`.

For example, in the [BasePersistence ](https://python-telegram-bot.readthedocs.io/en/stable/telegram.ext.basepersistence.html#telegram.ext.BasePersistence.get_user_data) class, there are methods that use and mention `(ext)_utils` in their docstrings. Does that mean the docstrings for these methods should be removed? If so, how would we document the use of these methods?

kennethcheo:
All mentions of `tg.(ext)._utils` modules should have been removed, but please let me know if I missed any.

The changes also follow the discussed scheme, and I hope I haven't altered any key phrasing or meanings. Let know me if anything could be worded better and I would be happy to edit it :)

kennethcheo:
The past 4 commits have very similar messages as I didn't really know what to name them after the first one. Maybe we can squash them into 1 commit after all documentation changes have been made?

Bibo-Joshi:
Thank you for the contribution @kennethcheo :)

PR #2700, opened 2021-10-05
change pylint error codes to messages

### LINKED ISSUE #2664: Pylint & Flake8: Error codes -> string comments
e.g. 

```
# pylint: disable=line-too-long
```
rather than

```
# pylint: disable=C0301
```

IMO this has low priority, but it may improve readability a bit - even though I fear that it will introduce a bunch of artificial line breaks. But whatever :D

Piraty:
### Checklist for PRs

Closes #2664


Bibo-Joshi:
mh, you are right. I don't know where exactly I got the idea from that this should be possible for flake8 as well, but it's indeed not. Thanks for the clarification :)

PR #2707, opened 2021-10-06
exact description of which command is not valid

kilomeow:
ValueError message contains wrong command


Bibo-Joshi:
closing due to inactivity & in favor of #2837

PR #2708, opened 2021-10-07
[REFACTOR]: Enum constants

Bibo-Joshi:
Hey. I'm aware that we didn't discuss this, but it felt like reasonable change. If you disagree, we can close :D 
Key points of this PR:

* group the constants in `tg.constants` into enums. This has the advantage of
    * better name space (`ParseMode.HTML` instead of `PARSEMODE_HTML`)
    * the enums are still instances of `int`/`str`, so no change there
    * the enums are iterable which may come in handy in some situations (see below)
    * we can control the representation of the enums (see below)
* Some constants didn't really fit into groups, so I left them as standalone constants
* For string enums, I made sure that the string representation reads something like `<ParseMode.HTML>`, rather than the default behavior of `<ParseMode.HTML: 'html'>`. This is because the actual value is of little interest and IMO it also encourages users to use the constants rather than plain strings.
* `tg.{ChatAction, ParseMode}` were removed in favor of `tg.constants.{ChatAction, ParseMode}`.

Some more changes that could be made but that I didn't make yet because I wasn't sure if we want that:

* For now I kept the constants like `Update.ALL_TYPES`. However, if you agree, I think we can remove them, as now one can just do `list(constants.UpdateType)`.
* In fact, one could think of dropping these constants alltogether and replacing e.g. `Chat.{PRIVATE, CHANNEL, …}` by `Chat.TYPES = tg.constants.ChatType`.

things that should be done in any case:

- [x] Update references to `telegram.ParseMode` if #2717 is merged before this
- [x] Add an enum for the `InputMedia*` types 
- [x] Add enums for Message/attachment types

### Checklist for PRs

Breaking changes:
- Constants in the module `constants` are now enums

### DIFF
--- a/telegram/_files/inputmedia.py
+++ b/telegram/_files/inputmedia.py
@@ -62,7 +62,7 @@ class InputMedia(TelegramObject):
             entities that appear in the caption, which can be specified instead of parse_mode.
         parse_mode (:obj:`str`, optional): Send Markdown or HTML, if you want Telegram apps to show
             bold, italic, fixed-width text or inline URLs in the media caption. See the constants
-            in :class:`telegram.ParseMode` for the available modes.
+            in :class:`telegram.constants.ParseMode` for the available modes.
 
     Attributes:
         type (:obj:`str`): Type of the input media.


Bibo-Joshi:
@harshil21 what do you think about the two points that I mentioned under "Some more changes that could be made but …"?

harshil21:
> `parse_mode: constants.ParseMode = None` tells your type checker that it's probably a bad idea to pass a plain string. But I wanted to hear your opinion on that

imo it should be as it is, because ultimately the value is actually a string, the user might think that ParseMode is some class they have to know how to initialize etc. Also whenever there's a new API update, and the user wants to use the new value, they'll have to switch back to a string anyway (until PTB updates).


Also skimming through that RTD I saw that we should add an `Enum` for location, see for e.g.:  
https://python-telegram-bot.readthedocs.io/en/enum-constants/telegram.bot.html#telegram.Bot.edit_message_live_location

PR #2717, opened 2021-10-08
Input media subclasses   init   handling refactor #2573

### LINKED ISSUE #2573: Refactor how InputMedia* subclasses handle inits
Currently all the subclasses handle the init completely on their own, while it would make more sense to move all attributes shared between the subclasses to `InputMedia`, i.e. call `super().__init__` like we do for `InlineQueryResult*`. This includes the attributes `caption_entities`, `parse_mode`, `type`, `media` & `caption`

See https://t.me/c/1494805131/16382 + replies for discussion.

eldbud:
* Refactored ```InputMedia.__init__``` so that it initializes the attributes common to all deriving classes.
* Add ```InputMedia._get_thumb_object``` static method
* Created ```_BaseMedium``` class, base class for media objects
* Created ```_BaseThumbedMedium``` class, base class for media objects that may have a thumbnail.
* Refactored Media classes so that they inherit from ```_BaseMedium```/```_BaseThumbedMedium```

closes #2573

### DIFF
--- a/docs/source/telegram.animation.rst
+++ b/docs/source/telegram.animation.rst
@@ -3,6 +3,9 @@
 telegram.Animation
 ==================
 
+.. Also lists methods of _BaseThumbedMedium, but not the ones of TelegramObject
+
 .. autoclass:: telegram.Animation
     :members:
     :show-inheritance:
+    :inherited-members: TelegramObject
--- a/docs/source/telegram.audio.rst
+++ b/docs/source/telegram.audio.rst
@@ -3,6 +3,9 @@
 telegram.Audio
 ==============
 
+.. Also lists methods of _BaseThumbedMedium, but not the ones of TelegramObject
+
 .. autoclass:: telegram.Audio
     :members:
     :show-inheritance:
+    :inherited-members: TelegramObject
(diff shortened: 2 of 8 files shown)


eldbud:
the tests here are failing due to the 
```
telegram/bot.py:232: error: unused 'type: ignore' comment
telegram/bot.py:233: error: unused 'type: ignore' comment
telegram/ext/extbot.py:162: error: unused 'type: ignore' comment
telegram/ext/extbot.py:163: error: unused 'type: ignore' comment
Found 4 errors in 2 files (checked 149 source files)
```
we discussed. any solution for this?

eldbud:
> Also, in line with #2687, the modules `base(thumbed)medium.py` should be prefixed with an underscore.

ah nice i didn't notice it was merged. Will be very happy to rename.

harshil21:
@eldbud the first 3 errors are pylint errors, that's why the `type: ignore` doesn't solve them and mypy complains about that. What you can do is do a `pylint: disable=used-before-assignment`. But it's better if you can address those issues pylint complains about

eldbud:
> Except for `test_official`, this LGTM now :)

test_official is fixed

PR #2762, opened 2021-11-03
Feat: Disable Example notifier run on V14

Poolitzer:
Hi, I had the idea to disable the example notifier for V14. I imagine we will make a few changes to the examples and we reduce the "spam" from this workflow. We could extend this to other notifiers but Im not sure thats needed.


Poolitzer:
nope this apparently doesn't work Im just going to drop this before it blows up more in my face :D

PR #2782, opened 2021-11-20
Invite link convenience functionality

Bibo-Joshi:
adds nome convenience functionality around chat invite links. nothing urgent.

### Checklist for PRs

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -3896,13 +3896,15 @@ class Bot(TelegramObject):
                 For timezone naive :obj:`datetime.datetime` objects, the default timezone of the
                 bot will be used.
             member_limit (:obj:`int`, optional): Maximum number of users that can be members of
-                the chat simultaneously after joining the chat via this invite link; 1-99999.
+                the chat simultaneously after joining the chat via this invite link;
+                1-:tg-const:`telegram.constants.ChatInviteLinkLimit.MEMBER_LIMIT`.
             timeout (:obj:`int` | :obj:`float`, optional): If this value is specified, use it as
                 the read timeout from the server (instead of the one specified during creation of
                 the connection pool).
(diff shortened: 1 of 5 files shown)


harshil21 (approved):
nice feature, but is there a reason why we don't use the other parameters of `ChatInviteLink` here too? Similar to how we do in [`send/edit_location`](https://github.com/python-telegram-bot/python-telegram-bot/blob/3164c0b42b0df7e523b449e559e291550a52413a/telegram/_bot.py#L1465)

Bibo-Joshi:
> nice feature, but is there a reason why we don't use the other parameters of `ChatInviteLink` here too? Similar to how we do in [`send/edit_location`](https://github.com/python-telegram-bot/python-telegram-bot/blob/3164c0b42b0df7e523b449e559e291550a52413a/telegram/_bot.py#L1465)

TBH, I hadn't really thought about it, but interesting idea! you'd do something like e.g.

```python
if isinstance(invite_link, ChatInviteLink):
    expire_date = expire_date or  invite_link.expire_date
```
? That could work. If we want to allow users to e.g. delete the `expire_data`, we'd have to maybe do something like
```python
def edit_chat_link(…, expire_date = DEFAULT_NONE):
    if isinstance(invite_link, ChatInviteLink):
         expire_date = expire_date or invite_link.expire_date
    expire_data = DefaultValue.get_value(expire_date)
```
such that `edit_chat_link(chat_link_instance, expire_date=None)` allows to explicitly set `expire_date` to `None`.
And we'd have to mind that one could want to override `member_limit` with `creates_join_request` and vice versa.

At first glance, I'm not sure if the added convenience is worth the added complexity of the method. What do you think?

harshil21:
ugh, in that case probably best to leave it like this. Not to mention documentation will also increase in complexity.

Poolitzer (approved):
Changes look good.

Issue #2505, opened 2021-05-05
Refactor handling of shortcut-bots


Bibo-Joshi:
A bunch of TG-Objects have `bot` attribute that is used for the shortcuts, e.g. `self.bot.send_message(…)`, which is dangerous in the rare edge cases where users manually instantiate those objects and forget to pass the `bot`. This will then lead to somewhat unprecise error messages `NoneType has no attribute 'send_message'`.
Moreover, `Message.bot` is undocumented, but not private, so people relying on auto-complete may be using it without knowing that it's not really intended for external use.

## Proposal:

Do something like

```python
TGObject:
    self._bot: Optional[Bot] = None

    def get_bot(self) -> Bot:
        if self._bot is None:
            raise RuntimeError('This object has no bot and shortcuts cant be used')
        return self._bot

    def set_bot(self, bot: Bot) -> None:
        self._bot = bot
```

This way we have helpful error messages without having to check `if self.bot is None` in every shortcut.

An alternate solution would be to make `TGObject.bot` a property that raises the error if `_bot is None`. But explicitly calling a method makes it more clear that something *could* go wrong, IMO.

---
Note: We briefly discussed this in dev-chat on 04-05-2021, see https://t.me/c/1494805131/15168

Poolitzer:
changing from `.bot` to `.get_bot()` could/should be considered breaking I guess, while using property setter wouldn't break anything.

I don't share the "more clear" mentality. Since its an undocumented property, raising this error is fine. Again, that this is even an problem happens only when someone outside of this library initiates their own objects, so they should then be fine with dealing with the error.


```suggestion
raise RuntimeError('This object was initiated without the bot parameter and therefore shortcuts can not be used')
```

Bibo-Joshi:
> changing from `.bot` to `.get_bot()` could/should be considered breaking I guess, while using property setter wouldn't break anything.

True. Though I don't see a reason to make an effort to have this non-breaking as this is not urgent at all and we're headed for a breaking release anyway

> I don't share the "more clear" mentality. Since its an undocumented property, raising this error is fine. 

What I mean is: If I access an attribute of an object (that's either documented or non-private & suggested by auto-complete), I don't expect that to raise an exception in any case. I'm fine with checking that the value is something that I expect, but having to do

```python
try:
   obj.attr
except:
   print('accessing attr raised an exception')
```
is something that I would argue is unnatural.

[...]

PR #2783, opened 2021-11-20
Less ignoring for mypy

Bibo-Joshi:
setup.cfg currently makes mypy shut up about a lot of things.
This tries to reduce that, which also lead to some actual improvements. Note that before #2505 this would have been way harder

### Checklist for PRs

None of those really matter here …

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -71,3 +71,6 @@ warn_unused_ignores = False
 
 [mypy-urllib3.*]
 ignore_missing_imports = True
+
+[mypy-apscheduler.*]
+ignore_missing_imports = True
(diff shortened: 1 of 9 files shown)


setup.cfg
```diff
@@ -70,6 +71,3 @@ warn_unused_ignores = False
 
 [mypy-urllib3.*]
 ignore_missing_imports = True
-
-[mypy-apscheduler.*]
-ignore_missing_imports = True
```
harshil21, line 75: Why is this removed? I ran `mypy telegram/` on my machine and now it complains about missing type hints / stubs

harshil21 (commented):
just a single comment

setup.cfg
Bibo-Joshi: mh, interesting that it doesn't complain during pre-commit … anyway, I reverted :)

harshil21 (approved):
LGTM! Deepsource has more false positives again..

PR #2796, opened 2021-11-26
Improve tests

harshil21:
Redo of #2658, which was reverted later because of PyCQA/pylint#5216. 

Also finally fixes a test which was almost always failing locally for me

### DIFF
--- a/telegram/_chat.py
+++ b/telegram/_chat.py
@@ -184,7 +184,7 @@ class Chat(TelegramObject):
 
     def __init__(
         self,
-        id: int,
+        id: int,  # pylint: disable=invalid-name
         type: str,
         title: str = None,
         username: str = None,
--- a/tests/test_bot.py
+++ b/tests/test_bot.py
@@ -593,7 +593,6 @@ class TestBot:
         )
         # Sometimes there can be a few seconds delay, so don't let the test fail due to that-
         if 1 <= abs(msg.poll.close_date.second - aware_close_date.second) < 5:
-            print('too much error!')
             msg.poll.close_date = msg.poll.close_date.replace(second=aware_close_date.second)
         assert msg.poll.close_date == aware_close_date.replace(microsecond=0)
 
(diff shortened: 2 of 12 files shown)


tests/test_bot.py
```diff
@@ -588,24 +583,27 @@ def test_send_close_date_default_tz(self, tz_bot, super_group_id):
...
             question=question,
             options=answers,
             close_date=close_date,
             timeout=60,
         )
-        assert message.poll.close_date == aware_close_date.replace(microsecond=0)
+        # Sometimes there can be a few seconds delay, so don't let the test fail due to that-
+        if 1 <= abs(msg.poll.close_date.second - aware_close_date.second) < 5:
+            msg.poll.close_date = msg.poll.close_date.replace(second=aware_close_date.second)
+        assert msg.poll.close_date == aware_close_date.replace(microsecond=0)
```
Bibo-Joshi, line 596: maybe just check that the timedelta `msg.poll.close_data - aware_close_date` is small?

telegram/_chat.py
```diff
@@ -184,7 +184,7 @@ class Chat(TelegramObject):
 
     def __init__(
         self,
-        id: int,
+        id: int,  # pylint: disable=invalid-name
```
Bibo-Joshi, line 187: do you know why `redefined-builtin` is not needed here? o.0

harshil21: nope, was wondering the same thing

harshil21: ah it's because there's a pylint ignore at the top of the file

tests/test_bot.py
harshil21: oh we can't do that since they are both in different timezones, hence they can have different minutes/hours. 

what I learned from this is that two datetimes can be equal even if their timezones are not equal.

Bibo-Joshi: As long as the *both* are aware, that shouldn't be a problem - since they are aware, you can convert them both to a common timezone. `(msg.poll.close_date - aware_close_date).total_seconds() <= 5` works fine for me …

telegram/_chat.py
Bibo-Joshi: ah okay 👍

PR #2805, opened 2021-12-04
Rearrange module level dunders and add public submodules

### LINKED ISSUE #2763: [DISCUSSION] Adding public submodules to `__all__`

Currently, in the v14 branch we for e.g do: `from telegram import error`. But if a user does `from telegram import *`, it does not import the public submodule `error`, when it should.

[PEP-8](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces) says that documented interfaces are considered public and `"To better support introspection, modules should explicitly declare the names in their public API using the __all__ attribute."`.

Additionally, PEP-8 also says that `__all__` should be placed [_above_](https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names) the imports.

Import and add public submodules to `__all__`, place `__all__` above imports.

#### Additional context

Original discussion: https://github.com/python-telegram-bot/python-telegram-bot/pull/2759#discussion_r740045558

harshil21:
Closes #2763


Bibo-Joshi:
Changes look good 👍 What would you think about adding a unit test for this? i.e. check for every file that's named `__init__.py` or doesn't have an underscore in the path (like `telegram/ext/filters.py`) that

* it has `__all__`
* `__all__` comes directly after the docstring
* for inits, that all non-underscore modules of that directory are included in `__all__`

I guess that could be done with a bit of regex. E.g. for the second one something like `^(\#[^\n]*\n)*((\"{3}[^\"]*\"{3})|)([\n ]*)__all__` could work

Not sure if this actually helps or just makes more problems 😄

harshil21:
> * it has `__all__`

hm, I specifically removed `__all__` for filters because iirc, it messed up with the docs (only rendered those in `__all__`) but maybe there's a setting for that. 

> * `__all__` comes directly after the docstring

actually I think this is something pylint should warn about. I'll open an issue upstream and see if they want to support it or not.

> * for inits, that all non-underscore modules of that directory are included in `__all__`

yes sounds reasonable.

Bibo-Joshi:
> > * it has `__all__`
> 
> hm, I specifically removed `__all__` for filters because iirc, it messed up with the docs (only rendered those in `__all__`) but maybe there's a setting for that.

to me that sounds like `__all__` was missing a lot of stuff 😬 

> > * `__all__` comes directly after the docstring
> 
> actually I think this is something pylint should warn about. I'll open an issue upstream and see if they want to support it or not.

nice idea 💪

harshil21:
> __all__ comes directly after the docstring

Opened https://github.com/PyCQA/pylint/issues/5471

tests/test_modules.py
```diff
@@ -0,0 +1,63 @@
...
+            mod = load_module(mod_path)
+            assert hasattr(mod, '__all__'), f"{folder}'s __init__ does not have an __all__!"
+
+            pub_mods = get_public_submodules_in_folder(folder)
+            cond = all(pub_mod in mod.__all__ for pub_mod in pub_mods)
+
+            assert cond, f"{path}'s __all__ should contain all public submodules ({pub_mods})!"
+            continue
+
+        if '_' in path:  # skip private modules
```
Bibo-Joshi, line 47: maybe rather `path.startswith('_')`?

Bibo-Joshi (commented):
nice 👍 about dunder names coming first: do you want to, wait for pylint, add a custom test temporarily or just merge? I'm open for all :)

tests/test_modules.py
harshil21: no, path is the full path (e.g. telegram/_files/something_i_cant_remember), `startswith` wouldn't cover that

harshil21:
> about dunder names coming first: do you want to, wait for pylint, add a custom test temporarily or just merge? I'm open for all :)

I think we can merge and then just update pylint dependency if they support it.

tests/test_modules.py
Bibo-Joshi: ah, makes sense

PR #2809, opened 2021-12-07
Api 5.5

### LINKED ISSUE #2808: [API] Api 5.5
Api 5.5 is here. Rejoice.

- [x] Doc note: Bots are now allowed to contact users who sent a join request to a chat where the bot is an administrator with the can_invite_users administrator right – even if the user never interacted with the bot before.

 - [x] Doc note/potential shortcut: Added support for mentioning users by their ID in inline keyboards. This will only work in Telegram versions released after December 7, 2021. Older clients will display unsupported message.
- [x] New methods, nedds shortcuts at : Added the methods banChatSenderChat and unbanChatSenderChat for banning and unbanning channel chats in supergroups and channels.
- [x] `Chat` addition: Added the field has_private_forwards to the class Chat for private chats, which can be used to check the possibility of mentioning the user by their ID.
- [x] `Chat`/`Message` addition: Added the field has_protected_content to the classes Chat and Message.
- [x] `Message` addition:  Added the field is_automatic_forward to the class Message.

- [x] Evaluate if worth adding to channel post/forward method docs: Note: After this update it will become impossible to forward messages from some chats. Use the fields has_protected_content in the classes Message and Chat to check this.

- [ ] Evaluate if worth adding to channel post: Note: After this update users are able to send messages on behalf of channels they own. Bots are expected to use the field sender_chat in the class Message to correctly support such messages. Edit: Decided to not add it.

Bibo-Joshi:
closes #2808

### Checklist for PRs

* New classes:
    
* Added new shortcuts:
    - [x] In `Chat` & `User` for all methods that accept `chat/user_id`
    
* If relevant:

    - [x] Added new filters for new message (sub)types
    - [x] Updated the Bot API version number in all places: `README.rst` and `README_RAW.rst` (including the badge), as well as `telegram.constants.BOT_API_VERSION`

### DIFF
--- a/telegram/bot.py
+++ b/telegram/bot.py
@@ -2409,6 +2409,58 @@ class Bot(TelegramObject):
         return result  # type: ignore[return-value]
 
     @log
+    def ban_chat_sender_chat(
+        self,
+        chat_id: Union[str, int],
+        sender_chat_id: int,
+        until_date: Union[int, datetime] = None,
+        timeout: ODVInput[float] = DEFAULT_NONE,
+        api_kwargs: JSONDict = None,
+    ) -> bool:
+        """
+        Use this method to ban a channel chat in a supergroup or a channel. The owner of the chat
+        will not be able to send messages and join live streams on behalf of the chat, unless it
+        is unbanned first. The bot must be an administrator in the supergroup or channel for this
+        to work and must have the appropriate administrator rights.
+
+         .. versionadded:: 13.9
+
+        Args:
+            chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target group or username
+                of the target supergroup or channel (in the format ``@channelusername``).
+            sender_chat_id (:obj:`int`): Unique identifier of the target sender chat
+            until_date (:obj:`int` | :obj:`datetime.datetime`, optional): Date when the sender chat
+                will be unbanned, unix time. If the chat is banned for more than 366 days or less
+                than 30 seconds from the current time they are considered to be banned forever.
+                For timezone naive :obj:`datetime.datetime` objects, the default timezone of the
+                bot will be used.
+            timeout (:obj:`int` | :obj:`float`, optional): If this value is specified, use it as
+                the read timeout from the server (instead of the one specified during creation of
+                the connection pool).
+            api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
+                Telegram API.
+
+        Returns:
+            :obj:`bool`: On success, :obj:`True` is returned.
+
+        Raises:
+            :class:`telegram.error.TelegramError`
+
+        """
+        data: JSONDict = {'chat_id': chat_id, 'sender_chat_id': sender_chat_id}
+
+        if until_date is not None:
+            if isinstance(until_date, datetime):
+                until_date = to_timestamp(
+                    until_date, tzinfo=self.defaults.tzinfo if self.defaults else None
+                )
+            data['until_date'] = until_date
+
+        result = self._post('banChatSenderChat', data, timeout=timeout, api_kwargs=api_kwargs)
+
+        return result  # type: ignore[return-value]
+
+    @log
     def unban_chat_member(
         self,
         chat_id: Union[str, int],
@@ -2449,6 +2501,43 @@ class Bot(TelegramObject):
             data['only_if_banned'] = only_if_banned
 
         result = self._post('unbanChatMember', data, timeout=timeout, api_kwargs=api_kwargs)
+
+        return result  # type: ignore[return-value]
+
+    @log
+    def unban_chat_sender_chat(
+        self,
+        chat_id: Union[str, int],
+        sender_chat_id: int,
+        timeout: ODVInput[float] = DEFAULT_NONE,
+        api_kwargs: JSONDict = None,
+    ) -> bool:
+        """Use this method to unban a previously banned channel in a supergroup or channel.
+        The bot must be an administrator for this to work and must have the
+        appropriate administrator rights.
+
+        .. versionadded:: 13.9
+
+        Args:
+            chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target chat or username
+                of the target supergroup or channel (in the format ``@channelusername``).
+            sender_chat_id (:obj:`int`): Unique identifier of the target sender chat.
+            timeout (:obj:`int` | :obj:`float`, optional): If this value is specified, use it as
+                the read timeout from the server (instead of the one specified during creation of
+                the connection pool).
+            api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
+                Telegram API.
+
+        Returns:
+            :obj:`bool` On success, :obj:`True` is returned.
+
+        Raises:
+            :class:`telegram.error.TelegramError`
+
+        """
+        data: JSONDict = {'chat_id': chat_id, 'sender_chat_id': sender_chat_id}
+
+        result = self._post('unbanChatSenderChat', data, timeout=timeout, api_kwargs=api_kwargs)
 
         return result  # type: ignore[return-value]
 
@@ -5499,10 +5588,14 @@ class Bot(TelegramObject):
     """Alias for :meth:`get_file`"""
     banChatMember = ban_chat_member
     """Alias for :meth:`ban_chat_member`"""
+    banChatSenderChat = ban_chat_sender_chat
+    """Alias for :meth:`ban_chat_sender_chat`"""
     kickChatMember = kick_chat_member
     """Alias for :meth:`kick_chat_member`"""
     unbanChatMember = unban_chat_member
     """Alias for :meth:`unban_chat_member`"""
+    unbanChatSenderChat = unban_chat_sender_chat
+    """Alias for :meth:`ban_chat_sender_chat`"""
     answerCallbackQuery = answer_callback_query
     """Alias for :meth:`answer_callback_query`"""
     editMessageText = edit_message_text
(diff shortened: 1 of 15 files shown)


telegram/chatjoinrequest.py
```diff
@@ -36,6 +36,12 @@ class ChatJoinRequest(TelegramObject):
 
     .. versionadded:: 13.8
 
+    Bots are now allowed to contact users who sent a join request to a chat where the bot is
+    an administrator with the can_invite_users administrator right – even if the user
+    never interacted with the bot before.
+
+    .. versionchanged:: 13.9
+
```
Bibo-Joshi, line 44:
```suggestion
    .. versionchanged:: 13.9
       Bots are now …

```

telegram/inline/inlinekeyboardbutton.py
```diff
@@ -45,6 +45,12 @@ class InlineKeyboardButton(TelegramObject):
 
           .. versionadded:: 13.6
 
+        * Added support for mentioning users by their ID in inline keyboards. This will only work
+          in Telegram versions released after December 7, 2021. Older clients will display
+          *unsupported message*.
+
+          .. versionadded:: 13.9
+
```
Bibo-Joshi, line 53: We didn't change anything here, right? Maybe rather something like "since bot API 5.5 it's allowed to …"?

telegram/user.py
```diff
@@ -233,6 +234,24 @@ def mention_html(self, name: str = None) -> str:
...
+
+        Args:
+            name (:obj:`str`): The name used as a link for the user. Defaults to :attr:`full_name`.
+
+        Returns:
+            :obj:`InlineKeyboardButton`: InlineButton with url set to the user mention
+        """
+        if name:
+            return InlineKeyboardButton(text=name, url=f"tg://user?id={self.id}")
+        return InlineKeyboardButton(text=self.full_name, url=f"tg://user?id={self.id}")
```
Bibo-Joshi, line 253: both `mention_html` and `mention_markdown` have similar f-strings for the link. Is it worth extracting that to another helper function?

telegram/bot.py
```diff
@@ -579,6 +579,14 @@ def forward_message(
     ) -> Message:
         """Use this method to forward messages of any kind. Service messages can't be forwarded.
 
+        Note:
+            Since the release of Bot API 5.5 it can be impossible to forward messages from
+            some chats. Use the field :attr:`telegram.Message.has_protected_content` and
```
Bibo-Joshi, line 584:
```suggestion
            some chats. Use the attributes :attr:`telegram.Message.has_protected_content` and
```

telegram/bot.py
```diff
@@ -579,6 +579,14 @@ def forward_message(
...
+            :attr:`telegram.Chat.has_protected_content` to check this.
+
+            As a workaround, it is still possible to use :meth:`copy_message`. However, this
+            behaviour is undocumented and might be changed by Telegram.
```
Bibo-Joshi, line 588: wut? this is stupid 🤦

telegram/bot.py
```diff
@@ -2408,6 +2416,45 @@ def ban_chat_member(
...
+    def ban_chat_sender_chat(
+        self,
+        chat_id: Union[str, int],
+        sender_chat_id: int,
+        timeout: ODVInput[float] = DEFAULT_NONE,
+        api_kwargs: JSONDict = None,
+    ) -> bool:
+        """
+        Use this method to ban a channel chat in a supergroup or a channel. Until the chat is
+        unbanned, the owner of the banned chat won't be able to send messages on behalf of *any of
```
Bibo-Joshi, line 2429:
```suggestion
        unbanned, the owner of the banned chat won't be able to send messages on behalf of *any of
```
`**any of their channels**`

telegram/chat.py
```diff
@@ -433,6 +457,52 @@ def ban_member(
             revoke_messages=revoke_messages,
         )
 
+    def ban_sender_chat(
```
Bibo-Joshi, line 460: what about another shortcut `ban_chat` for `bot.ban_chat_sender_chat(sender_chat_id=self.id,…)` so that one can do `message.sender_chat.ban`? same for unban

maybe you have suggestions for better namings. `banChatSenderChat` is a bit weird though - would have named it `banSenderChat` …

telegram/ext/filters.py
```diff
@@ -1965,16 +1965,16 @@ def remove_chat_ids(self, chat_id: SLT[int]) -> None:
 
     class sender_chat(_ChatUserBaseFilter):
         # pylint: disable=W0235
-        """Filters messages to allow only those which are from a specified sender chats chat ID or
+        """Filters messages to allow only those which are from a specified sender chat's chat ID or
```
Bibo-Joshi, line 1968: 👍 

We should add new filters for the new message attributes, too

telegram/message.py
```diff
@@ -1795,6 +1817,14 @@ def forward(
 
         For the documentation of the arguments, please see :meth:`telegram.Bot.forward_message`.
 
+        Note:
+            Since the release of Bot API 5.5 it can be impossible to forward messages from
+            some chats. Use the field :attr:`telegram.Message.has_protected_content` and
```
Bibo-Joshi, line 1822:
```suggestion
            some chats. Use the attributes :attr:`telegram.Message.has_protected_content` and
```

tests/test_bot.py
```diff
@@ -1012,6 +1012,16 @@ def test(url, data, *args, **kwargs):
         assert tz_bot.ban_chat_member(2, 32, until_date=until)
         assert tz_bot.ban_chat_member(2, 32, until_date=until_timestamp)
 
+    def test_ban_chat_sender_chat(self, monkeypatch, bot):
+        def test(url, data, *args, **kwargs):
```
Bibo-Joshi, line 1016:
```suggestion
        def make_assertion(url, data, *args, **kwargs):
```

telegram/bot.py
harshil21: yep changed this locally didn't commit yet

telegram/bot.py
harshil21: indeed, guess its not impossible after all

telegram/ext/filters.py
harshil21: yep doing that today

telegram/chat.py
harshil21: > what about another shortcut `ban_chat` for `bot.ban_chat_sender_chat(sender_chat_id=self.id,…)` so that one can do `message.sender_chat.ban`? same for unban

In `chat.py`? What if we add this shortcut in `message` instead? So one can do: `message.ban_chat()` which will call `bot.ban_chat_sender_chat(chat_id=self.chat_id, sender_chat_id=self.sender_chat.id)`

Bibo-Joshi: `Chat.bat_chat` was the idea, yes. I don't think we have any shortcut that passes along attributes of the current classes attributes - e.g. we also don't have `message.ban_chat_member`.
__if__ we want something like that, then I'd argue that it would be better to have it in `Update` (`update.ban_chat_sender_chat = bot.bat_chat_sender_chat(chat_id=update.effective_chat.id, sender_chat_id=update.effective_message.chat.id`). Would you be okay with postponing this idea? We can discuss it separately and if decide to go for it, we can also apply it for other cases that fit this scheme.

harshil21: on second thought, no I guess this isn't a good idea, there will be too many shortcuts / ambiguity / complexity. I'll just go with your initial suggestion

telegram/inline/inlinekeyboardbutton.py
Poolitzer: Well we usually relate to the PTB version and not the API version in the docs, not sure if I would want to break with that.

Bibo-Joshi: mh, my point is more that this works just fine with older versions of our lib. The only thing that change was that Telegram started accepting more kinds of input - when they rolled out API 5.5, not when we released v13.9 :)

telegram/chat.py
Bibo-Joshi: all right :)

telegram/user.py
Poolitzer: I dont think so, not for V13. We can discuss this for V14 imo.

telegram/inline/inlinekeyboardbutton.py
Poolitzer: hm, then maybe just the date is more fitting, then a random version number?

Bibo-Joshi: Or both 🙆🏻‍♂️

telegram/inline/inlinekeyboardbutton.py
```diff
@@ -76,12 +87,18 @@ class InlineKeyboardButton(TelegramObject):
             be launched when the user presses the button. This type of button must always be
             the ``first`` button in the first row.
         pay (:obj:`bool`, optional): Specify :obj:`True`, to send a Pay button. This type of button
-            must always be the ``first`` button in the first row.
+            must always be the ``first`` button in the first row and can only be used in invoice
```
Bibo-Joshi, line 90:
```suggestion
            must always be the `first` button in the first row and can only be used in invoice
```

tests/test_bot.py
```diff
@@ -1012,6 +1012,16 @@ def test(url, data, *args, **kwargs):
         assert tz_bot.ban_chat_member(2, 32, until_date=until)
         assert tz_bot.ban_chat_member(2, 32, until_date=until_timestamp)
 
+    def test_ban_chat_sender_chat(self, monkeypatch, bot):
```
Bibo-Joshi, line 1015:
```suggestion
    def test_ban_chat_sender_chat(self, monkeypatch, bot):
        # For now we just test that we pass the correct data to TG
```

Bibo-Joshi (commented):
IISC apart from the open comments, only updating the API version is missing, right?

Poolitzer (approved):
LGTM!

PR #2817, opened 2021-12-13
ChM.CREATOR -> Owner & use enums for dynamic types

Bibo-Joshi:
* renames `ChatMember.CREATOR` to `OWNER` since for some reason that's what TG named it even though the `type` is `'creator'` …
* and also `ChatMember.KICKED` -> `ChatMember.BANNED`
* for classes where we didn't hard code the type, try to resolve the type to the appropriate enum

### Checklist for PRs

* New classes:
    
* Added new shortcuts:
    
* If relevant:

### DIFF
--- a/examples/chatmemberbot.py
+++ b/examples/chatmemberbot.py
@@ -50,7 +50,7 @@ def extract_status_change(
         old_status
         in [
             ChatMember.MEMBER,
-            ChatMember.CREATOR,
+            ChatMember.OWNER,
             ChatMember.ADMINISTRATOR,
         ]
         or (old_status == ChatMember.RESTRICTED and old_is_member is True)
@@ -59,7 +59,7 @@ def extract_status_change(
         new_status
         in [
             ChatMember.MEMBER,
-            ChatMember.CREATOR,
+            ChatMember.OWNER,
             ChatMember.ADMINISTRATOR,
         ]
         or (new_status == ChatMember.RESTRICTED and new_is_member is True)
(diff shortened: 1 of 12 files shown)


Bibo-Joshi:
The tests fail, because:

* enum members have an attribute `__objclass__` which contains class such that `inspect` works correctly - see https://bugs.python.org/issue19263
* `BP.replace/insert_bot` doesn't handle types, so a warning is raised and the test fails

Note that the warning will also be issued when persisting other objects that use the new enum constants, e.g. `InlineQueryResult*` or `Poll` - it just didn't show yet. Having a warning raised by the classes PTB ships seems subobtimal. But so does special casing for enums … I see different options here:

* we're okay with the warning and go ahead with this
* we're not okay with the warning and don't use the enum members as constants
* we're not okay with the warning and add special casing for enums to `replace/insert`
* we postpone until we've had a more general discussion on `replace/insert_bot`

TBH I'm somewhat in favor of the last 😄

harshil21 (approved):
Changes look fine. Do you want to apply this to the other enums like `BotCommandScope` too?

Bibo-Joshi:
For those types of classes we have hard-coded the type, e.g.

https://github.com/python-telegram-bot/python-telegram-bot/blob/74c427047f60719834356c0445d13921470868ea/telegram/_botcommandscope.py#L174

`Poll`, `Chat` and `MessageEntity` were the only classes with a `type` attribute that don't have a subclass for every type …

Poolitzer (approved):
LGTM, +1 for defering, its not important right now imo

PR #2823, opened 2021-12-18
Add `Dispatcher.add_handlers`

### LINKED ISSUE #2821: [FEATURE] Adding multiple handlers at one

Calling `add_handler` multiple times maybe become tedious, so this proposes a new method to add multiple handlers at once.

A new method `Dispatcher.add_handlers` (trailing `s`!) with the following syntax:

```python
dispatcher.add_handlers(list/sequence_of_handlers, group)
```

should be a substitute for

```python
for handler in list_of_handlers:
    dispatcher.add_handler(handler, group=group)
```
and
```python
dispatcher.add_handler({1: [list_1], 2: [list_2]})
```
should be a shortcut for
```python
for group, handlers in my_dict.items():
    for handler in handlers:
        dispatcher.add_handler(handler, group=group)
```
Passing the `group` argument in this case should throw a `ValueError`

-

### Additional context

Proposed in  #2813

If you are interested in PRing for this, please leave a comment here.
For the tests to add, you can get inspiration e.g. [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/d064a5d8137d73cc814e3a2f77334b162f5a2844/tests/test_dispatcher.py#L407-L414).

harshil21:
Closes #2821 
Also removes redundant `Dispatcher.groups` attribute (as pointed out by @david-shiko)

### Checklist for PRs

### DIFF
--- a/examples/deeplinking.py
+++ b/examples/deeplinking.py
@@ -87,7 +87,7 @@ def deep_linked_level_3(update: Update, context: CallbackContext.DEFAULT_TYPE) -
     )
 
 
-def deep_link_level_3_callback(update: Update, context: CallbackContext.DEFAULT_TYPE) -> None:
+def deep_link_level_3_button(update: Update, context: CallbackContext.DEFAULT_TYPE) -> None:
     """Answers CallbackQuery with deeplinking url."""
     bot = context.bot
     url = helpers.create_deep_linked_url(bot.username, USING_KEYBOARD)
@@ -114,30 +114,24 @@ def main() -> None:
     # https://core.telegram.org/bots#deep-linking
 
     # Register a deep-linking handler
-    dispatcher.add_handler(
-        CommandHandler("start", deep_linked_level_1, filters.Regex(CHECK_THIS_OUT))
-    )
+    level_1 = CommandHandler("start", deep_linked_level_1, filters.Regex(CHECK_THIS_OUT))
 
-    # This one works with a textual link instead of an URL
-    dispatcher.add_handler(CommandHandler("start", deep_linked_level_2, filters.Regex(SO_COOL)))
+    # This one works with a textual link instead of a URL
+    level_2 = CommandHandler("start", deep_linked_level_2, filters.Regex(SO_COOL))
 
     # We can also pass on the deep-linking payload
-    dispatcher.add_handler(
-        CommandHandler("start", deep_linked_level_3, filters.Regex(USING_ENTITIES))
-    )
+    level_3 = CommandHandler("start", deep_linked_level_3, filters.Regex(USING_ENTITIES))
 
     # Possible with inline keyboard buttons as well
-    dispatcher.add_handler(
-        CommandHandler("start", deep_linked_level_4, filters.Regex(USING_KEYBOARD))
-    )
+    level_4 = CommandHandler("start", deep_linked_level_4, filters.Regex(USING_KEYBOARD))
 
     # register callback handler for inline keyboard button
-    dispatcher.add_handler(
-        CallbackQueryHandler(deep_link_level_3_callback, pattern=KEYBOARD_CALLBACKDATA)
-    )
+    level_3_button = CallbackQueryHandler(deep_link_level_3_button, pattern=KEYBOARD_CALLBACKDATA)
+
+    start_handler = CommandHandler("start", start)
 
     # Make sure the deep-linking handlers occur *before* the normal /start handler.
-    dispatcher.add_handler(CommandHandler("start", start))
+    dispatcher.add_handlers([level_1, level_2, level_3, level_4, level_3_button, start_handler])
 
     # Start the Bot
     updater.start_polling()
--- a/examples/deeplinking.py
+++ b/examples/deeplinking.py
@@ -87,7 +87,7 @@ def deep_linked_level_3(update: Update, context: CallbackContext.DEFAULT_TYPE) -
     )
 
 
-def deep_link_level_3_button(update: Update, context: CallbackContext.DEFAULT_TYPE) -> None:
+def deep_link_level_3_callback(update: Update, context: CallbackContext.DEFAULT_TYPE) -> None:
     """Answers CallbackQuery with deeplinking url."""
     bot = context.bot
     url = helpers.create_deep_linked_url(bot.username, USING_KEYBOARD)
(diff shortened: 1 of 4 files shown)


telegram/ext/_dispatcher.py
```diff
@@ -573,11 +570,43 @@ def add_handler(self, handler: Handler[UT, CCT], group: int = DEFAULT_GROUP) ->
...
             self.handlers[group] = []
-            self.groups.append(group)
-            self.groups = sorted(self.groups)
 
         self.handlers[group].append(handler)
 
+    def add_handlers(
+        self,
+        handlers: Union[Sequence[Handler], Dict[int, Sequence[Handler]]],
+        group: int = DEFAULT_GROUP,
```
Bibo-Joshi, line 579: Maybe using `DefaultValue` here is safer - that way we know for user that the user didn't pass `group = DEFAULT_GROUP`

telegram/ext/_dispatcher.py
```diff
@@ -573,11 +570,43 @@ def add_handler(self, handler: Handler[UT, CCT], group: int = DEFAULT_GROUP) ->
...
+
+        Args:
+            handlers (Sequence[:obj:`telegram.ext.Handler`] | \
+                Dict[int, Sequence[:obj:`telegram.ext.Handler`]]): \
+                Specify a sequence of handlers *or* a dictionary where the keys are groups and
+                values are handlers.
+            group (:obj:`int`, optional): Specify which group the sequence of ``handlers``
+                should be added to. Defaults to ``0``.
+
+        .. seealso:: :meth:`add_handler`
```
Bibo-Joshi, line 593: move before the argument list

telegram/ext/_dispatcher.py
```diff
@@ -573,11 +570,43 @@ def add_handler(self, handler: Handler[UT, CCT], group: int = DEFAULT_GROUP) ->
...
-            self.groups = sorted(self.groups)
 
         self.handlers[group].append(handler)
 
+    def add_handlers(
+        self,
+        handlers: Union[Sequence[Handler], Dict[int, Sequence[Handler]]],
+        group: int = DEFAULT_GROUP,
+    ) -> None:
+        """Register multiple handlers at once. Priority/order of handlers matters.
```
Bibo-Joshi, line 581:
```suggestion
        """Registers multiple handlers at once. The order of the handlers in the passed sequence(s) matters. See :meth:`add_handler` for details.
```

telegram/ext/_dispatcher.py
```diff
@@ -573,11 +570,43 @@ def add_handler(self, handler: Handler[UT, CCT], group: int = DEFAULT_GROUP) ->
...
+                values are handlers.
+            group (:obj:`int`, optional): Specify which group the sequence of ``handlers``
+                should be added to. Defaults to ``0``.
+
+        .. seealso:: :meth:`add_handler`
+        """
+        if isinstance(handlers, dict) and group != DEFAULT_GROUP:
+            raise ValueError('The `group` argument can only be used with a sequence of handlers.')
+
+        if isinstance(handlers, list):
```
Bibo-Joshi, line 598: this will fail, if you pass a tuple or another `Sequence` type. A quick solution would be to just swap to

```python
if isinstance(handlers, dict):
    ...
else:
```

Or, if you want to make it as general as possible, you could use `typing.Mapping` instead of `typing.Dict` - then you'd have to use use either `instance(handlers, collections.abc.Sequence)` or `… abc.Mapping)`.
But I'm fully okay with annotation as `Dict` for the mapping - I'd also be okay with annotating just as `Union[Tuple, List]` instead of `Sequence`.

telegram/ext/_dispatcher.py
```diff
@@ -482,7 +479,7 @@ def process_update(self, update: object) -> None:
         handled = False
         sync_modes = []
 
-        for group in self.groups:
+        for group in sorted(self.handlers):  # lower -> higher groups
```
Bibo-Joshi, line 482: sorting on every `process_update` seems like a bad idea, even if one usually only has a small number of groups. Let's sort `self.handlers` in `add_handler` - maybe we can even just insert new groups directly in the right order? 🤔

tests/test_dispatcher.py
```diff
@@ -413,6 +413,60 @@ def test_groups(self, dp):
...
+        photo_update = Update(2, message=Message(2, None, None, photo=True))
+        dp.update_queue.put(self.message_update)  # Putting updates in the queue calls the callback
+        dp.update_queue.put(photo_update)
+        sleep(0.1)  # sleep is required otherwise there is random behaviour
+
+        # Test if handler was added to correct group with correct order-
+        assert (
+            self.count == 2
+            and len(dp.handlers[1]) == 3
+            and dp.handlers[1][0].filters.name == 'filters.TEXT'
```
Bibo-Joshi, line 433: why not just `handlers[1][0] is msg_handler_set_count`?

tests/test_dispatcher.py
```diff
@@ -413,6 +413,60 @@ def test_groups(self, dp):
...
+        user_update = Update(3, message=Message(3, None, None, from_user=User(1, 's', True)))
+        voice_update = Update(4, message=Message(4, None, None, voice=True))
+        dp.update_queue.put(user_update)
+        dp.update_queue.put(voice_update)
+        sleep(0.1)
+
+        assert (
+            self.count == 4
+            and len(dp.handlers[1]) == 5
+            and dp.handlers[1][-1].filters.name == 'filters.VOICE'
```
Bibo-Joshi, line 456: see above

examples/deeplinking.py
```diff
@@ -114,30 +114,24 @@ def main() -> None:
...
-    dispatcher.add_handler(
-        CallbackQueryHandler(deep_link_level_3_callback, pattern=KEYBOARD_CALLBACKDATA)
-    )
+    level_3_button = CallbackQueryHandler(deep_link_level_3_button, pattern=KEYBOARD_CALLBACKDATA)
+
+    start_handler = CommandHandler("start", start)
 
     # Make sure the deep-linking handlers occur *before* the normal /start handler.
-    dispatcher.add_handler(CommandHandler("start", start))
+    dispatcher.add_handlers([level_1, level_2, level_3, level_4, level_3_button, start_handler])
```
Bibo-Joshi, line 134: I'd vote not to change examples for this.

telegram/ext/_dispatcher.py
```diff
@@ -573,11 +570,43 @@ def add_handler(self, handler: Handler[UT, CCT], group: int = DEFAULT_GROUP) ->
...
+        .. seealso:: :meth:`add_handler`
+        """
+        if isinstance(handlers, dict) and group != DEFAULT_GROUP:
+            raise ValueError('The `group` argument can only be used with a sequence of handlers.')
+
+        if isinstance(handlers, list):
+            for handler in handlers:
+                self.add_handler(handler, group)
+
+        elif isinstance(handlers, dict):
```
Bibo-Joshi, line 602:
```suggestion
        else:
```

Bibo-Joshi (changes requested):

telegram/ext/_dispatcher.py
```diff
@@ -573,11 +571,53 @@ def add_handler(self, handler: Handler[UT, CCT], group: int = DEFAULT_GROUP) ->
...
+                    self.add_handler(handler, handler_group)
+
+        elif isinstance(handlers, (list, tuple)):
+            for handler in handlers:
+                self.add_handler(handler, DefaultValue.get_value(group))
+
+        else:
+            raise ValueError(
+                "The `handlers` argument must be a sequence of handlers or a "
+                "dictionary where the keys are groups and values are handlers."
```
Bibo-Joshi, line 618:
```suggestion
                "dictionary where the keys are groups and values are sequences of handlers."
```

Bibo-Joshi (commented):
Thanks for the updates! one minor nitpick.
@Poolitzer are you interested in reviewing as well or are you okay if we merge?
Mentioning @david-shiko as well, since you proposed this feature :)

telegram/ext/_dispatcher.py
```diff
@@ -573,11 +571,53 @@ def add_handler(self, handler: Handler[UT, CCT], group: int = DEFAULT_GROUP) ->
...
+        if isinstance(handlers, dict) and not isinstance(group, DefaultValue):
+            raise ValueError('The `group` argument can only be used with a sequence of handlers.')
+
+        if isinstance(handlers, dict):
+            for handler_group, grp_handlers in handlers.items():
+                if not isinstance(grp_handlers, (list, tuple)):
+                    raise ValueError(f'Handlers for group {handler_group} must be a list or tuple')
+
+                for handler in grp_handlers:
+                    self.add_handler(handler, handler_group)
```
Poolitzer, line 609: we dont do a check before this if handler is actually an instance of Handler, which we do in the `add_handler` function. I would add this.

Poolitzer (changes requested):
Cool addition I like it. I would just add one check.

telegram/ext/_dispatcher.py
harshil21: but wouldn't that be redundant? It would raise an error in `add_handler` anyway.

Poolitzer: yeah, sure. no idea how I missed you literally call add_handler there. Nevermind this.

Poolitzer (approved):

Issue #2732, opened 2021-10-14
[FEATURE] Purge persistent chat_data from empty entries


ria4:
### What kind of feature are you missing? Where do you notice a shortcoming of PTB?

When using persistent chat_data, there's no method for purging entries which have been emptied. The chat_id key will always remain in the chat_data defaultdict. I believe it's the same for user_data.

Because of the `Dispatcher.__update_persistence()` implementation, even deleting the keys with `del chat_data[k]` won't remove them from the persistence file. See: https://github.com/python-telegram-bot/python-telegram-bot/blob/bc7c422a11a60a3064601aac3fd5a26fd9b45ae9/telegram/ext/dispatcher.py#L665-L669

### Describe the solution you'd like

- Add a method to the Dispatcher class for removing a key from the persistence file, with the key as parameter. It should not matter to the method whether the related value is empty or not.
- Add another method for purging all orphaned keys from the persistence file.

There probably should be 4 methods, 2 for `chat_data` and 2 for `user_data`.

ria4:
I agree on your first preference. Updating `BasePersistence.update_chat/user_data` with additional behaviour for data deletion seems clumsy.

I'm less sure about the second item. I like the consistency of making `user/chat_data` read-only, since `get/update` methods have already been defined and should definitely be preferred over direct editing of the `defaultdict`. But there's always a risk that restricting access on the object makes it really hard to accomplish operations we do not expect right now. Also, this would be a breaking change.

Finally, I believe it's definitely worth to offer an additional `purge_chat_data()` rather than to rely on a very unclear and accident-prone `drop_chat_data(<no id here, which actually means every id associated with empty values...>)`.

Bibo-Joshi:
> But there's always a risk that restricting access on the object makes it really hard to accomplish operations we do not expect right now.

I don't think this could be an issue. What we would effectively be doing with any solution would be to control how keys are deleted from the dict. All other dict-operations would remain intact and deleting wouldn't even be disallowed - just moved at most.
The worst you would get is some semantic asymmetries, e.g. that `user_data.get(key, None)` would still work while `user_data.pop(key, None` wouldn't.

> Also, this would be a breaking change.

This is actually not an issue atm - we're working towards v14, wich will be breaking anyway, which is why we don't try hard to make other changes non-breaking :)

PR #2842, opened 2022-01-03
Add move_chat_data and move_user_data to Dispatcher

DonalDuck004:
I have added move_chat_data and move_user_data to telegram/ext/dispatcher.py, this methods allow to move (and copy) the user/chat_data content for one chat to another chat.
This can be useful when, for example, a group becomes a supergroup

### DIFF
--- a/telegram/ext/dispatcher.py
+++ b/telegram/ext/dispatcher.py
@@ -653,14 +653,14 @@ class Dispatcher(Generic[CCT, UD, CD, BD]):
                 self.groups.remove(group)
 
     def move_chat_data(self, old_id: int, new_id: int, update_values: bool = False, copy_values: bool = False) -> None:
-        """Move chat_data content for old_id to new_id
+        """Move chat_data content for old_id to new_id.
+
         Args:
             old_id (:obj:`int`, required): The old chat_id
             new_id (:obj:`int`, required): The new chat_id
             update_values (:obj:`bool`, optional): if True, dict.update will be used else, an assigment operation will be performed. Set to False by default
             copy_values (:obj:`bool`, optional): if False, the user_data of old_id will be deleted. Set to True by default
-	"""
-
+        """
         if not isinstance(old_id, int) or not isinstance(new_id, int):
             raise ValueError("old_id, and new_id must be integers")
 
@@ -675,14 +675,14 @@ class Dispatcher(Generic[CCT, UD, CD, BD]):
         self.__update_persistence()
 
     def move_user_data(self, old_id: int, new_id: int, update_values: bool = False, copy_values: bool = False) -> None:
-        """Move user_data content from old_id to new_id
+        """Move user_data content for old_id to new_id
+
         Args:
             old_id (:obj:`int`, required): The old user_id
             new_id (:obj:`int`, required): The new user_id
             update_values (:obj:`bool`, optional): if True, dict.update will be used else, an assigment operation will be performed. Set to False by default
             copy_values (:obj:`bool`, optional): if False, the user_data of old_id will be deleted. Set to False by default
-	"""
-
+        """
         if not isinstance(old_id, int) or not isinstance(new_id, int):
             raise ValueError("old_id, and new_id must be integers")
 
--- a/telegram/ext/dispatcher.py
+++ b/telegram/ext/dispatcher.py
@@ -652,49 +652,57 @@ class Dispatcher(Generic[CCT, UD, CD, BD]):
                 del self.handlers[group]
                 self.groups.remove(group)
 
-    def move_chat_data(self, old_id: int, new_id: int, update_values: bool = False, copy_values: bool = False) -> None:
+    def move_chat_data(self,
+                       old_id: int,
+                       new_id: int,
+                       update_values: bool = False,
+                       copy_values: bool = False) -> None:
         """Move chat_data content for old_id to new_id.
 
         Args:
             old_id (:obj:`int`, required): The old chat_id
             new_id (:obj:`int`, required): The new chat_id
-            update_values (:obj:`bool`, optional): if True, dict.update will be used else, an assigment operation will be performed. Set to False by default
+            update_values (:obj:`bool`, optional): if True, dict.update will be used else, an override operation will be performed. Set to False by default
             copy_values (:obj:`bool`, optional): if False, the user_data of old_id will be deleted. Set to True by default
         """
         if not isinstance(old_id, int) or not isinstance(new_id, int):
-            raise ValueError("old_id, and new_id must be integers")
+            raise ValueError("old_id and new_id must be integers")
 
         if update_values:
-            self.chat_data.setdefault(new_id, {}).update(self.chat_data.get(old_id, {}))
+            self.chat_data[new_id].update(self.chat_data[old_id])
         else:
-            self.chat_data[new_id] = self.chat_data.get(old_id, {})
+            self.chat_data[new_id] = self.chat_data[old_id]
 
         if copy_values is False:
-            self.chat_data.pop(old_id, None)
+            del self.chat_data[old_id]
 
-        self.__update_persistence()
+        self.update_persistence()
 
-    def move_user_data(self, old_id: int, new_id: int, update_values: bool = False, copy_values: bool = False) -> None:
+    def move_user_data(self,
+                       old_id: int,
+                       new_id: int,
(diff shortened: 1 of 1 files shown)


telegram/ext/dispatcher.py
```diff
@@ -652,6 +652,50 @@ def remove_handler(self, handler: Handler, group: int = DEFAULT_GROUP) -> None:
                 del self.handlers[group]
                 self.groups.remove(group)
 
+    def move_chat_data(self, old_id: int, new_id: int, update_values: bool = False, copy_values: bool = False) -> None:
+        """Move chat_data content for old_id to new_id.
+
+        Args:
+            old_id (:obj:`int`, required): The old chat_id
+            new_id (:obj:`int`, required): The new chat_id
+            update_values (:obj:`bool`, optional): if True, dict.update will be used else, an assigment operation will be performed. Set to False by default
```
Bibo-Joshi, line 661: maybe `override` is a more speaking name?

telegram/ext/dispatcher.py
```diff
@@ -652,6 +652,50 @@ def remove_handler(self, handler: Handler, group: int = DEFAULT_GROUP) -> None:
...
+            copy_values (:obj:`bool`, optional): if False, the user_data of old_id will be deleted. Set to True by default
+        """
+        if not isinstance(old_id, int) or not isinstance(new_id, int):
+            raise ValueError("old_id, and new_id must be integers")
+
+        if update_values:
+            self.chat_data.setdefault(new_id, {}).update(self.chat_data.get(old_id, {}))
```
Bibo-Joshi, line 668: `setdefault` and `get` are not necessary as `chat_data` is already a `defaultdict`

telegram/ext/dispatcher.py
```diff
@@ -652,6 +652,50 @@ def remove_handler(self, handler: Handler, group: int = DEFAULT_GROUP) -> None:
...
+        else:
+            self.chat_data[new_id] = self.chat_data.get(old_id, {})
+
+        if copy_values is False:
+            self.chat_data.pop(old_id, None)
+
+        self.__update_persistence()
```
Bibo-Joshi, line 675: Please use `update_persistence` instead

Bibo-Joshi (commented):
Hi. Thanks for your PR!

I like the idea of having a more straightforward way to handle chat migration than what we currently describe [in the wiki](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Storing-bot%2C-user-and-chat-related-data#chat-migration) 👍 
However, I'm not sure if I see the need for coyping instead of moving and for `user_data` related logic. Can you give use case examples for copying user/chat data to another ID and moving user data?

In any case, we currently don't accept new features for the master branch (except for API updates), so if we go forward with this, I'd have to ask you to rebase your PR onto the `v14` branch. 
Also this PR would probably benefit from the changes proposed in #2732 - just as a note …

Poolitzer:
I would like to suggest adding builder to do this "automatically" for the end-user in Dispatcher. If they call it, upon receiving a chat migrated update, the move_chat_data should be called.

Bibo-Joshi:
yes, a `Dispatcher.migrate_chat_data` is what came to my mind when seeing this PR. It could accept a `Message`. but let's first see, which use cases OP had in mind :)

DonalDuck004:
Copying chat_data can be useful in cases like this [https://pastebin.com/1UJfxKk0](https://pastebin.com/1UJfxKk0), but honestly I added move user_data just for linearity 😅, should I remove it?

Bibo-Joshi:
IISC the use case you for copying instead of moving chat data to a new chat_id is to apply some kind of configuration that the user has made in the "old_id" chat - did I get that right?
If so: fair enough, I can see the use case for `chat_data`. Still, I'm not entirely convinced if we should add a utility method even for the `chat_data` use case. My three reasons are

1. doing it manually is really not hard: `context.chat_data.update(context.dispatcher.chat_data[old_id])`
2. My impression is that it's a rare case that you want to apply _all_ of `chat_data[old_id]`. even in your mwe, you only really need `chat_data[old_id]['settings']`
3. with `update_value = True`, this will only work if `context.chat_data` is a dict. Since the introduction of `ContextTypes` (v13.6?), `context.chat_data` may be a user defined type, so "merging" the "old_id" data into that is not always possible. For chat migration, I would argue that simply replacing `dp.chat_data[new_id]` is enough since the main inteded use case would be to handle chat migration status messages - i.e. there's no data in `dp.chat_data[new_id]` anyway

For `user_data` I'm less convinced of the use case …

@DonalDuck004 can you understand my arguments? Am I overlooking something that makes `move_chat_data` in the current generality worth adding in your opinion?

DonalDuck004:
Effectively you are right, but I still think that a specific method to manage the migration of a chat, is not a bad idea ... Maybe renaming it giving it the name `migrate_chat_data` as you suggested. However, the final decision is yours

Bibo-Joshi:
> Effectively you are right, but I still think that a specific method to manage the migration of a chat, is not a bad idea

Yes, I'm completely on board with that 👍 But if you agree that the other use cases are rare/specific, I'd like to focus on a `migrate_chat_data` function :)
Would you like to adapt your PR for that/open a new one?

IMO, that method should do the following:
* accept either a `Message` object and compute `old/new_id` from it as shown in the wiki section or accept `old/new_id` directly. We can even support both.
* always override the data in `new_id` and discard the data in `old_id` entry, since merging is not always possible
* include a warning about this fact in the docstring

@Poolitzer am I missing anything?

Poolitzer:
yes I am fine with these three points, but I still had my builder idea which I think I still haven't conveyed properly.

After implementing this functionality, I would love to have [this wiki snippet](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Storing-bot%2C-user-and-chat-related-data#status-updates-sent-by-telegram) integrated into Dispatcher. My reasoning behind is that every bot which works in groups would need to implement this snippet in order to work properly. We can spare users of our library this integration and make it work behind the scenes, so people can't forget that this is a thing (I did in a plethora of my projects). Thus I would vote to make it a Builder Call, which is enabled by default and can be disabled if wanted. This will be enough for every project which just uses chat_data, which is great. Projects with more complex data structure need to make this snippet anyway and do their specific migrate actions but that's nothing we can help with apart from showing how to do it with the snippet I guess.

DonalDuck004:
> > Effectively you are right, but I still think that a specific method to manage the migration of a chat, is not a bad idea
> 
> Yes, I'm completely on board with that 👍 But if you agree that the other use cases are rare/specific, I'd like to focus on a `migrate_chat_data` function :) Would you like to adapt your PR for that/open a new one?
> 
> IMO, that method should do the following:
> 
> * accept either a `Message` object and compute `old/new_id` from it as shown in the wiki section or accept `old/new_id` directly. We can even support both.
> * always override the data in `new_id` and discard the data in `old_id` entry, since merging is not always possible
> * include a warning about this fact in the docstring
> 
> @Poolitzer am I missing anything?

I think opening a new PR is best, also because I have to point to the V14 branch

Bibo-Joshi:
@DonaldDuck004 all right, then I'll close this one :) looking forward to your PR!

@Poolitzer Okay, now I understand what you mean. We touched that idea in [dev chat](https://t.me/c/1494805131/97) back when I made GitHubBot Revised crash by converting the dev chat 😄 I do see why this might desirable, but I somewhat agree with Pieters argument here: making `Dispatcher` automatically take action based on updates is not really what `Dispatcher` is there for.

Even worse, if we add a handler for that via `add_handler`, the handler might interfere with the users handlers. Unlikely, but still.
Additionally, doing this is not even straight forward: When a chat migrates, you get _two_ messages, one with `migrate_to_chat_id` and one with `migrate_from_chat_id` (the order is not guaranteed). If you have a `MessageHandler(filters.all, …)` and it receives the `migrate_to_chat_id` update, `update.effective_chat.id` will still be the old id. Should `chat_data` already be the new entry? In case something inside `chat_data` depends one the `chat_id`, this is relevant. Converse case for `migrate_from_chat_id`.

Both things are surely no big deal in _most_ cases, but we'd still be making a number of assumptions that a user would need to be aware of when using an automated transition. In contrast adding `add_handler(MessageHandler(filters.chat_migrated, lambda u,c: c.dispatcher.migrate_chat_data(u.effective_message)))` is a one liner and the user knows exactly what they get.
Can you understand my line of thought?

PR #2843, opened 2022-01-04
Update a test and some formatting fixes

harshil21:


Bibo-Joshi:
Thanks for the PR! as mentioned, most of the deepsource failures are false positives and I've already reported them as such.
the only valid one that I see is 

image

which should be solvable by doing that import in a `if typing.TYPE_CHEKCING` clause - could you add this to this PR?

PR #2852, opened 2022-01-10
Add `Dispatcher/Persistence.drop_chat/user_data`

### LINKED ISSUE #2732: [FEATURE] Purge persistent chat_data from empty entries

harshil21:
Closes #2732 

### Checklist for PRs

Breaking changes:
- `dp.chat/user_data` is now read only
- `BasePersistence.drop_chat/user_data` is an abstractmethod

### DIFF
--- a/telegram/ext/_dictpersistence.py
+++ b/telegram/ext/_dictpersistence.py
@@ -361,6 +361,30 @@ class DictPersistence(BasePersistence):
         self._callback_data = (data[0], data[1].copy())
         self._callback_data_json = None
 
+    def drop_chat_data(self, chat_id: int) -> None:
+        """Will delete the specified key from the :attr:`chat_data`.
+
+        Args:
+            chat_id (:obj:`int`): The chat id to delete from the persistence.
+        """
+        if self._chat_data is None:
+            self._chat_data = defaultdict(dict)
+        else:
+            del self._chat_data[chat_id]
+            self._chat_data_json = None
+
+    def drop_user_data(self, user_id: int) -> None:
+        """Will delete the specified key from the :attr:`user_data`.
+
+        Args:
+            user_id (:obj:`int`): The user id to delete from the persistence.
+        """
+        if self._user_data is None:
+            self._user_data = defaultdict(dict)
+        else:
+            del self._user_data[user_id]
+            self._user_data_json = None
+
     def refresh_user_data(self, user_id: int, user_data: Dict) -> None:
         """Does nothing.
 
--- a/telegram/ext/_dictpersistence.py
+++ b/telegram/ext/_dictpersistence.py
@@ -364,6 +364,8 @@ class DictPersistence(BasePersistence):
     def drop_chat_data(self, chat_id: int) -> None:
         """Will delete the specified key from the :attr:`chat_data`.
 
+        .. versionadded:: 14.0
+
         Args:
             chat_id (:obj:`int`): The chat id to delete from the persistence.
         """
@@ -375,6 +377,8 @@ class DictPersistence(BasePersistence):
 
     def drop_user_data(self, user_id: int) -> None:
         """Will delete the specified key from the :attr:`user_data`.
+
+        .. versionadded:: 14.0
 
         Args:
             user_id (:obj:`int`): The user id to delete from the persistence.
--- a/telegram/ext/_dictpersistence.py
+++ b/telegram/ext/_dictpersistence.py
@@ -370,10 +370,9 @@ class DictPersistence(BasePersistence):
             chat_id (:obj:`int`): The chat id to delete from the persistence.
         """
         if self._chat_data is None:
-            self._chat_data = defaultdict(dict)
-        else:
-            del self._chat_data[chat_id]
-            self._chat_data_json = None
+            return
+        self._chat_data.pop(chat_id, None)
+        self._chat_data_json = None
 
     def drop_user_data(self, user_id: int) -> None:
         """Will delete the specified key from the :attr:`user_data`.
@@ -384,10 +383,9 @@ class DictPersistence(BasePersistence):
             user_id (:obj:`int`): The user id to delete from the persistence.
         """
         if self._user_data is None:
-            self._user_data = defaultdict(dict)
-        else:
-            del self._user_data[user_id]
-            self._user_data_json = None
+            return
+        self._user_data.pop(user_id, None)
+        self._user_data_json = None
 
     def refresh_user_data(self, user_id: int, user_data: Dict) -> None:
         """Does nothing.
--- a/telegram/ext/_dispatcher.py
+++ b/telegram/ext/_dispatcher.py
@@ -113,6 +113,24 @@ class Dispatcher(Generic[BT, CCT, UD, CD, BD, JQ, PT]):
             instance to pass onto handler callbacks.
         workers (:obj:`int`, optional): Number of maximum concurrent worker threads for the
             ``@run_async`` decorator and :meth:`run_async`.
+        chat_data (:obj:`types.MappingProxyType`): A dictionary handlers can use to store data for
+            the chat.
+
+            .. versionchanged:: 14.0
+                :attr:`chat_data` is now read-only
+
+            .. tip::
+               Manually modifying :attr:`chat_data` is almost never needed and unadvisable.
+
+        user_data (:obj:`types.MappingProxyType`): A dictionary handlers can use to store data for
+            the user.
+
+            .. versionchanged:: 14.0
+               :attr:`user_data` is now read-only
+
+            .. tip::
+               Manually modifying :attr:`user_data` is almost never needed and unadvisable.
+
         bot_data (:obj:`dict`): A dictionary handlers can use to store data for the bot.
         persistence (:class:`telegram.ext.BasePersistence`): Optional. The persistence class to
             store data that should be persistent over restarts.
@@ -145,7 +163,9 @@ class Dispatcher(Generic[BT, CCT, UD, CD, BD, JQ, PT]):
         'update_queue',
         'job_queue',
         '_user_data',
+        'user_data',
         '_chat_data',
+        'chat_data',
         'bot_data',
         '_update_persistence_lock',
         'handlers',
@@ -198,10 +218,17 @@ class Dispatcher(Generic[BT, CCT, UD, CD, BD, JQ, PT]):
                 stacklevel=stack_level,
             )
 
-        self._user_data: DefaultDict[int, UD] = defaultdict(self.context_types.user_data)
-        self._chat_data: DefaultDict[int, CD] = defaultdict(self.context_types.chat_data)
+        self._user_data: DefaultDict[int, UD]
+        self._chat_data: DefaultDict[int, CD]
+        # Read only mapping-
+        self.user_data: Mapping[int, UD]
+        self.chat_data: Mapping[int, CD]
+
+        self._set_user_data(defaultdict(self.context_types.user_data))
+        self._set_chat_data(defaultdict(self.context_types.chat_data))
         self.bot_data = self.context_types.bot_data()
-        self.persistence: Optional[BasePersistence] = None
+
+        self.persistence: Optional[BasePersistence]
         self._update_persistence_lock = Lock()
         if persistence:
             if not isinstance(persistence, BasePersistence):
@@ -213,11 +240,11 @@ class Dispatcher(Generic[BT, CCT, UD, CD, BD, JQ, PT]):
             self.persistence.set_bot(self.bot)
 
             if self.persistence.store_data.user_data:
-                self._user_data = self.persistence.get_user_data()
+                self._set_user_data(self.persistence.get_user_data())
                 if not isinstance(self._user_data, defaultdict):
                     raise ValueError("user_data must be of type defaultdict")
             if self.persistence.store_data.chat_data:
-                self._chat_data = self.persistence.get_chat_data()
+                self._set_chat_data(self.persistence.get_chat_data())
                 if not isinstance(self._chat_data, defaultdict):
                     raise ValueError("chat_data must be of type defaultdict")
             if self.persistence.store_data.bot_data:
@@ -282,27 +309,6 @@ class Dispatcher(Generic[BT, CCT, UD, CD, BD, JQ, PT]):
         cls.logger.debug('Setting singleton dispatcher as %s', val)
         cls.__singleton = weakref.ref(val) if val else None
 
-    @property
-    def chat_data(self) -> Mapping[int, CD]:
-        """
-        :obj:`types.MappingProxyType`: A dictionary handlers can use to store data for the chat.
-
-        .. versionchanged:: 14.0
-           :attr:`chat_data` is now read-only
-
-        """
-        return MappingProxyType(self._chat_data)
-
-    @property
-    def user_data(self) -> Mapping[int, UD]:
-        """:obj:`types.MappingProxyType`: A dictionary handlers can use to store data for the user.
-
-        .. versionchanged:: 14.0
-           :attr:`user_data` is now read-only
-
-        """
-        return MappingProxyType(self._user_data)
-
     @classmethod
     def get_instance(cls) -> 'Dispatcher':
         """Get the singleton instance of this class.
@@ -317,6 +323,18 @@ class Dispatcher(Generic[BT, CCT, UD, CD, BD, JQ, PT]):
         if cls.__singleton is not None:
             return cls.__singleton()  # type: ignore[return-value] # pylint: disable=not-callable
         raise RuntimeError(f'{cls.__name__} not initialized or multiple instances exist')
+
+    def _set_chat_data(self, data: DefaultDict[int, CD]) -> None:
+        """Used for assigning a new value to the underlying chat_data dictionary. Also updates
+        the read-only mapping."""
+        self._chat_data = data
+        self.chat_data = MappingProxyType(self._chat_data)
+
+    def _set_user_data(self, data: DefaultDict[int, UD]) -> None:
+        """Used for assigning a new value to the underlying user_data dictionary. Also updates
+        the read-only mapping."""
+        self._user_data = data
+        self.user_data = MappingProxyType(self._user_data)
 
     def _pooled(self) -> None:
         thr_name = current_thread().name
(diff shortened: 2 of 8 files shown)


telegram/ext/_dispatcher.py
```diff
@@ -631,6 +652,78 @@ def remove_handler(self, handler: Handler, group: int = DEFAULT_GROUP) -> None:
...
+        """Used for deleting a key from the :attr:`chat_data`.
+
+        .. versionadded:: 14.0
+
+        Args:
+            chat_id (:obj:`int`, optional): The chat id to delete from the persistence. The entry
+                will be deleted even if it is not empty. Mutually exclusive with
+                :paramref:`all_empty_entries`.
+            all_empty_entries (:obj:`bool`, optional): If :obj:`True`, all empty entries from
+                :attr:`chat_data` will be deleted. Mutually exclusive with :paramref:`chat_id`.
```
Bibo-Joshi, line 665: "empty" doesn't necessarily have a meaning when using a custom type for `chat_data`. Maybe we could use smth like "evaluates to False in boolean context i.e. ``bool(entry)=False``"? 🤔

telegram/ext/_dispatcher.py
```diff
@@ -631,6 +652,78 @@ def remove_handler(self, handler: Handler, group: int = DEFAULT_GROUP) -> None:
...
+
+        Raises:
+            :exc:`ValueError`: When both :paramref:`chat_id` and :paramref:`all_empty_entries` are
```
Bibo-Joshi, line 668: just wordering: how does `paramref` render while we don't have support for it?

telegram/ext/_dispatcher.py
```diff
@@ -631,6 +652,78 @@ def remove_handler(self, handler: Handler, group: int = DEFAULT_GROUP) -> None:
...
+                provided, or none of them are passed.
+        """
+        if chat_id and all_empty_entries:
+            raise ValueError("You must pass either `chat_id` or `all_empty_entries` not both.")
+
+        if chat_id is None and not all_empty_entries:
+            raise ValueError("You must pass either `chat_id` or `all_empty_entries`.")
```
Bibo-Joshi, line 675: `not (chat_id ^ all_emty_entries)`

telegram/ext/_dispatcher.py
```diff
@@ -631,6 +652,78 @@ def remove_handler(self, handler: Handler, group: int = DEFAULT_GROUP) -> None:
...
+        Args:
+            chat_id (:obj:`int`, optional): The chat id to delete from the persistence. The entry
+                will be deleted even if it is not empty. Mutually exclusive with
+                :paramref:`all_empty_entries`.
+            all_empty_entries (:obj:`bool`, optional): If :obj:`True`, all empty entries from
+                :attr:`chat_data` will be deleted. Mutually exclusive with :paramref:`chat_id`.
+
+        Raises:
+            :exc:`ValueError`: When both :paramref:`chat_id` and :paramref:`all_empty_entries` are
+                provided, or none of them are passed.
```
Bibo-Joshi, line 669: or chat_id is not present. Is that desired, though? my first instinct would be to debug log & pass

Bibo-Joshi (commented):
Thanks for the PR! didn't have a look at the tests yet.
do you want to add `Persistence.drop_*_data` is this PR as well?

telegram/ext/_dispatcher.py
```diff
@@ -282,6 +282,27 @@ def _set_singleton(cls, val: Optional['Dispatcher']) -> None:
...
+    @property
+    def chat_data(self) -> Mapping[int, CD]:
+        """
+        :obj:`types.MappingProxyType`: A dictionary handlers can use to store data for the chat.
+
+        .. versionchanged:: 14.0
+           :attr:`chat_data` is now read-only
+
+        """
+        return MappingProxyType(self._chat_data)
```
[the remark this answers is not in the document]
harshil21: hm so no property then

telegram/ext/_dispatcher.py
harshil21: it doesn't. It just shows the text ``` `:paramref:`...` ```

telegram/ext/_dispatcher.py
harshil21: yes. We should also change the param name then. Any suggestions?

telegram/ext/_dispatcher.py
Bibo-Joshi: then I'd prefer to stick to `` for now :)

telegram/ext/_dispatcher.py
harshil21: it would be more user-friendly if we detected that first directly isn't it

Bibo-Joshi: if we don't raise an exception, then the outcome is always that `chat_id not in dispatcher.chat_data` after calling the method. Otherwise the outcome depends on `chat_id in dispatcher.chat_data` … Personally I find the first one more user friendly …

telegram/ext/_dispatcher.py
Bibo-Joshi: mh, not easy … maybe `all_falsy`? `Falsy` doesn't seem to be an offical term (couldn't find it on docs.python.org), but it seems to at least be widely used …

telegram/ext/_dispatcher.py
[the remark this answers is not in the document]
harshil21: Turns out it might be necessary. Consider this snippet:

``` python
class B:
    def __init__(self):
        self._chat_data = defaultdict(list)
        self.chat_data = MappingProxyType(self._chat_data)


b = B()
b._chat_data = defaultdict(dict)
b._chat_data.update({1: {}})

print(b._chat_data, b.chat_data)
# prints: defaultdict(<class 'dict'>, {1: {}}) defaultdict(<class 'list'>, {})
```
This is why the tests are failing for me, the `self.chat_data` mapping does not update when `self._chat_data` is reassigned (if it simply updated, it works fine)

Anyway, the performance hit of recreating the object every time is small. Unless you have another idea to make it work?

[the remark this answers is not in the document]
Bibo-Joshi: ah okay. that's to be expected, I guess …  IISC the only place where we override `_chat_data` is while loading the persistence data and I see several possible solutions here:

1. just rebuild `self.chat_data = MPT(self._chat_data)` after overriding at this one place
2. make `self._chat_data` a property with a setter that does the rebuilding
3. think about if we really need to override `chat_data` on loading the persistence. The more I look at it, the stranger it seems to me that we do that instead of `update(get_chat_data())`. Switching to `update` should even be non-breaking, I guess (even though we might be able to able to simplify some internals of PP/DictPersistence and adjust the expected return types of `get_*_data` - can be done in a separate PR, though)

what do you think?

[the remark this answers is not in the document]
harshil21: > 1. just rebuild `self.chat_data = MPT(self._chat_data)` after overriding at this one place

might work for now, but might be annoying to keep doing that in the long term - specially in tests

> 2. make `self._chat_data` a property with a setter that does the rebuilding

yes, sounds like it could work well, even when users modify/reassign `dp._chat_data`.

> 3. think about if we really need to override `chat_data` on loading the persistence. The more I look at it, the stranger it seems to me that we do that instead of `update(get_chat_data())`. Switching to `update` should even be non-breaking, I guess (even though we might be able to able to simplify some internals of PP/DictPersistence and adjust the expected return types of `get_*_data` - can be done in a separate PR, though)

yep, makes sense to use `defaultdict` in one place and just update it than do it two other places and reassign it.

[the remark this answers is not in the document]
harshil21: ok I'm stupid... a property for `dp._chat_data` wouldn't work either, it just leads to the initial problem. I ended up adding methods `_set_chat/user_data` which would also update `dp.chat/user_data`

telegram/ext/_dispatcher.py
harshil21: didn't change the param name since it would probably cause confusion imo

telegram/ext/_dispatcher.py
```diff
@@ -297,6 +324,18 @@ def get_instance(cls) -> 'Dispatcher':
             return cls.__singleton()  # type: ignore[return-value] # pylint: disable=not-callable
         raise RuntimeError(f'{cls.__name__} not initialized or multiple instances exist')
 
+    def _set_chat_data(self, data: DefaultDict[int, CD]) -> None:
+        """Used for assigning a new value to the underlying chat_data dictionary. Also updates
+        the read-only mapping."""
+        self._chat_data = data
+        self.chat_data = MappingProxyType(self._chat_data)
```
Bibo-Joshi, line 331: This is basically what a setter for a `_chat_data` property would have to look like. though from https://github.com/python-telegram-bot/python-telegram-bot/pull/2852#discussion_r785294597 I rather had the impression that you wanted to go with `update` anyway.
looking at the current implementation, I'm even more in favor of that, as it would shorten the implementation & make it more readable …

telegram/ext/_dispatcher.py
```diff
@@ -631,6 +670,70 @@ def remove_handler(self, handler: Handler, group: int = DEFAULT_GROUP) -> None:
...
+
+        .. versionadded:: 14.0
+
+        Args:
+            chat_id (:obj:`int`, optional): The chat id to delete from the persistence. The entry
+                will be deleted even if it is not empty. Mutually exclusive with
+                ``all_empty_entries``.
+            all_empty_entries (:obj:`bool`, optional): If :obj:`True`, all entries which evaluates
+                to :obj:`False` in a boolean context in :attr:`chat_data` will be deleted.
+                Mutually exclusive with ``chat_id``.
```
Bibo-Joshi, line 684: Can you add an example along the lines "if `bool(chat_data[chat_id]) is False`, then it will be dropped"?

Actually, are we sure that we want this parameter at all? It might just lead to confusion and I don't think that we'd be using this method internally at all.
If an advanced user wants to clear "unsed" chat/user data, they can easily call `drop_chat/user_data` in a loop on their own with their own definiton of "empty".

maybe @Poolitzer can give his 2cts as well …

telegram/ext/_dispatcher.py
```diff
@@ -643,10 +746,10 @@ def update_persistence(self, update: object = None) -> None:
 
     def __update_persistence(self, update: object = None) -> None:
         if self.persistence:
-            # We use list() here in order to decouple chat_ids from self.chat_data, as dict view
+            # We use list() here in order to decouple chat_ids from self._chat_data, as dict view
             # objects will change, when the dict does and we want to loop over chat_ids
-            chat_ids = list(self.chat_data.keys())
-            user_ids = list(self.user_data.keys())
+            chat_ids = list(self._chat_data.keys())
+            user_ids = list(self._user_data.keys())
```
Bibo-Joshi, line 752: why user the privates  here? I don't see the need for it and the less we use them, the safer we are

Bibo-Joshi (commented):
Thanks for the updates! 

> do you want to add Persistence.drop_*_data is this PR as well?

what about this part?

telegram/ext/_dispatcher.py
harshil21: > This is basically what a setter for a _chat_data property would have to look like

yeah but the actual property would be something like:
``` python
@property
def _chat_data(self):
    if self.persistence:
        return self.persistence.get_chat_data()
    return defaultdict(self.context_types.chat_data)
```
which is again rebuilding a new object everytime

> looking at the current implementation, I'm even more in favor of that, as it would shorten the implementation & make it more readable …

on top of this, we can still do 3. from https://github.com/python-telegram-bot/python-telegram-bot/pull/2852#discussion_r785294597 I guess

telegram/ext/_dispatcher.py
harshil21: > Actually, are we sure that we want this parameter at all? It might just lead to confusion and I don't think that we'd be using this method internally at all. If an advanced user wants to clear "unsed" chat/user data, they can easily call `drop_chat/user_data` in a loop on their own with their own definiton of "empty".

True, thats what the OP wanted though. But wouldn't there be some overhead calling `dp.update_persistence()` on every iteration in the loop?

harshil21:
> > do you want to add Persistence.drop_*_data is this PR as well?
>
> what about this part?

Soon, just wanted this review process to be over first. After that is added, then in `dp.drop_chat/user_data`, we can just call `Persistence.drop_chat/user_data` instead of `dp.update_persistence` correct?

telegram/ext/_dispatcher.py
Bibo-Joshi: you mean this?

> Add another method for purging all orphaned keys from the persistence file.

IISC that rather means "drop all entries from the persistence file that are no longer present in `dispatcher.chat/user_data`. tbh, I don't see a big need for it as long as we make sure that we don't produce any "orphaned keys" - by calling `persistence.drop_*` in `dispatcher.drop_*_data`.

Bibo-Joshi:
> > > do you want to add Persistence.drop_*_data is this PR as well?
> > 
> > 
> > what about this part?
> 
> Soon, just wanted this review process to be over first. After that is added, then in `dp.drop_chat/user_data`, we can just call `Persistence.drop_chat/user_data` instead of `dp.update_persistence` correct?

jup 👍

telegram/ext/_dispatcher.py
Bibo-Joshi: mh, no, I rather had in mind something like

```python
@property
def _chat_data(self):
    return self.__chat_data

@_chat_data.setter
def __chat_data(self, value):
    self.__chat_data = value
    self.chat_data = MappingProxyType(self.__chat_data)
```

but yeah, then let's just go for 3 :)

telegram/ext/_dictpersistence.py
```diff
@@ -361,6 +361,34 @@ def update_callback_data(self, data: CDCData) -> None:
...
+    def drop_chat_data(self, chat_id: int) -> None:
+        """Will delete the specified key from the :attr:`chat_data`.
+
+        .. versionadded:: 14.0
+
+        Args:
+            chat_id (:obj:`int`): The chat id to delete from the persistence.
+        """
+        if self._chat_data is None:
+            self._chat_data = defaultdict(dict)
```
Bibo-Joshi, line 373: why not just `return`?

telegram/ext/_dictpersistence.py
```diff
@@ -361,6 +361,34 @@ def update_callback_data(self, data: CDCData) -> None:
...
+    def drop_user_data(self, user_id: int) -> None:
+        """Will delete the specified key from the :attr:`user_data`.
+
+        .. versionadded:: 14.0
+
+        Args:
+            user_id (:obj:`int`): The user id to delete from the persistence.
+        """
+        if self._user_data is None:
+            self._user_data = defaultdict(dict)
```
Bibo-Joshi, line 387: same

telegram/ext/_dispatcher.py
```diff
@@ -198,10 +218,17 @@ def __init__(
...
-        self.user_data: DefaultDict[int, UD] = defaultdict(self.context_types.user_data)
-        self.chat_data: DefaultDict[int, CD] = defaultdict(self.context_types.chat_data)
+        self._user_data: DefaultDict[int, UD]
+        self._chat_data: DefaultDict[int, CD]
+        # Read only mapping-
+        self.user_data: Mapping[int, UD]
+        self.chat_data: Mapping[int, CD]
+
+        self._set_user_data(defaultdict(self.context_types.user_data))
+        self._set_chat_data(defaultdict(self.context_types.chat_data))
```
Bibo-Joshi, line 228: did we decide on switching to using to `_chat._data.update(…)` instead of `_set_chat_data(…)`? I lost track :D

telegram/ext/_picklepersistence.py
```diff
@@ -392,6 +392,46 @@ def update_callback_data(self, data: CDCData) -> None:
...
+        """Will delete the specified key from the :attr:`chat_data` and depending on
+        :attr:`on_flush` save the pickle file.
+
+        .. versionadded:: 14.0
+
+        Args:
+            chat_id (:obj:`int`): The chat id to delete from the persistence.
+        """
+        if self.chat_data is None:
+            self.chat_data = defaultdict(self.context_types.chat_data)
```
Bibo-Joshi, line 405: same

telegram/ext/_picklepersistence.py
```diff
@@ -392,6 +392,46 @@ def update_callback_data(self, data: CDCData) -> None:
...
+        """Will delete the specified key from the :attr:`user_data` and depending on
+        :attr:`on_flush` save the pickle file.
+
+        .. versionadded:: 14.0
+
+        Args:
+            user_id (:obj:`int`): The user id to delete from the persistence.
+        """
+        if self.user_data is None:
+            self.user_data = defaultdict(self.context_types.user_data)
```
Bibo-Joshi, line 425: same

telegram/ext/_dispatcher.py
```diff
@@ -631,6 +672,34 @@ def remove_handler(self, handler: Handler, group: int = DEFAULT_GROUP) -> None:
...
+    def drop_chat_data(self, chat_id: int) -> None:
+        """Used for deleting a key from the :attr:`chat_data`.
+
+        .. versionadded:: 14.0
+
+        Args:
+            chat_id (:obj:`int`): The chat id to delete from the persistence. The entry
+                will be deleted even if it is not empty.
+        """
+        del self._chat_data[chat_id]
```
Bibo-Joshi, line 684: this will raise a `KeyError` if `chat_id` is not present - thought we decided on not doing that 🤔  I'd suggest `.pop(chat_id, None)` instead

telegram/ext/_dispatcher.py
```diff
@@ -631,6 +672,34 @@ def remove_handler(self, handler: Handler, group: int = DEFAULT_GROUP) -> None:
...
+    def drop_user_data(self, user_id: int) -> None:
+        """Used for deleting a key from the :attr:`user_data`.
+
+        .. versionadded:: 14.0
+
+        Args:
+            user_id (:obj:`int`): The user id to delete from the persistence. The entry
+                will be deleted even if it is not empty.
+        """
+        del self._user_data[user_id]
```
Bibo-Joshi, line 698: same

tests/test_dispatcher.py
```diff
@@ -879,6 +906,36 @@ def callback(update, context):
...
+        "c_id,expected",
+        [(321, {123: [], 222: "remove_me"}), (111, None)],
+        ids=["test chat_id removal", "test no key in data (error)"],
+    )
+    def test_drop_chat_data(self, dp, c_id, expected):
+        dp._chat_data.update({123: [], 321: {'not_empty': 'no'}, 222: "remove_me"})
+
+        if c_id is not None and c_id not in dp.chat_data:
+            with pytest.raises(KeyError):
+                dp.drop_chat_data(c_id)
```
Bibo-Joshi, line 919: See above

telegram/ext/_dictpersistence.py
harshil21: you mean set the attribute and return, or just return without setting the attribute? I don't really know what to do here

telegram/ext/_dispatcher.py
harshil21: The `.update()` thing was planned for another PR, which will also update the return types of `get_user/chat/_data`

telegram/ext/_dispatcher.py
harshil21: right

telegram/ext/_dictpersistence.py
Bibo-Joshi: Well, if `_chat_data is None`, then `chat_id` is certainly not in `chat_data`. I see no reason to build a `defaultdict` here … so just

```python
if self._chat_data is None:
    return
```

telegram/ext/_dispatcher.py
Bibo-Joshi: Ah, I was unclear again 😬  I meant that adjusting the documented return type of `get_*_data` and the implementations of `PP/DictPersistence` can be done in another PR. switching to `.update()` should work without that (since defaultdicts are dicts) and adding `_set_*_data` here just to remove it in the next PR again seems a bit superfluous to me

Bibo-Joshi (commented):
Thanks for the updates ❤️ Code LGTM! I'll merge v14 with the animation-test fix so that we can have another go ad the CI - maybe codecov also shows up again …

harshil21:
> Code LGTM

🎉 
I had already merged v14 with animation test fix. They just changed the animation values again...

PR #2854, opened 2022-01-11
Doc linkcode tuning

Bibo-Joshi:
make the `[source]` anchors point to the correct file on github wherever reasonably possible, including the correct line numbers and the correct blob. I chose the output of `git describe --tags` here as it

* makes sure that on dev branches we don't point to master
* makes sure to point directly to the corresponding tag when building from a release commit

A build of this branch is up at https://python-telegram-bot.readthedocs.io/en/doc-linkcode-tuning/

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -441,12 +441,12 @@ def autodoc_process_docstring(app: Sphinx, what, name: str, obj: object, options
 
 
 def _git_branch() -> str:
-    """Get's the current git branch if available or fall back to `master`"""
+    """Get's the current git sha if available or fall back to `master`"""
     try:
         output = subprocess.check_output(  # skipcq: BAN-B607
-            ["git", "show", "-s", "--pretty=%d", "HEAD"], stderr=subprocess.STDOUT
+            ["git", "describe", "--tags"], stderr=subprocess.STDOUT
         )
-        return output.decode().strip(' \n()').split(', ')[1].split('/')[1]
+        return output.decode().strip()
     except Exception:
         return 'master'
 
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -397,23 +399,91 @@ def autodoc_skip_member(app, what, name, obj, skip, options):
...
+
+
+git_branch = _git_branch()
+base_url = "https://github.com/python-telegram-bot/python-telegram-bot/blob/"
+
+
+def linkcode_resolve(_, info):
+    """See www.sphinx-doc.org/en/master/usage/extensions/linkcode.html"""
+    combined = '.'.join((info['module'], info['fullname']))
+    # special casing for ExtBot which is due to the special structure of extbot.rst
```
harshil21, line 461: ah mystery solved 👍🏼

telegram/ext/_extbot.py
```diff
@@ -202,7 +202,7 @@ def insert_callback_data(self, update: Update) -> None:
             *In place*, i.e. the passed :class:`telegram.Message` will be changed!
 
         Args:
-            update (:class`telegram.Update`): The update.
+            update (:class:`telegram.Update`): The update.
```
harshil21, line 205: ah this was in the referencing pr so I'd left it

harshil21 (approved):
Nice work! I definitely wouldn't have thought of using an unrelated function to achieve this.

telegram/ext/_extbot.py
Bibo-Joshi: ah yes, didn't think of that :D shouldn't hurt, though …

PR #2870, opened 2022-01-18
Remove `telegram.ReplyMarkup`

### LINKED ISSUE #2868: API Policy: `ReplyMarkup`
I just noticed that ptbs `telegram.ReplyMarkup` is in fact an invention of PTB, i.e. it's not defined by the official API docs. We should consider removing that class or at least making it private.

IISC it's currently only used for

* type hinting - that could easily be replaced by a type alias
* the static method `_check_keyboard_type` that's used in the inits of or `{Reply, Inline}KeyboardMarkup` - that could moved e.g. to `tg._utils`

Personally, I'd be in favor of just removing the class. @harshil21 IIRC you added the `_check_keyboard_type` method - what do you think?

harshil21:
Closes #2868 

### Checklist for PRs


telegram/_utils/markup.py
```diff
@@ -0,0 +1,39 @@
...
+
+Warning:
+    Contents of this module are intended to be used internally by the library and *not* by the
+    user. Changes to this module are not considered breaking changes and may not be documented in
+    the changelog.
+"""
+
+
+def check_keyboard_type(keyboard: object) -> bool:
+    """Checks if the keyboard provided is of the correct type - A list of lists."""
```
Bibo-Joshi, line 33:
```suggestion
    """Checks if the keyboard provided is of the correct type - A list of lists.
    Implicitly tested in the init-tests of `{Inline, Reply}KeyboardMarkup`
    """
```

Bibo-Joshi (commented):
nice :) just one tiny comment

PR #2873, opened 2022-01-21
Simplify persistence

### LINKED ISSUE #2871: Adjust BasePersistence as discussed in #2852
that is

* update documentation saying that `get_chat/user_data` may return a dict instead of a defaultdict
* simplify the implementations of `Dict/PicklePersistence` accordingly
* adjust tests if necessary

harshil21:
Closes #2871 
### Checklist for PRs

### DIFF
--- a/telegram/ext/_basepersistence.py
+++ b/telegram/ext/_basepersistence.py
@@ -398,7 +398,7 @@ class BasePersistence(Generic[UD, CD, BD], ABC):
     def get_user_data(self) -> Dict[int, UD]:
         """Will be called by :class:`telegram.ext.Dispatcher` upon creation with a
         persistence object. It should return the ``user_data`` if stored, or an empty
-        :obj:`defaultdict` or :obj:`dict`. In the latter case, the dictionary should produce values
+        :obj:`dict`. In the latter case, the dictionary should produce values
         corresponding to one of the following:
 
           * :obj:`dict`
@@ -417,7 +417,7 @@ class BasePersistence(Generic[UD, CD, BD], ABC):
     def get_chat_data(self) -> Dict[int, CD]:
         """Will be called by :class:`telegram.ext.Dispatcher` upon creation with a
         persistence object. It should return the ``chat_data`` if stored, or an empty
-        :obj:`defaultdict` or :obj:`dict`. In the latter case, the dictionary should produce values
+        :obj:`dict`. In the latter case, the dictionary should produce values
         corresponding to one of the following:
 
           * :obj:`dict`
--- a/telegram/ext/_basepersistence.py
+++ b/telegram/ext/_basepersistence.py
@@ -132,10 +132,10 @@ class BasePersistence(Generic[UD, CD, BD], ABC):
         update_bot_data = instance.update_bot_data
         update_callback_data = instance.update_callback_data
 
-        def get_user_data_insert_bot() -> Union[DefaultDict[int, UD], Dict[int, UD]]:
+        def get_user_data_insert_bot() -> Dict[int, UD]:
             return instance.insert_bot(get_user_data())
 
-        def get_chat_data_insert_bot() -> Union[DefaultDict[int, CD], Dict[int, CD]]:
+        def get_chat_data_insert_bot() -> Dict[int, CD]:
             return instance.insert_bot(get_chat_data())
 
         def get_bot_data_insert_bot() -> BD:
--- a/telegram/ext/_basepersistence.py
+++ b/telegram/ext/_basepersistence.py
@@ -19,7 +19,7 @@
 """This module contains the BasePersistence class."""
 from abc import ABC, abstractmethod
 from copy import copy
-from typing import Dict, Optional, Tuple, cast, ClassVar, Generic, DefaultDict, NamedTuple
+from typing import Dict, Optional, Tuple, cast, ClassVar, Generic, DefaultDict, NamedTuple, Union
 
 from telegram import Bot
 from telegram.ext import ExtBot
@@ -132,10 +132,10 @@ class BasePersistence(Generic[UD, CD, BD], ABC):
         update_bot_data = instance.update_bot_data
         update_callback_data = instance.update_callback_data
 
-        def get_user_data_insert_bot() -> DefaultDict[int, UD]:
+        def get_user_data_insert_bot() -> Union[DefaultDict[int, UD], Dict[int, UD]]:
             return instance.insert_bot(get_user_data())
 
-        def get_chat_data_insert_bot() -> DefaultDict[int, CD]:
+        def get_chat_data_insert_bot() -> Union[DefaultDict[int, CD], Dict[int, CD]]:
             return instance.insert_bot(get_chat_data())
 
         def get_bot_data_insert_bot() -> BD:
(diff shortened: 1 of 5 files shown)


telegram/ext/_basepersistence.py
```diff
@@ -398,50 +395,50 @@ def _insert_bot(self, obj: object, memo: Dict[int, object]) -> object:
         return obj
 
     @abstractmethod
-    def get_user_data(self) -> DefaultDict[int, UD]:
+    def get_user_data(self) -> Union[DefaultDict[int, UD], Dict[int, UD]]:
```
Bibo-Joshi, line 398: why the union? `defaultdict` is a subclass of `dict`, so that should be clear anyway … same at a few other places

If there's no compelling reason to keep `defaultdict` in type hints & comments, I'd also suggest to add `.. versionchanged::` directives along the lines "May now return ordinary dicts instead of defaultdicts"

telegram/ext/_picklepersistence.py
```diff
@@ -323,12 +315,11 @@ def update_user_data(self, user_id: int, data: UD) -> None:
...
             user_id (:obj:`int`): The user the data might have been changed for.
-            data (:obj:`dict` | :attr:`telegram.ext.ContextTypes.user_data`): The
-                :attr:`telegram.ext.Dispatcher.user_data` ``[user_id]``.
+            data (:obj:`dict`): The :attr:`telegram.ext.Dispatcher.user_data` ``[user_id]``.
         """
         if self.user_data is None:
-            self.user_data = defaultdict(self.context_types.user_data)
-        if self.user_data.get(user_id) == data:
+            self.user_data = {}
+        if self.user_data.get(user_id, None) == data:  # type: ignore[arg-type]
```
Bibo-Joshi, line 322:
```suggestion
        if self.user_data.get(user_id) == data:  # type: ignore[arg-type]
```
what's the `type ignore` for?

telegram/ext/_picklepersistence.py
```diff
@@ -342,12 +333,11 @@ def update_chat_data(self, chat_id: int, data: CD) -> None:
...
             chat_id (:obj:`int`): The chat the data might have been changed for.
-            data (:obj:`dict` | :attr:`telegram.ext.ContextTypes.chat_data`): The
-                :attr:`telegram.ext.Dispatcher.chat_data` ``[chat_id]``.
+            data (:obj:`dict`): The :attr:`telegram.ext.Dispatcher.chat_data` ``[chat_id]``.
         """
         if self.chat_data is None:
-            self.chat_data = defaultdict(self.context_types.chat_data)
-        if self.chat_data.get(chat_id) == data:
+            self.chat_data = {}
+        if self.chat_data.get(chat_id, None) == data:  # type: ignore[arg-type]
```
Bibo-Joshi, line 340: same

Bibo-Joshi (changes requested):
you are quick! :)

Please revert the changes to dispatcher (except maybe for `_chat_data` -> `chat_data`). they are not related to the PR and tbh they don't change much at all anyway …

telegram/ext/_picklepersistence.py
harshil21: When we add `None`, it complains that the default arg doesn't match type of `user_data` (`UD`). Anyway this is fixed in latest mypy version, which we probably should upgrade to soon

telegram/ext/_basepersistence.py
harshil21: I thought mypy would complain, since `dict` is not an instance of `defaultdict`. anyway will check with our current mypy version and latest version

Bibo-Joshi: mh, where would it complain? i.e. where do we still use that `get_user_data()` is a `defaultdict`?

telegram/ext/_picklepersistence.py
Bibo-Joshi: ah okay. feel free to upgrade, then. still, the `None` should not be needed 🤔

harshil21: > ah okay. feel free to upgrade, then.

we'll do that in another PR, there's false positives in constants module + other things

telegram/ext/_basepersistence.py
```diff
@@ -398,50 +395,56 @@ def _insert_bot(self, obj: object, memo: Dict[int, object]) -> object:
         return obj
 
     @abstractmethod
-    def get_user_data(self) -> DefaultDict[int, UD]:
+    def get_user_data(self) -> Dict[int, UD]:
         """Will be called by :class:`telegram.ext.Dispatcher` upon creation with a
         persistence object. It should return the ``user_data`` if stored, or an empty
-        :obj:`defaultdict`. In the latter case, the :obj:`defaultdict` should produce values
+        :obj:`defaultdict` or :obj:`dict`. In the latter case, the dictionary should produce values
```
Bibo-Joshi, line 401:
```suggestion
        :obj:`dict`. In the latter case, the dictionary should produce values
```

telegram/ext/_basepersistence.py
```diff
@@ -398,50 +395,56 @@ def _insert_bot(self, obj: object, memo: Dict[int, object]) -> object:
...
                 The restored user data.
         """
 
     @abstractmethod
-    def get_chat_data(self) -> DefaultDict[int, CD]:
+    def get_chat_data(self) -> Dict[int, CD]:
         """Will be called by :class:`telegram.ext.Dispatcher` upon creation with a
         persistence object. It should return the ``chat_data`` if stored, or an empty
-        :obj:`defaultdict`. In the latter case, the :obj:`defaultdict` should produce values
+        :obj:`defaultdict` or :obj:`dict`. In the latter case, the dictionary should produce values
```
Bibo-Joshi, line 420: .

telegram/ext/_dictpersistence.py
```diff
@@ -315,8 +312,8 @@ def update_user_data(self, user_id: int, data: Dict) -> None:
             data (:obj:`dict`): The :attr:`telegram.ext.Dispatcher.user_data` ``[user_id]``.
         """
         if self._user_data is None:
-            self._user_data = defaultdict(dict)
-        if self._user_data.get(user_id) == data:
+            self._user_data = {}
+        if self._user_data.get(user_id, None) == data:
```
Bibo-Joshi, line 316:
```suggestion
        if self._user_data.get(user_id) == data:
```

Deepsource actually complains about this, since `None` is the default for the second argument :)

telegram/ext/_dictpersistence.py
```diff
@@ -329,8 +326,8 @@ def update_chat_data(self, chat_id: int, data: Dict) -> None:
             data (:obj:`dict`): The :attr:`telegram.ext.Dispatcher.chat_data` ``[chat_id]``.
         """
         if self._chat_data is None:
-            self._chat_data = defaultdict(dict)
-        if self._chat_data.get(chat_id) == data:
+            self._chat_data = {}
+        if self._chat_data.get(chat_id, None) == data:
```
Bibo-Joshi, line 330: .

telegram/ext/_dictpersistence.py
harshil21: oh I thought it raised an error like `.pop()`

Bibo-Joshi (approved):
Do you want to merge or do you want to extend this PR with dropping `insert/replace_bot` as discussed? I'm fine with both :)

harshil21:
@Bibo-Joshi I think we can merge and probably do that in the asyncio branch since the implementation will likely change anyway

Bibo-Joshi:
@harshil21 on <code>asyncio</code>, we mainly change where we call the <code>update_*_data</code> methods, but we don&#x27;t change any internals of the persistence (except from making the methods <code>async</code>). I would like it better for those changes to be in a different PR. We can still merge this one as is though :)

harshil21:
okay, lets do it in another PR then 👍🏼

PR #2890, opened 2022-02-10
use sets instead of lists to store CH and PH commands

### LINKED ISSUE #2885: Rename CommandHandler.command to CH.commands (plural) …
… since it's always a list independent on whether the input was a single string or a list of strings. We might also think about making the attribute a set instead of a last, as checking `string in commands_set` is [significantly faster](https://stackoverflow.com/questions/2831212/python-sets-vs-lists) than `string in commands_list` and at least currently that's the only thing we use that attribute for

mehrfirouzian:
fix CommandHandler and PrefixHandler to use lists instead of sets

Closes #2885


Poolitzer:
Hi, are you working on #2885 with this?

telegram/ext/_commandhandler.py
```diff
@@ -71,7 +71,7 @@ class CommandHandler(Handler[Update, CCT]):
         ValueError: when command is too long or has illegal chars.
 
     Attributes:
-        command (:obj:`str` | Tuple[:obj:`str`] | List[:obj:`str`]):
+        command (:obj:`str` | Tuple[:obj:`str`] | List[:obj:`str`] | Set[:obj:`str`]):
```
Bibo-Joshi, line 74: In the current implementation, `CommandHandler.command` is actually always a `List` …

Bibo-Joshi (changes requested):
Hi. thanks for the PR! IISC so far you have only allowed for the `PrefixHandler.prefix/command` setters to accept sets.
I think I should clarify a bit what the intention of #2885 was:

* The main idea of the issue was to rename `CommandHandler.command` to `CommandHandler.commands` (plural) since that attribute is currently actually always a list. Extending this to `PrefixHandlers.prefix` is surely reasonable.
* Since I noticed that we only ever need `x in self.command`, my idea was to make `CommandHandler.commands` a set instead of a list. Accepting sets as input in addition is surely a logical extension. So 

Additionally, while reviewing your PR, I noticed that unlike `PrefixHandler.command`, `CommandHandler.command` doesn't have a setter. I.e. if you do `command_handler.command = 'NewCommand'`, it won't be converted to `[newcommand]`. I think we should fix that. So what I would ask you to do is:

* Make sure that sets are accepted as input for `Prefix/CommandHandler.command` and `PrefixHandler.prefix` - this includes type hints & documentation
* rename `CommandHandler.command` to `CommandHandler._commands` and add getter + setter proprties `CommandHandler.commands`. The getter should provide access to the set `CH._commands` and the setter should accept a single string or tuple/list/set and convert it as it is converted on `__init__`
* Apply analogous changes to `PrefixHandler`

telegram/ext/_commandhandler.py
mehrfirouzian: > Hi. thanks for the PR! IISC so far you have only allowed for the `PrefixHandler.prefix/command` setters to accept sets. I think I should clarify a bit what the intention of #2885 was:
> 
> * The main idea of the issue was to rename `CommandHandler.command` to `CommandHandler.commands` (plural) since that attribute is currently actually always a list. Extending this to `PrefixHandlers.prefix` is surely reasonable.
> * Since I noticed that we only ever need `x in self.command`, my idea was to make `CommandHandler.commands` a set instead of a list. Accepting sets as input in addition is surely a logical extension. So
> 
> Additionally, while reviewing your PR, I noticed that unlike `PrefixHandler.command`, `CommandHandler.command` doesn't have a setter. I.e. if you do `command_handler.command = 'NewCommand'`, it won't be converted to `[newcommand]`. I think we should fix that. So what I would ask you to do is:
> 
> * Make sure that sets are accepted as input for `Prefix/CommandHandler.command` and `PrefixHandler.prefix` - this includes type hints & documentation
> * rename `CommandHandler.command` to `CommandHandler._commands` and add getter + setter proprties `CommandHandler.commands`. The getter should provide access to the set `CH._commands` and the setter should accept a single string or tuple/list/set and convert it as it is converted on `__init__`
> * Apply analogous changes to `PrefixHandler`

Thanks for the review, I'm working on it. should I change the name in arguments from command to commands too?

mehrfirouzian:
I also noticed _command is being used in PrefixHandler only because there is command is present in CommandHandler which is its parent, do I need to use _commands in CommandHandler?

telegram/ext/_commandhandler.py
```diff
@@ -71,31 +71,35 @@ class CommandHandler(Handler[Update, CCT]):
         ValueError: when command is too long or has illegal chars.
 
     Attributes:
-        command (:obj:`str` | Tuple[:obj:`str`] | List[:obj:`str`]):
-            The command or list of commands this handler should listen for.
+        commands (Set[:obj:`str`]):
+            The set of command(s) this handler should listen for.
```
Bibo-Joshi, line 75: the property already has a docstring, so you can remove this documentation of the attribute

telegram/ext/_commandhandler.py
```diff
@@ -71,31 +71,35 @@ class CommandHandler(Handler[Update, CCT]):
...
             Limitations are the same as described here https://core.telegram.org/bots#commands
         callback (:obj:`callable`): The callback function for this handler.
         filters (:class:`telegram.ext.BaseFilter`): Optional. Only allow updates with these
             Filters.
         run_async (:obj:`bool`): Determines whether the callback will run asynchronously.
+
+            .. versionchanged:: 14.0
+            :attr:`commands` is now a set and can be assinged after construction of model.
```
Bibo-Joshi, line 83: see above

telegram/ext/_commandhandler.py
```diff
@@ -155,6 +159,23 @@ def collect_additional_context(
...
+            Set[:obj:`str`]
+        """
+        return self._commands
+
+    @commands.setter
+    def commands(self, command: Union[str, List[str], Set[str]]) -> None:
+        if isinstance(command, str):
+            self._commands = {command.lower()}
+        else:
+            self._commands = {i.lower() for i in command}
```
Bibo-Joshi, line 177: THis is the same logic that's used on `__init__`. Could you extract it to a new protected method (e.g. `_parse_commands()`) that you call both here and on `__init__`? Then you can also move the regexing from `__init__` to that function. THat will make sure that also the setter only accepts valid bot commands (please add a unit test for this)

telegram/ext/_commandhandler.py
```diff
@@ -51,7 +51,7 @@ class CommandHandler(Handler[Update, CCT]):
         attributes to :class:`telegram.ext.CallbackContext`. See its docs for more info.
 
     Args:
-        command (:obj:`str` | Tuple[:obj:`str`] | List[:obj:`str`]):
+        command (:obj:`str` | Tuple[:obj:`str`] | List[:obj:`str`] | Set[:obj:`str`]):
```
Bibo-Joshi, line 54: please add a "versionchanged" directive saying that it now also accepts sets

telegram/ext/_commandhandler.py
```diff
@@ -216,78 +237,84 @@ class PrefixHandler(CommandHandler):
...
+        commands (Set[:obj:`str`]):
+            The set of command(s) that this handler should listen for.
         callback (:obj:`callable`): The callback function for this handler.
         filters (:class:`telegram.ext.BaseFilter`): Optional. Only allow updates with these
             Filters.
         run_async (:obj:`bool`): Determines whether the callback will run asynchronously.
 
+            .. versionchanged:: 14.0
+                :attr:`prefix` has been changed to :attr:`prefixes` and accepts sets as well.
+                :attr:`command` has been changed to :attr:`commands` and accepts sets as well.
```
Bibo-Joshi, line 251: Please move the "changed to" notes to a `versionchanged` directive right before the `Args:`. Please move the "now accepts sets" to a `versionchanged` directive in the documentation of the respective arguments

telegram/ext/_commandhandler.py
```diff
@@ -216,78 +237,84 @@ class PrefixHandler(CommandHandler):
...
+            self._prefixes = {prefix.lower()}
         else:
-            self._prefix = prefix
+            self._prefixes = set(prefix)
         self._build_commands()
 
-    @property  # type: ignore[override]
-    def command(self) -> List[str]:  # type: ignore[override]
+    @property
+    def commands(self) -> Set[str]:
```
Bibo-Joshi, line 297: Is this override still necessary? IISC, this does exatly the same as `CommandHandler.commands`

tests/test_commandhandler.py
```diff
@@ -185,6 +185,22 @@ def test_command_list(self):
         assert is_match(handler, make_command_update('/star'))
         assert not is_match(handler, make_command_update('/stop'))
 
+    def test_edit_commands(self):
+        """A command handler which its commands are changed should change as well"""
+        handler = CommandHandler(['test', 'star'], self.callback_basic)
+        assert is_match(handler, make_command_update('/test'))
+        handler.commands = "help"
+        assert is_match(handler, make_command_update('/help'))
```
Bibo-Joshi, line 193:
```suggestion
        assert is_match(handler, make_command_update('/help'))
        assert not is_match(handler, make_command_update('/test))
```

tests/test_commandhandler.py
```diff
@@ -185,6 +185,22 @@ def test_command_list(self):
...
+        handler.commands = ['exit', 'happy']
+        assert is_match(handler, make_command_update('/happy'))
```
Bibo-Joshi, line 195:
```suggestion
        assert is_match(handler, make_command_update('/happy'))
        assert not is_match(handler, make_command_update('/help'))
```

Bibo-Joshi (changes requested):
hi, thanks for the updates! I left a few comments below

Also please add a `.. versionchanged::` directive just before the `Args:` of `CommandHandler` saying that `command` was replaced by `commands`

telegram/ext/_commandhandler.py
mehrfirouzian: Sure, I first tried to do it like this
`self.commands = command`
But it wouldn't work when PrefixHandler tried to call it because it has a different setter than CommandHandler with the same name. But I'll make a function for it.

telegram/ext/_commandhandler.py
mehrfirouzian: yes, because this class uses a different setter than the parent one and needs a property to set it.

telegram/ext/_commandhandler.py
mehrfirouzian: I noticed we don't have CommandHandler command limitations for PrefixHandler, so illegal commands in prefixhandler should work properly shouldn't they?

Bibo-Joshi: > I noticed we don't have CommandHandler command limitations for PrefixHandler, so illegal commands in prefixhandler should work properly shouldn't they?

yes, since `PrefixHandler` is not limited to the commands recognized by Telegram. Actually the fact that `PrefixHandler` is a subclass of `CommandHandler` is more or less an implementation detail …

telegram/ext/_commandhandler.py
Bibo-Joshi: IIRC using `@CommandHandler.commands.setter` instead of `@commands.setter` should do the trick and then you don't need to re-implement the property

telegram/ext/_commandhandler.py
```diff
@@ -71,33 +78,29 @@ class CommandHandler(Handler[Update, CCT]):
...
-            The command or list of commands this handler should listen for.
-            Limitations are the same as described here https://core.telegram.org/bots#commands
         callback (:obj:`callable`): The callback function for this handler.
         filters (:class:`telegram.ext.BaseFilter`): Optional. Only allow updates with these
             Filters.
         run_async (:obj:`bool`): Determines whether the callback will run asynchronously.
+
+            .. versionchanged:: 14.0
+                :attr:`commands` is now a set and can be assinged after construction of model.
+
```
Bibo-Joshi, line 88: please remove

telegram/ext/_commandhandler.py
```diff
@@ -50,10 +50,17 @@ class CommandHandler(Handler[Update, CCT]):
         When setting ``run_async`` to :obj:`True`, you cannot rely on adding custom
         attributes to :class:`telegram.ext.CallbackContext`. See its docs for more info.
 
+    .. versionchanged:: 14.0
+        :attr:`commands` is now a set and can be assinged after construction of model.
```
Bibo-Joshi, line 54: The attribute ``command`` was replaced by :attr:`commands`

telegram/ext/_commandhandler.py
```diff
@@ -123,13 +126,13 @@ def check_update(
                 and message.text
                 and message.get_bot()
             ):
-                command = message.text[1 : message.entities[0].length]
+                command = message.text[1: message.entities[0].length]
```
Bibo-Joshi, line 129: please revert - this is the reason the `pre-commit` test is failing

telegram/ext/_commandhandler.py
```diff
@@ -155,6 +158,29 @@ def collect_additional_context(
...
+            Set[:obj:`str`]
+        """
+        return self._commands
+
+    @commands.setter
+    def commands(self, command: SLTS) -> None:
+        self._parse_commands(command)
+        for comm in self._commands:
+            if not re.match(r'^[\da-z_]{1,32}$', comm):
+                raise ValueError(f'Command `{comm}` is not a valid bot command')
```
Bibo-Joshi, line 182: when this is raised, `self._commands` has already been changed - it should be the other way around. You can even add a unit test along the lines

1. bulid commandhandler with valid commands
2. try to assign invalid commands
3. assert that the commands where not changed

what we can do to make this easier here & also make the init a bit cleaner is:

1. make `_parse_commands` return the parsed commands instead of assigning them directly to `self._commands`. It can be converted into a static method
2. apply the regexing for valid commands directly within `_parse_commands`
3. do `self._commands = self._parse_commands(command)` both here and in `__init__`.

telegram/ext/_commandhandler.py
```diff
@@ -223,71 +261,56 @@ class PrefixHandler(CommandHandler):
...
-
-    @command.setter
-    def command(self, command: Union[str, List[str]]) -> None:
-        if isinstance(command, str):
-            self._command = [command.lower()]
-        else:
-            self._command = command
+    @CommandHandler.commands.setter  # type: ignore[attr-defined,misc]
+    def commands(self, command: SLTS) -> None:
+        self._parse_commands(command)
```
Bibo-Joshi, line 307: IISC this doesn't rebuilt `self._combinations`, but that's needed. I guess it should be `self._build_commands` instead. If you want to reduce code duplication some more, you can add a method `_slts_to_set` that returns `{arg.lower()} if isinstance(arg, str) else {item.lower() for item in arg}`. This can be a static method of `CommandHandler` or even just a module level function. You can then use it in `CH._parse_commands`, in the setter of `prefixes` and here.

telegram/ext/_commandhandler.py
```diff
@@ -197,11 +223,23 @@ class PrefixHandler(CommandHandler):
         When setting ``run_async`` to :obj:`True`, you cannot rely on adding custom
         attributes to :class:`telegram.ext.CallbackContext`. See its docs for more info.
 
+    .. versionchanged:: 14.0
+        :attr:`prefixes` and :attr:`commands` are always a set and can be
+        assigned with sets as well.
```
Bibo-Joshi, line 228:
```suggestion
        The attributes ``prefix`` and ``command`` where replaced by :attr:`prefixes` and :attr:`commands`, respectively.
```

Bibo-Joshi (changes requested):

PR #2893, opened 2022-02-10
Remove `BP.insert/replace_bot`

### LINKED ISSUE #2882: Drop BasePersistence.insert/replace_bot
As discussed in the dev chat before, just writing down so that we don't forget

* Remove `insert/replace_bot`. Remove corresponding notes and instead leave a not saying that bot instances should not be serialized and instead `persistence.bot` *may* be used to re-insert bots into data.
* Use `pickle.Un/Pickler` to do the above for `PicklePersistence` - mock up at https://nekobin.com/lifojagumo.py
* Make sure that `deepcopy(TelegramObject)` keep the bot instance - add note about that to documentation
* Make sure that `pickle.dumps(TelegramObject)` ignores the bot instance - add note about that to documentation

Reference for developers: we had discussions in this eg [here](https://t.me/c/1494805131/21603) and up

harshil21:
Closes #2882 

### Checklist for PRs

Breaking changes:
- `insert/replace_bot` was removed so user may have to use `set_bot(bot)` before using shortcuts.

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -125,10 +125,14 @@ class Bot(TelegramObject):
         considered equal, if their :attr:`bot` is equal.
 
     Note:
-        Most bot methods have the argument ``api_kwargs`` which allows to pass arbitrary keywords
-        to the Telegram API. This can be used to access new features of the API before they were
-        incorporated into PTB. However, this is not guaranteed to work, i.e. it will fail for
-        passing files.
+        * Most bot methods have the argument ``api_kwargs`` which allows passing arbitrary keywords
+          to the Telegram API. This can be used to access new features of the API before they are
+          incorporated into PTB. However, this is not guaranteed to work, i.e. it will fail for
+          passing files.
+        * Bots should not be serialized since if you for e.g. change the bots token, then your
+          serialized instance will not reflect that change. Hence, we raise a
+          :exc:`pickle.PicklingError` upon serialization.
+
 
     .. versionchanged:: 14.0
 
@@ -138,6 +142,7 @@ class Bot(TelegramObject):
         * Removed the deprecated ``defaults`` parameter. If you want to use
           :class:`telegram.ext.Defaults`, please use the subclass :class:`telegram.ext.ExtBot`
           instead.
+        * Attempting to pickle a bot instance will now raise :exc:`pickle.PicklingError`.
 
     Args:
         token (:obj:`str`): Bot's unique authentication.
@@ -197,7 +202,9 @@ class Bot(TelegramObject):
         return hash(self.bot)
 
     def __reduce__(self) -> NoReturn:
-        """Called by pickle.dumps() and only when bot instance is present in the data structure."""
+        """
+        Called by pickle.dumps(). Serializing bots is unadvisable, so we try to forbid pickling.
+        """
         raise pickle.PicklingError('Bot objects cannot be pickled!')
 
     def to_dict(self) -> JSONDict:
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -190,6 +190,25 @@ class Bot(TelegramObject):
                 private_key, password=private_key_password, backend=default_backend()
             )
 
+    def __eq__(self, other: object) -> bool:
+        return self.bot == other
+
+    def __hash__(self) -> int:
+        return hash(self.bot)
+
+    def __reduce__(self) -> NoReturn:
+        """Called by pickle.dumps() and only when bot instance is present in the data structure."""
+        raise pickle.PicklingError('Bot objects cannot be pickled!')
+
+    def to_dict(self) -> JSONDict:
+        """See :meth:`telegram.TelegramObject.to_dict`."""
+        data: JSONDict = {'id': self.id, 'username': self.username, 'first_name': self.first_name}
+
+        if self.last_name:
+            data['last_name'] = self.last_name
+
+        return data
+
     # TODO: After https://youtrack.jetbrains.com/issue/PY-50952 is fixed, we can revisit this and
     # consider adding Paramspec from typing_extensions to properly fix this. Currently a workaround
     def _log(func: Any):  # type: ignore[no-untyped-def] # skipcq: PY-D0003
@@ -5598,25 +5617,6 @@ class Bot(TelegramObject):
 
         result = self._post('copyMessage', data, timeout=timeout, api_kwargs=api_kwargs)
         return MessageId.de_json(result, self)  # type: ignore[return-value, arg-type]
-
-    def to_dict(self) -> JSONDict:
-        """See :meth:`telegram.TelegramObject.to_dict`."""
-        data: JSONDict = {'id': self.id, 'username': self.username, 'first_name': self.first_name}
-
-        if self.last_name:
-            data['last_name'] = self.last_name
-
-        return data
-
-    def __eq__(self, other: object) -> bool:
-        return self.bot == other
-
-    def __hash__(self) -> int:
-        return hash(self.bot)
-
-    def __reduce__(self) -> NoReturn:
-        """Called by pickle.dumps() and only when bot instance is present in the data structure."""
-        raise pickle.PicklingError('Bot objects cannot be pickled!')
 
     # camelCase aliases
     getMe = get_me
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -130,9 +130,8 @@ class Bot(TelegramObject):
           incorporated into PTB. However, this is not guaranteed to work, i.e. it will fail for
           passing files.
         * Bots should not be serialized since if you for e.g. change the bots token, then your
-          serialized instance will not reflect that change. Hence, we raise a
-          :exc:`pickle.PicklingError` upon serialization.
-
+          serialized instance will not reflect that change. Trying to pickle a bot instance will
+          raise :exc:`pickle.PicklingError`.
 
     .. versionchanged:: 14.0
 
@@ -195,28 +194,9 @@ class Bot(TelegramObject):
                 private_key, password=private_key_password, backend=default_backend()
             )
 
-    def __eq__(self, other: object) -> bool:
-        return self.bot == other
-
-    def __hash__(self) -> int:
-        return hash(self.bot)
-
     def __reduce__(self) -> NoReturn:
         """Called by pickle.dumps(). Serializing bots is unadvisable, so we forbid pickling."""
         raise pickle.PicklingError('Bot objects cannot be pickled!')
-
-    def __deepcopy__(self, memodict: dict) -> 'Bot':
-        """Bot objects can't be deepcopied since they are not pickable, so just return self."""
-        return self
-
-    def to_dict(self) -> JSONDict:
-        """See :meth:`telegram.TelegramObject.to_dict`."""
-        data: JSONDict = {'id': self.id, 'username': self.username, 'first_name': self.first_name}
-
-        if self.last_name:
-            data['last_name'] = self.last_name
-
-        return data
 
     # TODO: After https://youtrack.jetbrains.com/issue/PY-50952 is fixed, we can revisit this and
     # consider adding Paramspec from typing_extensions to properly fix this. Currently a workaround
@@ -5626,6 +5606,21 @@ class Bot(TelegramObject):
 
         result = self._post('copyMessage', data, timeout=timeout, api_kwargs=api_kwargs)
         return MessageId.de_json(result, self)  # type: ignore[return-value, arg-type]
+
+    def to_dict(self) -> JSONDict:
+        """See :meth:`telegram.TelegramObject.to_dict`."""
+        data: JSONDict = {'id': self.id, 'username': self.username, 'first_name': self.first_name}
+
+        if self.last_name:
+            data['last_name'] = self.last_name
+
+        return data
+
+    def __eq__(self, other: object) -> bool:
+        return self.bot == other
+
+    def __hash__(self) -> int:
+        return hash(self.bot)
 
     # camelCase aliases
     getMe = get_me
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -22,6 +22,7 @@
 
 import functools
 import logging
+import pickle
 from datetime import datetime
 
 from typing import (
@@ -5611,6 +5612,9 @@ class Bot(TelegramObject):
 
     def __hash__(self) -> int:
         return hash(self.bot)
+
+    def __reduce__(self):
+        raise pickle.PicklingError('Bot objects cannot be pickled!')
 
     # camelCase aliases
     getMe = get_me
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -202,9 +202,7 @@ class Bot(TelegramObject):
         return hash(self.bot)
 
     def __reduce__(self) -> NoReturn:
-        """
-        Called by pickle.dumps(). Serializing bots is unadvisable, so we try to forbid pickling.
-        """
+        """Called by pickle.dumps(). Serializing bots is unadvisable, so we forbid pickling."""
         raise pickle.PicklingError('Bot objects cannot be pickled!')
 
     def to_dict(self) -> JSONDict:
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -205,6 +205,10 @@ class Bot(TelegramObject):
         """Called by pickle.dumps(). Serializing bots is unadvisable, so we forbid pickling."""
         raise pickle.PicklingError('Bot objects cannot be pickled!')
 
+    def __deepcopy__(self, memodict: dict) -> 'Bot':
+        """Bot objects can't be deepcopied since they are not pickable, so just return self."""
+        return self
+
     def to_dict(self) -> JSONDict:
         """See :meth:`telegram.TelegramObject.to_dict`."""
         data: JSONDict = {'id': self.id, 'username': self.username, 'first_name': self.first_name}
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -38,6 +38,7 @@ from typing import (
     cast,
     Sequence,
     Any,
+    NoReturn,
 )
 
 try:
@@ -158,7 +159,7 @@ class Bot(TelegramObject):
         'private_key',
         '_bot_user',
         '_request',
-        'logger',
+        '_logger',
     )
 
     def __init__(
@@ -177,7 +178,7 @@ class Bot(TelegramObject):
         self._bot_user: Optional[User] = None
         self._request = request or Request()
         self.private_key = None
-        self.logger = logging.getLogger(__name__)
+        self._logger = logging.getLogger(__name__)
 
         if private_key:
             if not CRYPTO_INSTALLED:
@@ -3000,9 +3001,9 @@ class Bot(TelegramObject):
         )
 
         if result:
-            self.logger.debug('Getting updates: %s', [u['update_id'] for u in result])
+            self._logger.debug('Getting updates: %s', [u['update_id'] for u in result])
         else:
-            self.logger.debug('No new updates found.')
+            self._logger.debug('No new updates found.')
 
         return Update.de_list(result, self)  # type: ignore[return-value]
 
@@ -5613,7 +5614,8 @@ class Bot(TelegramObject):
     def __hash__(self) -> int:
         return hash(self.bot)
 
-    def __reduce__(self):
+    def __reduce__(self) -> NoReturn:
+        """Called by pickle.dumps() and only when bot instance is present in the data structure."""
         raise pickle.PicklingError('Bot objects cannot be pickled!')
 
     # camelCase aliases
--- a/telegram/_telegramobject.py
+++ b/telegram/_telegramobject.py
@@ -17,6 +17,7 @@
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """Base class for Telegram Objects."""
+from copy import deepcopy
 try:
     import ujson as json
 except ImportError:
@@ -53,15 +54,12 @@ class TelegramObject:
         _bot: Optional['Bot']
     # Adding slots reduces memory usage & allows for faster attribute access.
     # Only instance variables should be added to __slots__.
-    __slots__ = (
-        '_id_attrs',
-        '_bot',
-    )
+    __slots__ = ('_id_attrs', '_bot')
 
     # pylint: disable=unused-argument
     def __new__(cls, *args: object, **kwargs: object) -> 'TelegramObject':
         # We add _id_attrs in __new__ instead of __init__ since we want to add this to the slots
-        # w/o calling __init__ in all of the subclasses. This is what we also do in BaseFilter.
+        # w/o calling __init__ in all of the subclasses.
         instance = super().__new__(cls)
         instance._id_attrs = ()
         instance._bot = None
@@ -211,3 +209,14 @@ class TelegramObject:
         if self._id_attrs:
             return hash((self.__class__, self._id_attrs))
         return super().__hash__()
+
+    def __getstate__(self):
+        return self.to_dict()
+
+    def __setstate__(self, state):
+        self.__init__(**state)
+
+    def __deepcopy__(self, memodict):
+        copy = self.__class__(**deepcopy(self.__getstate__(), memodict))
+        copy.set_bot(self.get_bot())
+        return copy
--- a/telegram/_telegramobject.py
+++ b/telegram/_telegramobject.py
@@ -84,6 +84,62 @@ class TelegramObject:
                 f"`{item}`."
             ) from exc
 
+    def __eq__(self, other: object) -> bool:
+        # pylint: disable=no-member
+        if isinstance(other, self.__class__):
+            if self._id_attrs == ():
+                warn(
+                    f"Objects of type {self.__class__.__name__} can not be meaningfully tested for"
+                    " equivalence.",
+                    stacklevel=2,
+                )
+            if other._id_attrs == ():
+                warn(
+                    f"Objects of type {other.__class__.__name__} can not be meaningfully tested"
+                    " for equivalence.",
+                    stacklevel=2,
+                )
+            return self._id_attrs == other._id_attrs
+        return super().__eq__(other)
+
+    def __hash__(self) -> int:
+        # pylint: disable=no-member
+        if self._id_attrs:
+            return hash((self.__class__, self._id_attrs))
+        return super().__hash__()
+
+    def __getstate__(self) -> Dict[str, Union[str, object]]:
+        """
+        This method is used for pickling. We remove the bot attribute of the object since those
+        are not pickable.
+        """
+        # Adapted from self.to_dict(), unlike to_dict(), this is not recursive.
+        data = {}
+        attrs = {attr for cls in self.__class__.__mro__[:-1] for attr in cls.__slots__}
+        for key in attrs:
+            if key.startswith('_'):
+                continue
+
+            value = getattr(self, key, None)
+            if value is not None:
+                data[key] = value
+        return data
+
+    def __setstate__(self, state: dict) -> None:
+        """
+        This method is used for unpickling. The data, which is in the form a dictionary, is
+        converted back into a class.
+        """
+        self.__init__(**state)  # type: ignore[misc]
+        if '_bot' in state:
+            self._bot = state['_bot']
+
+    def __deepcopy__(self, memodict: dict) -> object:
+        """This method deepcopies the object and sets the bot on the newly created copy."""
+        copy = self.__class__(**deepcopy(self.__getstate__(), memodict))
+        copy.set_bot(self._bot)
+        return copy
+
     @staticmethod
     def _parse_data(data: Optional[JSONDict]) -> Optional[JSONDict]:
         return None if data is None else data.copy()
@@ -189,59 +245,3 @@ class TelegramObject:
             bot (:class:`telegram.Bot` | :obj:`None`): The bot instance.
         """
         self._bot = bot
-
-    def __eq__(self, other: object) -> bool:
-        # pylint: disable=no-member
-        if isinstance(other, self.__class__):
-            if self._id_attrs == ():
-                warn(
-                    f"Objects of type {self.__class__.__name__} can not be meaningfully tested for"
-                    " equivalence.",
-                    stacklevel=2,
-                )
-            if other._id_attrs == ():
-                warn(
-                    f"Objects of type {other.__class__.__name__} can not be meaningfully tested"
-                    " for equivalence.",
-                    stacklevel=2,
-                )
-            return self._id_attrs == other._id_attrs
-        return super().__eq__(other)
-
-    def __hash__(self) -> int:
-        # pylint: disable=no-member
-        if self._id_attrs:
-            return hash((self.__class__, self._id_attrs))
-        return super().__hash__()
-
-    def __getstate__(self) -> Dict[str, Union[str, object]]:
-        """
-        This method is used for pickling. We remove the bot attribute of the object since those
-        are not pickable.
-        """
-        # Adapted from self.to_dict(), unlike to_dict(), this is not recursive.
-        data = {}
-        attrs = {attr for cls in self.__class__.__mro__[:-1] for attr in cls.__slots__}
-        for key in attrs:
-            if key.startswith('_'):
-                continue
-
-            value = getattr(self, key, None)
-            if value is not None:
-                data[key] = value
-        return data
-
-    def __setstate__(self, state: dict) -> None:
-        """
-        This method is used for unpickling. The data, which is in the form a dictionary, is
-        converted back into a class.
-        """
-        self.__init__(**state)  # type: ignore[misc]
-        if '_bot' in state:
-            self._bot = state['_bot']
-
-    def __deepcopy__(self, memodict: dict) -> object:
-        """This method deepcopies the object and sets the bot on the newly created copy."""
-        copy = self.__class__(**deepcopy(self.__getstate__(), memodict))
-        copy.set_bot(self._bot)
-        return copy
--- a/telegram/_telegramobject.py
+++ b/telegram/_telegramobject.py
@@ -43,8 +43,10 @@ class TelegramObject:
     with the appropriate name, a :exc:`KeyError` will be raised.
 
     When objects of this type are pickled, the :class:`~telegram.Bot` attribute associated with the
-    object will be removed. However, when deepcopying the object, the :class:`~telegram.Bot`
-    attribute is *not* removed.
+    object will be removed. However, when copying the object via :func:`copy.deepcopy`, the copy
+    will have the *same* bot instance associated with it, i.e::
+
+        assert telegram_object.get_bot() is copy.deepcopy(telegram_object).get_bot()
 
     .. versionchanged:: 14.0
         ``telegram_object['from']`` will look up the key ``from_user``. This is to account for
@@ -113,17 +115,7 @@ class TelegramObject:
         This method is used for pickling. We remove the bot attribute of the object since those
         are not pickable.
         """
-        # Adapted from self.to_dict(), unlike to_dict(), this is not recursive.
-        data = {}
-        attrs = {attr for cls in self.__class__.__mro__[:-1] for attr in cls.__slots__}
-        for key in attrs:
-            if key.startswith('_'):
-                continue
-
-            value = getattr(self, key, None)
-            if value is not None:
-                data[key] = value
-        return data
+        return self._get_attrs(include_private=False, recursive=False)
 
     def __setstate__(self, state: dict) -> None:
         """
@@ -132,11 +124,35 @@ class TelegramObject:
         """
         self.__init__(**state)  # type: ignore[misc]
 
-    def __deepcopy__(self, memodict: dict) -> object:
+    def __deepcopy__(self: TO, memodict: dict) -> TO:
         """This method deepcopies the object and sets the bot on the newly created copy."""
         copy = self.__class__(**deepcopy(self.__getstate__(), memodict))
         copy.set_bot(self._bot)
         return copy
+
+    def _get_attrs(
+        self, include_private: bool = False, recursive: bool = False
+    ) -> Dict[str, Union[str, object]]:
+        data = {}
+
+        # We want to get all attributes for the class, using self.__slots__ only includes the
+        # attributes used by that class itself, and not its superclass(es). Hence, we get its MRO
+        # and then get their attributes. The `[:-1]` slice excludes the `object` class
+        attrs = {attr for cls in self.__class__.__mro__[:-1] for attr in cls.__slots__}
+        for key in attrs:
+            if not include_private and key.startswith('_'):
+                continue
+
+            value = getattr(self, key, None)
+            if value is not None:
+                if recursive and hasattr(value, 'to_dict'):
+                    data[key] = value.to_dict()
+                else:
+                    data[key] = value
+
+        if recursive and data.get('from_user'):
+            data['from'] = data.pop('from_user', None)
+        return data
 
     @staticmethod
     def _parse_data(data: Optional[JSONDict]) -> Optional[JSONDict]:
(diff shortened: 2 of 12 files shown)


telegram/_bot.py
```diff
@@ -5612,6 +5613,9 @@ def __eq__(self, other: object) -> bool:
     def __hash__(self) -> int:
         return hash(self.bot)
 
+    def __reduce__(self):
+        raise pickle.PicklingError('Bot objects cannot be pickled!')
+
```
Bibo-Joshi, line 5618: I'm not sure if this necessary. The default-implementation is already not picklable because locks are used in the request object (both with urllib3 and with httpx). This is also highly likely to be the case for any custom implementation of `BasePersistence` on v14.
Apart from that, users can still override this to eg skip the request object on pickle.dumps …

telegram/_bot.py
```diff
@@ -22,6 +22,7 @@
 
 import functools
 import logging
+import pickle
```
Bibo-Joshi, line 25: unused

telegram/_telegramobject.py
```diff
@@ -211,3 +209,14 @@ def __hash__(self) -> int:
         if self._id_attrs:
             return hash((self.__class__, self._id_attrs))
         return super().__hash__()
+
+    def __getstate__(self):
+        return self.to_dict()
```
Bibo-Joshi, line 214: please check how this treats types apart from int/str/float - e.g. IIRC `to_dict` converts datetimes to float timestamps and init doesn't re-convert 🤔

telegram/ext/_basepersistence.py
```diff
@@ -73,33 +73,30 @@ class BasePersistence(Generic[UD, CD, BD], ABC):
...
+    :meth:`update_bot_data` or :meth:`refresh_bot_data`.
 
     Note:
-         :meth:`replace_bot` and :meth:`insert_bot` are used *independently* of the implementation
-         of the ``update/get_*`` methods, i.e. you don't need to worry about it while
-         implementing a custom persistence subclass.
+        You should avoid saving :class:`telegram.Bot` instances, this is because if you change
+        the bots token, this may lead to e.g. ``Chat not found`` errors. To prevent this, the
+        persistence *may* automatically use :attr:`bot` to replace the bot. Otherwise, you must
+        manually set the bot with :meth:`set_bot` upon loading of the data.
```
Bibo-Joshi, line 95:
```suggestion
        You should avoid saving :class:`telegram.Bot` instances. This is because if you change
        e.g. the bots token, this won't propagate to the serialized instances and may lead exceptions.
        
        To prevent this, the implemtation may use :attr:`bot` to replace bot instances with a placeholder before serialization and insert :attr:`bot` back when loading the data. Since :attr:`bot` will be set when the process starts, this will be the up-to-date bot instance.
        
        If the persistence implementation does not take care of this, you should make sure not to store any bot instances in the data that will be persisted. E.g. in case of :class:`TelegramObject`, one may call :meth:`set_bot` to ensure that shortcuts like `Message.reply_text` are available.
```

telegram/ext/_basepersistence.py
```diff
@@ -73,33 +73,30 @@ class BasePersistence(Generic[UD, CD, BD], ABC):
...
 
     .. versionchanged:: 14.0
-        The parameters and attributes ``store_*_data`` were replaced by :attr:`store_data`.
+        * The parameters and attributes ``store_*_data`` were replaced by :attr:`store_data`.
+        * ``insert/replace_bot`` was dropped since it was error-prone and had considerable overhead
```
Bibo-Joshi, line 99:
```suggestion
        * ``insert/replace_bot`` was dropped. Serialization of bot instances now needs to be handled by the specific implementation - see above note.
```

telegram/ext/_picklepersistence.py
```diff
@@ -28,11 +28,50 @@
...
 )
 
+from telegram import Bot
 from telegram._utils.types import FilePathInput
+from telegram._utils.warnings import warn
 from telegram.ext import BasePersistence, PersistenceInput
 from telegram.ext._contexttypes import ContextTypes
 from telegram.ext._utils.types import UD, CD, BD, ConversationDict, CDCData
 
+_REPLACED_KNOWN_BOT = "known_bot"
```
Bibo-Joshi, line 38: should be something more unique in a final version :D

telegram/ext/_picklepersistence.py
```diff
@@ -28,11 +28,50 @@
...
+_REPLACED_UNKNOWN_BOT = "unknown_bot"
+
+
+class BotPickler(pickle.Pickler):
```
Bibo-Joshi, line 42:
```suggestion
class _BotPickler(pickle.Pickler):
```
same below

telegram/ext/_picklepersistence.py
```diff
@@ -41,15 +80,10 @@ class PicklePersistence(BasePersistence[UD, CD, BD]):
...
-        :attr:`~telegram.ext.BasePersistence.REPLACED_BOT` and insert the bot set with
-        :meth:`telegram.ext.BasePersistence.set_bot` upon loading of the data. This is to ensure
-        that changes to the bot apply to the saved objects, too. If you change the bots token, this
-        may lead to e.g. ``Chat not found`` errors. For the limitations on replacing bots see
-        :meth:`telegram.ext.BasePersistence.replace_bot` and
-        :meth:`telegram.ext.BasePersistence.insert_bot`.
+    Note:
+        :class:`PicklePersistence` will automatically replace :class:`telegram.Bot` instances with
+        :attr:`~BasePersistence.bot`. So, if you for e.g. change the bot's token, you do *not* need
+        to manually call :meth:`~BasePersistence.set_bot`.
```
Bibo-Joshi, line 86:
```suggestion
    Note:
        This implementation of `BasePersistence` uses the functionality of the pickle module to support serialization of bot instances. Specifically any reference to :attr:`bot` will be replaced by a placeholder before pickling and :attr:`bot` will be inserted back when loading the data.
```

telegram/ext/_picklepersistence.py
```diff
@@ -151,7 +185,11 @@ def __init__(
     def _load_singlefile(self) -> None:
         try:
             with self.filepath.open("rb") as file:
-                data = pickle.load(file)
+                if self.bot is None:
+                    data = pickle.load(file)
+                else:
+                    data = BotUnpickler(file, self.bot).load()
```
Bibo-Joshi, line 191: why not always `BotUpickler`? in case `self.bot is None`, `None` is still more helpful than `"known_bot"`, isn't it?

telegram/_telegramobject.py
```diff
@@ -17,6 +17,7 @@
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """Base class for Telegram Objects."""
+from copy import deepcopy
```
Bibo-Joshi, line 20: class docstring should have a paragraph on pickle & deepcopy behavior :)

telegram/ext/_picklepersistence.py
```diff
@@ -28,11 +28,50 @@
...
+
+class BotUnpickler(pickle.Unpickler):
+    def __init__(self, file, bot):
+        super().__init__(file)
+        self._bot = bot
+
+    def persistent_load(self, pid):
+        # print('in unpickler persistent_id')
+        # print(pid)
+        if pid == _REPLACED_KNOWN_BOT:
```
Bibo-Joshi, line 68:
```suggestion
        if isinstance(pid, str) and pid == _REPLACED_KNOWN_BOT:
```
same below - I recall that we had that problem in `insert/replace_bot` when some `__eq__` raises exceptions …

telegram/_bot.py
Bibo-Joshi: ah, in the mock-up I didn't want to import `telegram.Bot`, so I manually produced an exception.
Anyway, I'd also be okay with keeping it, just wanted to point out the culprits

telegram/_telegramobject.py
harshil21: would calling `de_json(states)` instead suffice?

telegram/ext/_picklepersistence.py
harshil21: I guess, just followed your [mock up ](https://nekobin.com/lifojagumo.py)

telegram/ext/_picklepersistence.py
harshil21: > insert/replace_bot when some __eq__ raises exceptions

im not aware of this problem and what how its related here?

and `persistent_load` is only called for bot instances (whenever return something other than `None` in persistent_id`) so this is not an issue?

telegram/_telegramobject.py
Bibo-Joshi: should work, I guess so … however again the recursive to_dict/de_json also takes some runtime, that's not really needed here. Maybe we can extract the logic of getting the relevant attributes from `to_dict()` and use it in both places?
in any case, test cases will have to cover such things :)

telegram/ext/_picklepersistence.py
Bibo-Joshi: should do the job :) the content doesn't really matter - we could also just use a uuid4. we'll just have to make sure that it's unlikely for someone to actually use that string elsewhere …

telegram/ext/_picklepersistence.py
Bibo-Joshi: xD

telegram/ext/_dictpersistence.py
```diff
@@ -39,19 +41,16 @@ class DictPersistence(BasePersistence):
...
-        :attr:`~telegram.ext.BasePersistence.REPLACED_BOT` and insert the bot set with
-        :meth:`telegram.ext.BasePersistence.set_bot` upon loading of the data. This is to ensure
-        that changes to the bot apply to the saved objects, too. If you change the bots token, this
-        may lead to e.g. ``Chat not found`` errors. For the limitations on replacing bots see
-        :meth:`telegram.ext.BasePersistence.replace_bot` and
-        :meth:`telegram.ext.BasePersistence.insert_bot`.
+        * This class does *not* implement a :meth:`flush` method, meaning that data managed by
+          ``DictPersistence`` is in-memory only and will be lost when the bot shuts down. This is,
+          because ``DictPersistence`` is mainly intended as starting point for custom persistence
+          classes that need to JSON-serialize the stored data before writing them to file/database.
```
Bibo-Joshi, line 47: This actually misinterprets the intention of `flush` - maybe I wrote this and didn't understand it at that point? :D Anyway, since we're at it:

```suggestion
        * Data managed by ``DictPersistence`` is in-memory only and will be lost when the bot shuts down. This is,
          because ``DictPersistence`` is mainly intended as starting point for custom persistence
          classes that need to JSON-serialize the stored data before writing them to file/database.
```

telegram/ext/_dictpersistence.py
```diff
@@ -39,19 +41,16 @@ class DictPersistence(BasePersistence):
...
+
+        * This implementation of :class:`BasePersistence` does **not** replace bot instances with a
+          placeholder before serialization. Hence, you should make sure not to
+          store any bot instances in the data that will be persisted. E.g. in case of
+          :class:`telegram.TelegramObject`, one may call :meth:`BasePersistence.set_bot` to ensure
+          that shortcuts like :meth:`telegram.Message.reply_text` are available.
```
Bibo-Joshi, line 53: I'm not sure if this comment is necessary here - `TelegramObject` is not directly json dumpable anyway. Maybe instead make it a comment about this class only handling data that can be serialized by `json.dump(s)`?

telegram/ext/_picklepersistence.py
```diff
@@ -26,14 +28,99 @@
...
+
+def _all_subclasses(cls: Type[TelegramObject]) -> Set[Type[TelegramObject]]:
+    """Gets all subclasses of the specified object, recursively. from
+    https://stackoverflow.com/a/3862957/9706202
+    """
+    subclasses = cls.__subclasses__()
+    return set(subclasses).union([s for c in subclasses for s in _all_subclasses(c)])
+
+
+def _reconstruct_to(cls: Type[TelegramObject], kwargs: dict) -> TelegramObject:
```
Bibo-Joshi, line 56:
```suggestion
TO = TypeVar('TO', bound=TelegramObject)

def _reconstruct_to(cls: Type[TO ], kwargs: dict) -> TO:
```

telegram/ext/_picklepersistence.py
```diff
@@ -26,14 +28,99 @@
...
+    """
+    This method is used for unpickling. The data, which is in the form a dictionary, is
+    converted back into a class. Functions the same as :meth:`TelegramObject.__setstate__`
+    """
+    obj = cls(**kwargs)
+    if '_bot' in kwargs:
+        obj.set_bot(kwargs['_bot'])
+    return obj
```
Bibo-Joshi, line 64: maybe 
```suggestion
    bot = kwargs.pop('_bot', None)
    obj = cls(**kwargs)
    obj.set_bot(bot)
    return obj
```
? avoids passing the redundant parameter to the constructor

telegram/ext/_picklepersistence.py
```diff
@@ -26,14 +28,99 @@
...
+
+
+def _custom_reduction(cls: TelegramObject) -> Tuple[Callable, Tuple[Type[TelegramObject], dict]]:
```
Bibo-Joshi, line 67: TO as above

telegram/ext/_picklepersistence.py
```diff
@@ -26,14 +28,99 @@
...
+    This method is used for pickling. The bot attribute is preserved so _BotPickler().persistent_id
+    works as intended.
+    """
+    # Adapted from self.to_dict(), unlike to_dict(), this is not recursive.
+    data = {}
+    attrs = {attr for cls in cls.__class__.__mro__[:-1] for attr in cls.__slots__}
+    for key in attrs:
+        value = getattr(cls, key, None)
+        if value is not None:
+            data[key] = value
```
Bibo-Joshi, line 78: why not just call `__getstate__` here? IMO the minor break of encapsulation is justifiable since it removes code duplication. I'm not even sure, if calling magic methods is considered calling protected methods 🤔

telegram/ext/_picklepersistence.py
```diff
@@ -26,14 +28,99 @@
...
+    """
+    subclasses = cls.__subclasses__()
+    return set(subclasses).union([s for c in subclasses for s in _all_subclasses(c)])
+
+
+def _reconstruct_to(cls: Type[TelegramObject], kwargs: dict) -> TelegramObject:
+    """
+    This method is used for unpickling. The data, which is in the form a dictionary, is
+    converted back into a class. Functions the same as :meth:`TelegramObject.__setstate__`
+    """
```
Bibo-Joshi, line 60:
```suggestion
    """
    This method is used for unpickling. The data, which is in the form a dictionary, is
    converted back into a class. Functions the same as :meth:`TelegramObject.__setstate__`
    This function should be kept in place for backwards compatibility even if the pickling logic is changed, since `_custom_reduction` places references to this function into the pickled data
    """
```

telegram/ext/_picklepersistence.py
```diff
@@ -26,14 +28,99 @@
...
+
+class _BotPickler(pickle.Pickler):
+    __slots__ = ('_bot',)
+
+    # Here we define a private dispatch_table, because we want to preserve the bot attribute of
+    # objects so persistent_id works as intended. Otherwise, the bot attribute is deleted in
+    # __getstate__, which used for regular pickling (via pickle.dumps(...))
+    dispatch_table = copyreg.dispatch_table.copy()  # type: ignore[attr-defined]
+    for obj in _all_subclasses(TelegramObject):
+        dispatch_table[obj] = _custom_reduction
```
Bibo-Joshi, line 91: is there a specific reason why this is not in `__init__`? And We only need this in py3.7, right? If `reducer_override` is respected in py3.8+, I would suggest to place this in an `if py_version < (3,8)` clause and implement the `reducer_override` method in addition.
My gut feeling is that the `reducer_override` approach may be a bit more stable than the "get all subclasses" logic

telegram/ext/_picklepersistence.py
```diff
@@ -26,14 +28,99 @@
...
+class _BotUnpickler(pickle.Unpickler):
+    __slots__ = ('_bot',)
+
+    def __init__(self, bot: Bot, *args: Any, **kwargs: Any):
+        self._bot = bot
+        super().__init__(*args, **kwargs)
+
+    def persistent_load(self, pid: str) -> Optional[Bot]:
+        """Replaces the bot with the current bot if known, else it is replaced by :obj:`None`."""
+        return self._bot if pid == _REPLACED_KNOWN_BOT else None
```
Bibo-Joshi, line 121: to increase backwards compatibility, we could adjust this check such that it also accepts the current constant `BasePersistence.REPLACED_BOT` (or whatever the current name is). We could even just use the current value for `_R_K_B`.
It would also be nice to have a unit test for the backwards compatibility - one could dump some test data with the current implementation, save it in `tests/data` and make sure that it's correctly unpickled with the new implementation

telegram/ext/_picklepersistence.py
```diff
@@ -191,12 +275,11 @@ def _dump_singlefile(self) -> None:
             'callback_data': self.callback_data,
         }
         with self.filepath.open("wb") as file:
-            pickle.dump(data, file)
+            _BotPickler(self.bot, file, protocol=pickle.HIGHEST_PROTOCOL).dump(data)
```
Bibo-Joshi, line 278: curiosity: why the `HIGHEST_PROTOCOL`?

tests/test_callbackdatacache.py
```diff
@@ -159,7 +159,8 @@ def test_process_callback_query(self, callback_data_cache, data, message, invali
         if invalid:
             callback_data_cache.clear_callback_data()
 
-        effective_message = Message(message_id=1, date=None, chat=None, reply_markup=out)
+        chat = Chat(1, 'private')
+        effective_message = Message(message_id=1, date=datetime.now(), chat=chat, reply_markup=out)
```
Bibo-Joshi, line 163: is this related to this PR?

tests/test_persistence.py
```diff
@@ -1633,6 +1345,50 @@ def test_save_on_flush_single_files(self, pickle_persistence, good_pickle_files)
...
+        self, bot, update, pickle_persistence, good_pickle_files, recwarn
+    ):
+        u = UpdaterBuilder().bot(bot).persistence(pickle_persistence).build()
+        dp = u.dispatcher
+        bot_id = None
+
+        def first(update, context):
+            nonlocal bot_id
+            bot_id = update.message.get_bot()
+            assert bot_id
```
Bibo-Joshi, line 1358: I don't understand this one. `assert bot_id is bot`, right? what does this test?

tests/test_persistence.py
```diff
@@ -1633,6 +1345,50 @@ def test_save_on_flush_single_files(self, pickle_persistence, good_pickle_files)
             conversations_test = dict(pickle.load(f))['conversations']
         assert conversations_test['name1'] == conversation1
 
+    def test_custom_pickler_unpickler(
```
Bibo-Joshi, line 1348: why does this test use handler logic? Wouldn't it suffice to test with somethling like

```python
persistence.update_chat_data(chat_id, data_containing_bot)
persistence.flush()
persistence = Persistence(…)
loaded_data = persistence.get_chat_data
# make assertions
```
?

tests/test_telegramobject.py
```diff
@@ -145,3 +147,33 @@ def test_subscription(self):
...
+
+        with pytest.raises(RuntimeError):
+            unpickled.get_bot()  # There should be no bot when we pickle TGObjects
+
+        assert unpickled.chat == chat
+        assert unpickled.from_user == user
+        assert unpickled.date == date
+        assert unpickled.photo[0] == photo
+
+        # Now lets test deepcopying
```
Bibo-Joshi, line 170: please test deepcopy in a separate test

Bibo-Joshi (changes requested):
Thanks for the work on this! I left a number of comments. 

Additionally, please elaborate in the docstring of `Bot` why bots should not be serialized and that for that reason pickling will raise an exception

telegram/_bot.py
```diff
@@ -188,6 +190,25 @@ def __init__(
...
+    def __reduce__(self) -> NoReturn:
+        """Called by pickle.dumps() and only when bot instance is present in the data structure."""
+        raise pickle.PicklingError('Bot objects cannot be pickled!')
+
+    def to_dict(self) -> JSONDict:
+        """See :meth:`telegram.TelegramObject.to_dict`."""
+        data: JSONDict = {'id': self.id, 'username': self.username, 'first_name': self.first_name}
+
+        if self.last_name:
+            data['last_name'] = self.last_name
```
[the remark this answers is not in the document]
harshil21: It is in fact tested + I only moved the function so a  codecov false positive maybe

telegram/_bot.py
```diff
@@ -188,6 +190,25 @@ def __init__(
...
             )
 
+    def __eq__(self, other: object) -> bool:
+        return self.bot == other
+
+    def __hash__(self) -> int:
+        return hash(self.bot)
+
+    def __reduce__(self) -> NoReturn:
+        """Called by pickle.dumps() and only when bot instance is present in the data structure."""
```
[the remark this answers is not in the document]
harshil21: Shown in the docstrings == Shown in the docs?

No, dunder methods aren't documented by default.

I can change it into a comment then

telegram/_bot.py
```diff
@@ -188,6 +190,25 @@ def __init__(
                 private_key, password=private_key_password, backend=default_backend()
             )
 
+    def __eq__(self, other: object) -> bool:
+        return self.bot == other
+
+    def __hash__(self) -> int:
+        return hash(self.bot)
```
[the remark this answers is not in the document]
harshil21: Yes

telegram/_telegramobject.py
```diff
@@ -81,6 +84,60 @@ def __getitem__(self, item: str) -> object:
...
+        # Adapted from self.to_dict(), unlike to_dict(), this is not recursive.
+        data = {}
+        attrs = {attr for cls in self.__class__.__mro__[:-1] for attr in cls.__slots__}
+        for key in attrs:
+            if key.startswith('_'):
+                continue
+
+            value = getattr(self, key, None)
+            if value is not None:
+                data[key] = value
```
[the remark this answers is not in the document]
harshil21: Yes, maybe the three methods `to_dict()`, `_custom_reduction` and `__getstate__` can all share the same code. Put this in a file in tg.utils?

telegram/ext/_picklepersistence.py
harshil21: I'll unify all 3 methods as said above

> I'm not even sure if...

Don't think so. We call `__mro__ ` in there and that's standard practice

telegram/ext/_picklepersistence.py
harshil21: > why not in `__init__`?

Well we don't need to generate the dispatch_table again and again on `__init__`, doing it once is fine imo. 

And yes, I guess it makes sense to do the `reduce_override` thing for future proofing

telegram/ext/_picklepersistence.py
harshil21: Good idea!

telegram/ext/_picklepersistence.py
harshil21: Provides performance increase for py3.8+

tests/test_persistence.py
harshil21: That was just testing if a bot is present in the message, but guess that's not needed, since get_bot() would fail when there is no bot anyway

tests/test_persistence.py
harshil21: > why use handler logic?

Wanted to be more sure that this would work, guess I can add this 'simpler' test in addition to the handle one

tests/test_telegramobject.py
harshil21: Well it does share the code, agh guess I'll make it into two

telegram/_bot.py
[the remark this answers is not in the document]
Bibo-Joshi: If it doesn't render on RTD, then you can keep it a docstring :) just wanted to make sure that it doesn't :)

telegram/ext/_picklepersistence.py
Bibo-Joshi: > > why not in `__init__`?
> 
> Well we don't need to generate the dispatch_table again and again on `__init__`, doing it once is fine imo.

What happens if I add a new subclass at runtime?

> And yes, I guess it makes sense to do the `reduce_override` thing for future proofing

tests/test_persistence.py
Bibo-Joshi: I see. My reasoning is to separate the unit tests from the integration tests. `TestBasePersistence` already has tests for the dispatcher integration, so IMO it would suffice to test the methods of `PicklePersistence` here. I'm okay with having the handler-tests here as well, but in that case please make it a new test. That way, when dispatcher integration is change, it will be more clear why a failing test is failing :)

telegram/_bot.py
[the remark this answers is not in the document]
Bibo-Joshi: I rather think that this missing coveraged just showed because you moved it :)

telegram/_telegramobject.py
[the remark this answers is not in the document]
Bibo-Joshi: The logic is unique to `TelegramObject` and needs access to `self`, so a method `TO._whatever_naming` would feel more natural to me. Especially, if we're fine with `_custom_reduction` calling `TO.__getstate__`

tests/test_callbackdatacache.py
Bibo-Joshi: Ah okay, that's fine then. Just feared that there might be some consequences of the changes that we hadn't seen yet

telegram/ext/_picklepersistence.py
harshil21: On second thought, this would not be possible, since `persistent_load` will only run when the data has been tagged by the `persistent_id` during pickling. So for backwards compatibility, in the automatic transition script, we should:

1. Subclass `Bot` to add back [`__reduce__`](https://github.com/python-telegram-bot/python-telegram-bot/pull/1994/files#diff-ca0526e3005b92a3844ee53d68688fe5a8c438332365a0efc3275a814cca5008L3814-L3816)
2. Ask user for their token, etc to make a `Bot`
3.  Use `BP.insert_bot` to find `BP.REPLACED_BOT` and replace with the bot in step 1
4. Pickle the data back, using `pickle.dumps`

But one problem I see with this is that when the user starts running their bot in v14, `persistent_id` will mark this as an unknown bot. To mitigate this, we could hardcode it so that it recognizes that this bot was for the transition only.

What do you think?

harshil21: Alternatively, we could add logic to `persistent_id` to handle `BP.REPLACED_BOT`, but the actual bot would be inserted only in the second bot restart...

Bibo-Joshi: Ah, right. Mh, if we can't provide a proper backwards compatibility, then I guess a transition script is more sane. For that I guess it would suffice the load the data with very code and dump it with v14 code, no?

```python
from v13_pp import PP as PPv13
from tg.ext import PP as PPv14

ppv13 = PPv13(...)
ppv13.set_bot(...)
ppv14 = PPv14(...)
ppv14.set_bot(...)
chat_data = ppv13.get_chat_data()
pp14.update_chat_data(...)
pp14.flush()

```

harshil21: Yes.. that would work too I supoose

harshil21:
Current approach of `__setstate__` and `__deepcopy__` fails when attribute names of an object ≠ the expected argument names (e.g. a class taking `attr` as argument, but the self variable is assigned as: `self._attr = attr`).

Also fails when there are extra args and the constructor does not accept **kwargs.
So will redesign this.

harshil21:
Another way to achieve deepcopying without actually defining `__deepcopy__`, would be to use `was_called_by` in `__getstate__` as follows:

``` python
import copy
...
if was_called_by(inspect.currentframe(), Path(copy.__file__)):
    return self._get_attrs(include_private=True, recursive=False)
return self._get_attrs(include_private=True, recursive=False, remove_bot=True)
```

but didn't use this since it feels a bit hacky and unpythonic, although its elegant.

telegram/_bot.py
```diff
@@ -123,10 +125,14 @@ class Bot(TelegramObject):
...
-        to the Telegram API. This can be used to access new features of the API before they were
-        incorporated into PTB. However, this is not guaranteed to work, i.e. it will fail for
-        passing files.
+        * Most bot methods have the argument ``api_kwargs`` which allows passing arbitrary keywords
+          to the Telegram API. This can be used to access new features of the API before they are
+          incorporated into PTB. However, this is not guaranteed to work, i.e. it will fail for
+          passing files.
+        * Bots should not be serialized since if you for e.g. change the bots token, then your
+          serialized instance will not reflect that change. Hence, we raise a
+          :exc:`pickle.PicklingError` upon serialization.
```
Bibo-Joshi, line 134:
```suggestion
          serialized instance will not reflect that change. Trying to pickle a bot instance will raise
          :exc:`pickle.PicklingError`.
```

telegram/_bot.py
```diff
@@ -188,6 +195,29 @@ def __init__(
...
+    def __hash__(self) -> int:
+        return hash(self.bot)
+
+    def __reduce__(self) -> NoReturn:
+        """Called by pickle.dumps(). Serializing bots is unadvisable, so we forbid pickling."""
+        raise pickle.PicklingError('Bot objects cannot be pickled!')
+
+    def __deepcopy__(self, memodict: dict) -> 'Bot':
+        """Bot objects can't be deepcopied since they are not pickable, so just return self."""
+        return self
```
Bibo-Joshi, line 210: I disagree with this. trying to deepcopy a bot instance should either fail or return an actual deep copy. If any of the other changes in this PR rely on this, we'll have to rethink them

telegram/_bot.py
```diff
@@ -188,6 +195,29 @@ def __init__(
...
+
+    def to_dict(self) -> JSONDict:
+        """See :meth:`telegram.TelegramObject.to_dict`."""
+        data: JSONDict = {'id': self.id, 'username': self.username, 'first_name': self.first_name}
+
+        if self.last_name:
+            data['last_name'] = self.last_name
```
Bibo-Joshi, line 217: This line was not covered by tests. Can you please either add a test case or revert the moving of this method?

telegram/_telegramobject.py
```diff
@@ -81,6 +86,110 @@ def __getitem__(self, item: str) -> object:
                 f"`{item}`."
             ) from exc
 
+    def __eq__(self, other: object) -> bool:
```
Bibo-Joshi, line 89: tbh I find this moving of methods rather irritating 😄 if you didn't change anything in them, I'd prefer it if they just stayed where they are. makes git blame more reliable & makes reviewing easier

telegram/_telegramobject.py
```diff
@@ -81,6 +86,110 @@ def __getitem__(self, item: str) -> object:
...
+        result = cls.__new__(cls)  # create a new instance
+        memodict[id(self)] = result  # save the id of the object in the dict
+
+        attrs = self._get_attrs(include_private=True)  # get all its attributes
+
+        for k in attrs:  # now we set the attributes in the deepcopied object
+            setattr(result, k, deepcopy(getattr(self, k), memodict))
+
+        result.set_bot(bot)  # Assign the bots back
+        self._bot = bot
```
Bibo-Joshi, line 142:
```suggestion
        self.set_bot(bot)
```

telegram/_telegramobject.py
```diff
@@ -81,6 +86,110 @@ def __getitem__(self, item: str) -> object:
...
+        This method is used for unpickling. The data, which is in the form a dictionary, is
+        converted back into a class. Should be modified in place.
+        """
+        for key, val in state.items():
+            setattr(self, key, val)
+
+    def __deepcopy__(self: TO, memodict: dict) -> TO:
+        """This method deepcopies the object and sets the bot on the newly created copy."""
+        bot = self._bot  # Save bot so we can set it after copying
+        self._bot = None  # set to None so it is not deepcopied
```
Bibo-Joshi, line 131:
```suggestion
        bot = self.get_bot()  # Save bot so we can set it after copying
        self.set_bot(None)  # set to None so it is not deepcopied
```

telegram/ext/_dictpersistence.py
```diff
@@ -60,6 +56,8 @@ class DictPersistence(BasePersistence):
         store_data (:class:`PersistenceInput`, optional): Specifies which kinds of data will be
             saved by this persistence instance. By default, all available kinds of data will be
             saved.
+
+            .. versionadded:: 14.0
```
Bibo-Joshi, line 60: IMO this is already covered by the versionchanged directive

telegram/ext/_dictpersistence.py
```diff
@@ -76,6 +74,8 @@ class DictPersistence(BasePersistence):
     Attributes:
         store_data (:class:`PersistenceInput`): Specifies which kinds of data will be saved by this
             persistence instance.
+
+            .. versionadded:: 14.0
```
Bibo-Joshi, line 78: same

telegram/ext/_picklepersistence.py
```diff
@@ -17,23 +17,119 @@
...
+    https://stackoverflow.com/a/3862957/9706202
+    """
+    subclasses = cls.__subclasses__()
+    return set(subclasses).union([s for c in subclasses for s in _all_subclasses(c)])
+
+
+def _reconstruct_to(cls: Type[TO], kwargs: dict) -> TO:
+    """
+    This method is used for unpickling. The data, which is in the form a dictionary, is
+    converted back into a class. Functions the same as :meth:`TelegramObject.__setstate__`.
```
Bibo-Joshi, line 63:
```suggestion
    converted back into a class. Works mostly the same as :meth:`TelegramObject.__setstate__`.
```

Why not use `TO.__setstate__`, then?

telegram/ext/_picklepersistence.py
```diff
@@ -17,23 +17,119 @@
...
+    This function should be kept in place for backwards compatibility even if the pickling logic
+    is changed, since `_custom_reduction` places references to this function into the pickled data.
+    """
+    bot = kwargs.pop('_bot', None)
+    obj = cls.__new__(cls)
+    for key, val in kwargs.items():
+        setattr(obj, key, val)
+    obj.set_bot(bot)
```
Bibo-Joshi, line 71: why is it necessary to set `bot` that way instead of via `setattr`?

telegram/ext/_picklepersistence.py
```diff
@@ -17,23 +17,119 @@
...
+
+
+class _BotPickler(pickle.Pickler):
+    __slots__ = ('_bot',)
+
+    def __init__(self, bot: Bot, *args: Any, **kwargs: Any):
+        self._bot = bot
+        # Here we define a private dispatch_table, because we want to preserve the bot attribute of
+        # objects so persistent_id works as intended. Otherwise, the bot attribute is deleted in
+        # __getstate__, which is used during regular pickling (via pickle.dumps(...))
```
Bibo-Joshi, line 91: please move this comment into the `if py_ver` cause

telegram/ext/_picklepersistence.py
```diff
@@ -17,23 +17,119 @@
...
+class _BotUnpickler(pickle.Unpickler):
+    __slots__ = ('_bot',)
+
+    def __init__(self, bot: Bot, *args: Any, **kwargs: Any):
+        self._bot = bot
+        super().__init__(*args, **kwargs)
+
+    def persistent_load(self, pid: str) -> Optional[Bot]:
+        """Replaces the bot with the current bot if known, else it is replaced by :obj:`None`."""
+        return self._bot if pid == _REPLACED_KNOWN_BOT else None
```
Bibo-Joshi, line 130: Can you raise `UnpicklingError` if the pid matches neither of `_R_(UN)KNOWN_BOT`? just to be sure …

tests/test_bot.py
```diff
@@ -241,6 +242,10 @@ def test_to_dict(self, bot):
         if bot.last_name:
             assert to_dict_bot["last_name"] == bot.last_name
 
+    def test_bot_pickling_error(self, bot):
+        with pytest.raises(pickle.PicklingError, match="Bot objects cannot be pickled"):
+            pickle.dumps(bot)
+
```
Bibo-Joshi, line 248: *if* the `__deepcopy__` part stays, we'll need a test

tests/test_persistence.py
```diff
@@ -1633,6 +1362,108 @@ def test_save_on_flush_single_files(self, pickle_persistence, good_pickle_files)
...
+        )
+        with pytest.raises(pickle.UnpicklingError, match=err_msg):
+            with open('pickletest_chat_data', 'rb') as f:
+                pickle.load(f)
+
+        # Test that our custom unpickler works as intended -- inserts the current bot
+        # We have to create a new instance otherwise unpickling is skipped
+        pp = PicklePersistence("pickletest", single_file=False, on_flush=False)
+        pp.bot = bot  # Set the bot
+        assert pp.get_chat_data()[12345]['current_bot']._bot is bot
```
Bibo-Joshi, line 1385:
```suggestion
        assert pp.get_chat_data()[12345]['current_bot'].get_bot() is bot
```

telegram/_telegramobject.py
```diff
@@ -81,6 +86,110 @@ def __getitem__(self, item: str) -> object:
...
+                # __dict__ has attrs from superclasses, so no need to put in the for loop below
+                data.update(self.__dict__)
+            except AttributeError:
+                pass
+        # We want to get all attributes for the class, using self.__slots__ only includes the
+        # attributes used by that class itself, and not its superclass(es). Hence, we get its MRO
+        # and then get their attributes. The `[:-1]` slice excludes the `object` class
+        for cls in self.__class__.__mro__[:-1]:
+            for key in cls.__slots__:
+                if not include_private and (key.startswith('_') or key == "bot"):
```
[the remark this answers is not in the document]
harshil21: correct, there is one passport related class which was using it, i'll modify that then

telegram/_bot.py
harshil21: This only comes into effect when trying to directly deepcopy a `Bot`. So if this is removed, then for e.g. context.user_data['bot_inst'] = Bot(...)` [will fail to save](https://github.com/python-telegram-bot/python-telegram-bot/pull/2893/files#diff-0f06c3fbe6411bfe0c9836f2e4f317ba382b8cac774aa2d1a015b857063ab4d9R421)

If the bot is in a `TelegramObject`, then it is removed, deepcopied and re-inserted back in. And thus removing this won't be an issue.

So we can either document saying that bots won't actually be deepcopied, or we remove this method and say that directly serializing bots is not possible

telegram/_telegramobject.py
harshil21: problem is `get_bot()` may raise an exception, so not calling that to be safe

telegram/ext/_picklepersistence.py
harshil21: there's a subtle difference, `__setstate__` seems to construct a new instance internally and should not return anything (modify in place is only allowed). Here, we have to construct a new instance manually. Plus, `TO.__setstate__` assigns to `self`, here we need it to assign to our new instance (`obj`)

telegram/ext/_picklepersistence.py
harshil21: I don't think I even need to pop the bot in the first place, will see

telegram/_telegramobject.py
```diff
@@ -81,6 +86,110 @@ def __getitem__(self, item: str) -> object:
...
+                    continue
+
+                value = getattr(self, key, None)
+                if value is not None:
+                    if recursive and hasattr(value, 'to_dict'):
```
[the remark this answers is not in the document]
harshil21: Can't do this since `InputFile` doesn't inherit from `TGO` and has a `to_dict` method...

telegram/ext/_picklepersistence.py
```diff
@@ -17,23 +17,121 @@
...
+        self._bot = bot
+        super().__init__(*args, **kwargs)
+
+    def persistent_load(self, pid: str) -> Optional[Bot]:
+        """Replaces the bot with the current bot if known, else it is replaced by :obj:`None`."""
+        if pid == _REPLACED_KNOWN_BOT:
+            return self._bot
+        if pid == _REPLACED_UNKNOWN_BOT:
+            return None
+        raise pickle.UnpicklingError("Found unknown persistent id when unpickling!")
```
harshil21, line 132: codecov gonna complain about this since I don't think it's possible to test this...

telegram/_telegramobject.py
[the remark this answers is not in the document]
Bibo-Joshi: right. forgot about that

telegram/_bot.py
Bibo-Joshi: good, then let's remove. documenting something about this is maybe not strictly necessary - deepcopying bots directly is probably an even more rare use case than pickling them directly. BTW: I guess the fact thet `deepcopy` and `pickle` use the same interface should be considered more an implementation detail than intended behavior, i.e. I'd always try to make a clear distinction between serializing in general, picking as specific serialization method & copying via `deepcopy`.

telegram/_telegramobject.py
Bibo-Joshi: ah right, forgot about that

telegram/ext/_picklepersistence.py
Bibo-Joshi: what I have in mind is something like 

```python
obj = cls.__new__(cls)
obj.__setstate__(kwargs)
```

IISC that should work just fine

Bibo-Joshi (approved):
Finally get's a LGTM from me 😃 Thanks for the effort you put into this! Is there anything left from your side? Otherwise we can merge.

harshil21:
@Bibo-Joshi Nope, all done and am happy with it now

PR #2922, opened 2022-03-07
Expand and fix a filter

### LINKED ISSUE #2907: v14-Filters: CamelCase vs SCREAMING_SNAIL_CASE
After a bit internal discussion, we came up with the following naming rule:

* For simple cases, we have `filters.CamelCase` as filter class that needs to be initialized by the user and if applicable `filters.CAMEL_CASE` with a pre-instantiated shortcut
* For cases where we have some kind of "nested" logic, i.e. subfilters that should be carried in the same namespace, we do something like `filters.CamelCase.{ALL, SUP_TYPE_1, SUP_TYPE_2}`. In this case `filters.CamelCase` may even be just a pure namespace class, i.e. not a subclass of `BaseFilter`

Accordingly, `filters.DOCUMENT` should be moved to `filters.Document.ALL`.

Additionally, `filters.STICKER` should be replaced by `filters.Sticker.{ALL, ANIMATED, STATIC, VIDEO}` with `filters.Sticker` being a pure namespace class.

It may be worth squashing this into the first filters-refactor commit on the v14 branch …

harshil21:
Closes #2907 

### Checklist for PRs


telegram/ext/filters.py
```diff
@@ -1096,18 +1095,30 @@ def __init__(self, values: SLT[int]):
...
 
     Examples:
         Use these filters like: ``filters.Document.MP3``,
-        ``filters.Document.MimeType("text/plain")`` etc. Or use just ``filters.DOCUMENT`` for all
-        document messages.
+        ``filters.Document.MimeType("text/plain")`` etc. Or just use ``filters.Document.ALL`` for
+        all document messages.
+
+    Caution:
+        ``filters.DOCUMENT`` itself is *not* a filter.
```
Bibo-Joshi, line 1108:
```suggestion
        ``filters.Document`` itself is *not* a filter, but just a convenience namespace.
```

telegram/ext/filters.py
```diff
@@ -1871,15 +1875,57 @@ def filter(self, message: Message) -> bool:
...
-class _Sticker(MessageFilter):
+class Sticker:
+    """Filters messages which contain a sticker.
+
+    Examples:
+        Use this filter like: ``filters.Sticker.VIDEO``. Or, just use ``filters.Sticker.ALL`` for
+        any type of sticker.
+
+    Caution:
+        ``filters.STICKER`` itself is *not* a filter.
```
Bibo-Joshi, line 1886:
```suggestion
        ``filters.Sticker`` itself is *not* a filter, but just a convenience namespace.
```

telegram/ext/filters.py
```diff
@@ -1871,15 +1875,57 @@ def filter(self, message: Message) -> bool:
...
+
+    class _Animated(MessageFilter):
+        __slots__ = ()
+
+        def filter(self, message: Message) -> bool:
+            return bool(message.sticker) and bool(message.sticker.is_animated)  # type: ignore
+
+    ANIMATED = _Animated(name="filters.Sticker.ANIMATED")
+    """Messages that contain :attr:`telegram.Message.sticker` and is an animated sticker."""
+
```
Bibo-Joshi, line 1908: maybe explicitly reference `Sticker.is_animated` here? same below for video/static

Bibo-Joshi (approved):
LGTM. let's merge v14 once asyncio is in there to check the CI, but that should be it :)

PR #2731, opened 2021-10-14
Asyncio

Bibo-Joshi:
Supersedes #2402. Closes #2288 when ready - which is not in the forseable future :D 

* ext-tests are deleted for now so that we can first concentrate on getting `telegram` to work
* any changes to v14 should be merged into this ASAP

### TODO

Let's use this as place to write everything down that comes to mind while working on this - will be a lot

<details>
<p>

- [x] move `PtbRequestBase` and `HttpxRequest` to a `tg.request` module/package. Maybe rename to `RequestBase` and `HTTPXRequest`
- [x] Improve handling of files in request - probably move that logic into a standalone method or even into `Bot`?
- [x] Update error code handling in request, use the enum from the std-lib
- [x] Check if we get `coroutine was never awaited` somewhere in the tests and also check if we can make those warnings fail the tests
- [x] check back with noam on whether the `timeout` parameter should be used for `write_timeout` as well in in `BaseRequest`
- [x] Check back with Noam about httpx connection pool limits: IISC on `pool_timeout` we want to create a new connection instead of dropping the request?!
- [x] Think about how we handle non-async callbacks for handlers and jobs. We might want to make sure that they run in a background thread - which is the default behavior of APscheduler

</p>
</details>

- [x] Think about what to do with exceptions in the long running tasks of updater. Currently we put them in the `update_queue`, which isn't really nice
- [x] persistence
- [ ] think about a replacement for `exception_event`
- [ ] a _lot_ of tests
  - [ ] For `JobQueue` test that weekdays are working as expected to avoid regression when we update APS - see [here](https://t.me/pythontelegrambotgroup/570673) and [here](https://github.com/agronholm/apscheduler/issues/465)
- [ ] check for added `# TODO` comments
- [x] update examples. for `chatmemberbot.py` we can also include the changes proposed [here](https://t.me/pythontelegrambotgroup/566931)

### Things that don't need to be done in this PR

If they are not, create issues for them!

- [ ] set socketops
- [ ] switch to a different library for webhooks - maybe starlette (backend of fastAPI)?
- [ ] Add transaction support to `BaspePersistence`, i.e. abstract methods `start/end_transaction`  so that we can do
 1. tell persistence that we're going to start transaction
 2. hand over data
 3. tell persistence that I'm finished

    

### Review

Thinks to look out for in a review

- [ ] double check if we need to update docs regarding accepted types
- [x] Check for unwanted changes (also in tests!) - these are likely to be there because merging v14 was a mess
[...]

PR #2925, opened 2022-03-08
Update Test Dependencies versions

harshil21:
Updates to our test suite:
black 22.3.0: out of beta
pytest 7.1.2: major release
mypy 0.950: fixes and improvements
pyupgrade 2.31.0: drops py 3.6 support + bug fixes (can't find a changelog)
pylint [2.13.7](https://github.com/PyCQA/pylint/releases/tag/v2.13.7): major release

Note about mypy:
- There are many ignores in Enums because of https://github.com/python/mypy/issues/12132 . Edit: Resolved in mypy [0.942](https://mypy-lang.blogspot.com/2022/03/mypy-0942-released.html).

Note about `pytest.approx`:
- datetimes are not supported by approx (https://github.com/pytest-dev/pytest/issues/8395), so we should not use them.

TODO:
~~Update `furo` dependency.[ Lots of new features added](https://github.com/pradyunsg/furo/blob/main/docs/changelog.md). Including edit on github button.~~ 
Completed in #2969 
- [x] Add `--enable=useless-suppression` to pylint args. This will warn about [unused pylint ignore comments](https://docs.pylint.org/en/v2.12.2/user_guide/message-control.html?highlight=useless-suppressin#detecting-useless-disables)
- [x] Update pylint to new version [2.13.7](https://github.com/PyCQA/pylint/releases/tag/v2.13.7)
- [x] Update mypy again to resolve above issue
- [x] In c3eda6f / #2937 we added the hard pinned click dependency for black in requirements-dev.txt and .pre-commit-config.yaml, revert this here when upgrading black to => 22.3.0
- [x] update `pytest-asyncio` and use the new functionality of auto-detecting `async` tests to remove the `@pytest.mark.asyncio`. Since some of the monkeypatched tests block with that, they'll have to be adjusted - see [@a3f15f9](https://github.com/python-telegram-bot/python-telegram-bot/pull/2731/commits/a3f15f9921d7c2bc4b4681b6ce149eb04a76a817)
- [x] update sphinx to [`4.5.0`](https://www.sphinx-doc.org/en/master/changes.html#release-4-5-0-released-mar-28-2022)
- [x] migrate to pre-commit.ci for pre-commit tests
- [x] maybe add `isort` to the pre-commit setup


telegram/constants.py
```diff
@@ -230,7 +230,7 @@ class ChatType(StringEnum):
     .. versionadded:: 14.0
     """
 
-    __slots__ = ()
+    __slots__ = ()  # type: ignore[misc]
```
Bibo-Joshi: just curious: what's mypy complaining about here?

tests/test_chatmemberupdated.py
```diff
@@ -127,7 +127,7 @@ def test_de_json_all_args(
 
         assert chat_member_updated.chat == chat
         assert chat_member_updated.from_user == user
-        assert pytest.approx(chat_member_updated.date == time)
+        assert chat_member_updated.date - time < datetime.timedelta(seconds=1)
```
Bibo-Joshi, line 131: should we use `abs` for these cases?

.pre-commit-config.yaml
```diff
@@ -16,14 +16,18 @@ repos:
...
     -   id: pylint
         files: ^(telegram|examples)/.*\.py$
         args:
           - --rcfile=setup.cfg
           #  run pylint across multiple cpu cores to speed it up-
-          - --jobs=0  # See https://pylint.pycqa.org/en/latest/user_guide/run.html?#parallel-execution to know more
+          # https://pylint.pycqa.org/en/latest/user_guide/run.html?#parallel-execution to know more
+          - --jobs=0
+          # Warns about unused pylint ignores:
+          - --enable=useless-suppression
```
Bibo-Joshi, line 27: if possible, please directly add this to `setup.cfg` so that it's also active when running pylint manually

telegram/constants.py
```diff
@@ -108,7 +108,7 @@ class CallbackQueryLimit(IntEnum):
     .. versionadded:: 14.0
     """
 
-    __slots__ = ()  # type: ignore[misc]
+    __slots__ = ()
```
Bibo-Joshi, line 111: okay, seems to have been fixed in later versions xD

telegram/ext/_application.py
```diff
@@ -466,7 +466,7 @@ async def start(self) -> None:
                 _logger.debug('Loop for updating persistence started')
 
             if self.job_queue:
-                await self.job_queue.start()
+                await self.job_queue.start()  # type: ignore[union-attr]
```
Bibo-Joshi, line 472: can we get rid of the comment by changing the above line to `if self.job_queue is not None`? same below

.pre-commit-config.yaml
```diff
@@ -67,3 +67,10 @@ repos:
         files: ^(telegram|examples|tests)/.*\.py$
         args:
           - --py37-plus
+-   repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+    -   id: isort
+        name: isort (python)
+        # compatibility with black
+        args: ["--profile", "black", "--filter-files", "-l", "-99"]
```
Bibo-Joshi, line 76: what does `"--filter-files", "-l", "-99"` do?

why is it necessary to set the profile again here instead of just in the pyproject.toml?

pyproject.toml
```diff
@@ -8,3 +8,7 @@ skip-string-normalization = true
 # see https://github.com/psf/black/issues/1778
 force-exclude = '^(?!/(telegram|examples|tests)/).*\.py$'
 include = '(telegram|examples|tests)/.*\.py$'
+
+[tool.isort]  # make sure that the config values are the same in pre-commit-config
+profile = "black"
+line_length = 99
```
Bibo-Joshi, line 14: is that still necessary even if we tell black to use 99 chars already 🤔 I don't recall that being the case in https://github.com/Bibo-Joshi/led-strip-controller where I first used isort …

requirements-dev.txt
```diff
@@ -8,10 +8,11 @@ flake8==4.0.1
...
 
 pytest==7.0.1
 pytest-asyncio==0.16.0
 
-flaky
-beautifulsoup4
-wheel
+flaky  # Used for flaky tests (flaky decorator)
+beautifulsoup4  # used in test_official for parsing tg docs
+wheel  # TODO ??
```
Bibo-Joshi, line 18: required for building the wheels for releases with `python setup.py bdist_wheel`, IIRC

pyproject.toml
harshil21: Nice _plug_. [This](https://pycqa.github.io/isort/docs/configuration/profiles.html#black) page told me that it takes 88 chars for `black` config. Not sure if it fetches our set line length from the black config or it needs to explicitly told.. will check

harshil21: okay it needs to be told the line length otherwise it uses 88

.pre-commit-config.yaml
Bibo-Joshi: > [This](https://pycqa.github.io/isort/docs/configuration/options.html#filter-files) is what filter-files does.

I found that as well, but what does "filtering files" mean? 🤔 

> `-l` sets the line length. I'll test without this and see if it uses pyproject.toml

ah, it's ell, not one 🤦‍♂️

harshil21: > I found that as well, but what does "filtering files" mean? 🤔

tbh I don't know either, removing

telegram/ext/_application.py
harshil21: didn't fix it.. here's the error: `error: Item "object" of the upper bound "Optional[JobQueue]" of type variable "JQ" has no attribute "start"`

pyproject.toml
Bibo-Joshi: shame … but okay :)

harshil21:
I had to add a disable `import-error`  for pylint in `setup.cfg` since for some reason, pylint complains about that in all of the examples when running via git hook. But when I run using `pre-commit run -a`, & `pylint examples/` pylint does not having any trouble importing...

tests/conftest.py
```diff
@@ -95,6 +89,11 @@ def env_var_2_bool(env_var: object) -> bool:
 # session. See https://github.com/pytest-dev/pytest-asyncio/issues/68 for more details.
 @pytest.fixture(scope='session')
 def event_loop(request):
+    # ever since ProactorEventLoop became the default in Win 3.8+, the app crashes after the loop
+    # is closed. Hence, we use SelectorEventLoop on Windows to avoid this. See
+    # https://github.com/python/cpython/issues/83413, https://github.com/encode/httpx/issues/914
+    if sys.version_info[0] == 3 and sys.version_info[1] >= 8 and sys.platform.startswith('win'):
+        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
Bibo-Joshi, line 96: Can you check if this is still necessary even after changing the bogous `monkeypatch`?

harshil21: you mean the `call_after` function? I'm not sure of what the relation/issue was with monkeypatching, was it due to the event loop being closed?

Bibo-Joshi: ah sorry, I just saw that you change a `monkeypatch` in one of the commits but seems to have been reverted. The issue I had before coming up with `call_after` was that I was patching some functions that were required do run for proper init/shutdown. It's possible using a different loop policy resolves that, but I'm unsure whether this is treating cause or just the symptoms.
At first glance, I think would prefer to fiddle with the loop as little as possible, both to keep the test suite maintainable and to test on the default setting on windows. If using `call_after` wherever suitable does not solve the issue, I'm of course fine with doing this.

Bibo-Joshi: > yeah, unfortunately `call_after` seems to only treat the symptoms. This error would happen on many tests, irrespective of the monkeypatches.

that's what I'm unsure about, as 

* IIRC `call_after` had fixed the issues for me when I tried upgrading `pytest-asyncio` without changing any tests outside of `test_application`
* This mwe https://github.com/encode/httpx/issues/914#issuecomment-1050772624 works fine for me in the default setting with the `WindowsProactorEventLoopPolicy`

I can try to have a closer look tonight

> Since this is a bug in cpython, maybe it would be beneficial to switch the event loop policy on windows to the selectoreventloop in our actual code? `httpx` might raise a warning in future versions too ([encode/httpx#914 (comment)](https://github.com/encode/httpx/issues/914#issuecomment-1051004818) ).

We can keep that in mind as an option, but I think trying something like that is a bit premature before getting some user feedback on v20.

> Another question is why this only showed up after upgrading `pytest-asyncio`, is because after >0.17.0, they no longer alter the event loop policy (https://github.com/pytest-dev/pytest-asyncio/blob/master/CHANGELOG.rst#0170-22-01-13). So in a way, the default setting on windows wasn't really tested until now :)

Ah, I didn't know that. good finding 👍

harshil21: > This mwe works fine for me in the default setting with the WindowsProactorEventLoopPolicy

hmm, interesting all of them seem to work for me too. However this causes the error:
``` python
from telegram import Bot
import asyncio

async def bot_initalize():
    bot = Bot("TOKEN")
    await bot.initialize()

asyncio.run(bot_initalize())
```
adding a `await bot.shutdown()` fixes it.. 

Hmm but still since there's no proper traceback for this it's hard to know exactly how many tests are causing this error. It's hard to know how and when exactly this is happening,,

Bibo-Joshi: > Hmm but still since there's no proper traceback for this it's hard to know exactly how many tests are causing this error, and when exactly this is happening,,

I ended up doing a bisection in terms of commenting out tests to see which one cause that behavior.

harshil21: > I ended up doing a bisection in terms of commenting out tests to see which one cause that behavior.

ha I had done the same thing

> jup, that's exactly what I used `call_after` for: make sure that `bot.shutdown` is actually called instead of just overriding it

but that is treating the symptom isn't it? I say to leave the event loop policy mod here, anyway tests run fine on win 3.7.. so

Bibo-Joshi: > but that is treating the symptom isn't it?

I tend to disagree. since calling `bot.shutdown` fixes the issue, this sounds to me like not properly closing the `httpx.AsyncClient` might the be issue. different event loops handling that better or worse is another thing, but ultimately it's better to do a clean shutdown

Bibo-Joshi: I did some digging and the problem seems to be indeed more deeply rooted than just some missing `bot.shutdown` calls. My observations/tries:

* the behavior is not deterministic. the tests sometimes fail, sometime they don't
* removing our workaround that makes the `event_loop` session scoped (thus using a fresh event loop for every test) didn't help
* I have a suspicion that there is an issue with `pytest.mark.parametrize` - at least there was one parametrized test that failed more often than when I removed the parametrization
* I inserted some prints into `HTTPXReq.init/shutdown`. on parametrized tests, the prints where sometimes mixed up, e.g. `init - init - shutdown - shutdown` instead of `init-shutdown-init-shutdown`. Also sometimes the `shutdown` prints where _after_ the `Event loop closed` error

I'm okay with keeping the workaround as is, but I'd vote to at least open an issue that gathers some details, links to the cpython issue etc

harshil21: > * the behavior is not deterministic. the tests sometimes fail, sometime they don't

I noticed that if you run just one test, that test won't fail, but you'll get the warning in the logs. However if you run two or more tests (where one or more causes the error), then it causes the last test to fail. But yes, it is hard to find the root cause, because if bot.shutdown() had some code which causes a problem, then running `test_request` should fail but it doesn't?

> * I have a suspicion that there is an issue with `pytest.mark.parametrize` - at least there was one parametrized test that failed more often than when I removed the parametrization

mostly related to what I said above (running two or more tests will fail the test suite)

> I'm okay with keeping the workaround as is, but I'd vote to at least open an issue that gathers some details, links to the cpython issue etc

Yeah, there is a small chance users can bump into this issue, specially since we still can't find the real reason why.

PR #2926, opened 2022-03-09
Docs/CSI for asyncio

harshil21:
Gathering doc fixes for asyncio

### Checklist for PRs


telegram/ext/_applicationbuilder.py
```diff
@@ -425,31 +423,95 @@ def request(self: BuilderType, request: BaseRequest) -> BuilderType:
...
+            :class:`ApplicationBuilder`: The same builder with the updated argument.
+        """
         self._request_param_check(name='write_timeout', get_updates=False)
         self._write_timeout = write_timeout
         return self
 
     def pool_timeout(self: BuilderType, pool_timeout: Optional[float]) -> BuilderType:
+        """Sets the connection pool's connection freeing timeout to be used for the
+        :paramref:`~telegram.request.HTTPXRequest.pool_timeout` parameter of
+        :attr:`telegram.Bot.request`. Will default to :obj:`None`.
```
Bibo-Joshi, line 506:
```suggestion
        :attr:`telegram.Bot.request`. Defaults to :obj:`None`.
```
?

telegram/ext/_application.py
```diff
@@ -113,12 +113,31 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ]):
...
+            :class:`Application` for the ``context`` argument of handler and job callbacks.
+
     Attributes:
         bot (:class:`telegram.Bot`): The bot object that should be passed to the handlers.
         update_queue (:class:`asyncio.Queue`): The synchronized queue that will contain the
             updates.
-        updater (:class:`telegram.ext.Updater`, optional): The updater used by this application.
-        job_queue (:class:`telegram.ext.JobQueue`): Optional. The :class:`telegram.ext.JobQueue`
+        updater (:class:`telegram.ext.Updater`): The updater used by this application.
+        job_queue (:class:`telegram.ext.JobQueue`): The :class:`telegram.ext.JobQueue`
```
Bibo-Joshi, line 140: both maye be `None`.

telegram/ext/_application.py
```diff
@@ -113,12 +113,31 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ]):
         * Initialization is now done through the :class:`telegram.ext.ApplicationBuilder`.
         * Removed the attribute ``groups``.
 
+    Args:
```
Bibo-Joshi, line 116: mh, why did you document the arguments? this class should not be instantiated manually but via the `ApplicationBuilder`

telegram/ext/_applicationbuilder.py
```diff
@@ -478,39 +546,104 @@ def get_updates_request(self: BuilderType, get_updates_request: BaseRequest) ->
     def get_updates_connection_pool_size(
         self: BuilderType, get_updates_connection_pool_size: int
     ) -> BuilderType:
+        """Sets the size of the connection pool to be used for the
+        :paramref:`~telegram.request.HTTPXRequest.connection_pool_size` parameter of
+        :attr:`telegram.Bot.request` for :meth:`telegram.Bot.get_updates`. Defaults to ``1``.
```
Bibo-Joshi, line 551: `Bot.request` is the BaseRequest instance for non-get_updates requests … Currently the get_updates instance is not exposed (IISC `Bot.request` is currently only really there because `File.download` accesses it).

same below

telegram/ext/_application.py
harshil21: oh, felt weird not to have an Args section. Maybe link to the ApplicationBuilder in the Args section? That way it becomes a bit more clear to the users

Bibo-Joshi: you mean like [this](https://github.com/python-telegram-bot/python-telegram-bot/blob/1bbe39df051bafd1b181262554b2ec6b3dafe733/telegram/ext/_application.py#L107-L114)? ;P

harshil21: yeah like move that Note to Args section

Bibo-Joshi: ah. mh, I don't see a big difference between it being a note and it bein in the `args` section … but I'm also not agains it. you can try and see what's possible here - not sure how well that plays with sphinx's layout …

telegram/ext/_application.py
```diff
@@ -738,33 +740,31 @@ async def __process_update_wrapper(self, update: object) -> None:
...
 
         for handlers in self.handlers.values():
             try:
                 for handler in handlers:
-                    check = handler.check_update(update)
-                    if not (check is None or check is False):
-                        if not context:
+                    check = handler.check_update(update)  # Should the handler handle this update?
+                    if not (check is None or check is False):  # if yes,
+                        if not context:  # build a context if not already built
```
Bibo-Joshi, line 762: IMO this is a bit overly verbose/close to stating the obvious … just my 2cts - you can keep it if you like.

harshil21: I know it's obvious to us, but not sure if it is obvious to someone new / unfamiliar to PTB

telegram/ext/_application.py
```diff
@@ -665,7 +747,8 @@ def __create_task(
         return task
 
     def __create_task_done_callback(self, task: asyncio.Task) -> None:
-        self.__create_task_tasks.discard(task)
+        """Used for handling asyncio exceptions. Unretrieved exceptions will be raised on exit"""
```
Bibo-Joshi, line 750: this docstring doesn't make much sense imo. let's just keep the comment

telegram/ext/_application.py
```diff
@@ -1223,7 +1307,7 @@ async def dispatch_error(
                     job=job,
                     coroutine=coroutine,
                 )
-                if not block or (
+                if not block or (  # If error handler has `block=False`, create a Task to run cb
```
Bibo-Joshi, line 1310: cb = ?

harshil21: callback. cb is used all over updater too

Bibo-Joshi: ok, makes sense

telegram/ext/_application.py
```diff
@@ -146,7 +146,7 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ]):
 
             .. seealso::
                 :meth:`add_handler`, :meth:`add_handlers`.
-        error_handlers (Dict[:obj:`callable`, :obj:`bool`]): A dict, where the keys are error
+        error_handlers (Dict[:term:`coroutine`, :obj:`bool`]): A dict, where the keys are error
```
Bibo-Joshi, line 149: shouldn't it be `coroutine function`? [this](https://docs.python.org/3/glossary.html#term-coroutine) vs [this](https://docs.python.org/3/glossary.html#term-coroutine-function)

telegram/ext/_basepersistence.py
```diff
@@ -146,7 +146,7 @@ def __init__(
 
     @property
     def update_interval(self) -> float:
-        """:obj:`int`, optional: Time (in seconds) that the :class:`~telegram.ext.Application`
+        """:obj:`int`: Time (in seconds) that the :class:`~telegram.ext.Application`
```
Bibo-Joshi, line 149:
```suggestion
        """:obj:`float`: Time (in seconds) that the :class:`~telegram.ext.Application`
```

telegram/ext/_callbackqueryhandler.py
```diff
@@ -64,18 +65,20 @@ class CallbackQueryHandler(Handler[Update, CCT]):
...
+        callback (:term:`coroutine`): The callback function for this handler. Will be called when
+            :meth:`check_update` has determined that an update should be processed by this handler.
+            Callback signature::
+
+                async def callback(update: Update, context: CallbackContext)
 
             The return value of the callback is usually ignored except for the special case of
             :class:`telegram.ext.ConversationHandler`.
-        pattern (:obj:`str` | `Pattern` | :obj:`callable` | :obj:`type`, optional):
+        pattern (:obj:`str` | `Pattern` | :term:`coroutine` | :obj:`type`, optional):
```
Bibo-Joshi, line 76: no, the pattern-callable should not be async

telegram/ext/_callbackqueryhandler.py
```diff
@@ -93,8 +96,8 @@ class CallbackQueryHandler(Handler[Update, CCT]):
             :meth:`telegram.ext.Application.process_update`. Defaults to :obj:`True`.
 
     Attributes:
-        callback (:obj:`callable`): The callback function for this handler.
-        pattern (`Pattern` | :obj:`callable` | :obj:`type`): Optional. Regex pattern, callback or
+        callback (:term:`coroutine`): The callback function for this handler.
+        pattern (`Pattern` | :term:`coroutine` | :obj:`type`): Optional. Regex pattern, callback or
```
Bibo-Joshi, line 100: .

telegram/ext/_updater.py
```diff
@@ -266,7 +265,7 @@ async def _start_polling(
 
         self._logger.debug('Updater started (polling)')
 
-        await self._bootstrap(
+        await self._bootstrap(  # Makes sure no webhook is set by calling delete_webhook
```
Bibo-Joshi, line 268:
```suggestion
        # the bootstrapping phase does two things:
        # 1) make sure there is no webhook set
        # 2) apply drop_pending_updates
        await self._bootstrap(
```

telegram/ext/_updater.py
```diff
@@ -568,6 +568,11 @@ async def _bootstrap(
         ip_address: str = None,
         max_connections: int = 40,
     ) -> None:
+        """Entry point for handling webhooks. :meth:`start_polling` calls this to delete any
+        present webhook. :meth:`start_webhook` calls this to set a webhook using
+        :meth:`telegram.Bot.set_webhook. If there are unsuccessful attempts, it will be retried as
+        specified by :paramref:`max_retries`.
+        """
```
Bibo-Joshi, line 575:
```suggestion
        """Prepares the setup for fetching updates: delete or set the webhook and drop pending updates if appropriate. If there are unsuccessful attempts, this will retry as
        specified by :paramref:`max_retries`.
        """
```

telegram/ext/_callbackdatacache.py
```diff
@@ -248,7 +248,7 @@ def process_message(self, message: Message) -> None:
 
         Note:
             Checks :attr:`telegram.Message.via_bot` and :attr:`telegram.Message.from_user` to check
-            if the reply markup (if any) was actually sent by this caches bot. If it was not, the
+            if the reply markup (if any) was actually sent by this bot cache. If it was not, the
```
Bibo-Joshi, line 251: caches can't send messages - bots can ;)

telegram/ext/_conversationhandler.py
```diff
@@ -304,7 +320,7 @@ def __init__(
         self._name = name
         self._map_to_parent = map_to_parent
 
-        self.timeout_jobs: Dict[Tuple[int, ...], 'Job'] = {}
+        self.timeout_jobs: Dict[Tuple[int, ...], 'Job'] = {}  # TODO: figure out purpose of this
```
Bibo-Joshi, line 323: it keeps track of the jobs that are scheduled for conversation timeouts so we can abort them if a new update comes in in time

telegram/ext/_conversationhandler.py
```diff
@@ -542,6 +557,8 @@ async def _initialize_persistence(
                 'persistence!'
             )
 
+        # Here we will fill the self._conversations variable with conversations from persistence,
+        # and this includes child conversations.
```
Bibo-Joshi, line 561: maybe just include "this handler and it's child conversations" in the docstring?

telegram/ext/_conversationhandler.py
```diff
@@ -564,6 +581,7 @@ async def _initialize_persistence(
         return self._conversations
 
     def _get_key(self, update: Update) -> Tuple[int, ...]:
+        """Gets the chat/user/(inline)message id of the chat this update is associated with."""
```
Bibo-Joshi, line 584: not quite. the conversation key is a combination of those. maybe just "Build the conversation key associated with the update"?

telegram/ext/_conversationhandler.py
harshil21: okay, so as I understand it, whenever a new update comes in, the previous job is cancelled and removed, and the new one is added in this dict. So shouldn't the max size be 1? Meaning it doesn't have to be a dict?

Bibo-Joshi: well, for each conversation key (e.g. for each user), there is a different job

telegram/ext/_application.py
```diff
@@ -1316,6 +1316,8 @@ async def process_error(
             job (:class:`telegram.ext.Job`, optional): The job that caused the error.
 
                 .. versionadded:: 14.0
+            coroutine (:term:`coroutine function`): The coroutine that caused the error. Only
+                present if the coroutine was run with :attr:`block=False <Handler.block>`.
```
Bibo-Joshi, line 1320: * the argument is optional.
* "only present" doesn't really make sense for an argument. this should instead go into the docs of `CallbackContext.coroutine`. suggestion for the text: `Only present in error handlers if the error was caused by a coroutine run with App.create_task or a handler callback with block=False` (the fact that block=false uses app.create_task doesn't have to be documented here IMO)

telegram/ext/_conversationhandler.py
```diff
@@ -781,7 +785,7 @@ async def handle_update(  # type: ignore[override]
             else:
                 block = DefaultValue.get_value(handler.block)
 
-        try:
+        try:  # Now determine if the handler in the current state should be handled asynchronously
```
Bibo-Joshi, line 788: no, we already determined that above. we now just act depending on the result ;)

PR #2946, opened 2022-04-09
hashtags fixed in markdownV2

srezasm:
# Fixed hashtags for markdownV2 format.
Apparently, markdownV2 doesn't handle hashtags with the preceding '\\' character. Through a chat with one of the telegram assistants I figured that the problem will be fixed by replacing the '#' with '\%23\'. each '\\' is just a backslash and '%23' represents the '#' character in HTML format.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -99,6 +99,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Sascha <https://github.com/saschalalala>`_
 - `Shelomentsev D <https://github.com/shelomentsevd>`_
 - `Simon Schürrle <https://github.com/SitiSchu>`_
+- `SReza SMohseni <https://github.com/srezasm>`_
 - `sooyhwang <https://github.com/sooyhwang>`_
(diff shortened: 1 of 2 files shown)


harshil21:
> Apparently, markdownV2 doesn't handle hashtags with the preceding '\' character.

but it does? `\` is used to escape the `#` symbol, I just tested it and it works fine. Also note that `\` is used as an escape character in strings in python, so you need to make your string raw by prepending `r` e.g.: `r"escaping \#` .

srezasm:
> but it does? `\` is used to escape the `#` symbol, I just tested it and it works fine. Also note that `\` is used as an escape character in strings in python, so you need to make your string raw by prepending `r` e.g.: `r"escaping \#` .

the question about the mentioned problem: https://t.me/BotTalk/637111
the answer: https://t.me/BotTalk/637192

harshil21:
The answer also points out the [reason](https://t.me/BotTalk/637200). Since html url encoding is only needed for raw requests you must use that 'escaping' you pointed out. However the dependency of our library already takes care of this so this isn't required.

PR #2948, opened 2022-04-10
to_true shortcut for ChatPermissions

Poolitzer:
### Checklist for PRs

* New classes:
    
* Added new shortcuts:
    
* If relevant:

This was a spontaneous idea and quickly done by me to avoid other work. No proper issue to close here.

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -4802,7 +4802,8 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         Use this method to restrict a user in a supergroup. The bot must be an administrator in
         the supergroup for this to work and must have the appropriate admin rights. Pass
         :obj:`True` for all boolean parameters in :class:`telegram.ChatPermissions` to lift
-        restrictions from a user.
+        restrictions from a user, see
+        :meth:`telegram.ChatPermissions.all_true` for a shortcut.
 
         Args:
             chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target chat or username
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -4832,11 +4832,11 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
             api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
                 Telegram API.
 
-            Returns:
-                :obj:`bool`: On success, :obj:`True` is returned.
+        Returns:
+            :obj:`bool`: On success, :obj:`True` is returned.
 
-            Raises:
-                :class:`telegram.error.TelegramError`
+        Raises:
+            :class:`telegram.error.TelegramError`
         """
         data: JSONDict = {
             'chat_id': chat_id,
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -4801,9 +4801,9 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         """
         Use this method to restrict a user in a supergroup. The bot must be an administrator in
         the supergroup for this to work and must have the appropriate admin rights. Pass
-        :obj:`True` for all boolean parameters in :class:`telegram.ChatPermissions` to lift
-        restrictions from a user, see
-        :meth:`telegram.ChatPermissions.all_true` for a shortcut.
+        :obj:`True` for all boolean parameters to lift restrictions from a user.
+
+        .. seealso:: :meth:`telegram.ChatPermissions.all_permissions`
 
         Args:
             chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target chat or username
@@ -4832,11 +4832,11 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
             api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
                 Telegram API.
 
-        Returns:
-            :obj:`bool`: On success, :obj:`True` is returned.
+            Returns:
+                :obj:`bool`: On success, :obj:`True` is returned.
 
-        Raises:
-            :class:`telegram.error.TelegramError`
+            Raises:
+                :class:`telegram.error.TelegramError`
         """
         data: JSONDict = {
             'chat_id': chat_id,
--- a/telegram/_chatpermissions.py
+++ b/telegram/_chatpermissions.py
@@ -128,6 +128,8 @@ class ChatPermissions(TelegramObject):
         This method returns the ChatPermissions object with all attributes set to :obj:`bool`.
         This is e.g. useful when unrestricting a ChatMember with
         :meth:`telegram.Bot.restrict_chat_member`.
+
+        .. versionadded:: 14.0
         """
         # we generate the object so we can set the attributes to True dynamically in case
         # there are more added later.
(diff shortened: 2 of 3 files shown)


tests/test_chatpermissions.py
```diff
@@ -124,3 +124,13 @@ def test_equality(self):
 
         assert a != d
         assert hash(a) != hash(d)
+
+    def test_all_true(self):
+        f = ChatPermissions()
+        t = ChatPermissions.all_true()
+        # if the dirs are the same, the attributes will all be there
+        assert dir(f) == dir(t)
```
harshil21, line 132: maybe also add a `assert f != t`?

Poolitzer: sure, doesn't hurt.

Poolitzer: ah no, that gives me an assertionerror. So that == apparently doesn't care if the attributes Value is different.

harshil21: so in `all_true` we either need also add logic to dynamically update `self._id_attrs`, or just abandon dynamic approach and hardcode the `True` values and make it a `classmethod`

Poolitzer: okay so I think also dynamically updating the id sucks design wise. So I would go with hardcode and classmethod.

telegram/_chatpermissions.py
```diff
@@ -121,3 +121,14 @@ def __init__(
             self.can_invite_users,
             self.can_pin_messages,
         )
+
+    @classmethod
+    def all_true(cls) -> 'ChatPermissions':
```
harshil21, line 126: so do we need a `all_false` method too? Also I did prefer the clarity of `unrestrict` better. More suggestions:

`allow_all`/`disallow_all`
`all_permissions`/`no_permissions`

Poolitzer: > so do we need a all_false method too?

That is currently the assumed default by telegram, but sure, we can add that.

> Also I did prefer the clarity of unrestrict better

I don't like it because it is implicit design. If Telegram ever changes the behaviour (doubtfully, still), we would have a legacy method name dragging around which will make no sense if used in a different project.

> more suggestions

If you'd really like to rename the method, I like `all_permissions`/`no_permissions` best.

harshil21: >  I like all_permissions/no_permissions best.

let's go for this then

tests/test_chatpermissions.py
```diff
@@ -124,3 +124,15 @@ def test_equality(self):
...
+        # now we just need to check that all attributes are True. _id_attrs returns all values,
+        # if a new one is added without defaulting to True, this will fail
+        for key in t._id_attrs:
```
harshil21, line 135: imo we should test with `.__slots__` since we won't update `_id_attrs` when a new attribute is added

Poolitzer: Ah, backwards compatibility? Or whats the reason for not updating.

harshil21: yep backwards compatibility

Poolitzer: yay backwards compatibility

telegram/_bot.py
```diff
@@ -4351,7 +4351,8 @@ async def restrict_chat_member(
         """
         Use this method to restrict a user in a supergroup. The bot must be an administrator in
         the supergroup for this to work and must have the appropriate admin rights. Pass
-        :obj:`True` for all boolean parameters to lift restrictions from a user.
+        :obj:`True` for all boolean parameters to lift restrictions from a user, see
+        :meth:`telegram.ChatPermissions.all_true` for a shortcut.
```
Bibo-Joshi, line 4355:
```suggestion
        :obj:`True` for all boolean parameters to lift restrictions from a user.
        
    .. seealso:: :meth:`telegram.ChatPermissions.all_true`
```

telegram/_chatpermissions.py
```diff
@@ -121,3 +121,23 @@ def __init__(
             self.can_invite_users,
             self.can_pin_messages,
         )
+
+    @classmethod
+    def all_permissions(cls) -> 'ChatPermissions':
+        """
+        This method returns the ChatPermissions object with all attributes set to :obj:`True`.
```
Bibo-Joshi, line 128:
```suggestion
        This method returns an object with all attributes set to :obj:`True`.
```

telegram/_chatpermissions.py
```diff
@@ -121,3 +121,23 @@ def __init__(
...
+        This is e.g. useful when unrestricting a ChatMember with
```
Bibo-Joshi, line 129: either "chat member" or link to the class

telegram/_chatpermissions.py
```diff
@@ -121,3 +121,23 @@ def __init__(
...
+        :meth:`telegram.Bot.restrict_chat_member`.
+
+        .. versionadded:: 14.0
+        """
+        return cls(True, True, True, True, True, True, True, True)
+
+    @classmethod
+    def no_permissions(cls) -> 'ChatPermissions':
+        """
+        This method returns the ChatPermissions object with all attributes set to :obj:`False`.
```
Bibo-Joshi, line 139:
```suggestion
        This method returns an object with all attributes set to :obj:`False`.
```

Bibo-Joshi (commented):
just had a quick glance and let some tipicking. Still, I'd like to come back to this only after we've merged asyncio ;)

telegram/_chatpermissions.py
Poolitzer: why an undefined an? This is the object where the attribute belongs to... If you dont like repeating ChatPermission, I would vote for either "returns the object" or "returns this object".

telegram/_chatpermissions.py
Poolitzer: I like the second one.

telegram/_chatpermissions.py
Bibo-Joshi: IMO it's obvious enough that it would return an instance of the class this method is a classmethod of … I'd be okay with ``a :class:`ChatPermisisons` object``

Poolitzer: so actually, object is not good, since everything in python is an object. But instance sounds great, will do that.

telegram/_chatpermissions.py
```diff
@@ -139,6 +139,6 @@ def no_permissions(cls) -> 'ChatPermissions':
         This method returns an :class:`ChatPermisisons` instance
         with all attributes set to :obj:`False`.
 
-        .. versionadded:: 14.0
+        .. versionadded:: 2.0
```
harshil21, line 142:
``` suggestion
        .. versionadded:: 20.0
```

telegram/_bot.py
```diff
@@ -4831,11 +4832,11 @@ async def restrict_chat_member(
...
 
-        Returns:
-            :obj:`bool`: On success, :obj:`True` is returned.
+            Returns:
+                :obj:`bool`: On success, :obj:`True` is returned.
 
-        Raises:
-            :class:`telegram.error.TelegramError`
+            Raises:
+                :class:`telegram.error.TelegramError`
```
harshil21, line 4839: does that render correctly?

Poolitzer: yes, but its still wrong.

Bibo-Joshi (approved):
LGTM. @harshil21 do you have any comments left? If not, you can go ahead and merge IMO

PR #2956, opened 2022-04-17
API 6.0

harshil21:
html diff in the api page:

https://t.me/c/1161850314/3674
invite link to the chat: https://t.me/joinchat/AAAAAEVAbcr5AkmFd_dAIw
Also adds `de_json` to `(Inline)KeyboardButton`, `WebhookInfo`

### Checklist for PRs

- [x] Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)
- [x] Created new or adapted existing unit tests
- [ ] Added myself alphabetically to `AUTHORS.rst` (optional)
- [x] Added new classes & modules to the docs

### If the PR contains API changes (otherwise, you can delete this passage)

* New classes:
    - [x] Added `self._id_attrs` and corresponding documentation
    - [x] `__init__` accepts `**_kwargs`
    
* Added new shortcuts:
    - [x] In `Chat` & `User` for all methods that accept `chat/user_id`
    - [x] In `Message` for all methods that accept `chat_id` and `message_id`
    - [ ] For new `Message` shortcuts: Added `quote` argument if methods accepts `reply_to_message_id`
    - [ ] In `CallbackQuery` for all methods that accept either `chat_id` and `message_id` or `inline_message_id`
    
* If relevant:

    - [x] Added new constants at `telegram.constants` and shortcuts to them as class variables
    - [ ] Added new handlers for new update types
    - [x] Added new filters for new message (sub)types
    - [x] Added or updated documentation for the changed class(es) and/or method(s)
    - [x] Updated the Bot API version number in all places: `README.rst` and `README_RAW.rst` (including the badge), as well as `telegram.constants.BOT_API_VERSION`
    - [ ] Added logic for arbitrary callback data in `tg.ext.Bot` for new methods that either accept a `reply_markup` in some form or have a return type that is/contains `telegram.Message`

PR #2978, opened 2022-04-27
Adjust bot method defaults

### LINKED ISSUE #2900: Handling of defaults for bot methods
Currently we have a few places, where the default value in the signature specifies the default value specified by Telegram. This is both the case for bot methods and Telegram classes. E.g. in `send_poll`:

https://github.com/python-telegram-bot/python-telegram-bot/blob/92cb6f3ae8d5c3e49b9019a9348d4408135ffc95/telegram/bot.py#L5141-L5147

and in `ReplyKeyboardMarkup`:

https://github.com/python-telegram-bot/python-telegram-bot/blob/92cb6f3ae8d5c3e49b9019a9348d4408135ffc95/telegram/replykeyboardmarkup.py#L87-L92

More importantly, we pass those values to TG. This isn't really necessary, since we'd get the same behavior when not passing those parameters. This has two effects:

1. More params are getting passed -> more traffic than strictly necessary. admittedly, probably barely noticable, but still.
2. In case TG changes some defaults, the change will not be reflected in PTB until we change the signature.

While the first one is clearly a downside, one *might* argue that the second one does have some advantages in terms of stability. Personally, I'd rather comply with the current state of the bot API, but I'm open to discussing this.

In any case, I think we should do this uniformly, i.e. either reflect the default value in the signature everywhere or nowhere & try to unit test this in `test_official`. E.g. Currently e.g. the `emoji` parameter of `send_dice` has `None` as default value instead of `'🎲'`:

https://github.com/python-telegram-bot/python-telegram-bot/blob/92cb6f3ae8d5c3e49b9019a9348d4408135ffc95/telegram/bot.py#L5315-L5326

In case we go for removing the default values, some bot methods internals may be simplified.
Either change would be non-breaking since the current defaults in the signatures are up to date with the defaults documented in the API.

harshil21:
Closes #2900 

To be merged after #2956


tests/test_official.py
```diff
@@ -77,15 +78,26 @@ def check_method(h4):
...
         ), f'Param {param.name!r} of method {method.__name__!r} requirement mismatch!'
-        checked.append(parameter[0])
+
+        # Now we will check that we don't pass default values if the parameter is not required.
+        # parameter[2] can either be Required/Yes or a paragraph where the first word has that info
+        telegram_param_required = is_parameter_required_by_tg(tg_parameter[2])
+
+        if not telegram_param_required:
+            default_arg_none = check_defaults_type(param)
+            assert default_arg_none, f"Param {param.name!r} of {method.__name__!r} should be None"
```
Bibo-Joshi, line 98: why does it matter if the parameter is required by TG? IISC it should suffice to check that all parameters that are being passed to TG are either required or have `(Default)None` as default value, no?

telegram/_forcereply.py
```diff
@@ -68,7 +68,7 @@ class ForceReply(TelegramObject):
 
     def __init__(
         self,
-        selective: bool = False,
+        selective: bool = None,
```
Bibo-Joshi, line 71: shouldn't the init be also changed to `self.selective = selective`, then?

telegram/_replykeyboardmarkup.py
```diff
@@ -41,12 +41,12 @@ class ReplyKeyboardMarkup(TelegramObject):
             each represented by an Array of :class:`telegram.KeyboardButton` objects.
         resize_keyboard (:obj:`bool`, optional): Requests clients to resize the keyboard vertically
             for optimal fit (e.g., make the keyboard smaller if there are just two rows of
-            buttons). Defaults to :obj:`False`, in which case the custom keyboard is always of the
+            buttons). Defaults to :obj:`None`, in which case the custom keyboard is always of the
```
Bibo-Joshi, line 44: I think I would prefer to keep this as is, i.e. document what TG does when the parameter is not passed. (BTW otherwise, we'd have to change also all the other docstrings)

telegram/_replykeyboardremove.py
```diff
@@ -57,7 +57,7 @@ class ReplyKeyboardRemove(TelegramObject):
 
     __slots__ = ("selective", "remove_keyboard")
 
-    def __init__(self, selective: bool = False, **_kwargs: Any):
+    def __init__(self, selective: bool = None, **_kwargs: Any):
```
Bibo-Joshi, line 60: see above

telegram/_replykeyboardmarkup.py
```diff
@@ -88,9 +88,9 @@ class ReplyKeyboardMarkup(TelegramObject):
     def __init__(
         self,
         keyboard: Sequence[Sequence[Union[str, KeyboardButton]]],
-        resize_keyboard: bool = False,
-        one_time_keyboard: bool = False,
-        selective: bool = False,
+        resize_keyboard: bool = None,
+        one_time_keyboard: bool = None,
+        selective: bool = None,
```
Bibo-Joshi, line 93: see above about the `bool` conversion

Bibo-Joshi (approved):

PR #3007, opened 2022-05-07
Fix in docs string

KnorpelSenf:
Did not read your template, sorry, just wanted to fix this one word.

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -3716,7 +3716,7 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
 
         If you'd like to make sure that the Webhook request comes from Telegram, Telegram
         recommends using a secret path in the URL, e.g. https://www.example.com/<token>. Since
-        nobody else knows your bot's token, you can be pretty sure it's us.
+        nobody else knows your bot's token, you can be pretty sure it's them.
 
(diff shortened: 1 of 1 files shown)


telegram/_bot.py
```diff
@@ -3716,7 +3716,7 @@ async def set_webhook(
 
         If you'd like to make sure that the Webhook request comes from Telegram, Telegram
         recommends using a secret path in the URL, e.g. https://www.example.com/<token>. Since
-        nobody else knows your bot's token, you can be pretty sure it's us.
+        nobody else knows your bot's token, you can be pretty sure it's them.
```
harshil21, line 3719:
```suggestion
        nobody else knows your bot's token, you can be pretty sure it's Telegram.
```
I would change it to `Telegram` since it makes more clear who it is, rather than `us` (PTB or Telegram?) and `them` (Telegram or the attackers?)

Bibo-Joshi: I'm fine with both.

KnorpelSenf: Please @ me when you have agreed on something, I'm fine with both, but slightly prefer `them` because it's been mentioned several times in the sentence before that we're talking about Telegram.

PR #3010, opened 2022-05-08
Doc fixes

Bibo-Joshi:
Gathering doc fixes to be merged before the next release

PRs to mention in the commit message:

- [ ] #3007
- [ ] #3012
- [ ] #3067 
- [ ] #3082
- [ ] #3081


harshil21:
I just noticed that after #3035, parameter referencing doesn't work for arguments under "Keyword Args:". Might be worth looking into later.

Bibo-Joshi:
> I just noticed that after #3035, parameter referencing doesn't work for arguments under "Keyword Args:". Might be worth looking into later.

Damn. But it appears to be rather easy to fix. From [the docs](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#confval-napoleon_use_keyword), https://github.com/sphinx-doc/sphinx/issues/2379 and https://github.com/sphinx-doc/sphinx/pull/2381 it appears that the only difference is that instead of `:param:` and `:type:`, the roles `:kwarg:` and `:kwtype:` are used. Adding a new capturing group for that [here](https://github.com/sqlalchemyorg/sphinx-paramlinks/blob/acedb03149e3f87ff599174b033754c2f58f1c95/sphinx_paramlinks/sphinx_paramlinks.py#L102-L103) and adjusting the functions `cvt` & `secondary_cvt` accordingly seems to do the trick in a first test. I'll try and PR for that upstream :)

PR #3011, opened 2022-05-08
Use ammaraskar/sphinx-action for nice error messages in docs CI

Bibo-Joshi:
let's see how well that works :)


Bibo-Joshi:
Welp, it was worth a try.

PR #3012, opened 2022-05-08
add Linkcheck to doc ci

### LINKED ISSUE #2974: [sphinx build] Adding linkcheck test
In my asyncio "review", I saw that somebody inserted a 404 link into the docstring. Apart from daring to do that to our docs (shame on them), existing links can break as well. SO I researched a bit and found the `linkcheck` build option of sphinx. We can run this like once a week. I am quite sure we have to parse the output since it doesn't break when it finds a non working link, but that shouldn't be too much work _TM_. From me messing around with it, we should definitely exclude our changelist file since it has too many github links and it doesn't matter too much if those break IMO.

Buildname linkcheck: https://www.sphinx-doc.org/en/master/man/sphinx-build.html

Config: https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-the-linkcheck-builder

After a short internal discussion, we want to different test triggers:
 - [ ] once/twice per month, on the whole docs
 - [ ] on a PR/push, only on the changed files.

Bibo-Joshi:
closes #2974 

runs the job every 2nd month. Ignores a number of URLs where it makes sense IMO. Also fixes a bunch.
Doesn't run on PRs for now. If we want that, we have two options:

- run on all files. After I ignored all the gh pr/issue links, that's actually not too bad performance wise. Would be okay with doing that.
- if we want to run only on changed files, we'd have to somehow map changed files to the corresponding rst files. IMO not worth the trouble.

Example for a failing test at https://github.com/python-telegram-bot/python-telegram-bot/runs/6339916713?check_suite_focus=true

- [ ] remove the `on: push` config from the workflow which is currently just there for demo purposes

### DIFF
--- a/.github/workflows/docs-linkcheck.yml
+++ b/.github/workflows/docs-linkcheck.yml
@@ -3,9 +3,6 @@ on:
   schedule:
     # First day of month at 05:46 in every 2nd month
     - cron: '46 5 1 */2 *'
-  push:
-    branches:
-      - linkcheck
 
 jobs:
   test-sphinx-build:
--- a/.github/workflows/docs-linkcheck.yml
+++ b/.github/workflows/docs-linkcheck.yml
@@ -1,13 +1,11 @@
 name: Test Documentation Build
 on:
-  pull_request:
-    branches:
-      - master
-      - doc-fixes
+  schedule:
+    # First day of month at 05:46 in every 2nd month
+    - cron: '46 5 1 */2 *'
   push:
-    branches: 
-      - master
-      - doc-fixes
+    branches:
+      - linkcheck
 
 jobs:
   test-sphinx-build:
--- a/.github/workflows/docs-linkcheck.yml
+++ b/.github/workflows/docs-linkcheck.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   test-sphinx-build:
-    name: test-sphinx-build
+    name: test-sphinx-linkcheck
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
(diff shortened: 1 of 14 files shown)


Poolitzer:
I don't think checking every link in every PR is going to help, potentially loads of checking at once. Just the month run would be fine. We shall see if that gets/is rate limited already anyway.

telegram/ext/_application.py
```diff
@@ -658,9 +658,10 @@ def run_webhook(
 
         If :paramref:`cert`
         and :paramref:`key` are not provided, the webhook will be started directly on
-        ``http://listen:port/url_path``, so SSL can be handled by another
+        ````http://listen:port/url_path````, so SSL can be handled by another
         application. Else, the webhook will be started on
-        ``https://listen:port/url_path``. Also calls :meth:`telegram.Bot.set_webhook` as required.
+        ````https://listen:port/url_path````. Also calls :meth:`telegram.Bot.set_webhook` as
```
harshil21, line 663: does that render right?

.github/workflows/docs-linkcheck.yml
```diff
@@ -0,0 +1,35 @@
...
+    strategy:
+      matrix:
+        python-version: [3.7]
+        os: [ubuntu-latest]
+      fail-fast: False
+    steps:
+      - uses: actions/checkout@v3
+      - name: Initialize vendored libs
+        run:
+          git submodule update --init --recursive
```
harshil21, line 23: no need of this

Bibo-Joshi:
If the two of you are okay with the rest of the changes, I'll remove the `on: push` and merge :)

harshil21 (approved):
looks good!

Poolitzer (approved):
LGTM

PR #3017, opened 2022-05-09
Fix `CCT.DEFAULT_TYPE` type hint

### LINKED ISSUE #3008: [Type Hinting] Pylance can't infer type of `CallbackContext.DEFAULT_TYPE`
### Steps to Reproduce

1. Start typing `context.bot.` in vscode with pylance
2. No type hints/autocomplete is available since pylance cannot infer the type of `CallbackContext.DEFAULT_TYPE`

The same applies for rest: `UD`, `CD`, `BD` (user/chat/bot_data)

### Expected behaviour

Pylance could maybe infer the type and give autocomplete suggestions? But I wouldn't expect it to.

### Actual behaviour

It doesn't infer the type correctly.

Possible solution: Remove `CallbackContext.DEFAULT_TYPE` and add a `CallbackContextType`  (or similar naming) where that would be simply `CallbackContext[ExtBot, Dict, Dict, Dict]`. We could keep this in the same file too.

### Operating System

Linux, Ubuntu 22.04

```shell
python-telegram-bot 20.0a0
Bot API 6.0
Python 3.10.4 (main, Apr  2 2022, 09:04:19) [GCC 11.2.0]
```

### Relevant log output

```python
-
```

### Additional Context

Originally reported here: https://t.me/pythontelegrambottalk/218000

harshil21:
Closes #3008 by changing the import to `ContextTypes.DEFAULT_TYPE`.

Also use `float()` in `timerbot.py` for setting the timer.

TODO:
- [ ] Update the wiki after release of v20.0a1
- [ ] Update transition script for people upgrading from v20.0a0

### DIFF
--- a/telegram/ext/_callbackcontext.py
+++ b/telegram/ext/_callbackcontext.py
@@ -34,7 +34,7 @@ from typing import (
 from telegram._callbackquery import CallbackQuery
 from telegram._update import Update
 from telegram.ext._extbot import ExtBot
-from telegram.ext._utils.types import BD, BT, CD, UD  # pylint: disable=unused-import
+from telegram.ext._utils.types import BD, BT, CD, UD
 
 if TYPE_CHECKING:
     from asyncio import Queue
@@ -301,12 +301,12 @@ class CallbackContext(Generic[BT, UD, CD, BD]):
             if chat:
                 self._chat_id_and_data = (
                     chat.id,
-                    application.chat_data[chat.id],  # pylint: disable=protected-access
+                    application.chat_data[chat.id],
                 )
             if user:
                 self._user_id_and_data = (
                     user.id,
-                    application.user_data[user.id],  # pylint: disable=protected-access
+                    application.user_data[user.id],
                 )
         return self
 
@@ -336,12 +336,12 @@ class CallbackContext(Generic[BT, UD, CD, BD]):
         if job.chat_id:
             self._chat_id_and_data = (
                 job.chat_id,
-                application.chat_data[job.chat_id],  # pylint: disable=protected-access
+                application.chat_data[job.chat_id],
             )
         if job.user_id:
             self._user_id_and_data = (
                 job.user_id,
-                application.user_data[job.user_id],  # pylint: disable=protected-access
+                application.user_data[job.user_id],
             )
         return self
 
--- a/telegram/ext/_callbackcontext.py
+++ b/telegram/ext/_callbackcontext.py
@@ -38,9 +38,10 @@ from telegram.ext._utils.types import BD, BT, CD, UD
 
 if TYPE_CHECKING:
     from asyncio import Queue
+    from typing import Any
 
-    from telegram.ext import Application, Job, JobQueue
-    from telegram.ext._utils.types import CCT, JQ
+    from telegram.ext import Application, Job, JobQueue  # noqa: F401
+    from telegram.ext._utils.types import CCT
 
 _STORING_DATA_WIKI = (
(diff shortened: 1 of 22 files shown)


telegram/ext/_callbackcontext.py
```diff
@@ -321,12 +301,12 @@ def from_update(
             if chat:
                 self._chat_id_and_data = (
                     chat.id,
-                    application.chat_data[chat.id],  # pylint: disable=protected-access
```
Bibo-Joshi, line 324: how come this is not needed anymore?

telegram/ext/_callbackcontext.py
```diff
@@ -104,26 +104,6 @@ class CallbackContext(Generic[BT, UD, CD, BD]):
 
     """
 
-    if TYPE_CHECKING:
-        DEFAULT_TYPE = CallbackContext[  # type: ignore[misc]  # noqa: F821
-            ExtBot, Dict, Dict, Dict
```
Bibo-Joshi, line 109: maybe add a seealso instead that points to `ContextTypes.DEFAULT_TYPE`?

telegram/ext/_contexttypes.py
```diff
@@ -53,6 +53,19 @@ class ContextTypes(Generic[CCT, UD, CD, BD]):
 
     """
 
+    DEFAULT_TYPE = CallbackContext["ExtBot", Dict, Dict, Dict]  # type: ignore[misc]
+    """Shortcut for the type annotation for the `context` argument that's correct for the
+    default settings, i.e. if :class:`telegram.ext.ContextTypes` is not used.
+
```
Bibo-Joshi, line 59: should we maybe still add a comment that this may not work with all type checkers? 🤔

telegram/ext/_callbackcontext.py
harshil21: I was surprised too, maybe something in pylint internals stopped it from detecting this before? idk

telegram/ext/_contexttypes.py
harshil21: Don't think its needed as long we're supporting the popular type checkers

telegram/ext/_callbackcontext.py
Bibo-Joshi: okay. interesting :D

telegram/ext/_contexttypes.py
```diff
@@ -53,6 +53,19 @@ class ContextTypes(Generic[CCT, UD, CD, BD]):
 
     """
 
+    DEFAULT_TYPE = CallbackContext["ExtBot", Dict, Dict, Dict]  # type: ignore[misc]
+    """Shortcut for the type annotation for the `context` argument that's correct for the
```
Bibo-Joshi, line 57:
```suggestion
    """Shortcut for the type annotation for the ``context`` argument that's correct for the
```

Bibo-Joshi (approved):
code LGTM. did you check with both pyright and mypy if they can work with the changes? did you try autocompletion also in both VSC + pylance & pycharm?

telegram/ext/_contexttypes.py
```diff
@@ -53,6 +53,19 @@ class ContextTypes(Generic[CCT, UD, CD, BD]):
 
     """
 
+    DEFAULT_TYPE = CallbackContext["ExtBot", Dict, Dict, Dict]  # type: ignore[misc]
```
Bibo-Joshi, line 56: actually, I think we can just use `dict` here instead of `typing.Dict` since we don't need to specify the types for keys & values

PR #3019, opened 2022-05-09
Rename `Handler` to `BaseHandler`

Bibo-Joshi:
### Checklist for PRs

TODO:
- [ ] Update wiki after release of v20.0a1
- [ ] Update transition script for this

### DIFF
--- a/docs/source/telegram.ext.basehandler.rst
+++ b/docs/source/telegram.ext.basehandler.rst
@@ -1,0 +1,6 @@
+telegram.ext.BaseHandler
+========================
+
+.. autoclass:: telegram.ext.BaseHandler
+    :members:
+    :show-inheritance:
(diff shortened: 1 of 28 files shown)


docs/source/telegram.ext.basehandler.rst
```diff
@@ -0,0 +1,6 @@
+telegram.ext.BaseHandler
```
harshil21, line 1: if the file was renamed instead of deleted, the git blame history would be saved right?

Bibo-Joshi: mh, I thought pycharms renaming thingy would use `git mv` …

Bibo-Joshi: force pushed a commit where i manually used `git mv`, but that didn't help either. afterwards I also found https://stackoverflow.com/questions/2314652 :D

PR #3026, opened 2022-05-12
Fix: Change string to bool

Poolitzer:
### Checklist for PRs

I found this while adapting the V13 tests. According to our type hinting this should be more correct, lets see if the CI agrees.


tests/test_chatphoto.py
```diff
@@ -83,7 +83,7 @@ async def test_get_and_download(self, bot, chat_photo):
 
         new_file = await bot.get_file(chat_photo.small_file_id)
 
-        assert isinstance(new_file.file_id, str) and new_file.file_id != ""
+        assert new_file.unique_file_id == chat_photo.small_file_unique_id
```
Bibo-Joshi, line 86:
```suggestion
        assert new_file.file_unique_id == chat_photo.small_file_unique_id
```

Bibo-Joshi (changes requested):

PR #3035, opened 2022-05-14
Introduce keyword-only arguments in Bot methods

### LINKED ISSUE #3016: Introduce keyword-only arguments for auxiliary arguments of Bot methods
# Current situation

All bot methods have arguments `{read, write, connect, pool}_timeout` and `api_kwargs`, which are added by PTB and do not correspond to arguments defined by the Bot API.
When a bot API update introduces new arguments to a method, we currently have to add them *after* the arguments mentioned above to ensure backwards compatibility for people that pass the arguments as positional arguments. This leads to constructs like

https://github.com/python-telegram-bot/python-telegram-bot/blob/d917404de1b6f6fd7e11ed2301bb89d68904d6cf/telegram/_bot.py#L557-L574

which is not very nice and may lead to people passing values to the wrong parameters if they rely on the order defined by the bot API.

# Proposed solution

Make the arguments `{read, write, connect, pool}_timeout` and `api_kwargs` [keyword-only arguments](https://peps.python.org/pep-3102/).

# ToDo

A PR for this issue should include:

* Update the definitions of all methods of the `Bot` class that have these argumemts
* The docstrings should be updated to reflect the order of the signature
* all corresponding methods of the `ExtBot` class should be updated as well
* as should all shortcut methods like `Message.reply_text`, `Chat.send_message` etc
* It should be checked if the rendered HTML documentation includes the `*` in the displayed signature. If it does not, the docstrings should include some other indicator that these arguments are keyward-only arguments. How this should look may be discussed at a later stage
* The unit tests must be updated accordingly. Running the test suit to see which tests fail after the changes should be a good start. Additionally, we should make sure that the auxiliary functions that currently test the signature definition of shortcut methods are still working as expected. This is a bit more intricate, so the development team can help with that.

# Additional Info

We have a [contribution guide](https://python-telegram-bot.readthedocs.io/en/v20.0a0/contributing.html)

tal66:
Hi, this closes #3016. 
Made `{read, write, connect, pool}_timeout` , `api_kwargs` keyword-only arguments.
The docs generated looked fine to me, and the tests passed without much changing - did I miss anything?

(btw, happened to notice in `test_message.py`, that some args were in the wrong order - for example, line 347. just added argument names there)

Edit: saw the pre-commit formatting error (4 lines), will fix in the next commit after your comments


telegram/_user.py
```diff
@@ -650,17 +661,18 @@ async def send_document(
         disable_notification: DVInput[bool] = DEFAULT_NONE,
         reply_to_message_id: int = None,
         reply_markup: ReplyMarkup = None,
-        read_timeout: ODVInput[float] = DEFAULT_NONE,
-        write_timeout: ODVInput[float] = 20,
-        connect_timeout: ODVInput[float] = DEFAULT_NONE,
-        pool_timeout: ODVInput[float] = DEFAULT_NONE,
         parse_mode: ODVInput[str] = DEFAULT_NONE,
         thumb: FileInput = None,
         api_kwargs: JSONDict = None,
```
harshil21, line 666: `api_kwargs` should be moved to keyword only section for `User.send_document`

telegram/_user.py
```diff
@@ -988,6 +1001,11 @@ async def send_video(
         caption_entities: Union[List["MessageEntity"], Tuple["MessageEntity", ...]] = None,
         filename: str = None,
         protect_content: ODVInput[bool] = DEFAULT_NONE,
+        *,
```
harshil21, line 1004: same here, `api_kwargs` for `User.send_video` should be moved under the asterisk.

tests/test_chat.py
```diff
@@ -28,8 +28,8 @@
 def chat(bot):
     return Chat(
         TestChat.id_,
-        TestChat.title,
-        TestChat.type_,
+        title=TestChat.title,
+        type=TestChat.type_,
```
harshil21, line 32: Any specific reason why this was changed? A failing test?

harshil21 (changes requested):
Thanks for the PR! Overall looks great! Some comments:

1. imo, the `quote` argument of the shortcut methods in `Message`, `Chat`, `User` should also be keyword only since its added by PTB. But lets wait and see what the other maintainers think.
2. About tests: 

    -  In `test_bot.py`, one could add a test to check all bot methods and see if there are exactly 5 keyword arguments by checking their [`kind`](https://docs.python.org/3.12/library/inspect.html#inspect.Parameter.kind) and [`name`](https://docs.python.org/3.12/library/inspect.html#inspect.Parameter.name) (`{read, write, connect, pool}_timeout` and `api_kwargs`). One could [retrieve](https://docs.python.org/3.12/library/inspect.html#inspect.getmembers) all public bot methods by using [`iscoroutinefunction`](https://docs.python.org/3.12/library/inspect.html#inspect.iscoroutinefunction) and checking if the method doesn't start with an underscore. 
    - The same would be need to done for the shortcuts, i.e in `test_{chat, user, message}`

tests/test_chat.py
tal66: because it's not in the same order:

`telegram._chat.Chat def __init__(self,
             id: int,
             type: str,
             title: Optional[str] = None,
...
`

tal66:
2nd commit addressed your some of your comments. 

I'll add the test you asked tomorrow (UTC time).

telegram/_bot.py
```diff
@@ -157,6 +157,9 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
           :class:`telegram.ext.Defaults`, please use the subclass :class:`telegram.ext.ExtBot`
           instead.
         * Attempting to pickle a bot instance will now raise :exc:`pickle.PicklingError`.
+        * The following are now keyword-only arguments in Bot methods:
+          {read, write, connect, pool}_timeout, api_kwargs. Use a named argument for those,
```
Bibo-Joshi, line 161:
```suggestion
          ``{read, write, connect, pool}_timeout``, ``api_kwargs``. Use a named argument for those,
```

telegram/_bot.py
```diff
@@ -562,14 +568,15 @@ async def send_message(
         disable_notification: DVInput[bool] = DEFAULT_NONE,
         reply_to_message_id: int = None,
         reply_markup: ReplyMarkup = None,
+        allow_sending_without_reply: ODVInput[bool] = DEFAULT_NONE,
+        entities: Union[List["MessageEntity"], Tuple["MessageEntity", ...]] = None,
+        protect_content: ODVInput[bool] = DEFAULT_NONE,
+        *,
```
Bibo-Joshi, line 574: if we have to move these anyway, I think I'd like to put them into the order [defined by telegram](https://core.telegram.org/bots/api#sendmessage). But it's not a priority to me.

telegram/_bot.py
```diff
@@ -7187,6 +7429,7 @@ async def log_out(
     @_log
     async def close(
         self,
+        *,
         read_timeout: ODVInput[float] = DEFAULT_NONE,
         write_timeout: ODVInput[float] = DEFAULT_NONE,
         connect_timeout: ODVInput[float] = DEFAULT_NONE,
```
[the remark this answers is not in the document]
Bibo-Joshi: No, I don't think there's a reason - probably just forgotten. @tal66 if you could add it, that would be great :)

telegram/_chat.py
```diff
@@ -66,7 +66,11 @@ class Chat(TelegramObject):
     considered equal, if their :attr:`id` is equal.
 
     .. versionchanged:: 20.0
-        Removed the deprecated methods ``kick_member`` and ``get_members_count``.
+
+        * Removed the deprecated methods ``kick_member`` and ``get_members_count``.
+        * The following are now keyword-only arguments in Bot methods:
+          {read, write, connect, pool}_timeout, api_kwargs. Use a named argument for those,
```
Bibo-Joshi, line 72:
```suggestion
          ``{read, write, connect, pool}_timeout``, ``api_kwargs``. Use a named argument for those,
```

telegram/_message.py
```diff
@@ -85,13 +85,17 @@ class Message(TelegramObject):
...
+
+        * The arguments and attributes ``voice_chat_scheduled``, ``voice_chat_started`` and
+          ``voice_chat_ended``, ``voice_chat_participants_invited`` were renamed to
+          :paramref:`video_chat_scheduled`/:attr:`video_chat_scheduled`,
+          :paramref:`video_chat_started`/:attr:`video_chat_started`,
+          :paramref:`video_chat_ended`/:attr:`video_chat_ended` and
+          :paramref:`video_chat_participants_invited`/:attr:`video_chat_participants_invited`,
+          respectively, in accordance to Bot API 6.0.
+        * The following are now keyword-only arguments in Bot methods:
+          {read, write, connect, pool}_timeout, api_kwargs. Use a named argument for those,
```
Bibo-Joshi, line 97: see above

telegram/_user.py
```diff
@@ -62,6 +62,12 @@ class User(TelegramObject):
     Objects of this class are comparable in terms of equality. Two objects of this class are
     considered equal, if their :attr:`id` is equal.
 
+    .. versionchanged:: 20.0
+
+        The following are now keyword-only arguments in Bot methods:
+        {read, write, connect, pool}_timeout, api_kwargs. Use a named argument for those,
```
Bibo-Joshi, line 68: see above

Bibo-Joshi (commented):
Thanks for the nice PR! I left a new nitpicks below.

> 1. imo, the `quote` argument of the shortcut methods in `Message`, `Chat`, `User` should also be keyword only since its added by PTB. But lets wait and see what the other maintainers think.

I guess the need to separate that from the other arguments is not *as* big, but I'm okay with the idea. We should then however make *all* PTB-added arguments keyword-only. We have have few more of those, e.g. `InlineQurey.answer(…, auto_pagination=True)` and `Bot.send_location(…, location=…)`. For the methods of `Bot`, those can be seen [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/5e924014debbb6c696ee33aac1bfe477ba51e87b/tests/test_official.py#L45-L50). For the shortcut methods, that would be all the arguments that are passed `check_shortcut_signature(…, addition_kwargs=[…])` in the tests, e.g. [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/5e924014debbb6c696ee33aac1bfe477ba51e87b/tests/test_message.py#L719-L721).
Inserting a `print(shortcut, additional_kwargs)` [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/5e924014debbb6c696ee33aac1bfe477ba51e87b/tests/conftest.py#L490) & then running the test suite once should reveal all of them.

Corresponding `.. versionchanged:: 20.0` directives should be added.

> 2. About tests: 
>
>     -  In `test_bot.py`, one could add a test to check all bot methods and see if there are exactly 5 keyword arguments by checking their [`kind`](https://docs.python.org/3.12/library/inspect.html#inspect.Parameter.kind) and [`name`](https://docs.python.org/3.12/library/inspect.html#inspect.Parameter.name) (`{read, write, connect, pool}_timeout` and `api_kwargs`). One could [retrieve](https://docs.python.org/3.12/library/inspect.html#inspect.getmembers) all public bot methods by using [`iscoroutinefunction`](https://docs.python.org/3.12/library/inspect.html#inspect.iscoroutinefunction) and checking if the method doesn't start with an underscore. 
>     - The same would be need to done for the shortcuts, i.e in `test_{chat, user, message}`

I think we can get away more easily if we make all ptb-added parameters kwargs-only:

* For checking the signature of bot methods, we can just check in `test_official` that all parameters not documented by TG are are kwargs-only.
* For checking shortcut signatures, it would suffice to ensure that the `additional_kwargs` are kwargs-only. all other parameters are matched against the Bot signature anyway.

@harshil21 what do you think?

@tal66 I'm aware that unit tests can be tedious to wrap your head around if you're not entirely familiar with the code base. If you have troubles with updating tests, please don't hesitate to reach out either here or in the dev-group on Telgeram!

harshil21:
> make all PTB-added arguments keyword-only

agree.

>  we can just check in `test_official`

sure, modifying that test would be easier than starting from scratch

tal66:
Hi, for the tests I wanted to use `inspect.getfullargspec`, but for some reason it doesn't work as expected from inside `test_official` (thought of using it inside `check_method`).

For example, `print(inspect.getfullargspec(telegram.Bot.send_message))` gives ```FullArgSpec(args=[], varargs='args', varkw='kwargs', defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={'return': <class 'telegram._message.Message'>})``` with args and kwonlyargs empty `[]` (even though inspect.signature does work). And it's the same for all other methods.
Any idea why?

tal66:
well, I tried some stuff and this works: `[p.name for p in inspect.signature(method).parameters.values() if p.kind == inspect.Parameter.POSITIONAL_OR_KEYWORD]`. nice. but it's still weird that `getfullargspec` doesn't.

Bibo-Joshi:
Hi. `getfullargspec` can not "see through" decorators (at least not in all cases IIRC) and all the bot methods have a decorator for logging. `inspect.signature` is indeed the recommended method for inspecting the signature :)

telegram/_bot.py
```diff
@@ -1161,6 +1159,9 @@ async def send_document(
                    Accept :obj:`bytes` as input.
 
         Keyword Args:
+            filename (:obj:`str`, optional): Custom file name for the document, when uploading a
+                new file. Convenience parameter, useful e.g. when sending files generated by the
+                :obj:`tempfile` module.
```
harshil21, line 1164:
``` suggestion
                :obj:`tempfile` module.

                .. versionadded:: 13.1

File: telegram/_inline/inlinequery.py [hunk 11]
```diff
@@ -123,13 +129,14 @@ async def answer(
         next_offset: str = None,
         switch_pm_text: str = None,
         switch_pm_parameter: str = None,
+        current_offset: str = None,
+        *,
+        auto_pagination: bool = False,
```
harshil21, line 135: can you also change the Args: to Keyword Args in the docs for this method (InlineQuery.answer())?

telegram/_inline/inlinequery.py
```diff
@@ -43,6 +43,12 @@ class InlineQuery(TelegramObject):
     Note:
         In Python :keyword:`from` is a reserved word use :paramref:`from_user` instead.
 
+    .. versionchanged:: 20.0
+
+        * The following are now keyword-only arguments in Bot methods:
+          ``{read, write, connect, pool}_timeout``, ``api_kwargs``, ``auto_pagination``. Use a named argument
```
harshil21:
``` suggestion
          ``{read, write, connect, pool}_timeout``, ``api_kwargs``, :paramref:`answer.auto_pagination`. Use a named argument
```
since its being used in only one place I guess we can directly reference it where

tests/test_official.py
```diff
@@ -123,6 +123,23 @@ def check_method(h4):
 
     assert (sig.parameters.keys() ^ checked) - ignored == set()
 
+    check_extra_params_are_keywordonly(method.__name__, checked, name)
+
+
+def check_extra_params_are_keywordonly(method_name, api_params, name_in_api):
+    for c in [telegram.Bot, telegram.Chat, telegram.Message, telegram.User]:
```
harshil21, line 130: a bit futile to test anything other than Bot since they mostly have different names anyway. We would need the test [hinrich](https://github.com/python-telegram-bot/python-telegram-bot/pull/3035#pullrequestreview-973159961) mentioned for the shortcut methods :)

tal66: ok, moved shortcut methods to a separate test `test_official.test_shortcut_methods__extra_params_are_keyword_only`. 
commit in a few minutes

Bibo-Joshi: no, not quite. To test that shortcut methods behavo correctly, we don't need to change anything in `test_official`. It suffices to adapt [this function](https://github.com/python-telegram-bot/python-telegram-bot/blob/298c5fab3b8655ebae4416341335fd9c05f2ce5a/tests/conftest.py#L471-L530) which is used in all the tests of shortcut methods. Please adapt that function such that it checks that the `additional_kwargs` are keyword-only arguments of the `shortcut`. 

The overall ideas is:

* `test_official` ensures that the positional arguments of bot methods correspond to the offical bot API and that any additional arguments are keyword-only
* `check_shortcut_signature` ensures that the signatures of the shortcut methods match the signatures of the bot methods. because the bot methods are already checked agains `test_official`, this also ensures that the shortcut methods match the official signatures

Bibo-Joshi: BTW after removing the loop over the different classes, this method can be simplified by just adding the logic of the function to the body of check_method :)

tal66: I see, changing it. question: I did a rebase master locally and the hashes changed, does it matter to you if i'll use `git push -f` or to stash and return to the previous flow?

tal66: `additional_kwargs` doesn't include all the timeout args, filename, etc..
so i'm also checking that kw args in bot_method_sig.parameters are kw args in shortcut. 

you definitely have too many shortcut methods.. wasn't there a more maintainable way to define those?

Bibo-Joshi: > `additional_kwargs` doesn't include all the timeout args, filename, etc.. so i'm also checking that kw args in bot_method_sig.parameters are kw args in shortcut.

you're right, thanks 👍 

> you definitely have too many shortcut methods.. wasn't there a more maintainable way to define those?

I do think that these shortcuts are one of the strengths of PTB 🤓

Bibo-Joshi:
I took the liberty to further adjust `check_shortcut_signature`, which also revealed one more necessary change in `CallbackQuery.edit_message_caption` :)
With this, I don't think there's anything left to add from my side. If @harshil21 gives a LGTM as well, we can merge :)

tests/conftest.py
```diff
@@ -504,6 +504,10 @@ def check_shortcut_signature(
     # shortcuts to return more specific types than the bot method, but it's only annotations after
     # all
     for kwarg in effective_shortcut_args:
+        expected_kind = shortcut_sig.parameters[kwarg].kind
+        if bot_sig.parameters[kwarg].kind != expected_kind:
```
harshil21, line 508:
```suggestion
        expected_kind = bot_sig.parameters[kwarg].kind
        if shortcut_sig.parameters[kwarg].kind != expected_kind:
```
shouldn't this be the other way around? Bot signature should be the 'correct' one, not the other way around

Bibo-Joshi: Eagle eyes as usual :)

harshil21 (approved):
LGTM! If there are some docs to updated, that can also be done later on

PR #3037, opened 2022-05-16
Drop support for `ujson`

Bibo-Joshi:
Let me try to wrap up what lead us here:

* Noam suggested to switch from `ujson` to `orjson` as optional 3rd party json library, because the latter has serveral benefits
* orjson turned out to not play well with httpx out of the box, because `orjson.dumps` returns bytes instead of a string
* investigating the different possibilities lead us to the conclusion that it may be better (in terms of maintenance and code design) if PTB did not offer a built-in support for a 3rd party json library

Hence, this PR removes support for `ujson`.
We don't really know how many users are actually using the `ujson` feature currently and we also don't have any benchmarks on how `ujson` actually affects the performance. _If_ users do request the option to use a custom json library, there are a few things that we can do and that we need to consider. I'm writing them down below in some detail for future reference.

* Where is `ujson` currently used, i.e. where could users want to use a 3rd party library instead?
  1. `DictPersisence`: This is no problem. `BasePersistence` is already an ABC, so any user can go ahead and implement a custom version of `DictPersistence` with a 3rd party json library
  2. a small part in the passport-decryption logic. This is a problem, but IMHO a neglectible one b/c passports are basically dead anyways
  3. `WebhookHandler`: No problem. Users can implement a custom webhook solution that uses a 3rd party json lib
  4. `TelegramObject.to_json`: No problem. This method is not actually used anywhere.
  5. Networking backend. This is the actually interesting point. Let's get into that
* Networking backend: JSON DE-coding: This is currently not possible to customize. If needed, we can easily fix that by making `BaseRequest._parse_json_payload` a public method. Then custom implementations can override it to use a 3rd party json lib.
* Networking backend: JSON EN-coding: This is already mostly doable already: in `do_request` just do something like `{name: my_json_lib.dumps(value) for name, value in request_data.parameters.items()}`. However, the big performance advantage of some json libraries comes from the fact that they can readily json-dump a larger variety of objects, i.e. not only lists, dicts, strings & numbers but also e.g. `datetime` and `dataclass` objects. So a user might want to skip the `TelegramObject.to_dict()` step. To support this, we could consider to add a propert `RequestData.raw_parameters` (or similar naming) through which we expose exactly those values.

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -42,10 +42,6 @@ from typing import (
     no_type_check,
 )
 
-try:
-    import ujson as json
-except ImportError:
-    import json  # type: ignore[no-redef]  # noqa: F723
 
 try:
     from cryptography.hazmat.backends import default_backend
@@ -4571,10 +4567,7 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         if start_parameter is not None:
             data["start_parameter"] = start_parameter
         if provider_data is not None:
-            if isinstance(provider_data, str):
-                data["provider_data"] = provider_data
-            else:
-                data["provider_data"] = json.dumps(provider_data)
+            data["provider_data"] = provider_data
         if photo_url is not None:
             data["photo_url"] = photo_url
         if photo_size is not None:
--- a/README.rst
+++ b/README.rst
@@ -137,7 +137,6 @@ Optional Dependencies
 PTB can be installed with optional dependencies:
 
 * ``pip install python-telegram-bot[passport]`` installs the `cryptography>=3.0 <https://cryptography.io>`_ library. Use this, if you want to use Telegram Passport related functionality.
-* ``pip install python-telegram-bot[json]`` installs the `ujson>=4.0.0 <https://pypi.org/project/ujson/>`_ library. It will then be used for JSON de- & encoding, which can bring speed up compared to the standard `json <https://docs.python.org/3/library/json.html>`_ library.
 * ``pip install python-telegram-bot[socks]`` installs ``httpx[socks]``. Use this, if you want to work behind a Socks5 server.
 
 Quick Start
--- a/README_RAW.rst
+++ b/README_RAW.rst
@@ -135,7 +135,6 @@ Optional Dependencies
 ``python-telegram-bot-raw`` can be installed with optional dependencies:
 
 * ``pip install python-telegram-bot[passport]`` installs the `cryptography <https://cryptography.io>`_ library. Use this, if you want to use Telegram Passport related functionality.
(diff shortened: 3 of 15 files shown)


tsnoam:
@Bibo-Joshi Thanks!

This sums up our conversation very well.

harshil21 (approved):

telegram/_bot.py
```diff
@@ -4571,10 +4567,7 @@ async def send_invoice(
         if start_parameter is not None:
             data["start_parameter"] = start_parameter
         if provider_data is not None:
-            if isinstance(provider_data, str):
-                data["provider_data"] = provider_data
-            else:
-                data["provider_data"] = json.dumps(provider_data)
+            data["provider_data"] = provider_data
```
Poolitzer, line 4570: if this can only be a string now for some reason, you would need to edit the doc string + type hints.

Poolitzer (changes requested):
Actually found a mistake \o/

telegram/_bot.py
harshil21: it can be a string or an object which is json serializable, so its correct?

Poolitzer: then at least the doc is wrong, which says it would be encoded, and we should definitely note down the change in the doc anyway. But is there not a better type hint to say that an object needs to be json serializable?

Bibo-Joshi: The encoding is done by the networking backend, same as for the other telegram objects that are accepted as parameter values. If that wasn't the case, the tests would fail :)

Poolitzer: So you are saying the same error would appear if I were to supply a non serializable object, just with a different traceback?

Bibo-Joshi: yes. after all, everything that we pass to TG has to be JSON-dumped at some point. The new `BaseRequest` + `RequestData` just makes it more clear, where that's supposed to happen so that we no longer need it here.

Poolitzer: ah okay then

Poolitzer (approved):

MiguelX413:
I would definitely appreciate an easy way to customize the json backend as a fan of ujson. I would like to be able to keep using it in my Telegram bots.

Bibo-Joshi:
Thanks for reaching out @MiguelX413. I would like to have this change published in a release before making a decision on this topic, so that we can get a better impression of the impact of this PR on the community. I hope that you can understand my reasoning.

Would you mind opening a feature request issue so that we don't forget about this?
For the time being, v20.0a0 still supports ujson and v20.0a1 with this change in it will still allow customizing the json ENcoding behavior as detailed above.

MiguelX413:
> Thanks for reaching out @MiguelX413. I would like to have this change published in a release before making a decision on this topic, so that we can get a better impression of the impact of this PR on the community. I hope that you can understand my reasoning.

That makes sense to me. @Bibo-Joshi 
> Would you mind opening a feature request issue so that we don't forget about this? For the time being, v20.0a0 still supports ujson and v20.0a1 with this change in it will still allow customizing the json ENcoding behavior as detailed above.

Sure thing.

PR #3042, opened 2022-05-17
EasyConversationHandler

123yagel:
Hey!
I found myself struggling a lot with `ConversationHandler` state machine, so I came up with new idea that I didn't found out there: Programming bot logic with _regular functions_ that waits for input from the user!
I implemented POC (MVP actually) here, and want to open a discussion about it. What relevant, what less relevant, and improving implementation.

### Example
```Python
async def collect_number(context):
    """
    Collects an int from the user
    """
    while True:
        update = await context.get_update()
        if update.message.text.isdigit():
            return int(update.message.text)
        await update.message.reply_text('This is not a number. Enter a number:')
```
And we can use this function in other funtions:
```Python
    await context.last_update.message.reply_text(f'How many apples?')
    how_many = await collect_number(context)
```

Adding handler:
```Python
    application.add_handler(EasyConversationHandler(siblings_callback, first_message_filter=Regex('/siblings')))
```
See the full example in `examples/easyconversationbot.py`

### Checklist for PRs
I didn't checked all of this yet, but will do after some discussion and CR

* Added new shortcuts:
    
* If relevant:

    - [ ] Link new and existing constants in docstrings instead of hard coded number and strings
    - [ ] Add new message types to `Message.effective_attachment`
      - [ ] Add the handlers to the warning loop in the `ConversationHandler`

### DIFF
--- a/examples/easyconversationbot.py
+++ b/examples/easyconversationbot.py
@@ -51,10 +51,9 @@ async def collect_number(context):
     """
     while True:
         update = await context.get_update()
-        if update.message.text.isdigit() or True:
+        if update.message.text.isdigit():
             return int(update.message.text)
         await update.message.reply_text('This is not a number. Enter a number:')
-        continue
 
 
 async def collect_list(context, list_name):
--- a/examples/easyconversationbot.py
+++ b/examples/easyconversationbot.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# pylint: disable=unused-argument, wrong-import-position
+# pylint: disable=wrong-import-position
 # This program is dedicated to the public domain under the CC0 license.
 
 """
@@ -90,7 +90,7 @@ async def print_data(update, context):
     await update.message.reply_text(f"Your children: {children}")
 
 
-async def print_help(update, context_unused):
+async def print_help(update, _):
     """Callback for /help command"""
     await update.message.reply_text('Type /siblings, /children or /print')
 
--- a/examples/easyconversationbot.py
+++ b/examples/easyconversationbot.py
@@ -82,6 +82,7 @@ async def children_callback(context):
 
 # Simple command handlers callbacks
 
+
 async def print_data(update, context):
     """Callback for /print command"""
     siblings = context.user_data.get(SIBLINGS_KEY, "Not set yet")
--- a/examples/easyconversationbot.py
+++ b/examples/easyconversationbot.py
@@ -4,7 +4,8 @@
 
 """
 Simple example of EasyConversationBot.
-The bot let user add family members and print them, in simple, functional way - without state machines.
+The bot let user add family members and print them, in simple, functional way
+ - without state machines.
 """
 
 import logging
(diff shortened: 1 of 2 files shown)


123yagel:
pre-commit.ci autofix

Poolitzer:
Hello there \o/

First of all, thank you very much for your PR. You found a problem for yourself, went ahead and fixed it for yourself, that is great. And even better, you decided to contribute it back to our project, so again, thanks!

Sadly, I am afraid that we don't want another ConversationHandler type in our main repository. This wouldn't fit into the architecture/design of PTB we aim for.

That doesn't mean all your effort is in vain though! We very much want to redirect you to our [ptbcontrib](https://github.com/python-telegram-bot/ptbcontrib/) repository! We made that exactly for code ideas like yours: Something which extends the functionality of PTB, but might be too special/outside of the architecture to be part of the main code repo.
We can then link to your module from the wiki and the examples.

Lastly, a small suggestion: If you have an idea as this for a future project, maybe make an issue first and get the opinion of the maintainers. You can get a better outline for the idea and know if they are going to accept it or not before you put a lot of work into it.

123yagel:
Ok, Thanks you for the reply - I understand and appreciate the high standards of pdb.

I have learned a lot on the way, and already refactor my bot to use this (with more simple fixes and new name), so my work isn't for nothing :)

In this opportunity, May I ask for an implementation advice?
As you can see in the code, I gave the callback the same `CallbackContext` but with some added attributes. What type of object should I give instead? I want that the programmer can use
* `.replay_X()`,
*  `.get_update()`,
*  and use the context regular attributes like `.user_data`.
Should I inherit from `CallbackContext` at all? How do I "link" all of the `reply_X()` method of message?

Bibo-Joshi:
> In this opportunity, May I ask for an implementation advice?

Hi. TBH the main issue I see with your implementation is that it does not play well with the overall setup of the PTB framework, especially the handlers.

After the conversation starts in a chat, the `EasyConversationHandler` will grap *every* update from that chat regardless of whether that update will actually be used in the internal logic. That prevents other handlers (added *after* the conversation) from handling them. Any logic that one could add to `EasyConversationHandler` in order to adjust that would more or less bring us back to a full finite state machine setup as `ConversationHandler` already is. Providing `per_{chat, user, message}` arguments as `ConversationHandler` would only improve that marginally as well.

Moreover the logic of "fetching updates from a queue as long I need" feels like a re-implementation of the whole `Application.run_polling` - which already fetches the updates from the `update_queue` and processes them one by one.

The framework that PTB provides is certainly not the only one that one can imagine and using a custom logic for processing the updates may very well be suited for your use case. I don't see it integrated in a handler-based setup though and as pool already said not in the main library at all. For your custom implementation I would recommend to decouple your design from the handler setup and work directly with the `Updater` (in case you want to use that class to fetch the updates from Telegram). If you want to use the `CallbackContext` setup along with `bot/chat/user_data`, then you should at least skip `Application.{run_polling, process_update}`  and instead manually build the `context` argument via `CallbackContext.from_update`.

PR #3045, opened 2022-05-18
Start pulling apart CH & PH

### LINKED ISSUE #2885: Rename CommandHandler.command to CH.commands (plural) …

Bibo-Joshi:
Closes #2885. supersedes & closes #2890

in hindsight the review process of #2890 revealed to me that it might be better to

* properly split PrefixHandler from CommandHandler
* not have the commends updatable via setters. For now I completely removed that. If users actually use that & ask for it, we cann add proper method `update_commands()` etc. IMO

### Checklist for PRs

### DIFF
--- a/telegram/ext/_prefixhandler.py
+++ b/telegram/ext/_prefixhandler.py
@@ -16,11 +16,11 @@
 #
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
-"""This module contains the CommandHandler and PrefixHandler classes."""
-import re
+"""This module contains the PrefixHandler class."""
+import itertools
 from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, TypeVar, Union
 
-from telegram import MessageEntity, Update
+from telegram import Update
 from telegram._utils.defaultvalue import DEFAULT_TRUE
 from telegram._utils.types import SCT, DVInput
 from telegram.ext import filters as filters_module
@@ -33,20 +33,44 @@ if TYPE_CHECKING:
 RT = TypeVar("RT")
 
 
-class CommandHandler(BaseHandler[Update, CCT]):
-    """BaseHandler class to handle Telegram commands.
+class PrefixHandler(BaseHandler[Update, CCT]):
+    """BaseHandler class to handle custom prefix commands.
 
-    Commands are Telegram messages that start with ``/``, optionally followed by an ``@`` and the
-    bot's name and/or some additional text. The handler will add a :obj:`list` to the
-    :class:`CallbackContext` named :attr:`CallbackContext.args`. It will contain a list of strings,
-    which is the text following the command split on single or consecutive whitespace characters.
+    This is an intermediate handler between :class:`MessageHandler` and :class:`CommandHandler`.
+    It supports configurable commands with the same options as :class:`CommandHandler`. It will
+    respond to every combination of :attr:`prefix` and :attr:`command`. It will add a :obj:`list`
+    to the :class:`CallbackContext` named :attr:`CallbackContext.args`. It will contain a list of
+    strings, which is the text following the command split on single or consecutive whitespace
+    characters.
+
+    Examples:
+
+        Single prefix and command:
+
+        .. code:: python
+
+            PrefixHandler("!", "test", callback)  # will respond to '!test'.
+
+        Multiple prefixes, single command:
+
+        .. code:: python
+
+            PrefixHandler(["!", "#"], "test", callback)  # will respond to '!test' and '#test'.
+
+        Multiple prefixes and commands:
+
+        .. code:: python
+
+            PrefixHandler(
+                ["!", "#"], ["test", "help"], callback
+            )  # will respond to '!test', '#test', '!help' and '#help'.
+
 
     By default, the handler listens to messages as well as edited messages. To change this behavior
     use :attr:`~filters.UpdateType.EDITED_MESSAGE <telegram.ext.filters.UpdateType.EDITED_MESSAGE>`
-    in the filter argument.
 
     Note:
-        * :class:`CommandHandler` does *not* handle (edited) channel posts.
+        * :class:`PrefixHandler` does *not* handle (edited) channel posts.
 
     Warning:
         When setting :paramref:`block` to :obj:`False`, you cannot rely on adding custom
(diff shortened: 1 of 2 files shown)


telegram/ext/_prefixhandler.py
```diff
@@ -0,0 +1,185 @@
...
+
+    By default, the handler listens to messages as well as edited messages. To change this behavior
+    use :attr:`~filters.UpdateType.EDITED_MESSAGE <telegram.ext.filters.UpdateType.EDITED_MESSAGE>`
+
+    Note:
+        * :class:`PrefixHandler` does *not* handle (edited) channel posts.
+
+    Warning:
+        When setting :paramref:`block` to :obj:`False`, you cannot rely on adding custom
+        attributes to :class:`telegram.ext.CallbackContext`. See its docs for more info.
```
Poolitzer, line 77: hm, what is this warning doing here? Is it saying that the custom .args attribute which this handler adds wont be there?

Poolitzer (approved):
Question:

telegram/ext/_commandhandler.py
```diff
@@ -98,12 +103,13 @@ def __init__(
         super().__init__(callback, block=block)
 
         if isinstance(command, str):
-            self.command = [command.lower()]
+            commands = frozenset((command.lower(),))
         else:
-            self.command = [x.lower() for x in command]
-        for comm in self.command:
+            commands = frozenset(x.lower() for x in command)
+        for comm in command:
```
harshil21, line 109: if `command` is a string, this will iterate through the string, which is wrong

telegram/ext/_prefixhandler.py
```diff
@@ -0,0 +1,185 @@
...
+
+RT = TypeVar("RT")
+
+
+class PrefixHandler(BaseHandler[Update, CCT]):
+    """BaseHandler class to handle custom prefix commands.
+
+    This is an intermediate handler between :class:`MessageHandler` and :class:`CommandHandler`.
+    It supports configurable commands with the same options as :class:`CommandHandler`. It will
+    respond to every combination of :attr:`prefix` and :attr:`command`. It will add a :obj:`list`
```
harshil21, line 41:
```suggestion
    respond to every combination of :attr:`prefix` and :attr:`commands`. It will add a :obj:`list`
```

telegram/ext/_prefixhandler.py
```diff
@@ -0,0 +1,185 @@
...
+    .. versionchanged:: 20.0
+
+        * Removed the attributes ``command`` and ``prefix``. Instead, the new :attr:`commands`
+          contains all commands that this handler listens to as :class:`frozenset`, which includes
+          the prefixes.
+        * Updating the prefixes and commands this handler listens to is no longer possible.
+
+    Args:
+        prefix (:obj:`str` | Collection[:obj:`str`]):
+            The prefix(es) that will precede :attr:`command`.
```
harshil21, line 88:
```suggestion
            The prefix(es) that will precede :attr:`commands`.
```

telegram/ext/_prefixhandler.py
```diff
@@ -0,0 +1,185 @@
...
+            :class:`telegram.ext.ConversationHandler`.
+        filters (:class:`telegram.ext.filters.BaseFilter`, optional): A filter inheriting from
+            :class:`telegram.ext.filters.BaseFilter`. Standard filters can be found in
+            :mod:`telegram.ext.filters`. Filters can be combined using bitwise
+            operators (``&`` for :keyword:`and`, ``|`` for :keyword:`or`, ``~`` for :keyword:`not`)
+        block (:obj:`bool`, optional): Determines whether the return value of the callback should
+            be awaited before processing the next handler in
+            :meth:`telegram.ext.Application.process_update`. Defaults to :obj:`True`.
+
+    Attributes:
```
harshil21, line 107: missing commands attribute

telegram/ext/_prefixhandler.py
```diff
@@ -0,0 +1,185 @@
...
+
+        * Removed the attributes ``command`` and ``prefix``. Instead, the new :attr:`commands`
+          contains all commands that this handler listens to as :class:`frozenset`, which includes
+          the prefixes.
+        * Updating the prefixes and commands this handler listens to is no longer possible.
+
+    Args:
+        prefix (:obj:`str` | Collection[:obj:`str`]):
+            The prefix(es) that will precede :attr:`command`.
+        command (:obj:`str` | Collection[:obj:`str`]):
```
harshil21, line 89: rename here too?

telegram/ext/_prefixhandler.py
```diff
@@ -0,0 +1,185 @@
...
+        command: SCT[str],
+        callback: HandlerCallback[Update, CCT, RT],
+        filters: filters_module.BaseFilter = None,
+        block: DVInput[bool] = DEFAULT_TRUE,
+    ):
+
+        super().__init__(
+            callback=callback,
+            block=block,
+        )
```
harshil21, line 132:
```suggestion
        super().__init__(callback=callback, block=block)
```
i like it short

harshil21 (changes requested):
tests should be updated + .rst file renaming needed

telegram/ext/_commandhandler.py
```diff
@@ -52,9 +52,15 @@ class CommandHandler(BaseHandler[Update, CCT]):
...
         attributes to :class:`telegram.ext.CallbackContext`. See its docs for more info.
 
+    .. versionchanged:: 20.0
+
+        * Renamed the attribute ``command`` to :attr:`commands`, which now is always a
+          :class:`frozenset`
+        * Updating the commands this handler listens to is no longer possible.
+
     Args:
         command (:obj:`str` | Collection[:obj:`str`]):
```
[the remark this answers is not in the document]
Bibo-Joshi: In either case we only cover either the singular or the plural. IIRC we sticked with the singular in such cases for now, e.g. for filters.user

telegram/ext/_commandhandler.py
```diff
@@ -98,12 +103,13 @@ def __init__(
         super().__init__(callback, block=block)
 
         if isinstance(command, str):
-            self.command = [command.lower()]
+            commands = frozenset((command.lower(),))
```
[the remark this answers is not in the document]
Bibo-Joshi: Does that make a difference? 🤔

telegram/ext/_commandhandler.py
Bibo-Joshi: I really shouldn't try to finish a PR while in a hurry ... Sorry, I thought I had already changed that

telegram/ext/_commandhandler.py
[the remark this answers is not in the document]
harshil21: a  tiny one maybe

telegram/ext/_commandhandler.py
Bibo-Joshi: Turns out I forgot to push a commit …

telegram/ext/_prefixhandler.py
Bibo-Joshi: When you have handlers in multiple groups and you do `context.custom_arg = foo` in a lower group, you cannot rely on `context.custom_arg` being available in the higher group. We have this warning in all handlers.

Issue #3038, opened 2022-05-16
Adjust weekday behavior of `JobQueue.run_daily`


Bibo-Joshi:
# Current situation

Currently, 0 correspondigs to monday and 6 to sunday. This does not match the the behavior of [cron](https://en.wikipedia.org/wiki/Cron), which is the de-facto standard for many scheduling related things in IT. In cron, we have 0=sunday, 6=saturday. Moreover, the APScheduler library, which we currently use as backend for the `JobQueue` will also adapt the sun-sat approach in the next major release 4.0 - cf https://github.com/agronholm/apscheduler/issues/495

# Proposed change

We should make the transition to a sun-sat based numbering scheme. This has the advantages of
* complying with cron
* having one less thing to worry about for users when we switch to APS 4.0 at some point in the future

An implementation could for example use a mapping like [here](https://github.com/agronholm/apscheduler/issues/286#issuecomment-449273964) to pass the correct values to APS.
Accepting the weekdays as `mon, tue` etc instead of integers could also be added as additional support, but is not mandatory IMO.

# A PR for this issue should

* Implement a conversion of the input such that we accept 0=sun, 6=sat, but pass the correct values to APS
* add a `.. versionchanged:: 20.0` directive to the docstring of `run_daily` and the `days` parameter to explain the changed behavior
* Add a unit test to `test_jobqueue.py` that ensures that this works as expected
* Make `JobQueue.run_daily` issue a warning that tells users about the changed behavior and recommends them to double check that they pass the correct days. This warning should be issued via the `warn` function from `telegram._utils.warnings`. The unit test should ensure that this warning is issued correctly. See [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/298c5fab3b8655ebae4416341335fd9c05f2ce5a/tests/test_telegramobject.py#L103-L114) for an example on how warnings can be checked in unit tests.

# Additional Info

We have a [contribution guide](https://python-telegram-bot.readthedocs.io/en/v20.0a0/contributing.html)

jcnecio:
Hi, 

So does the proposed change above mean if I passed in `days=[0,1,2,3]`  in `JobQueue.run_daily` it should resolve to `"sun,mon,tue,wed"` as being passed below?

https://github.com/python-telegram-bot/python-telegram-bot/blob/1f7efe4519e1ee82da684d390b7b108e61067877/telegram/ext/_jobqueue.py#L461

Bibo-Joshi:
[...]

PR #3046, opened 2022-05-18
Aligned JobQueue.run_daily `days` parameter with cron weekdays notation (#3038)

jcnecio:
Changes included in the PR:
- Updated docstring of `JobQueue.run_daily` that the `days` parameter range 0-6 is now sunday-saturday.
- Added a warning message to double check if the parameter is still correct.
- Added a test case to check if the new number scheme is correctly scheduled.
- Updated test cases where the warning should be visible on call.

### Checklist for PRs

### DIFF
--- a/telegram/ext/_jobqueue.py
+++ b/telegram/ext/_jobqueue.py
@@ -28,6 +28,7 @@ from apscheduler.job import Job as APSJob
 from apscheduler.schedulers.asyncio import AsyncIOScheduler
 
 from telegram._utils.types import JSONDict
+from telegram._utils.warnings import warn
 from telegram.ext._extbot import ExtBot
 from telegram.ext._utils.types import JobCallback
 
@@ -419,7 +420,10 @@ class JobQueue:
                 (:obj:`datetime.time.tzinfo`) is :obj:`None`, the default timezone of the bot will
                 be used.
             days (Tuple[:obj:`int`], optional): Defines on which days of the week the job should
-                run (where ``0-6`` correspond to monday - sunday). Defaults to ``EVERY_DAY``
+                run (where ``0-6`` correspond to sunday - saturday). Defaults to ``EVERY_DAY``
+
+                .. versionchanged:: 20.0
+                    Changed day of the week mapping of 0-6 from monday-sunday to sunday-saturday.
             data (:obj:`object`, optional): Additional data needed for the callback function.
                 Can be accessed through :attr:`Job.data` in the callback. Defaults to
                 :obj:`None`.
@@ -447,18 +451,24 @@ class JobQueue:
             queue.
 
         """
+        warn(
+            "Prior to v20.0 the `days` parameter was aligned to that of cron's weekday scheme "
+            "we recommend double checking if the passed value is correct.",
+            stacklevel=2,
+        )
         if not job_kwargs:
             job_kwargs = {}
 
         name = name or callback.__name__
         job = Job(callback=callback, data=data, name=name, chat_id=chat_id, user_id=user_id)
 
+        cron_mapping = ["sun", "mon", "tue", "wed", "thu", "fri", "sat"]
         j = self.scheduler.add_job(
             job.run,
             name=name,
             args=(self.application,),
             trigger="cron",
-            day_of_week=",".join([str(d) for d in days]),
+            day_of_week=",".join([cron_mapping[d] for d in days]),
             hour=time.hour,
             minute=time.minute,
             second=time.second,
--- a/tests/test_jobqueue.py
+++ b/tests/test_jobqueue.py
@@ -303,7 +303,11 @@ class TestJobQueue:
         scheduled_time = job_queue.jobs()[0].next_t.timestamp()
         assert scheduled_time == pytest.approx(expected_time)
 
-    async def test_run_daily(self, job_queue):
+    async def test_run_daily(self, job_queue, recwarn):
+        expected_warning = (
+            "Prior to v20.0 the `days` parameter was aligned to that of cron's weekday scheme we "
+            "recommend double checking if the passed value is correct."
+        )
         delta, now = 1, dtm.datetime.now(pytz.utc)
         time_of_day = (now + dtm.timedelta(seconds=delta)).time()
         expected_reschedule_time = (now + dtm.timedelta(seconds=delta, days=1)).timestamp()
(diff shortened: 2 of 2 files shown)


tests/test_jobqueue.py
```diff
@@ -313,6 +317,27 @@ async def test_run_daily(self, job_queue):
         assert self.result == 1
         scheduled_time = job_queue.jobs()[0].next_t.timestamp()
         assert scheduled_time == pytest.approx(expected_reschedule_time)
+        assert len(recwarn) == 1
+        assert str(recwarn[0].message) == expected_warning
+
+    async def test_run_daily_days_of_week(self, job_queue, recwarn):
```
Bibo-Joshi, line 323:
```suggestion
    @pytest.mark.parametrize('weekday', (0,1,2,3,4,5,6))
    async def test_run_daily_days_of_week(self, job_queue, recwarn, weekday):
```

With this the test will automatically be run for each of the six weekdays so that we can check that the behavior works as expected for all of them :) you can use the `weekday` variable in the test, e.g. as `job_queue.run_daily(…, day=[weekday])`

Bibo-Joshi (changes requested):
Thanks for the nice PR! It looks good overall 👍 The comments I left below are mostly just small stuff :)

telegram/ext/_jobqueue.py
```diff
@@ -419,7 +420,10 @@ async def callback(context: CallbackContext)
                 (:obj:`datetime.time.tzinfo`) is :obj:`None`, the default timezone of the bot will
                 be used.
             days (Tuple[:obj:`int`], optional): Defines on which days of the week the job should
-                run (where ``0-6`` correspond to monday - sunday). Defaults to ``EVERY_DAY``
+                run (where ``0-6`` correspond to sunday - saturday). Defaults to ``EVERY_DAY``
```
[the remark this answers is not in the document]
jcnecio: No worries, I actually thought that there's a constant named EVERY_DAY which is 0-6

telegram/ext/_jobqueue.py
```diff
@@ -447,18 +451,24 @@ async def callback(context: CallbackContext)
             queue.
 
         """
+        warn(
+            "Prior to v20.0 the `days` parameter was aligned to that of cron's weekday scheme "
+            "we recommend double checking if the passed value is correct.",
```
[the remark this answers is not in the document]
jcnecio: Good catch :)

telegram/ext/_jobqueue.py
```diff
@@ -447,18 +451,24 @@ async def callback(context: CallbackContext)
...
+            stacklevel=2,
+        )
         if not job_kwargs:
             job_kwargs = {}
 
         name = name or callback.__name__
         job = Job(callback=callback, data=data, name=name, chat_id=chat_id, user_id=user_id)
 
+        cron_mapping = ["sun", "mon", "tue", "wed", "thu", "fri", "sat"]
```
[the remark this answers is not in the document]
jcnecio: I was having doubts on where to put this, thanks!

tests/test_jobqueue.py
```diff
@@ -313,6 +317,27 @@ async def test_run_daily(self, job_queue):
         assert self.result == 1
         scheduled_time = job_queue.jobs()[0].next_t.timestamp()
         assert scheduled_time == pytest.approx(expected_reschedule_time)
+        assert len(recwarn) == 1
+        assert str(recwarn[0].message) == expected_warning
```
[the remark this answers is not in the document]
jcnecio: tried changing stacklevel to 1 it gave an error wrong stacklevel, 
stacklevel 2 worked 🤷

tests/test_jobqueue.py
jcnecio: I wanted to do this inside the test, but I just cant get a way to get the next item in the jobs list, thanks for this!

also, is there a better way to handle the weekday of the same day? I ended up with this as the offset
```python

        offset = (weekday + 6 - now.weekday()) % 7
        offset = offset if offset > 0 else 7
```

telegram/ext/_jobqueue.py
```diff
@@ -51,6 +51,7 @@ class JobQueue:
     """
 
     __slots__ = ("_application", "scheduler", "_executor")
+    CRON_MAPPING = ["sun", "mon", "tue", "wed", "thu", "fri", "sat"]
```
Bibo-Joshi, line 54:
```suggestion
    _CRON_MAPPING = ("sun", "mon", "tue", "wed", "thu", "fri", "sat")
```
let's make it protected, as it's not intended to be used by users directly :) making it a tuple emphasizes that it's not entended to be changed.

Bibo-Joshi (changes requested):
just one nitpick :) please also merge the master branch into yours - we changed the pre-commit settings a bit so we should make sure that your branch passes the updated version as well

tests/test_jobqueue.py
Bibo-Joshi: I think that's fine. one can probably drop the second line (weekday + day >= 6, now.weekday() <= 6, so the difference will never be negative), but I'm also okay with keeping it. it's just tests 😄

telegram/ext/_jobqueue.py
jcnecio: thanks @Bibo-Joshi  for the review, I learned a lot already in my first open source PR

jcnecio: Does the pr need to be from my forks master?

Bibo-Joshi: mh, no, you don't need a new branch or pr. see the "Resolve any merge conflicts that arise." step in the [contribution guide](https://docs.python-telegram-bot.org/en/v20.0a0/contributing.html#instructions-for-making-a-code-change)

jcnecio: oh, update my branch's history, I pushed it now, luckily no conflicts at all

Bibo-Joshi (approved):

telegram/ext/_jobqueue.py
```diff
@@ -447,6 +453,11 @@ async def callback(context: CallbackContext)
             queue.
 
         """
+        warn(
+            "Prior to v20.0 the `days` parameter was not aligned to that of cron's weekday scheme."
+            "We recommend double checking if the passed value is correct.",
+            stacklevel=2,
+        )
```
Poolitzer, line 460: What about a comment here, reminding us that we remove this in the next major version?

jcnecio: is this enough?
```python
        # TODO: After v20.0, we should remove the this warning.
```

Bibo-Joshi: should be enough IMO, yes :)

Bibo-Joshi:
Thank you for the contribution! :)

PR #3052, opened 2022-05-19
Add webappbot example

Bibo-Joshi:
Static html page should ofc be included in the ptb homepage once this is merged.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -52,6 +52,12 @@ A basic example on how to set up a custom error handler.
 ### [`chatmemberbot.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/chatmemberbot.py)
 A basic example on how `(my_)chat_member` updates can be used.
 
+### [`webappbot.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/webappbot.py)
+A basic example of how [Telegram WebApps](https://core.telegram.org/bots/webapps) can be used.
+Use in combination with [`webappbot.html`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/webappbot.html).
+For your convenience, this file is hosted by the PTB team such that you don't need to host it yourself.
+Uses the [`iro.js`](https://iro.js.org) JavaScript library to showcase a userinterface that is hard to achieve with native Telegram functionality.
+
 ### [`contexttypesbot.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/contexttypesbot.py)
 This example showcases how `telegram.ext.ContextTypes` can be used to customize the `context` argument of handler and job callbacks.
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -56,7 +56,7 @@ A basic example on how `(my_)chat_member` updates can be used.
 A basic example of how [Telegram WebApps](https://core.telegram.org/bots/webapps) can be used.
 Use in combination with [`webappbot.html`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/webappbot.html).
 For your convenience, this file is hosted by the PTB team such that you don't need to host it yourself.
-Uses the [`iro.js`](https://iro.js.org) JavaScript library to showcase a userinterface that is hard to achieve with native Telegram functionality.
+Uses the [`iro.js`](https://iro.js.org) JavaScript library to showcase a user interface that is hard to achieve with native Telegram functionality.
 
 ### [`contexttypesbot.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/contexttypesbot.py)
 This example showcases how `telegram.ext.ContextTypes` can be used to customize the `context` argument of handler and job callbacks.
--- a/examples/webappbot.py
+++ b/examples/webappbot.py
@@ -3,8 +3,8 @@
 # This program is dedicated to the public domain under the CC0 license.
 
 """
-Simple example of Telegram WebApps. The static website for this website is hosted by the PTB team
-for your convenience.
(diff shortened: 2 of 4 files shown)


examples/webappbot.py
```diff
@@ -0,0 +1,75 @@
...
+)
+logger = logging.getLogger(__name__)
+
+
+# Define a few command handlers. These usually take the two arguments update and
+# context.
+async def start(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    """Send a message with a button that opens a the web app."""
+    await update.message.reply_text(
+        "Please press the button below to open the WebApp.",
```
harshil21, line 42:
```suggestion
        "Please press the button below to choose a color from the WebApp.",
```

examples/webappbot.py
```diff
@@ -0,0 +1,75 @@
...
+        reply_markup=ReplyKeyboardMarkup.from_button(
+            KeyboardButton(
+                text="Click Me!",
```
harshil21, line 45:
```suggestion
                text="Open the color picker",
```

examples/README.md
```diff
@@ -52,6 +52,12 @@ A basic example on how to set up a custom error handler.
 ### [`chatmemberbot.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/chatmemberbot.py)
 A basic example on how `(my_)chat_member` updates can be used.
 
+### [`webappbot.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/webappbot.py)
+A basic example of how [Telegram WebApps](https://core.telegram.org/bots/webapps) can be used.
+Use in combination with [`webappbot.html`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/webappbot.html).
+For your convenience, this file is hosted by the PTB team such that you don't need to host it yourself.
+Uses the [`iro.js`](https://iro.js.org) JavaScript library to showcase a userinterface that is hard to achieve with native Telegram functionality.
```
harshil21, line 59:
```suggestion
Uses the [`iro.js`](https://iro.js.org) JavaScript library to showcase a user interface that is hard to achieve with native Telegram functionality.
```

examples/webappbot.py
```diff
@@ -0,0 +1,75 @@
...
+from telegram.ext import Application, CommandHandler, ContextTypes, MessageHandler, filters
+
+# Enable logging
+logging.basicConfig(
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s", level=logging.INFO
+)
+logger = logging.getLogger(__name__)
+
+
+# Define a few command handlers. These usually take the two arguments update and
```
harshil21, line 37:
```suggestion
# Define a `start` command handler. These usually take the two arguments update and
```

examples/webappbot.py
```diff
@@ -0,0 +1,75 @@
...
+                text="Click Me!",
+                web_app=WebAppInfo(url="https://python-telgeram-bot.org/static/webappbot"),
+            )
+        ),
+    )
+
+
+async def web_app_data(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    """Print the received data and remove the button."""
+    data = json.loads(update.effective_message.web_app_data.data)
```
harshil21, line 54:
```suggestion
    # Here we use json to load the data, since we sent the data as a string (see webapp.html)
    data = json.loads(update.effective_message.web_app_data.data)
```

harshil21 (changes requested):
nice! The website isn't exactly mobile friendly though, you have to pull up and make it 'expanded' to see the rest of the interface, which we could do without.

Bibo-Joshi:
> The website isn't exactly mobile friendly though, you have to pull up and make it 'expanded' to see the rest of the interface, which we could do without. 

Good point. I could either reduce the padding around the color picker or have the app automatically expanded via Telegram.WebApp.expand(). Which do you prefer?

examples/webappbot.py
```diff
@@ -0,0 +1,75 @@
...
+# Define a few command handlers. These usually take the two arguments update and
+# context.
+async def start(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    """Send a message with a button that opens a the web app."""
+    await update.message.reply_text(
+        "Please press the button below to open the WebApp.",
+        reply_markup=ReplyKeyboardMarkup.from_button(
+            KeyboardButton(
+                text="Click Me!",
+                web_app=WebAppInfo(url="https://python-telgeram-bot.org/static/webappbot"),
```
harshil21, line 46:
```suggestion
                web_app=WebAppInfo(url="https://python-telegram-bot.org/static/webappbot"),
```

Bibo-Joshi:
I tried around for quite a while to get the container to adjust to the screen size or at least just be smaller, but I failed :/ So I just made the webapp window expand … This can ofc always be fine tuned later on

harshil21 (approved):
Okay. will need backport to v13 after merge

Bibo-Joshi:
Also works fine with the static file hosted at ptb.org/statis/webappbot. merging.

Scolymus:
Is this example working?

I'm trying to run it but I recieve an error "AttributeError: type object 'ContextTypes' has no attribute 'DEFAULT_TYPE'". In fact, python does not recognize "from telegram import __version_info__" and I had to delete the check for version type. In any case, I have a doubt about that version. In the python file it says it needs "20, 0, 0, "alpha", 1". However, the last pre version avalaible in pip install is 20.0a0. The branch for webapi does not exist neither...

Poolitzer:
If you dont delete the try except, it should/would have told you that the installed PTB version does in fact not correspond with the one you need to run the example. That one isn't released yet, so you can either install the latest "version" of this library from master or you take care of the errors (I think it is CallbackContext instead of ContextTypes).

PR #3053, opened 2022-05-19
Remove `InputFile.is_image`

### LINKED ISSUE #3020: Deprecation: `imghdr` module is being deprecated in Python 3.11
The `imghdr` module is being deprected in `3.11` and will be removed in `3.13`. See [PEP 594](https://peps.python.org/pep-0594/#imghdr) for details. 

We currently use it only in one area: https://github.com/python-telegram-bot/python-telegram-bot/blob/be8f4f7aad9c3ded333950d00ecde57dbdda59c2/telegram/_files/inputfile.py#L108

Alternatives would be:
1. Just use the `mimetype` module? We were using only the `mimetype` module until [@b83a659](https://github.com/python-telegram-bot/python-telegram-bot/commit/b83a659) where we used a 'hack' to detect the type from the byte stream. This was then [replaced](https://github.com/python-telegram-bot/python-telegram-bot/commit/5dc1e4cac19cdd34d7cad4688afa0277a0c8a436) by the `imghdr` module which did the same thing better.
2. Use a library called [filetype](https://github.com/h2non/filetype.py) to detect images. This was [recommended](https://peps.python.org/pep-0594/#deprecated-modules) by the PEP.
3. Just copy the python [source](https://github.com/python/cpython/blob/3.11/Lib/imghdr.py) of the `imghdr` module into ours?

Bibo-Joshi:
closes #3020

### Checklist for PRs


telegram/_files/inputfile.py
```diff
@@ -82,39 +85,13 @@ def __init__(
         ):
```
harshil21, line 85: above this line (line 84), when can `obj.name` be an `int`?

telegram/_files/inputfile.py
```diff
@@ -35,7 +34,11 @@ class InputFile:
     """This object represents a Telegram InputFile.
 
     .. versionchanged:: 20.0
-        The former attribute ``attach`` was renamed to :attr:`attach_name`.
+
+        * The former attribute ``attach`` was renamed to :attr:`attach_name`.
+        * Method ``is_image`` was removed. If you pass :obj:`bytes` to :paramref:`obj` and would
+          like to have the mime type automatically guessed, please pass :paramref:`filename`
```
harshil21, line 40: isn't the `filename` already guessed from [earlier](https://github.com/python-telegram-bot/python-telegram-bot/blob/64de8966adc69da230d5206f6d14b61c499373e4/telegram/_files/inputfile.py#L86)?

telegram/_files/inputfile.py
```diff
@@ -35,7 +34,11 @@ class InputFile:
     """This object represents a Telegram InputFile.
 
     .. versionchanged:: 20.0
-        The former attribute ``attach`` was renamed to :attr:`attach_name`.
+
+        * The former attribute ``attach`` was renamed to :attr:`attach_name`.
+        * Method ``is_image`` was removed. If you pass :obj:`bytes` to :paramref:`obj` and would
```
harshil21, line 39: If I understand the python [docs](https://docs.python.org/3.12/library/mimetypes.html#mimetypes.guess_type) correctly, we can just pass `guess_type(obj)` since it says it supports a path like object from python 3.8 onwards? That perhaps increases accuracy?

telegram/_files/inputfile.py
Bibo-Joshi: only if you pass a file-like object, i.e. `open('file', 'rb')` instead of plain bytes (`Path(…).read_bytes()`)

telegram/_files/inputfile.py
Bibo-Joshi: Don't think that this would work. [path-like](https://docs.python.org/3.12/glossary.html#term-path-like-object) != [file-like](https://docs.python.org/3/glossary.html?highlight=glossary#term-file-object)

telegram/_files/inputfile.py
Bibo-Joshi: Apparently that's the case for `InputFile(subprocess.Popen(…).stdout)` - see `test_subprocess_pip` in `test_inputfile`

telegram/_files/inputfile.py
harshil21: I just tested passing a `Path` object to `guess_type` and it works. Probably cause a `path` like object extends from `PurePath` like it said in the [docs](https://docs.python.org/3.12/library/os.html#os.PathLike)

harshil21: ah but we have it open()ed already, so this won't work hmm

Bibo-Joshi: my gut feeling is that `InputFile` should represent the raw data to be uploaded 🤔 We could try and accept path objects for the bot methods/`parse_file_input` though, but that's out of scope for this PR IMO. See also https://github.com/python-telegram-bot/python-telegram-bot/projects/7#card-79049723

harshil21 (approved):

PR #2741, opened 2021-10-21
Update _message.py - issue #2606

nivramam:
closes #2606

### Checklist for PR
- [x] modified `message.photo = None` instead of an empty list `[]`
- [ ] Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)
- [x] Created new or adapted existing unit tests
- [ ] Added myself alphabetically to `AUTHORS.rst` (optional)

PR #3057, opened 2022-05-20
Empty lists

Bibo-Joshi:
Addresses https://github.com/python-telegram-bot/python-telegram-bot/pull/2741#issuecomment-1001740609 and the respective item of #2633.

I also made some smaller code changes, hence this is a standalone PR instead of being included in the doc fixes pr.
In addition to updating the docs of `Message`, I also made corresponding changes to `Poll` and `Game`. Note that there are also classes like `InlineQueryResult*` which have a `caption_entities` attribute. However, those classes are used as input for API methods and having those attributes always be list would make problems - we'd have to somehow decide when to pass an empty list to TG and when to skip the parameter. So I left them as is.

### Checklist for PRs

### DIFF
--- a/telegram/_games/game.py
+++ b/telegram/_games/game.py
@@ -63,8 +63,9 @@ class Game(TelegramObject):
             game message. Can be automatically edited to include current high scores for the game
             when the bot calls :meth:`telegram.Bot.set_game_score`, or manually edited
             using :meth:`telegram.Bot.edit_message_text`.
-        text_entities (List[:class:`telegram.MessageEntity`]): Optional. Special entities that
+        text_entities (List[:class:`telegram.MessageEntity`]):  Special entities that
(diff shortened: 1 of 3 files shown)


telegram/_games/game.py
```diff
@@ -63,8 +63,9 @@ class Game(TelegramObject):
             game message. Can be automatically edited to include current high scores for the game
             when the bot calls :meth:`telegram.Bot.set_game_score`, or manually edited
             using :meth:`telegram.Bot.edit_message_text`.
-        text_entities (List[:class:`telegram.MessageEntity`]): Optional. Special entities that
+        text_entities (List[:class:`telegram.MessageEntity`]):  Special entities that
```
harshil21, line 66:
```suggestion
        text_entities (List[:class:`telegram.MessageEntity`]): Special entities that
```

harshil21 (approved):

PR #3065, opened 2022-05-23
Feat: Don't add signal handlers on Windows

Poolitzer:
### Checklist for PRs

### DIFF
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -548,7 +548,7 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AbstractAsyncContextManager)
         allowed_updates: List[str] = None,
         drop_pending_updates: bool = None,
         close_loop: bool = True,
-        stop_signals: DVInput[Sequence[int]] = DEFAULT_NONE,
+        stop_signals: ODVInput[Sequence[int]] = DEFAULT_NONE,
     ) -> None:
         """Convenience method that takes care of initializing and starting the app,
         polling updates from Telegram using :meth:`telegram.ext.Updater.start_polling` and
@@ -647,7 +647,7 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AbstractAsyncContextManager)
         ip_address: str = None,
         max_connections: int = 40,
         close_loop: bool = True,
-        stop_signals: DVInput[Sequence[int]] = DEFAULT_NONE,
+        stop_signals: ODVInput[Sequence[int]] = DEFAULT_NONE,
     ) -> None:
         """Convenience method that takes care of initializing and starting the app,
         polling updates from Telegram using :meth:`telegram.ext.Updater.start_webhook` and
@@ -737,7 +737,7 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AbstractAsyncContextManager)
     def __run(
         self,
         updater_coroutine: Coroutine,
-        stop_signals: DVInput[Sequence[int]],
+        stop_signals: ODVInput[Sequence[int]],
         close_loop: bool = True,
     ) -> None:
         # Calling get_event_loop() should still be okay even in py3.10+ as long as there is a
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -548,7 +548,7 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AbstractAsyncContextManager)
         allowed_updates: List[str] = None,
         drop_pending_updates: bool = None,
         close_loop: bool = True,
-        stop_signals: Optional[Sequence[int]] = None,
+        stop_signals: DVInput[Sequence[int]] = DEFAULT_NONE,
     ) -> None:
         """Convenience method that takes care of initializing and starting the app,
         polling updates from Telegram using :meth:`telegram.ext.Updater.start_polling` and
@@ -616,10 +616,6 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AbstractAsyncContextManager)
         def error_callback(exc: TelegramError) -> None:
             self.create_task(self.process_error(error=exc, update=None))
 
-        if stop_signals is None:
-            if platform.system() != "windows":
-                stop_signals = (signal.SIGINT, signal.SIGTERM, signal.SIGABRT)
-
         return self.__run(
             updater_coroutine=self.updater.start_polling(
                 poll_interval=poll_interval,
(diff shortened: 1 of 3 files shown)


telegram/ext/_application.py
```diff
@@ -711,6 +716,10 @@ def run_webhook(
                 "Application.run_webhook is only available if the application has an Updater."
             )
 
+        if stop_signals is None:
+            if platform.system() != "windows":
```
Bibo-Joshi, line 720:
```suggestion
        if stop_signals is None and platform.system() != "windows":
```
see deepsource

tests/test_application.py
```diff
@@ -1680,3 +1680,28 @@ async def raise_method(*args, **kwargs):
...
         assert len(recwarn) == 0
+
+    @pytest.mark.dev
+    def test_signal_handlers(self, app, monkeypatch):
+        # this test should make sure that signal handlers are set by default on Linux + Mac,
+        # and not on Windows.
+
+        def signal_handler_test(*args, **kwargs):
+            # args[0] is the signal, [1] the callback
+            assert args[0] in (signal.SIGINT, signal.SIGTERM, signal.SIGABRT)
```
Bibo-Joshi, line 1691: This doesn't test anything on windows, because it's never called. suggestion: something like `received_signals.append(args[0])` and then later on

```
if windows:
    assert received_signals == []
else:
    assert received_signals = [signal.SIGINT, signal.SIGTERM, signal.SIGABRT]
```

tests/test_application.py
```diff
@@ -1680,3 +1680,28 @@ async def raise_method(*args, **kwargs):
...
+
+        loop = asyncio.get_event_loop()
+        monkeypatch.setattr(loop, "add_signal_handler", signal_handler_test)
+
+        async def abort_app(context):
+            raise KeyboardInterrupt
```
Bibo-Joshi, line 1697: nice 👍 then we can maybe even refactor the other tests to also run on windows with this trick :)

Bibo-Joshi (changes requested):

tests/test_application.py
Poolitzer: I am fine with that, this is supposed to test the default. If the raise test later fails at one CI run, we know we can switch the defaults back.

Bibo-Joshi: the problem is that it wouldn't fail if we pass a signal on windows because the signal would still be one of the ones that you specified here.

Poolitzer: no but then it would raise the error, as it does below...

Bibo-Joshi: the current CI failures are in `test_run_stop_signal_warning_windows` which expects a warning to be issues on windows and has to be adapted. In the current state doing `run(stop_signal=(Signal.SIGINT,))` would not raise an exception because you mocked the method that raises `NotImplementedError`.

Bibo-Joshi: > It does not though? It expects a totally different warning? And it gets it, but for some reason expects it to not be there?

Fair enough: It expects `stop_signal=None` *not* to issue a warning, but it still does due to https://github.com/python-telegram-bot/python-telegram-bot/pull/3065#discussion_r879858634.

tests/test_application.py
```diff
@@ -1680,3 +1680,28 @@ async def raise_method(*args, **kwargs):
                 app.run_webhook(close_loop=False, stop_signals=None)
 
         assert len(recwarn) == 0
+
+    @pytest.mark.dev
```
harshil21, line 1684: to remove

harshil21 (changes requested):
nice. hinrich reviewed pretty much everything, so only one comment from me-

tests/test_application.py
Poolitzer: Wupps.

tests/test_application.py
Bibo-Joshi: this is still open :) i.e.

* `test_run_stop_signal_warning_windows` needs to be adapted and 
* IMO it would still be good to explicitly check that we don't pass signals on windows

Bibo-Joshi (changes requested):
Thanks for the updates! Can you add tests for the missing coverage?

telegram/ext/_application.py
```diff
@@ -736,17 +737,21 @@ def _raise_system_exit() -> NoReturn:
...
         # See the docs of get_event_loop() and get_running_loop() for more info
         loop = asyncio.get_event_loop()
 
+        if stop_signals is DEFAULT_NONE and platform.system() != "Windows":
+            stop_signals = (signal.SIGINT, signal.SIGTERM, signal.SIGABRT)
+
         try:
-            for sig in stop_signals or []:
-                loop.add_signal_handler(sig, self._raise_system_exit)
+            if not isinstance(stop_signals, DefaultValue):
```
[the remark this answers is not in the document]
Poolitzer: Im sorry, do you mean in tests? because the default is very much tested?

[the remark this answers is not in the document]
Bibo-Joshi: well, now codecov is happy …

tests/test_application.py
```diff
@@ -1680,3 +1680,37 @@ async def raise_method(*args, **kwargs):
...
+                # this should be empty, but still called once.
+                assert args == []
+
+            monkeypatch.setattr(loop, "add_signal_handler", signal_handler_test_windows)
+            # this should run through, and we expect the signals to be empty
+            app.job_queue.run_once(abort_app, 2)
+            # pass false so it is not default
+            app.run_webhook(
+                port=49152, webhook_url="example.com", close_loop=False, stop_signals=False
+            )
```
Bibo-Joshi, line 1716: sorry, but this still doesn't do it IMO. `stop_signals=False` doesn't make sense because we do in fact want to test the default case. Moreover, `add_signal_handler` is never called (as intended), so the `assert` in the mocked function doesn't test anything. Granted, it raises an AssertionError if a code change accidently passes a signal on windows (only without `stop_signal=False` though), but that's really hard to read from this test. Additionally, it feels unncessary to have the above tests do nothing on windows (which again is not obvious from the tests) and only this if-clause actually test something.

What speaks against implementing the test as suggested in https://github.com/python-telegram-bot/python-telegram-bot/pull/3065/files#r879864569 ?

Poolitzer: > What speaks against implementing the test as suggested 

me forgetting you suggested that.

tests/test_application.py
Bibo-Joshi: it's back :D

Poolitzer: its like weed.

Poolitzer:
I did yup

PR #3066, opened 2022-05-25
Add backslash character to be escaped in escape_markdown()

rikkaneko:
### Checklist for PRs

#### This PR just add the backslash character `\` to the `escape_chars` in `telegram/helpers.py`. This is my first PR too :)  

I found that Telegram will return error if the message text contain the backslash character `\`, in **Markdown V2** mode.  
`escape_markdown()` does not handle this special symbol.  Backslash character `\` **does not need** to be escaped in V1.
For example, this sample Markdown V2 link `*\\\*` would become `\*\\\\*` after `escape_markdown()`.  However, Telegram will reply with error since the `\` is incorrectly escaped. The correct result should be `\*\\\\\\\*` with escaped `\`.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -89,6 +89,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Poolitzer <https://github.com/Poolitzer>`_
 - `Pranjalya Tiwari <https://github.com/Pranjalya>`_
 - `Rahiel Kasim <https://github.com/rahiel>`_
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
Hi. This has already been taken care of in #3055 - note that line 66 now starts with a backslash. I should have mentioned that [in the chat](https://t.me/pythontelegrambotgroup/593634?thread=592435), sorry 😬

rikkaneko:
Oh, I do not see this update. I will make another PR if found other bug or improvement.

PR #3068, opened 2022-05-26
Fix a bug with non-blocking entry points

Bibo-Joshi:
See https://t.me/pythontelegrambotgroup/595029?thread=595002 for discussion. The problem is that `_update_state` doesn't do anything if the new state is `None` (because `return None` means "stay in the current state") so that `_conversations.get(key)` returns the `PendingState`.

### Checklist for PRs

### DIFF
--- a/telegram/ext/_conversationhandler.py
+++ b/telegram/ext/_conversationhandler.py
@@ -708,8 +708,13 @@ class ConversationHandler(BaseHandler[Update, CCT]):
             # check if future is finished or not
             if state.done():
                 res = state.resolve()
-                self._update_state(res, key)
-                state = self._conversations.get(key)
+                if res is None:
+                    # Special case if the error was raised in a non-blocking entry-point
+                    self._conversations.pop(key, None)
+                    state = None
+                else:
+                    self._update_state(res, key)
+                    state = self._conversations.get(key)
 
             # if not then handle WAITING state instead
             else:
--- a/telegram/ext/_conversationhandler.py
+++ b/telegram/ext/_conversationhandler.py
@@ -107,7 +107,6 @@ class PendingState:
             _logger.exception(
                 "Task function raised exception. Falling back to old state %s",
                 self.old_state,
-                exc_info=exc,
             )
             return self.old_state
 
@@ -717,7 +716,6 @@ class ConversationHandler(BaseHandler[Update, CCT]):
                     self._conversations.pop(key, None)
                     state = None
                 else:
-                    print("resolved stat", res)
                     self._update_state(res, key)
                     state = self._conversations.get(key)
 
--- a/telegram/ext/_conversationhandler.py
+++ b/telegram/ext/_conversationhandler.py
@@ -711,8 +711,15 @@ class ConversationHandler(BaseHandler[Update, CCT]):
 
             # check if future is finished or not
             if state.done():
-                state = state.resolve()
-                self._update_state(state, key)
+                res = state.resolve()
+                # Special case if an error was raised in a non-blocking entry-point
+                if state.old_state is None and state.task.exception():
+                    self._conversations.pop(key, None)
+                    state = None
+                else:
+                    print("resolved stat", res)
+                    self._update_state(res, key)
+                    state = self._conversations.get(key)
 
             # if not then handle WAITING state instead
             else:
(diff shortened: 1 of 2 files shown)


telegram/ext/_conversationhandler.py
```diff
@@ -708,8 +708,13 @@ def check_update(self, update: object) -> Optional[_CheckUpdateType]:
             # check if future is finished or not
             if state.done():
                 res = state.resolve()
-                self._update_state(res, key)
-                state = self._conversations.get(key)
+                if res is None:
+                    # Special case if the error was raised in a non-blocking entry-point
```
harshil21, line 712: so this bug only happens for non blocking entry points and not for handlers in states or fallbacks?

Bibo-Joshi: Yes. if we're already in a non-None state, then `self._update_state(res, key)` works as expected even if `res` is the old state. We also have a test case for that (right above the newly added one) which covers that. unfortunately it did not cover exceptions in non-blocking entrypoints.

harshil21 (approved):
good catch by the OP

CedricFauth:
I tested out the current release 20.0a0. Every non-blocking callback inside state or fallback that returns None has that issue not only entry points I think. Maybe that is already fixed somewhere else. Currently, when I am returning None in any state I get the PendingState... error.

Bibo-Joshi:
@CedricFauth thanks for the info! I've pushed more changes that should address that. Could you check if those fix the problems for you? You can install from this branch via `pip install git+https://github.com/python-telegram-bot/python-telegram-bot.git@ch-entry-point-bug`.

CedricFauth:
@Bibo-Joshi Yes but no :/ You fixed my previous problem. Returning None now works inside the conversation handler but a new problem occurred: 
I am unable to leave a conversation (return ConversationHandler.END). The next time I send a message to the bot (after ending a conversation) it says: `Selecting conversation (..., ...) with state -1` (which is the END state so the bot should have left the conversation before but does not). This ends in my fallback handler getting called every time I send a message. The user is not able to do anything because the conversation stays in state -1 forever.
This is only the case when using `ConversationHandler(block=False)` and the current branch `ch-enrty-point-bug`. In the pre-release it never happened.
Hope I could help you find the problem.

Bibo-Joshi:
Yeah, I should have properly run the tests before pushing 🤦‍♂️ hopefully it's now better :)

CedricFauth:
Seems to work now, thanks!

harshil21 (commented):
So I was testing all combinations of `block=False/True` of `ConversationHandler` with the handlers in `entry_points`, `states` and `fallbacks`, and noticed that when `ConvHandler(..., block=False)`, the exception thrown from one entry point isn't discarded correctly, and when another entry point is run, the same exception is printed again.

MRE: https://nekobin.com/modoxayexo

Steps to reproduce:
1. Run bot, send `/start`. You'll see `in start` printed along with the exception.
2. Now send `/start2`. You'll see the same exception printed again, and then `in start2` is printed to console.
3. State is correctly switched to `GENDER` though.

Expected behavior:
In step 2, the exception should not be printed again. This is indeed the behaviour when `ConvHandler(..., block=True)`.

Bibo-Joshi:
> State is correctly switched to GENDER though.

Just to double check: your observation is only about error prints & not about wrong behavior right?

The difference is the heading of the traceback: The first (triggered by `/start`) will say ` No error handlers are registered, logging exception.`. If you have a custom error handler, this one won't show (unless your error handler explicitly logs it again).
The second one (striggered by the next update) will say `Task function raised exception. Falling back to old state None`, which is logged by `PendingState.resolve()`. IMO those two long entries serve different purposes: The first informing you about an error in the callback & the second one about what's going on state-wise in the ConversationHandler. the traceback being printed twice may surely be seen as a bit overly verbose in the default case, though I guess it's not really harmful either.

telegram/ext/_conversationhandler.py
```diff
@@ -708,8 +712,14 @@ def check_update(self, update: object) -> Optional[_CheckUpdateType]:
...
             if state.done():
                 res = state.resolve()
-                self._update_state(res, key)
-                state = self._conversations.get(key)
+                # Special case if an error was raised in a non-blocking entry-point
+                if state.old_state is None and state.task.exception():
+                    self._conversations.pop(key, None)
+                    state = None
+                else:
+                    print("resolved stat", res)
```
harshil21, line 720: to remove

harshil21 (approved):
Did some more testing, and everything seems to run fine now

Issue #3054, opened 2022-05-20
[FEATURE] Maintain ujson support.


MiguelX413:
### What kind of feature are you missing? Where do you notice a shortcoming of PTB?

On a bot that is used enough, the workload of encoding and decoding json when using the telegram API may be a relevant factor to bot performance.

### Describe the solution you'd like

Maintaining support for ujson as an optional dependency to accelerate json decoding and encoding.

### Additional context

I am making this based on https://github.com/python-telegram-bot/python-telegram-bot/pull/3037#issuecomment-1132516188

annihilatorrrr:
It's faster than the traditional json of python as i noticed in some of the operations, i did used ujson all over the repo and if it's available in the machine why shouldn't we use it.

WisdomPill:
why not use orjson? it should be even faster afaik 

Have a look at the [repo](https://github.com/ijl/orjson)

MiguelX413:
> why not use orjson? it should be even faster afaik
> 
> Have a look at the [repo](https://github.com/ijl/orjson)

That is addressed in https://github.com/python-telegram-bot/python-telegram-bot/pull/3037#issue-1237599033 @WisdomPill

PR #3072, opened 2022-05-30
Make json en-/decoding customizable

Bibo-Joshi:
There has already been some feed back on #3037 in #3054, so if we do an a1 release soonish, then IMO it's probably good to make the json logic customizable already.

This PR does the minimal effort solution for that, i.e. make the payload-parsing method of `BaseRequest` public. IMO making the raw parameters available (instead of the `to_dict()` versions) can be done at a later stage if there is a request for it. For now the use case should be relatively small as `TGObject` classes are no dataclasses yet.

### Checklist for PRs

### DIFF
--- a/telegram/request/_baserequest.py
+++ b/telegram/request/_baserequest.py
@@ -68,6 +68,11 @@ class BaseRequest(
         finally:
             await request_object.shutdown()
 
+    Tip:
+        JSON encoding and decoding is done with the standard libraries :mod:`json` by default.
+        To use a custom library for this, you can override :meth:`parse_json_payload` and implement
+        custom logic to encode the keys of :attr:`telegram.request.RequestData.parameters`.
+
     .. versionadded:: 20.0
     """
 
@@ -163,7 +168,7 @@ class BaseRequest(
             connect_timeout=connect_timeout,
             pool_timeout=pool_timeout,
         )
-        json_data = self._parse_json_response(result)
+        json_data = self.parse_json_payload(result)
         # For successful requests, the results are in the 'result' entry
         # see https://core.telegram.org/bots/api#making-requests
         return json_data["result"]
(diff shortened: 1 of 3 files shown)


telegram/request/_requestdata.py
```diff
@@ -64,6 +64,11 @@ def parameters(self) -> Dict[str, Union[str, int, List, Dict]]:
     def json_parameters(self) -> Dict[str, str]:
         """Gives the parameters as mapping of parameter name to the respective JSON encoded
         value.
+
+        Tip:
+            By default, this property uses the standard libraries :func:`json.dumps`.
+            To use a custom library for JSON encoding, you can directly encode the keys of
+            :attr:`parameters` - note that string valued keys should not be JSON encoded.
```
harshil21, line 71: ah clever, so instead of making `RequestParameter` public, just tell the users to encode `self.parameters` 👍🏽

telegram/request/_requestdata.py
```diff
@@ -96,7 +101,13 @@ def parametrized_url(self, url: str, encode_kwargs: Dict[str, Any] = None) -> st
 
     @property
     def json_payload(self) -> bytes:
-        """The parameters as UTF-8 encoded JSON payload."""
+        """The parameters as UTF-8 encoded JSON payload.
+
```
harshil21, line 105:
```suggestion
        """The :attr:`json_parameters` as UTF-8 encoded JSON payload.
```
?

Bibo-Joshi: I'll link `parameters` - makes it a bot more clear what's going on IMO :)

telegram/request/_baserequest.py
```diff
@@ -68,6 +68,11 @@ class BaseRequest(
         finally:
             await request_object.shutdown()
 
+    Tip:
+        JSON encoding and decoding is done with the standard libraries :mod:`json` by default.
```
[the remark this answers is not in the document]
Bibo-Joshi: damn you, search & replace ✊

harshil21 (approved):

PR #3077, opened 2022-06-01
Feature/convinience message property

### LINKED ISSUE #3075: [FEATURE] Add convenience property `id` to `Message`
## Current situation

Telegram currently offers attribute `message_id` to `Message`. But this is inconsistent with the rest of their API, e.g. `Chat.id`, `User.id`, `InlineQuery.id`, etc. We can offer a convenience property `Message.id` which will make it more consistent and make it less prone to errors, e.g. when someone forgets it's `message_id` and not `id`. 

## Proposed change

Add the property `id` to `Message` which just returns `self.message_id`. A similar example can be found [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/_message.py#L570-L573).

## A PR for this issue should

- Add a docstring along with a `.. versionadded:: 20.0` directive to the function.
- Add an unit test to `test_message.py` that ensures that this works as expected and is the same as `Message.message_id`.

## Additional Info

We have a [contribution guide](https://python-telegram-bot.readthedocs.io/en/v20.0a0/contributing.html)

dglitxh:
### Checklist for PRs

closes #3075

### DIFF
--- a/telegram/_message.py
+++ b/telegram/_message.py
@@ -576,7 +576,7 @@ class Message(TelegramObject):
     def id(self) -> int:  # pylint: disable=invalid-name
         """
         :obj:`int`: Shortcut for :attr:`message_id`.
-        
+
             .. versionadded:: 20.0
         """
         return self.message_id
--- a/telegram/_message.py
+++ b/telegram/_message.py
@@ -575,11 +575,10 @@ class Message(TelegramObject):
     @property
     def id(self) -> int:  # pylint: disable=invalid-name
         """
-
         .. versionchanged:: 20.0
-            :obj: `int` : Convenience property. This is to reduce errors when users
-            use Message.id instead of Message.message_id which is inconsistent with most
-            of the other properties which follow a similar pattern.
+          :obj: `int` : Convenience property. This is to reduce errors when users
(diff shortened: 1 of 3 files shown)


telegram/_message.py
```diff
@@ -572,6 +572,17 @@ def chat_id(self) -> int:
...
         return self.chat.id
 
+    @property
+    def id(self) -> int:  # pylint: disable=invalid-name
+        """
+        .. versionchanged:: 20.0
+          :obj: `int` : Convenience property. This is to reduce errors when users
+          use Message.id instead of Message.message_id which is inconsistent with most
+          of the other properties which follow a similar pattern.
+
```
harshil21, line 582: no need to mention the reason for this addition imo. We can just have something like:

```suggestion
        :obj:`int`: Shortcut for :attr:`message_id`.
        
            .. versionadded:: 20.0
```

harshil21 (commented):
Thanks for the quick PR! Just left one comment below-

telegram/_message.py
dglitxh: That's great! I was having a bit of an issue with that.

harshil21:
pre-commit.ci run

PR #3090, opened 2022-06-10
Mermaid example diagrams

cyc8:
Add Mermaid diagrams for better visualization of the examples and to make use of version control. It will be also easier to edit the diagram if the examples change.


Bibo-Joshi:
Hi. Thanks for the PR! Having the state diagram in a text-like format would indeed be nice. Can you explain how one would generate pictures with mermaid and/or how one could embedd those diagrams into the respective examples pages (e.g. [here](http://localhost:63342/python-telegram-bot/docs/build/html/examples.conversationbot.html#state-diagram))?
So far I haven't worked much with diagram-generation-software, although one of our emeritus maintainers suggested `graphviz` on a few occasions …

cyc8:
Hi @Bibo-Joshi, I've used Mermaid since it is supported and rendered in Github in the Markdown view. To include it is as easy as having a code block like ` ```mermaid ---- mermaid code here ---- ``` ` You can read more about it in the [Github Blog Post](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/). Mermaid is based on JavaScript, I hope that is fine.
- The diagram is a flow chart, I've included a line to link the [documentation of flow charts](https://mermaid-js.github.io/mermaid/#/flowchart) 
- To include it on the documentation site you can make a request to the Mermaid API or host it by yourself, [here is the full documentation ](https://mermaid-js.github.io/mermaid/#/n00b-gettingStarted?id=_3-calling-the-javascript-api)

Poolitzer:
The sphinx plugin should be enough, no need to self host an API.

harshil21 (commented):
Looks great, thanks for the PR. 

> Can you explain how one would generate pictures

I can see that we can download it as svg/png at https://mermaid.live

> The sphinx plugin should be enough

Agreed. Other options would be to use svg/png, but that would mean updating those files too after changing the mermaid code -- double work.

----
Can you also apply similar changes by integrating the [sphinxcontrib-mermaid](https://github.com/mgaitan/sphinxcontrib-mermaid) sphinx extension?

You'll have to add it to the `conf.py` and `requirements-docs.txt`. I guess we can just use these files as the reference by doing: `.. mermaid:: examples/conversationbot.md` in the `docs/examples.conversationbot.rst` file, and then build the docs locally by following our [contribution guide](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/.github/CONTRIBUTING.rst).

Bibo-Joshi:
If we can integrate the diagrams into sphinx directly, then we should do that and ditch images completely, I agree. Just for completeness: I see that sphinx has built-in support for graphivz, i.e. we wouldn't need a plugin. I'm fine with using a plugin, though.

cyc8:
I've installed the sphinx mermaid extension and embedded it successfully in the docs. However, there are a few things to be aware of:
- The dark mode makes the lines hard to read. This is a known issue at Mermeid and I can only change the color of the lines, not the arrowheads ([see here](https://github.com/mermaid-js/mermaid/issues/1236)). A quick workaround could be to style Mermaid diagram backgrounds (`div` with class mermaid) to a lighter grey. Together with some `border-radius` this can look well integrated within the documentation.
- It would be good to add `height: auto` to the SVG inside the mermaid `div`, especially for the [nested conversation](https://docs.python-telegram-bot.org/en/v20.0a1/examples.nestedconversationbot.html) which seems too wide for the column

examples/conversationbot.mmd
```diff
@@ -0,0 +1,22 @@
...
+    H --> J("(location)"):::userInput
+    I --> |You seem a bit paranoid!| K[\" "/]
+    J --> |Maybe I can visit...| K
+    K --> |"Tell me about yourself..."| L(("BIO")):::state
+    L --> M("(text)"):::userInput
+    M --> |"Thanks and bye!"| termination((" ")):::termination
+    classDef userInput  fill:#2a5279, color:#ffffff, stroke:#ffffff
+    classDef state fill:#222222, color:#ffffff, stroke:#ffffff
+    classDef firstState fill:#009c11, stroke:#42FF57, color:#ffffff
+    classDef termination fill:#bb0007, stroke:#E60109, color:#ffffff
```
Bibo-Joshi, line 22: please add an empty line at the end of the file - same for all other files where the circle with the dash shows in the [diff overview](https://github.com/python-telegram-bot/python-telegram-bot/pull/3090/files)

Bibo-Joshi:
Hi, thanks for the updates!

Personally I don't see a problem with darkmode - IMO the lines are rather well visible. I'm attaching screenshots below so that the others can see without having to build the docs :) 

Having the nestedconversationbot chart a bit wider would be nice, I agree. Don't recall exactly how one would modify the css here. If it's a small additional snippet for the sphinx config, we can do that. if it's a bigger effort, I'm also okay with user having to zoom in :D 

I have some other comments on the layout:

* in the nested conv. diagram, go directly to termination instead of listing "End" as state
* the diagram for `conversationbot2.py` is missing some sort of exit. 
* IMHO the legend is not really necessary. I'm okay with keeping it, but would suggest to do the following (if possible)
  * rename `First State` to `Entry Point` & `Termination` to `End`
  * adjust the size of entry point & end to be no bigger than `State`
  * make the "input -> state" left to right instead of top to bottom and list enrty point & end below that - saves some space
* the bright background of the legend and the nested conversation seems off in both bright & dark mode. If possible, please remove.
* in the nested example, it would be nice to have the nested conversation and the main conversation more vertically aligned. I guess that mermaid computes this automatically and that it's hard to fine tune that - just listing for completeness :)

<details><summary>Screenshots</summary>

image
image

</details>

Bibo-Joshi:
Thanks for the new updates! I built the docs from your branch locally and everything looks fine to me now :) If I get a LGTM from pool or harshil, we can merge :)

harshil21:
can we remove the old `.png` diagrams in the examples folder then?

Bibo-Joshi:
Thank you for the nice cnotribution! :)

PR #3106, opened 2022-06-19
Add `is_premium` attribute to `User`

MarcoBuster:
### Checklist for PRs

* New classes:
    - [x] `__init__` accepts `**_kwargs`


MarcoBuster:
The official documentation it's not here yet, but `User` has a new optional attribute `is_premium` which is present and `True` if and only if the user has an active Telegram Premium Subscription.

PR #3120, opened 2022-06-25
Drop pre-commit dependencies from reqs-dev

Bibo-Joshi:
Just a suggestion.

Pros:

- Less maintenance work keeping `reqs-dev` and `.pre-commit` in sync
- Less noise from dependabot
- `pre-commit` still works as expected

Cons:

- People that want to use the command line tools like `black` and `isort` don't have the directly availbale via `pip install -r reqs-dev.txt`. Not sure how many people actually do that though, plus which of them are not familiar enough with our setup to just get the version from the pre-commit config …


Poolitzer:
The workflow also notified on `requirements.txt` to keep the additional dependencies in sync with the mypy hook.

Additionally, I would probably add a line in `Contributing.rst`, saying that one can install the pre-commit runs manually and run them on the files if desired. Much like the black line we have, maybe append to that.

harshil21 (commented):
but wouldn't this just make dependabot not notify us of any new releases for the deleted dev-reqs? maybe a better option is to [ignore](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) minor and patch versions for those?

we can also change the dependabot [separator](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pull-request-branch-nameseparator) to "-" to conform to our style more.

fwiw, I'm still for automating updating pre-commit's additional_dependencies on dependabot PR's using a custom `.yml` script, but I don't know if/how that's possible

Poolitzer:
@harshil21 It should be very easy if we cant tell dependabot about it to write our own for that actually.

harshil21:
@Poolitzer yeah we can&#x27;t even tell <code>pre-commit</code> to update them automatically, see https://github.com/pre-commit/pre-commit/issues/1351

Bibo-Joshi:
> The workflow also notified on `requirements.txt` to keep the additional dependencies in sync with the mypy hook, I would probably still include that, but I'm also fine with it fully gone, whatever you prefer. Just wanted to point it out.

Thanks for the catch! Had overlooked that and will revert acoordingly

> Additionally, I would probably add a line in `Contributing.rst`, saying that one can install the pre-commit runs manually and run them on the files if desired. Much like the black line we have, maybe append to that.

Jup, good idea 👍 

> but wouldn't this just make dependabot not notify us of any new releases for the deleted dev-reqs? 

pre-commit.ci takes care of that - see #3085

> we can also change the dependabot [separator](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pull-request-branch-nameseparator) to "-" to conform to our style more.

With the slash, Pycharm automatically groups the branches in a directory style. I like that very much tbh, but in the end it's just a ide-specific gimmick. I'm fine with either way

> fwiw, I'm still for automating updating pre-commit's additional_dependencies on dependabot PR's using a custom `.yml` script, but I don't know if/how that's possible

that would indeed be nice! I found https://github.com/dependabot/dependabot-core/issues/2040 on this - FYI.
having a workflow that triggers on PRs that change `reqs.txt` and search-replaces in `.pre-commit.yml`  should probably suffice … Not a top priority IMO though :D

.github/workflows/pre-commit_dependencies_notifier.yml
```diff
@@ -15,5 +14,5 @@ jobs:
       - name: running the check
         uses: Poolitzer/notifier-action@master
         with:
-          notify-message: Hey! Looks like you edited the (dev) requirements or the pre-commit hooks. I'm just a friendly reminder to keep the pre-commit hook versions in sync with the dev requirements and the additional dependencies for the hooks in sync with the requirements :)
+          notify-message: Hey! Looks like you edited the requirements or the pre-commit hooks. I'm just a friendly reminder to keep the pre-commit hook versions in sync with the dev requirements and the additional dependencies for the hooks in sync with the requirements :)
```
harshil21, line 17:
```suggestion
          notify-message: Hey! Looks like you edited the requirements or the pre-commit hooks. I'm just a friendly reminder to keep the additional dependencies for the hooks in sync with the requirements :)
```

harshil21 (approved):

Issue #3111, opened 2022-06-21
[Documentation] Add note on editing with reply markups


Bibo-Joshi:
We should add note to all `Bot.edit_*` methods that states the limitation about editing messages from the official docs:

> Please note, that it is currently only possible to edit messages without `reply_markup` or with inline keyboards.

This should be done via the `Note:` role of sphinx as done e.g. [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/11007c17154e874716cdb99f6fdcd04c1724c8df/telegram/_message.py#L84)

A contribution in this direction is very welcome! A PR for this should be made against the `doc-fixes` branch. Please also check out the [contribution guide](https://docs.python-telegram-bot.org/en/v20.0a1/contributing.html), especially the part on how to [build the documentation locally](https://docs.python-telegram-bot.org/en/v20.0a1/contributing.html#documenting).

abhinav1912:
As a first contribution, this seems like a manageable task. Should I pick this?

Bibo-Joshi:
Hi, thanks for your interest! Someone already PRed for this issue - see #3121. But #3110 and #3109 are possible alternatives :)

Bibo-Joshi:
closed by #3121

PR #3121, opened 2022-06-25
Note on editing with reply markups

ExalFabu:
Hi, i would like to help and contribute on #3111 
Following the [official documentation](https://core.telegram.org/bots/api#updating-messages), the methods i updated are: 
- `Bot.edit_message_text`
- `Bot.edit_message_caption`
- `Bot.edit_message_media`

I am not adding the note on the method `Bot.edit_message_reply_markup` since I don't think it makes sense in the first place, I hope I'm not wrong.

Screenshots:
edit_message_text

edit_message_caption

edit_message_media

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -3374,8 +3374,8 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         """
         Use this method to edit text and game messages.
 
-        Note: 
-            It is currently only possible to edit messages without 
+        Note:
+            It is currently only possible to edit messages without
             :paramref:`reply_markup` or with inline keyboards
 
         Args:
@@ -3470,8 +3470,8 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         """
         Use this method to edit captions of messages.
 
-        Note: 
-            It is currently only possible to edit messages without 
+        Note:
+            Please note, that it is currently only possible to edit messages without
             :paramref:`reply_markup` or with inline keyboards
 
         Args:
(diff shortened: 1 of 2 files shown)


telegram/_bot.py
```diff
@@ -3374,6 +3374,10 @@ async def edit_message_text(
         """
         Use this method to edit text and game messages.
 
+        Note:
+            It is currently only possible to edit messages without
+            :paramref:`reply_markup` or with inline keyboards
```
Bibo-Joshi, line 3379:
```suggestion
            :attr:`telegram.Message.reply_markup` or with inline keyboards.
```

Same in the other places

telegram/_bot.py
```diff
@@ -3466,6 +3470,10 @@ async def edit_message_caption(
         """
         Use this method to edit captions of messages.
 
+        Note:
+            Please note, that it is currently only possible to edit messages without
```
Bibo-Joshi, line 3474:
```suggestion
            It is currently only possible to edit messages without
```

to keep it unified :)

Bibo-Joshi (changes requested):
Hi, thanks for the PR! I left two comments below.

> I am not adding the note on the method `Bot.edit_message_reply_markup` since I don't think it makes sense in the first place, I hope I'm not wrong.

Why do yout think that it doesn't make sense to have the note there? people can still try to edit messages without reply markup of with `ReplyKeyboardMarkup` …

ExalFabu:
> Why do yout think that it doesn't make sense to have the note there? people can still try to edit messages without reply markup of with `ReplyKeyboardMarkup` …

As i understood it, the meaning of the note specifies that you can only edit the text of a message if it does not contain a reply markup, since that method is supposed to be used to edit the reply markup I thought it was kind of unnecessary. 
Anyway I added it since I don't disagree with the "better safe than sorry" philosophy

Bibo-Joshi:
Thank you for the contribution! :)

PR #3126, opened 2022-06-27
Add `post_shutdown`

### LINKED ISSUE #3105: post_shutdown method in addition to post_init

i would like to be able to await coroutines before of shutting down the bot after application.run_polling().

it could be done implementing a post_shutdown method in addition to the already existing post_init method

It is especially useful for closing a db connection in aiosqlite ( await db.close() )

_No response_

### Additional context

_No response_

ExalFabu:
This should resolve #3105 
I strictly followed #3078, I hope I've not missed any documentation that was required to change as well

Anyway, as I mentioned in the issue i am struggling with these test that i added:  
- `test_run_polling_post_shutdown`
- `test_run_webhook_post_shutdown`

Here's a screenshot of the error (it's the same for both tests)
image

I don't know if I've messed up the test itself or I'm missing something

### DIFF
--- a/telegram/ext/_applicationbuilder.py
+++ b/telegram/ext/_applicationbuilder.py
@@ -135,6 +135,7 @@ class ApplicationBuilder(Generic[BT, CCT, UD, CD, BD, JQ]):
         "_persistence",
         "_pool_timeout",
         "_post_init",
+        "_post_shutdown",
         "_private_key",
         "_private_key_password",
         "_proxy_url",
@@ -178,6 +179,7 @@ class ApplicationBuilder(Generic[BT, CCT, UD, CD, BD, JQ]):
         self._concurrent_updates: DVInput[Union[int, bool]] = DEFAULT_FALSE
         self._updater: ODVInput[Updater] = DEFAULT_NONE
         self._post_init: Optional[Callable[[Application], Coroutine[Any, Any, None]]] = None
+        self._post_shutdown: Optional[Callable[[Application], Coroutine[Any, Any, None]]] = None
 
     def _build_request(self, get_updates: bool) -> BaseRequest:
         prefix = "_get_updates_" if get_updates else "_"
@@ -271,6 +273,7 @@ class ApplicationBuilder(Generic[BT, CCT, UD, CD, BD, JQ]):
             persistence=persistence,
             context_types=DefaultValue.get_value(self._context_types),
             post_init=self._post_init,
+            post_shutdown=self._post_shutdown,
             **self._application_kwargs,  # For custom Application subclasses
         )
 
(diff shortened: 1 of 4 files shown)


telegram/ext/_applicationbuilder.py
```diff
@@ -928,6 +931,42 @@ async def post_init(application: Application) -> None:
...
+
+        Tip:
+            This can be used for custom shutdown logic that requires to await coroutines, e.g.
+            closing a database connection
+
+        Example:
+            .. code::
+
+                async def post_shutdown(application: Application) -> None:
+                    await db_connection.close()
```
Bibo-Joshi, line 950:
```suggestion
                    await application.bot_data['database'].close()
```

just an idea - makes it more obvious why we might need the `application` parameter

Bibo-Joshi (changes requested):
Thanks for the PR! I left two comments below

ExalFabu:
I fixed and removed that duplicated line of code and the tests i wrote passed.
Though it seems another test broke: `test_run_error_in_application`, only when the error is simulated inside `application.initialize()`

And for what I understand it's really not an issue, since at this point of the execution the updater hasn't been initialized yet, and that's why `app.shutdown()` doesn't call `updater.shutdown()`

If my assumptions are correct, i may fix the test asserting `app._initialized` and `app.updater._initilized` as false, removing the need to monkeypatch the methods
Alternatively i can assert the two different cases inside a if statement
Something like `assert set(shutdowns) == ( {"application", "updater"} if method == "start" else {"application"} ) `

Bibo-Joshi:
The bad boy is this here:

https://github.com/python-telegram-bot/python-telegram-bot/blob/f1d03393def74bfc09f48556ea2b20e8c6098189/telegram/ext/_application.py#L374-L376

this tells `Application.shutdown` to just exit if the app wasn't properly initialized and hence `Updater.shutdown` is not called.
While we can't be 100% sure that `Updater.initialize` wasn't called by `Application.initialize`, I'm okay with that. Catching all errors that can happen during startup is a kind of a sisyphus task, so a best effort suffices IMO.

telegram/ext/_applicationbuilder.py
```diff
@@ -928,6 +931,42 @@ async def post_init(application: Application) -> None:
         self._post_init = post_init
         return self
 
+    def post_shutdown(
+        self: BuilderType, post_shutdown: Callable[[Application], Coroutine[Any, Any, None]]
+    ) -> BuilderType:
+        """
+        Sets a callback to be executed by :meth:`Application.run_polling` and
+        :meth:`Application.run_webhook` *after* executing :meth:`Application.shutdown` but
+        *before* executing :meth:`Updater.shutdown` respectively.
```
harshil21, line 940:
```suggestion
        :meth:`Application.run_webhook` *after* executing :meth:`Updater.shutdown`
        and :meth:`Application.shutdown`.
```

harshil21 (commented):
just one comment

telegram/ext/_applicationbuilder.py
ExalFabu: Oops, missed that comment

Is this up to me to commit? If yes let me know, i don't want to hold this PR just for this (sorry)

harshil21: yeah, you can just commit my suggestion (click the button above)

harshil21 (approved):

Bibo-Joshi:
Thank you for the contribution :)

PR #3127, opened 2022-06-28
Update default values for `concurrent_updates` and `connection_pool_size`

### LINKED ISSUE #3125: Adjust defaults for `concurrent_updates` and `conenction_pool_size`
`concurrent_updates` defaults to 4096, while `connection_pool_size` defaults to 128. If there is more than 128 updates being processed in parallel, this will cause pool timeouts rather quickly (unless the pool timeout is set to a very high value).
I suggest to update the default values such that the connection pool size is bigger than the number of updates that are allowed to be processed in parallel. It could actually be bigger (not equal), as there can be even more concurrent requests spawned by e.g. jobs or `asyncio/App.create_task`.
Moreover, the documentation of should be extended a bit more to give hints on this connection between `concurrent_updates`, `connection_pool_size` and `pool_timeout`.

Bibo-Joshi:
closes #3125

### Checklist for PRs


Poolitzer (approved):
Changes look good to me

PR #3147, opened 2022-07-08
Add `api_kwargs` to `log_out`

Bibo-Joshi:
### Checklist for PRs

### DIFF
--- a/tests/test_bot.py
+++ b/tests/test_bot.py
@@ -982,6 +982,7 @@ class TestBot:
     @pytest.mark.asyncio
     async def test_answer_web_app_query(self, bot, monkeypatch):
         params = False
+
         # For now just test that our internals pass the correct data
 
         async def make_assertion(url, request_data: RequestData, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


tests/test_bot.py
```diff
@@ -2964,3 +2965,18 @@ def test_coroutine_functions(self):
             "The following methods should be defined as coroutine functions: "
             f"{','.join(non_coroutine_functions)} "
         )
+
+    def test_api_kwargs_present(self):
```
harshil21, line 2969: why not additionally check for the timeout kwargs? Or are we already testing that somewhere

harshil21 (commented):
didn't I point this out in some PR? Maybe the change was never implemented or was lost in a rebase

tests/test_bot.py
Bibo-Joshi: Don't recall that :D but I'll extend the tests 👍

tests/test_bot.py
```diff
@@ -2945,22 +2956,35 @@ def test_camel_case_bot(self):
                 not_available_camelcase_functions.append(function_name)
```
harshil21, line 2956: can `test_camel_case_bot` also make use of the decorator

harshil21 (approved):

PR #3148, opened 2022-07-08
New rate limiting mechanism

### LINKED ISSUE #3018: New Mechanism for Avoiding Flood Limits
supercedes & closes #2139

Below I write down the different thoughts that I had on how we can introduce a new rate limiting mechanism. Pretty much everything is up for discussion. What do you want to have designed differently? What did I miss?

# Feature/Mechanism-Wishlist 

## Integration into PTB

One of the major shortcomings of `MessageQueue` was that it was not properly integrated with the framework. IMO that's one of the main points that should be considered. I image that the setup be such that something like

```python
ExtBot(…, rate_limiter=RateLimiter(…))
```
or
```python
Application.bulder().token('token').rate_limiter(RateLimiter(…)).build()
```
suffices. That is, `ExtBot` would have a (protected?) attribute holding the rate limiter and all requests would automatically go through that.
Note that I wrote `ExtBot` - this functionality should be purely for `telegram.ext` and not in `telegram`.

`getUpdates` should not get passed through the rate limiter.

## Abstract Interface

I'd like to make this a pluggable interface similar to `BaseRequest`. that way we give users a chance to implement custom, elaborate logic on their own. Such an interface class will probably just have once method - maybe `process_request` or so that `Bot` will call.

However, I'm somewhat unsure what this method should do/return. I see two options:

1. Return the actual return value of the api request. In this case, `ExtBot` would probably do something like `await self._rate_limiter.process_request(self._post(…))` and in `process_update` one could do somethling like `async with self._semaphore: await coroutine`
2. Alternatively, `procless_update` could return an async context manager, such that `ExtBot` would do `async with self._rate_limiter.process_request(…): await self._post(…)`

I guess the second one might be a bit harder for python-novices to implement. Then again, doing that is a rather advanced task anyway.
The first one may also allow for a bit more flexibility in the implementations. OTOH, the context manager approach separates the `Bot` from the `RateLimiter` a bit more clearly, which I potentially like.

### Open question: Retries

One thing that one can think of is whether the rate limiter should be allowed/expected to retry a request in case a `RetryAfter` exception is raised. My gut feeling is that it shouldn't, as it violates the chain of command in the PTB framework. But I'm not completely sure.
In case we want to allow for retries, we'd have to provide way for that to be possible, e.g. by passing a tuple `(method, kwargs)` such that the rate limiter can call `method(**kwargs)` as often as needed.

## Per-Request Decisions/Priorities

IMO it's important to try and allow for requests to be processed differently depending on 

* the endpoint (`answerInlineQuery`, `sendMessage`, etc)
* the parameters passed to that endpoint
* maybe even additional arguments that can be passed on a per-call basis.

This would allow to e.g. give answering inline queries a higher priority than sending messages and giving batch-notifications a lower priority than other messages (there even was a Discussions entry requesting something like this at some point, but I can't find it now)

So a signature of `process_request` could look like

```python
async def process_request(self, (coroutine ,)endpoint: str, data: Dict[str, Any], rate_limit_kwargs: Dict[str, Any]):
    ...
``` 

Here IMO the `data` should *not* be the `RequestData` that we pass to `BaseRequest` but basically the plain parameters that are passed to the bot method (after inserting `DefaultValues`.

# Naming

Maybe `BaseRateLimiter`?

# Possible built-in implementation

We could provide a basic implementation out of the box. For that we could use a 3rd party dependency like `aiolimiter` or `pyrate-limiter`. Both are published under the MIT license, so that shouldn't be a problem. `aiolimiter` has only one stable release but to me it looks bit bit cleaner, with a small, well defined feature set - though that may just be a first impression.
I imagine that such an implementation could be relatively close to what `MessageQueue` formerly did, i.e. enforce a global delay and a per-group/chat delay. One could think of providing that in a priority-queue style fashion such that user can specify a priority though the `rate_limit_kwargs`.

# Implementation

The main task of calling `process_request` in the right spot should not be too difficult. The main issue that I see is to keep the logic away from `tg.Bot`. For that I have a few ideas:

## Part 1
For passing `endpoint` and `data` it should suffice to split up `tg.Bot._post` a bit so that `ExtBot` has a chance to link into the spot between inserting `DefaultValues` and calling `Bot.request._post`.

## Part 2
For passing `rate_limit_kwargs`, that's a bit more tricky. We'd have to add a new parameter to all methods of `ExtBot`, but we don't want to add them to `Bot`. 

One straightforward option would be to split up `Bot.send_message` into `Bot._send_message` and `Bot.send_message`, where only the protected one has the additional parameter. `Bot.send_message` would then just pass everything along to `_send_message` and `ExtBot.send_message` would call `Bot._send_message` directly. This would result in *many* additional loc in `tg.Bot`. Additionally, this doesn't play well with shortcut methods.

The other idea that is to piggyback on `api_kwargs`: We add a parameter `rate_limit_kwargs` to `ExtBot.send_message`, but call `super().send_message(…, api_kwargs=api_kwargs |= {secret_key: rate_limit_kwargs)`. Then before passing everything to the rate limiter, we extract them as `rate_limit_kwargs = api_kwargs.pop(secret_key, {})`. This looks like a dirty hack at first glance, but the longer I think about it, the more I actually like it:

1. `api_kwargs` doesn't interfere with any of the other parameters, so it's rather safe to use for that
2. keeps `Bot` and `ExtBot` apart without introducing endless boilerplate
3. allows the shortcut methods to also have a paremeter `rate_limit_kwargs` - we can just throw an error if they are used with a `Bot` instead of an `ExtBot`

Bibo-Joshi:
closes #3018

Some comments:

* I went with the `aiolimiter` library despite https://github.com/python-telegram-bot/python-telegram-bot/issues/3018#issuecomment-1144030359. AFAIK the performance problems are only relevant for *very* high load and the built-in implementation is only supposed to be a minimal effort reference implementation …
* I introduced `aiolimiter` as optional requirement, since using the `AIORateLimiter` is purely opt-in. This design goes in the direction that I'd like to discuss in the context of https://github.com/python-telegram-bot/python-telegram-bot/projects/7#card-77767364
* I did not add `rate_imit_args` to shortcut methods. IMO that would add too much of a connection between tg and tg.ext and if users want to pass `rate_limit_args` they can always use the standard methods instead.

### Checklist for PRs


Bibo-Joshi:
I'll have to try & check how I can get the tests to work on unix. otherwise, this ready for a first round of review imo

Bibo-Joshi:
test comment

Bibo-Joshi:
@Bibo-Joshi test reply

Bibo-Joshi:
got tests to work on ubuntu. Skipped them on macOS, b/c I don't think that it's really worth the effort 😬

bharel:
@Bibo-Joshi @harshil21 I created an [asyncio limiter](https://pypi.org/project/asynciolimiter/) that doesn't have the httpx issue. In fact, the issue doesn't have anything to do with httpx at all. The interface is similar although a tad different, but it is considerably more efficient. You're welcome to test it and tell me if it's better for you :-)

telegram/ext/_extbot.py
```diff
@@ -121,12 +212,101 @@ def __init__(
...
+    ) -> Union[bool, JSONDict, None]:
+        rate_limit_args = self._extract_rl_kwargs(data)
+        if not self.rate_limiter and rate_limit_args is not None:
+            raise ValueError(
+                "`rate_limit_args` can only be used if a `ExtBot.rate_limiter` is set."
+            )
+
+        # getting updates should not be rate limited!
+        if endpoint == "getUpdates" or not self.rate_limiter:
+            return await super()._do_post(
```
harshil21, line 266: shouldn't we call `self.get_updates` here? Hmm, that would lead to infinite recursion.. or I guess it's not supposed to be used here?

telegram/ext/_extbot.py
```diff
@@ -121,12 +212,101 @@ def __init__(
...
+        return api_kwargs
+
+    @classmethod
+    def _extract_rl_kwargs(cls, data: Optional[JSONDict]) -> Optional[RLARGS]:
+        """Extracts the `rate_limit_args` from `data` if it exists."""
+        if not data:
+            return None
+        return data.pop(cls.__RL_KEY, None)
+
+    async def _do_post(
```
harshil21, line 248: Can you add a comment saying that the call hierarchy is `Bot._post, ExtBot._do_post`?

tests/test_ratelimiter.py
```diff
@@ -0,0 +1,356 @@
...
+
+        assert len(self.rl_received) == 2
+        assert self.rl_received[0] == ("getMe", {}, None)
+        assert self.rl_received[1] == (
+            "setMyCommands",
+            dict(commands=[BotCommand("test", "test")], language_code="en"),
+            (43, "test-1"),
+        )
+        assert len(self.request_received) == 4
+        assert self.request_received[0][1]["url"].endswith("getMe")
```
harshil21, line 127: can you add a comment on what the indexes refer to?

tests/test_ratelimiter.py
```diff
@@ -0,0 +1,356 @@
...
+
+        async with standard_bot:
+            await standard_bot.set_my_commands(
+                commands=[BotCommand("test", "test")], language_code="en"
+            )
+        async with rl_bot:
+            await rl_bot.set_my_commands(
+                commands=[BotCommand("test", "test")],
+                language_code="en",
+                rate_limit_args=(43, "test-1"),
```
harshil21, line 116: shouldn't we also test passing api_kwargs here and then checking it's presence in `BaseRateLimiter.data`?

harshil21 (changes requested):
Finally we have a working rate limiter! Didn't test out the changes locally _yet_. Also have a gut feeling that there's probably a bug here and there which I didn't catch

telegram/ext/_aioratelimiter.py
```diff
@@ -0,0 +1,242 @@
...
+class AIORateLimiter(BaseRateLimiter[int]):
+    """
+    Implementation of :class:`~telegram.ext.BaseRateLimiter` using the library
+    `aiolimiter <https://aiolimiter.readthedocs.io/>`_.
+
+    Important:
+        If you want to use this class, you must install PTB with the optional requirement
+        ``rate-limiter``, i.e. ``pip install python-telegram-bot[rate-limiter]``.
+
+    The rate limiting is applied by combining to throttles and :meth:`process_request` roughly
```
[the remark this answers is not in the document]
Bibo-Joshi: mh, I mean to say that there are two context managers that apply rate limiting. Maybe "combining two levels of throttling"/"The rate limits are applied in two levels"?

telegram/ext/_aioratelimiter.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        max_retries = rate_limit_args or self._max_retries
+
+        group: Union[int, str, bool] = False
+        chat = False
+        chat_id = data.get("chat_id")
+        if chat_id is not None:
+            chat = True
+        if (isinstance(chat_id, int) and chat_id < 0) or isinstance(chat_id, str):
+            # string chat_id only works for channels and supergroups
```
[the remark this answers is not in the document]
Bibo-Joshi: mh, I hadn't thought about that. You're right, we should make sure that this works as well 👍

telegram/ext/_aioratelimiter.py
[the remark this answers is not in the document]
harshil21: oh. Two levels of throttling sounds good yeah

telegram/ext/_aioratelimiter.py
```diff
@@ -0,0 +1,242 @@
...
+            except RetryAfter as exc:
+                if i == max_retries:
+                    self._logger.exception(
+                        "Rate limit hit after maximum of %d retries", max_retries, exc_info=exc
+                    )
+                    raise exc
+
+                sleep = exc.retry_after + 0.1
+                self._logger.info("Rate limit hit. Retrying after %f seconds", sleep)
+                # Make sure we don't allow other requests to be processed
```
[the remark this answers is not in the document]
Bibo-Joshi: Mh, right. Currently this applied globally. I'd be willing to let that fall under "this is a minimal effort reference implementation" though, as a per-group lock would require an additional cache …

telegram/ext/_extbot.py
Bibo-Joshi: Sorry, what's not supposed to be used here? `_post` and `_do_post` are the methods that are called by every bot method internally, including `getUpdates`. Here we just make sure that `getUpdates` isn't rate limited

tests/test_ratelimiter.py
Bibo-Joshi: yes, we probably should :)

telegram/ext/_aioratelimiter.py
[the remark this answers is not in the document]
harshil21: hmm okay. We should document this case explicitly since it's kind of a let down (yeah we shouldn't hit rate limits in the first place but still)

telegram/ext/_extbot.py
harshil21: nvm, I got a bit confused, all good now

tests/test_ratelimiter.py
Bibo-Joshi: test for ptbgithubbridge - ignore me

.github/CONTRIBUTING.rst
```diff
@@ -26,7 +26,7 @@ Setting things up
 
    .. code-block:: bash
 
-      $ pip install -r requirements.txt -r requirements-dev.txt
+      $ pip install -r requirements-all.txt
```
Bibo-Joshi, line 29: test for ptb github bridge - ignore me

Bibo-Joshi, line 29: test for ptb github bridge - ignore me

harshil21 (approved):
much awaited addition!

PR #3151, opened 2022-07-09
small test improvement

harshil21:


Bibo-Joshi (commented):
Maybe we can keep the `parametrize` approach but reduce the explicit list to `list(ChatAction)`?

harshil21:
hmm okay, but then I'll move the unknown action part of the test to a new one, so it isn't tested on every run

PR #3167, opened 2022-07-23
Improve `InvalidToken` Error Message

harshil21:
Deletes the `Bot._validate_token` step since bot tokens could change in the future. Now whenever an `InvalidToken` exception is raised, the token is printed along with it so the user can see where they went wrong.

----
#### Original proposal:

Improves PTB's verification of bot tokens and also adds a better error message along the way. Suppose your bot token is `123456:AAF5_-y.ert`. So the error message now shows:

```
E telegram.error.InvalidToken: Illegal character found in bot token: 
E 123456:AAF5_-y.ert
E               ^
```
We could also add another check where we assert that the no. of chars after `:` is `34` or `35`. I say `34` since in https://core.telegram.org/bots/#creating-a-new-bot it's 34, but I have only seen it be 35 though. I didn't add this since it might be too strict and could have false positives in the future?
 
### Checklist for PRs

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -22,6 +22,7 @@ import asyncio
 import functools
 import logging
 import pickle
+import string
 from contextlib import AbstractAsyncContextManager
 from datetime import datetime
 from types import TracebackType
@@ -420,13 +421,30 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
 
     @staticmethod
     def _validate_token(token: str) -> str:
-        """A very basic validation on token."""
-        if any(x.isspace() for x in token):
-            raise InvalidToken()
+        """Performs a validation on token by verifying the correct syntax and giving meaningful
+        error messages when it's incorrect.
+
+        .. versionchanged:: 20.0
+           Now returns a more descriptful error message.
+
+        """
+        # To pinpoint where exactly there is a illegal character, we multiply the current index to
+        # a whitespace and then add a caret (^) for easy identification. This is similar to what
+        # Python does with SyntaxErrors.
+        for idx, char in enumerate(token):
+            if char not in tuple(string.ascii_letters + string.digits) + ("-", "_", ":"):
+                err_msg = "Illegal character found in the bot token: "
+                raise InvalidToken(f"{err_msg}\n{token}\n{' ' * idx}^")
 
         left, sep, _right = token.partition(":")
-        if (not sep) or (not left.isdigit()) or (len(left) < 3):
-            raise InvalidToken()
+        if not sep:
+            raise InvalidToken("Missing `:` separator in the token")
+
+        if not left.isdigit():
+            raise InvalidToken("All characters left of the `:` separator must be digits!")
+
+        if len(left) < 3:
+            raise InvalidToken("Number of digits left of the `:` separator must be greater than 3")
 
         return token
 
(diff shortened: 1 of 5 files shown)


Bibo-Joshi (commented):
I like the ideas! but I'd really like to avoid false positives here. Even the current checks make assertions that are not build on explicitly documented behavior:

* token not containing spaces is not documented (but that's sure to make sense)
* part before `:` being all digits is not documented (that's almost sure as well, as it's the bots user id)
* part before `:` being >= 3 is not documented. I've never seen user ids shorter than 5  I think, but still …

checking for invalid characters in addition seems a bit risky to me, tbh … unless ofc I missed some documentation on that …

harshil21:
> * token not containing spaces is not documented (but that's sure to make sense)

yeah, we were already checking that before, so it's safe by now.

> * part before `:` being >= 3 is not documented. I've never seen user ids shorter than 5  I think, but still …

yeah I'm not sure why we were checking for that. Since it's just the bot id I'd be okay for deleting this check altogether.

> checking for invalid characters in addition seems a bit risky to me, tbh … unless ofc I missed some documentation on that …

Yeah, I'll open an issue in the bot api repo for this soon and see if I get a concrete answer.

harshil21:
Well so the reply received in the issue above is not what I expected.. but I have another idea:

We could suggest to the user what went wrong with their token _after_ receiving an error from `getMe`. This should de-risk it enough and make it safe. The error shown would be along the lines:

```
E telegram.error.InvalidToken: Not Found
E
E PTB suggests the following issue with the token:
E Illegal character found in bot token: 
E 123456:AAF5_-y.ert
E               ^
```
What do you think?

Bibo-Joshi:
TBH, the answer makes me rather question if it's worth to keep the `_validate_token` step at all …

telegram/request/_baserequest.py
```diff
@@ -312,20 +312,24 @@ async def _request_wrapper(
...
             #   2) correct tokens but non-existing method, e.g. api.tg.org/botTOKEN/unkonwnMethod
             # We can basically rule out 2) since we don't let users make requests manually
             # TG returns 401 Unauthorized for correctly formatted tokens that are not valid
-            raise InvalidToken(message)
-        if code == HTTPStatus.BAD_REQUEST:
+
+            # retrieve token from the url-
+            start = url[url.find("/bot") + 4 :]
+            token = start[: start.find("/")]
+            raise InvalidToken(message + f"\nThe token: {token} was rejected by the server.")
```
Bibo-Joshi, line 327: doing this at the request level seems rather unclean to me, tbh, especially since we have to extract the token from the url again. what would you rather think about doing something like

```
try:
    ...
except InvalidToken as exc:
    raise InvalidToken(f"The token {token} was …") from exc
```

in `Bot._post` - or even just once in `Bot.initialize`? I guess the latter should already suffice …

telegram/_bot.py
```diff
@@ -197,7 +197,9 @@ def __init__(
         private_key: bytes = None,
         private_key_password: bytes = None,
     ):
-        self.token = self._validate_token(token)
+        if not token:
+            raise InvalidToken("You must pass the token you received from https://t.me/Botfather ")
```
Bibo-Joshi, line 201:
```suggestion
            raise InvalidToken("You must pass the token you received from https://t.me/Botfather!")
```

telegram/_bot.py
```diff
@@ -372,7 +374,10 @@ async def initialize(self) -> None:
             return
 
         await asyncio.gather(self._request[0].initialize(), self._request[1].initialize())
-        await self.get_me()
+        try:
+            await self.get_me()
+        except InvalidToken as exc:
+            raise InvalidToken(f"The token {self.token} was rejected by the server.") from exc
```
Bibo-Joshi, line 380:
```suggestion
        try:
            await self.get_me()
        except InvalidToken as exc:
            raise InvalidToken(f"The token `{self.token}` was rejected by the server.") from exc
```

I guess a CSI-style comment on why we do this (at all, here and nowhere else) might be good

PR #3168, opened 2022-07-23
Test python 3.11 & Change Enum Behaviour 

harshil21:
Tests if py 3.11 works with the lib or not (the Python team really wants everyone to test it out). Not to be merged atm.

Also changes:
- Made str(StringEnum) & str(IntEnum) return consistent across python versions
- Also modify IntEnum's `__repr__` to match `StringEnum`'s

### DIFF
--- a/telegram/_utils/enum.py
+++ b/telegram/_utils/enum.py
@@ -23,21 +23,21 @@ Warning:
     user. Changes to this module are not considered breaking changes and may not be documented in
     the changelog.
 """
-import enum
+import enum as _enum
 import sys
 from typing import Type, TypeVar, Union
 
 _A = TypeVar("_A")
 _B = TypeVar("_B")
-_Enum = TypeVar("_Enum", bound=enum.Enum)
+_Enum = TypeVar("_Enum", bound=_enum.Enum)
 
 
-def get_member(_enum: Type[_Enum], value: _A, default: _B) -> Union[_Enum, _A, _B]:
-    """Tries to call ``enum(value)`` to convert the value into an enumeration member.
+def get_member(enum_cls: Type[_Enum], value: _A, default: _B) -> Union[_Enum, _A, _B]:
+    """Tries to call ``enum_cls(value)`` to convert the value into an enumeration member.
     If that fails, the ``default`` is returned.
     """
     try:
-        return _enum(value)
+        return enum_cls(value)
     except ValueError:
         return default
 
@@ -45,7 +45,7 @@ def get_member(_enum: Type[_Enum], value: _A, default: _B) -> Union[_Enum, _A, _
 # Python 3.11 and above has a different output for mixin classes for IntEnum, StrEnum and IntFlag
 # see https://docs.python.org/3.11/library/enum.html#notes. We want e.g. str(StrEnumTest.FOO) to
 # return "foo" instead of "StrEnumTest.FOO", which is not the case < py3.11
-class StringEnum(str, enum.Enum):
+class StringEnum(str, _enum.Enum):
     """Helper class for string enums where ``str(member)`` prints the value, but ``repr(member)``
     gives ``EnumName.MEMBER_NAME``.
     """
@@ -60,7 +60,7 @@ class StringEnum(str, enum.Enum):
 
 
 # Apply the __repr__ modification and __str__ fix to IntEnum
-class IntEnum(enum.IntEnum):
+class IntEnum(_enum.IntEnum):
     """Helper class for int enums where ``str(member)`` prints the value, but ``repr(member)``
     gives ``EnumName.MEMBER_NAME``.
     """
--- a/telegram/_utils/enum.py
+++ b/telegram/_utils/enum.py
@@ -23,24 +23,21 @@ Warning:
     user. Changes to this module are not considered breaking changes and may not be documented in
     the changelog.
 """
+import enum
 import sys
-from enum import Enum, IntEnum
 from typing import Type, TypeVar, Union
-
-if sys.version_info >= (3, 11):
-    from enum import StrEnum
 
 _A = TypeVar("_A")
 _B = TypeVar("_B")
-_Enum = TypeVar("_Enum", bound=Enum)
+_Enum = TypeVar("_Enum", bound=enum.Enum)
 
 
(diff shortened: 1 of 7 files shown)


harshil21:
3.11 tests are failing due to https://github.com/python/cpython/issues/94601, which has been fixed and will be included in 3.11b5. Built docs locally without any issues.

harshil21:
So the root cause of the bug might not have been the issue I linked, but the changed behavior when mixing `str` and `Enum` in py3.11. See https://docs.python.org/3.11/library/enum.html#notes. 

Hence I adjusted the behavior for 3.11. Also noticed that we show the Enum value on `__repr__` for `IntEnum` derived classes. Is that intended behavior?

Bibo-Joshi:
About allowed failures: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

IISC that may lead to some unexpected UI stuff (https://github.com/community/community/discussions/15452, https://github.com/actions/toolkit/issues/399), but whatever …

Regarding `repr`, `format` and `str`: I think I've changed my perspective a bit by now. Consider the following example

```python
def pprint(arg: int | str) -> str:
    return "Value is" + str(arg)  # explicitly cast in case it's an int
```

Now `pprint(ChatType.PRIVATE)` would yield `"Value is ChatType.PRIVATE"` which I'd argue is unexpected. So I'm now leaning towards having `StringEnum == str.__str__`. As no user has complained about that so far, I'm also okay with keeping it as is.

In any case
* `StringEnum` and `IntEnum` should behave the same
* `repr` should give unambiguous output and hence `ChatType.PRIVATE` is still derisable here IMO

harshil21:
Waiting for https://github.com/actions/python-versions/pull/183 to be merged so the 3.11 tests run...

telegram/_utils/enum.py
```diff
@@ -23,9 +23,13 @@
     user. Changes to this module are not considered breaking changes and may not be documented in
     the changelog.
 """
-from enum import Enum
+import sys
+from enum import Enum, IntEnum
```
Bibo-Joshi, line 27: `import enum` and then use ase `enum.(IntEnum)` -> fixes the "redef" thingy below

telegram/_utils/enum.py
```diff
@@ -41,12 +45,49 @@ def get_member(enum: Type[_Enum], value: _A, default: _B) -> Union[_Enum, _A, _B
...
-    """Helper class for string enums where the value is not important to be displayed on
+# Python 3.11 and above has a different output for mixin classes for IntEnum, StrEnum and IntFlag
+# see https://docs.python.org/3.11/library/enum.html#notes. We want e.g. str(StrEnumTest.FOO) to
+# return "foo" instead of "StrEnumTest.FOO", which is not the case < py3.11
+if sys.version_info < (3, 11):
+
+    class StringEnum(str, Enum):
+        """Helper class for string enums where the value is not important to be displayed on
+        stringification.
+        """
```
Bibo-Joshi, line 56:
```suggestion
        """Helper class for string enums where ``str(member)`` prints the value, but ``repr(member)`` gives ``EnumName.MEMBER_NAME```.
        """
```

or something along those lines … same for `IntEnum`

telegram/_utils/enum.py
```diff
@@ -41,12 +45,49 @@ def get_member(enum: Type[_Enum], value: _A, default: _B) -> Union[_Enum, _A, _B
...
+
+        def __repr__(self) -> str:
+            return f"<{self.__class__.__name__}.{self.name}>"
+
+        def __str__(self) -> str:
+            return str.__str__(self)
+
+else:
+
+    class StringEnum(StrEnum):
```
Bibo-Joshi, line 68: does the if-else actually save something here? feels a bit like just declaring `StringEnum` manually would be enough, as we have to modify the buit-in `StrEnum` anyways. Once 3.10 is dropped, that changes, ofc

telegram/constants.py
```diff
@@ -25,7 +25,9 @@
 those classes.
 
 .. versionchanged:: 20.0
-    Since v20.0, most of the constants in this module are grouped into enums.
+
+    * Most of the constants in this module are grouped into enums.
+    * ``str(<Enum>)`` now returns the value of the enum instead of the name of the enum.
```
Bibo-Joshi, line 30:
```suggestion
Most of the constants in this module are grouped into enums.
``str(<Enum>)`` now the value of the enum, while ``repr(<Enum>)`` returns thef the name of the enum.
```

differences between pre-releases don't need to be documented

telegram/_utils/enum.py
harshil21: ah wow

telegram/constants.py
harshil21: oh right I forgot, Enum's have been here for so long already

Bibo-Joshi (approved):
lgtm. rerunning the failing tests (test-official fails b/c of the api update, ofc)

telegram/_utils/enum.py
```diff
@@ -23,30 +23,54 @@
...
 from typing import Type, TypeVar, Union
 
 _A = TypeVar("_A")
 _B = TypeVar("_B")
-_Enum = TypeVar("_Enum", bound=Enum)
+_Enum = TypeVar("_Enum", bound=enum.Enum)
 
 
-def get_member(enum: Type[_Enum], value: _A, default: _B) -> Union[_Enum, _A, _B]:
+def get_member(_enum: Type[_Enum], value: _A, default: _B) -> Union[_Enum, _A, _B]:
```
tsnoam, line 35: 1. Nitpick: usually prefixing an argument with _ means that it's unused. In cases where there is an attempt to avoid name collisions it's customary to put the _ at the end.
2. The docstring needs updating.

And thinking as I write: maybe it would be more "clean" to do `import enum as _enum` at the top of the file.

harshil21: Doesn't prefixing with `_` means it's "private"? Afaik, unused is just `_`. So leave `import enum` as is?

tsnoam: Funny how the same notation means different things.

When an argument to a function is prefixed with _ it means it is unused.

When a function (or an import) is named prefixed with _ it means it is private.

PR #3172, opened 2022-07-25
Fix customwebhookbot.py example.

mylogin:
Use port in `setWebhook`, use a self-signed certificate for custom port, port changed to one currently supported for webhooks, disable listen only loopback interface

### DIFF
--- a/examples/customwebhookbot.py
+++ b/examples/customwebhookbot.py
@@ -18,6 +18,7 @@ import html
 import logging
 from dataclasses import dataclass
 from http import HTTPStatus
+from pathlib import Path
 
 import uvicorn
 from starlette.applications import Starlette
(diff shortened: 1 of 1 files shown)


Poolitzer:
Hi, I do not see the need for this. The example in the current way suggests you have it behind a reverse proxy, which imo is the way to go anyway. But changing it to a direct serving is very easy, with the 7 lines change you suggest. I do not see the need to change the example though, both ways are fine.

Bibo-Joshi:
I agree with pool here. The example can only ever capture a subset of possible use cases and especially for webhook setups, users will have to make adaptions for their specific configuration. Still, thanks for reaching out and proposing the changes! That's always appreciated :)

PR #3218, opened 2022-08-31
PTB Docs Search Overhaul

harshil21:
Introduces **Search-as-you-type** and **Advanced Search** — the ability to search by the _type_ of object — Class, Attribute, Method, Parameter, or just the documentation in general. 

https://user-images.githubusercontent.com/37377066/187569700-9bdbd5c0-d43d-4c2b-94bd-01eb5d875a6e.mp4

## Where can I test this?

https://docs.python-telegram-bot.org/en/doc-search/

----
#### Note for reviewers:
- I made a [fork](https://github.com/harshil21/readthedocs-sphinx-search) of [readthedocs/readthedocs-sphinx-search](https://github.com/readthedocs/readthedocs-sphinx-search) and used that here, since the number of changes to both the theme and the javascript was drastic enough. You can review the changes [here](https://github.com/harshil21/readthedocs-sphinx-search/commit/ce048b6dfe9411983097fea16d4a14b43695f5fe). We can also discuss moving this to the PTB-organization if wanted.

### Things which can be improved further:

- Assign a keyboard shortcut for opening advanced search? Maybe also one to clear the input field.
- The theme is dark mode by default. We can try to support light mode if we get enough requests.
- Design / colors of all elements! I tried to improve this as much as possible, but I'm no css wizard.

#### Things which can't (really) be improved:

- Search results when there are typos in your query, or you're searching for two closely matching words. Since RTD's backend for search is ElasticSearch, one would have to follow the [Advanced Search Syntax](https://docs.readthedocs.io/en/stable/guides/advanced-search.html#search-query-syntax) to get a better result.


Bibo-Joshi:
Pfew, that's quite something that you've cooked up!
The search-as-you type is definitely nice. However, I'm not quite comfortable with the overall setup.

I feel like the things like toggling the result types, adjusting the theme (colors and light/dark mode) and mobile support are things that go beyond "I'm doing a little tweaking" and should rather be proposed to upstream. Maintaining this exclusively for/as part of PTB is not something that I would look forward to.
Recall that we switched from the rtd theme to furo exactly b/c we were starting to fiddle with small details in the theme :D

Bibo-Joshi:
I see your points, but have a different view point regarding the negatives & the positives … at least until support for mobile devices is enhanced (apart from the "search for" toggle being way off to the right, the search box is "striked through" by the blue separator bar, at least for me)

Bibo-Joshi:
LGTM :) @Poolitzer do you want to have a look as well?

Poolitzer:
Just from a quick look at it, the type button looks misaligned. And it doesn't say no results if I filter out existing results for my query.

image

Poolitzer:
Yes I like the CSS look now good job! I just realized going backwards/forwards one page does not show the search again, is that fixable? Even worse if you filter stuff out, all of that apparently generates a new page. Makes you think the browser broke :D

docs/requirements-docs.txt
```diff
@@ -1,5 +1,6 @@
 sphinx==5.1.1
 sphinx-pypi-upload
 furo==2022.6.21
+git+https://github.com/harshil21/furo-sphinx-search@main
```
Bibo-Joshi, line 4: can we pin a specific commit once we're happy (e.g. right before merge)? Makes me feel safer :D

.readthedocs.yml
```diff
@@ -21,6 +21,29 @@ python:
...
+      telegram.passport*.html: -7
+
+   ignore:
+      - coc.html
+      - bot_methods.html
+      # Defaults
+      - search.html
+      - search/index.html
+      - 404.html
+      - 404/index.html'
```
Bibo-Joshi, line 49: new line

Bibo-Joshi (changes requested):
I guess from my side everything's fine now, just two tiny nitpicks below. @Poolitzer do you have any change requests left?

Poolitzer:
Almost happy. Back and forth works great now, but if you

1. type something in the search bar
2. press enter so the "old" search site is opened
3. now press the search bar on the left again

The search is not triggered. Is that fixable? Then I am happy.

PR #1168, opened 2018-07-26
Bot API 4.0

Eldinnie:
- [x] Added support for Telegram Passport. See the official announcement on the blog and the manual for details. (#1174)
- - [x] Add class [PasportData](https://core.telegram.org/bots/api#passportdata)
- - [x] Add class [PassportFile](https://core.telegram.org/bots/api#passportfile)
- - [x] Add class[EncryptedPassportElement](https://core.telegram.org/bots/api#encryptedpassportelement)
- - [x] Add class [EncryptedCredentials](https://core.telegram.org/bots/api#encryptedcredentials)
- - [x] Add method [setPassportDataErrors](https://core.telegram.org/bots/api#setpassportdataerrors) to `bot`
- - [x] Add baseclass [PassportElementError](https://core.telegram.org/bots/api#setpassportdataerrors)
- - [x] Add subclasses [PassportElementErrorDataField](https://core.telegram.org/bots/api#passportelementerrordatafield), [PassportElementErrorFrontSide](https://core.telegram.org/bots/api#passportelementerrorfrontside), [PassportElementErrorReverseSide](https://core.telegram.org/bots/api#passportelementerrorreverseside), [PassportElementErrorSelfie](https://core.telegram.org/bots/api#passportelementerrorselfie), [PassportElementErrorFile](https://core.telegram.org/bots/api#passportelementerrorfile) and [PassportElementErrorFiles](https://core.telegram.org/bots/api#passportelementerrorfiles)

- [x] Added support for editing the media content of messages: 
- - [x] Add the method [editMessageMedia](https://core.telegram.org/bots/api#editmessagemedia) to bot
- - [x] add shortcut to Message
- - [x] Add classes [InputMediaAnimation](https://core.telegram.org/bots/api#inputmediaanimation), [inputMediaAudio](https://core.telegram.org/bots/api#inputmediaaudio), and [inputMediaDocument](https://core.telegram.org/bots/api#inputmediadocument).

- [x] Added the field `thumb` to the [Audio ](https://core.telegram.org/bots/api#audio) object to contain the thumbnail of the album cover to which the music file belongs. (#1184)
- - [x] Add thumb to Audio

- [x] Added support for attaching custom thumbnails to uploaded files. For animations, audios, videos and video notes, which are less than 10 MB in size, thumbnails are generated automatically. (#1184)
- - [x] Add thumb to video
- - [x] Add thumb to videonote

- [x] tg:// URLs now can be used in inline keyboard url buttons and text_link message entities. (no need for change)

[...]

PR #3223, opened 2022-09-06
Feat: Rewrite File.download into download_to_drive and download_to_object

Poolitzer:
TODO after merge: Update wiki and transition guide/script

### Checklist for PRs


Poolitzer:
I dont get codecov path, `test_download_file_obj` covers that, no?

https://github.com/python-telegram-bot/python-telegram-bot/blob/27a0fca6238e0200eb8e9b1b4250bd15e3639323/tests/test_file.py#L157-L167

telegram/_files/file.py
```diff
@@ -197,14 +187,66 @@ async def download(
...
+                read_timeout=read_timeout,
+                write_timeout=write_timeout,
+                connect_timeout=connect_timeout,
+                pool_timeout=pool_timeout,
             )
-        )
+            if self._credentials:
+                buf = decrypt(
+                    b64decode(self._credentials.secret), b64decode(self._credentials.hash), buf
+                )
```
Bibo-Joshi, line 247: This could be extracted to a protected method, since it's used in both `download_to_*` methods

telegram/_files/file.py
```diff
@@ -197,14 +187,66 @@ async def download(
...
+                :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
+            connect_timeout (:obj:`float` | :obj:`None`, optional): Value to pass to
+                :paramref:`telegram.request.BaseRequest.post.connect_timeout`. Defaults to
+                :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
+            pool_timeout (:obj:`float` | :obj:`None`, optional): Value to pass to
+                :paramref:`telegram.request.BaseRequest.post.pool_timeout`. Defaults to
+                :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
+
+        Returns:
+            :obj:`io.BufferedWriter`: The same object as :paramref:`out`.
```
Bibo-Joshi, line 228: Since this method modifies `out` in-place, I would vote not to return anything. Formerly, returning `out` made a bit of sense, since the method returned something in the other case. Now I don't see a need for that …

telegram/_files/file.py
```diff
@@ -197,14 +187,66 @@ async def download(
...
+        Download this file. paramref:`out` needs to be supplied with a
+        :obj:`io.BufferedWriter`, the file contents will be saved to that object using the
+        :obj:`out.write<io.BufferedWriter.write>` method.
+
+        .. versionchanged:: 20.0
+
+            * :paramref:`custom_path` parameter now also accepts :class:`pathlib.Path` as argument.
+            * Returns :class:`pathlib.Path` object in cases where previously a :obj:`str` was
+              returned.
+            * Splitting :paramref:`custom_path` and :paramref:`out` in two functions.
```
Bibo-Joshi, line 208: The former method ``download`` was split into :meth:`download_to_drive` and :meth:`download_to_obj`.

telegram/_files/file.py
```diff
@@ -197,14 +187,66 @@ async def download(
...
+
+
+        Args:
+            out (:obj:`io.BufferedWriter`): A file-like object. Must be opened for writing in
```
Bibo-Joshi, line 212: Re-Checking the python docs, maybe [BufferedIOBase](https://docs.python.org/3/library/io.html?highlight=bufferedwriter#io.BufferedIOBase) actually better represents what we mean here? Also the type hint should maybe be changed from `IO` to [`BytesIO`](https://docs.python.org/3/library/typing.html?highlight=typing%20io#typing.BinaryIO) ? 🤔

Bibo-Joshi (changes requested):
Thanks for the PR! I left some comments, some of which are not actually due to your changes :D 

IISC Codecov mainly complains about decrypting files with `self._credentials` not being covered.

Regarding naming, I'd like to also throw `download_to_file` (as alternative to `download_to_drive`) and `download_to_memory` (as alternative to `download_to_obj`) into the race.

telegram/_files/file.py
```diff
@@ -65,7 +67,8 @@ class File(TelegramObject):
             is supposed to be the same over time and for different bots.
             Can't be used to download or reuse the file.
         file_size (:obj:`str`): Optional. File size in bytes.
-        file_path (:obj:`str`): Optional. File path. Use :meth:`download` to get the file.
+        file_path (:obj:`str`): Optional. File path. Use e.g. :meth:`download_to_drive()` to get
```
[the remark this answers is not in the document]
Poolitzer: They do, and I actually prefer it (though the change was unintentional):

image

I think this makes it clear that it is a function

telegram/_files/file.py
Poolitzer: They say a typical implemention should not inherit, but wrap though..?

telegram/_files/file.py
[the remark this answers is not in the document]
Bibo-Joshi: braces are inserted automatically though, if you use `:meth:`. see [here](https://docs.python-telegram-bot.org/en/v20.0a4/telegram.file.html?highlight=file.download#telegram.File)

telegram/_files/file.py
Poolitzer: Sure, good point.

telegram/_files/file.py
Poolitzer: But its one function call both ways, what would be the upside of moving it to its own function?

telegram/_files/file.py
Bibo-Joshi: > A typical [BufferedIOBase](https://docs.python.org/3/library/io.html?highlight=bufferedwriter#io.BufferedIOBase) implementation should not inherit _from a [RawIOBase](https://docs.python.org/3/library/io.html?highlight=bufferedwriter#io.RawIOBase) implementation_, but wrap one, like [BufferedWriter](https://docs.python.org/3/library/io.html?highlight=bufferedwriter#io.BufferedWriter) and [BufferedReader](https://docs.python.org/3/library/io.html?highlight=bufferedwriter#io.BufferedReader) do.

and in the docs of `BufferedWriter` it says 

> It inherits [BufferedIOBase](https://docs.python.org/3/library/io.html?highlight=bufferedwriter#io.BufferedIOBase).

🤔

Poolitzer: Well, I don't know either xD

Poolitzer: I tried both ``issinstance`` and ``issubclass``, both are true, so we are good I guess.

telegram/_files/file.py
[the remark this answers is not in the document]
Poolitzer: ah thats the wrong type you fixed above, my bad.

Lauraki (approved):

telegram/_files/file.py
Bibo-Joshi: reducing code duplication, as always …

harshil21 (changes requested):

telegram/_files/file.py
```diff
@@ -143,32 +156,12 @@ async def download(
...
 
         Returns:
-            :class:`pathlib.Path` | :obj:`io.BufferedWriter`: The same object as :paramref:`out` if
-            specified. Otherwise, returns the filename downloaded to or the file path of the
-            local file.
-
-        Raises:
-            ValueError: If both :paramref:`custom_path` and :paramref:`out` are passed.
+            :class:`pathlib.Path`: Returns the filename downloaded to
+            or the file path of the local file.
```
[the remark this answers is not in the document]
Poolitzer: I think I will wait for local mode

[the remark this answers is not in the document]
Bibo-Joshi: ready to go 😉

telegram/_files/file.py
```diff
@@ -97,38 +102,46 @@ def __init__(
...
+        :obj:`pathlib.Path`, it will be saved to that path.
 
         Note:
-            * :paramref:`custom_path` and :paramref:`out` are mutually exclusive.
-            * If neither :paramref:`custom_path` nor :paramref:`out` is provided and
-              :attr:`file_path` is the path of a local file (which is the case when a Bot API
-              Server is running in local mode), this method will just return the path.
+            If :paramref:`custom_path` isn't provided and :attr:`file_path` is the path of a
+              local file (which is the case when a Bot API Server is running in local mode), this
+              method will just return the path.
```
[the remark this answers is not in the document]
Poolitzer: This one? I would remove the other one tbh, I like it as a note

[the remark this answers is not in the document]
harshil21: alright

Bibo-Joshi:
So I finally got around to review this again and I apologize for taking so long.
The new test look good to me, I understood what they do while reading them. However, it would be nice to have some comments that explain how you created the `image_de/encrypted` and the input for `FileCredentials`. That would have the benefits of a) being able to reproduce things and b) have even better understanding of what's going on.

> * Did it make sense that I did this whole encrypted file way? It feels to me like the natural way to test this, but maybe there is a better way?

LGTM me. 

>   * If that makes sense to use, shouldn't we use this in different tests?

That would mean tuching passport tests. grrrrrrr. I'm okay with only touching them if absolutely necessary xD

> * Should we rename the hash param in decrypt, it shadows a builtin function.

Okay with me, can be a separate PR though.

> * Why does EncryptedPassportElement take credentials argument?

No clue, goes back to #1168. IISC removing should be fine, maybe in the same PR as the hash param

> * Does it make sense not to apply the decryption on local files? Because thats what currently happens...

It does not, actually 🤔 Even for local files, they should be encrypted, I guess. Can you fix that + add tests?

telegram/_files/file.py
```diff
@@ -131,19 +133,23 @@ async def download_to_drive(
             If :paramref:`custom_path` isn't provided and :attr:`file_path` is the path of a
               local file (which is the case when a Bot API Server is running in local mode), this
               method will just return the path.
+            If, however, :attr:`telegram.PassportFile.get_file` is the source of this object,
```
harshil21, line 136:
```suggestion
            If, :attr:`telegram.PassportFile.get_file` is the source of this object,
```

Also the note should say that this is only done if `custom_path` isn't provided

harshil21 (commented):
nice work, didn't look at the tests too closely:

telegram/_files/file.py
Poolitzer: Oh I thought that `custom_path` isn't provided + local file would be clear because it continues the line above. That is why I wrote however, so I do not have to repeat the conditions. Since they are the same.

telegram/_files/file.py
```diff
@@ -35,18 +35,22 @@
...
-    link expires, a new one can be requested by calling :meth:`telegram.Bot.get_file`.
+    This object represents a file ready to be downloaded. The file can be e.g. downloaded with
+    :attr:`download_to_drive`. It is guaranteed that the link will be valid for at least 1 hour.
+    When the link expires, a new one can be requested by calling :meth:`telegram.Bot.get_file`.
 
     Objects of this class are comparable in terms of equality. Two objects of this class are
     considered equal, if their :attr:`file_unique_id` is equal.
 
+    .. versionchanged:: 20.0:
+        ``download`` was split into :meth:`download_to_drive` and :meth:`download_to_object`.
```
Bibo-Joshi, line 46:
```suggestion
        ``download`` was split into :meth:`download_to_drive` and :meth:`download_to_memory`.
```

Just pitching my suggestions once more :D For my feeling, "memory" gives you a better idea of what this is doing. with "object", I have to think about what this naming is trying to tell me …

telegram/_files/file.py
```diff
@@ -220,9 +271,19 @@ async def download_as_bytearray(self, buf: bytearray = None) -> bytearray:
...
         else:
-            buf.extend(await self.get_bot().request.retrieve(self._get_encoded_url()))
+            if self._credentials:
+                buf.extend(
+                    self._prepare_decrypt(
+                        await self.get_bot().request.retrieve(self._get_encoded_url())
+                    )
+                )
+            else:
+                buf.extend(await self.get_bot().request.retrieve(self._get_encoded_url()))
```
Bibo-Joshi, line 286: Can we remove a bit of code duplication here as well?

```
bytes_data = ... if local_file else
if self._credentials:
    bytes_data = decrypt(byates_data)
buf.extend(bytes_data)`
```

Bibo-Joshi (commented):
Thanks for the updates & the improvements for local encrypted files!

examples/conversationbot.py
```diff
@@ -23,7 +23,7 @@
 except ImportError:
     __version_info__ = (0, 0, 0, 0, 0)  # type: ignore[assignment]
 
-if __version_info__ < (20, 0, 0, "alpha", 1):
+if __version_info__ < (20, 0, 0, "alpha", 5):
```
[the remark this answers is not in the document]
Poolitzer: Maybe in the release page? But it's hard to say when a change breaks an example post those changes

telegram/_files/file.py
Poolitzer: uh pfff okay.

examples/conversationbot.py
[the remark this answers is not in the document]
Bibo-Joshi: I'm mainly talking about our plans to make the next release a beta release. Only if we get another API update before #3249 is mergable, the next release would be calld 20.0a5 … Mind you, this check would still be correct, since `(20, 0, 0, "alpha", 5) < (20, 0, 0, "beta", 0)`, but it would probably still confuse users xD I'll add a note o the "how to release page" as reminder :)

Bibo-Joshi (approved):
precommit and doc-build are failing 🤔 
otherwise LGTM 🙂

harshil21 (approved):
LGTM

Issue #2698, opened 2021-10-04
[DISCUSSION] Dataclasses


Bibo-Joshi:
The idea to use dataclasses has been discussed a number of times (see #2056), more recently in #2491 and there is also a connection to #2607. So here I am opening an issue to write down some thoughts.

This is rather for from being my final judgement on the matter and I'm aware that it will be a while until we properly discuss this. So more of a reminder for future-me.

## Motivation
We have a lot of boilerplate code. especially the inits of classes like `Message` take up a lot of place, along with `__slots__` and `de/to_json`.
There are ways to hide those away by using 3rd pary libraries (that inspired `dataclasses`) or the built-in `dataclasses`. Here is what such approaches can possibly do for us:

* removing boilerplate inits
* removing boilerplate slots
* Solving big parts of #2491 without any work at all
* Making TelegramObjects immutable. I'm really not sure if we want that, but it's at least an interesting thought given that incoming updates represent a static information and changing them doesn't make sense at all.

**Edits:** 
* inspired by #2747 dataclasses could also be used for `ext.Defaults` and `CallbackContext`, especially making them immutable
* dataclasses would (as side effect) solve large parts of microsoft/pylance-release#1983
* with dataclasses, we could get rid of the `__new__` workaround in `TelegramObject`: The reason for that was that we didn't want to have to call `super().__init__` in all the subclasses. if `TO` is a dataclass as well, then the auto-generated inits of the subclasses would automatically do that for us

## A word on 3rd party libs
A number of 3rd party libraries that offer similar functionality as `dataclasses`, most notably `attrs` and `pydantic` but also [some others](https://pydantic-docs.helpmanual.io/benchmarks/). I know that we try to keep our dependencies at a minimum and I'm fully backing that mentality. Still, let me list some pros & cons that I see of possible usage of `attrs` & `pydantic` - compared to using `dataclasses`

* ➕ Automated handling of (nested!) to/de_json logic. `pydantic` has this built-in, `attrs` can do this together with the `cattrs` library
* ➕/➖ Data validation. `pydantic` is heavily inspired by that. This means that e.g. `self.text = text` is replaced by `self.text = str(text)` in `pydantic`, but apparently also [slows it down](https://stefan.sofa-rockers.org/2020/05/29/attrs-dataclasses-pydantic/)
[...]

PR #3234, opened 2022-09-11
Readable string representations for TelegramObject

Bibo-Joshi:
Another part of #2698. As per https://github.com/python-telegram-bot/python-telegram-bot/issues/2698#issuecomment-1241181888 doesn't add a "pprint" variant yet. IMO we can add this if we/users actually see a need for that :)

Note that once #3233 is merged, this will also include `api_kwargs` in the representation. IMO that's desirable. Once could however do a special case and exclude that if it's an empty dictionary. edit: I did that

### Checklist for PRs


Poolitzer (approved):
You formulated it as a question if we want to drop empty api_kwargs and if I read this correctly you just decided to go ahead with this. I like it, and the rest of the code change!

tests/test_telegramobject.py
```diff
@@ -284,3 +284,45 @@ def test_deepcopy_subclass_telegram_obj(self, bot):
...
+            "TGO(api_kwargs={'foo': 'bar'}, dict_attr={BotCommand(command='command_1', "
+            "description='description_1'): BotCommand(command='command_2', "
+            "description='description_2')}, int_attr=42, "
+            "list_attr=[BotCommand(command='command_1', description='description_1'), "
+            "BotCommand(command='command_2', description='description_2')], "
+            "string_attr='string', to_attr=BotCommand(command='command', "
+            "description='description'))"
+        )
+        assert str(TGO(api_kwargs={"foo": "bar"})) == expected_with_api_kwargs
+        assert str(TGO(api_kwargs={"foo": "bar"})) == expected_with_api_kwargs
```
harshil21, line 328:
```suggestion
        assert str(TGO(api_kwargs={"foo": "bar"})) == repr(TGO(api_kwargs={"foo": "bar"})) == expected_with_api_kwargs
```
note that str() assertion was repeated twice

tests/test_telegramobject.py
```diff
@@ -284,3 +284,45 @@ def test_deepcopy_subclass_telegram_obj(self, bot):
...
+        expected_without_api_kwargs = (
+            "TGO(dict_attr={BotCommand(command='command_1', description='description_1'): "
+            "BotCommand(command='command_2', description='description_2')}, int_attr=42, "
+            "list_attr=[BotCommand(command='command_1', description='description_1'), "
+            "BotCommand(command='command_2', description='description_2')], "
+            "string_attr='string', to_attr=BotCommand(command='command', "
+            "description='description'))"
+        )
+        assert str(TGO()) == expected_without_api_kwargs
+        assert repr(TGO()) == expected_without_api_kwargs
```
harshil21, line 316:
```suggestion
        assert str(TGO()) == repr(TGO()) == expected_without_api_kwargs
```
single line is better?

harshil21 (changes requested):
btw, should we include the documentation of TGObject's dunder methods? I think we should since it makes it more clear what we can do with the object as a whole.

telegram/_telegramobject.py
```diff
@@ -100,8 +103,26 @@ def _apply_api_kwargs(self) -> None:
...
-        return str(self.to_dict())
+    def __repr__(self) -> str:
+        """
+        * `__repr__` goal is to be unambiguous
+        * `__str__` goal is to be readable
+        * `str()` calls `__repr__`, if `__str__` is not defined
+
+        In our case "unambiguous" and "readable" largely coincide, so we can use the same logic.
+
+        Prints the object in the form `ClassName(arg=value, arg=value, ...)`
```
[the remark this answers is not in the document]
Bibo-Joshi: which advantage do you see in explicitly documenting `__repr__` and `__str__` over just stating their effect? I.e. we also only say "objects of this type are comparable in terms of equality` instead of documenting `__eq__` so far … 🤔

tests/test_telegramobject.py
Bibo-Joshi: wouldn't that make it harder to see which of the two assertions fails, if one fails?

harshil21: py 3.10+ shows where the assertion fails with carets, so it's okay to do this I guess

telegram/_telegramobject.py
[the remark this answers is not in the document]
harshil21: 1. Its technical documentation, so if a user wanted to for e.g. override methods, it would be clear what exactly is being overriden and what effect its causing

2. They define the behaviour of classes more clearly, specially for those who prefer that over a summary above.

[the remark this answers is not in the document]
Bibo-Joshi: Fair enough. I'll document Str and repr here then. The other dunder methods can go into another pr or just doc-fixes

tests/test_telegramobject.py
Bibo-Joshi: mh, I'll keep it in two lines for now … py3.9 will reach EOL only in years

telegram/_telegramobject.py
```diff
@@ -100,8 +103,26 @@ def _apply_api_kwargs(self) -> None:
...
+        """
+        as_dict = self._get_attrs(recursive=False, include_private=False)
+
+        if not self.api_kwargs:
+            # Drop api_kwargs from the representation, if empty
+            as_dict.pop("api_kwargs", None)
```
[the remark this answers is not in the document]
Bibo-Joshi: I went ahead and made it ignore all values that have length zero & also documented that.

harshil21 (approved):
LGTM

PR #3252, opened 2022-09-20
Handle non-string data correctly with regex pattern in CQH

Bibo-Joshi:
See https://t.me/pythontelegrambotgroup/626511?thread=626501 for discussion

### Checklist for PRs

- [ ] ~Documented code changes according to the [CSI standard](https://standards.mousepawmedia.com/en/stable/csi.html)~

### DIFF
--- a/telegram/ext/_callbackqueryhandler.py
+++ b/telegram/ext/_callbackqueryhandler.py
@@ -130,6 +130,7 @@ class CallbackQueryHandler(BaseHandler[Update, CCT]):
             :obj:`bool`
 
         """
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
> change LGTM, DeepSource complains though, would replacing the variable with an underscore fix it?

ah, that again … changing the name of the variable would mean that one can't rely on calling `some_handler.collect_additional_context` with keyword arguments anymore. I have reported that as false positive already a couple of times …

harshil21 (approved):

PR #3261, opened 2022-09-28
Add chat_id(s) and username(s) parameters to ChatJoinRequestHandler 

### LINKED ISSUE #3215: [Suggestion] Add chat(s) parameter to ChatJoinRequestHandler
This param should allow to filter out chats which will be handled by the ChatJoinRequestHandler, much like the pattern argument of the CallbackQueryHandler. It should allow "username" strings as well as ids and if set, the handler should check if the incoming update is from that chat.

For first time contributors, check how CallbackQueryHandler implements the pattern argument in check_update: https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/_callbackqueryhandler.py#L123

miles170:
`chat_id` and `username` parameters should allow filtering requests to allow only those from a specified chat ID or username.

Closes #3215.

### Checklist for PRs

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -70,6 +70,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Loo Zheng Yuan <https://github.com/loozhengyuan>`_
 - `LRezende <https://github.com/lrezende>`_
 - `macrojames <https://github.com/macrojames>`_
+- `miles <https://github.com/miles170>`_
 - `Matheus Lemos <https://github.com/mlemosf>`_
 - `Michael Elovskikh <https://github.com/wronglink>`_
 - `Mischa Krüger <https://github.com/Makman2>`_
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -70,9 +70,9 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Loo Zheng Yuan <https://github.com/loozhengyuan>`_
 - `LRezende <https://github.com/lrezende>`_
 - `macrojames <https://github.com/macrojames>`_
-- `miles <https://github.com/miles170>`_
 - `Matheus Lemos <https://github.com/mlemosf>`_
 - `Michael Elovskikh <https://github.com/wronglink>`_
+- `miles <https://github.com/miles170>`_
 - `Mischa Krüger <https://github.com/Makman2>`_
 - `naveenvhegde <https://github.com/naveenvhegde>`_
 - `neurrone <https://github.com/neurrone>`_
--- a/telegram/ext/_chatjoinrequesthandler.py
+++ b/telegram/ext/_chatjoinrequesthandler.py
@@ -60,7 +60,6 @@ class ChatJoinRequestHandler(BaseHandler[Update, CCT]):
             .. versionadded:: 20.0
         username (:obj:`str` | Collection[:obj:`str`], optional): Filters requests to allow only
             those which are asking to join the specified username(s).
-            Leading ``'@'`` in username will be discarded.
 
             .. versionadded:: 20.0
         block (:obj:`bool`, optional): Determines whether the return value of the callback should
--- a/telegram/ext/_chatjoinrequesthandler.py
+++ b/telegram/ext/_chatjoinrequesthandler.py
@@ -45,11 +45,11 @@ class ChatJoinRequestHandler(BaseHandler[Update, CCT]):
             The return value of the callback is usually ignored except for the special case of
             :class:`telegram.ext.ConversationHandler`.
         chat_id (:obj:`int` | Collection[:obj:`int`], optional): Filters requests to allow only
-            those which are from specified chat ID(s).
+            those which are asking to join the specified chat ID(s).
 
             .. versionadded:: 20.0
         username (:obj:`str` | Collection[:obj:`str`], optional): Filters requests to allow only
-            those which are from specified username(s).
+            those which are asking to join the specified username(s).
 
             .. versionadded:: 20.0
         block (:obj:`bool`, optional): Determines whether the return value of the callback should
--- a/telegram/ext/_chatjoinrequesthandler.py
+++ b/telegram/ext/_chatjoinrequesthandler.py
@@ -44,12 +44,12 @@ class ChatJoinRequestHandler(BaseHandler[Update, CCT]):
 
             The return value of the callback is usually ignored except for the special case of
             :class:`telegram.ext.ConversationHandler`.
-        chat_id (SCT[:obj:`int`], optional): Filters requests to allow only
-            those which are from a specified chat ID.
+        chat_id (:obj:`int` | Collection[:obj:`int`], optional): Filters requests to allow only
+            those which are from specified chat ID(s).
 
             .. versionadded:: 20.0
-        username (SCT[:obj:`str`], optional): Filters requests to allow only
-            those which are from a specified username.
+        username (:obj:`str` | Collection[:obj:`str`], optional): Filters requests to allow only
+            those which are from specified username(s).
 
             .. versionadded:: 20.0
         block (:obj:`bool`, optional): Determines whether the return value of the callback should
@@ -58,14 +58,6 @@ class ChatJoinRequestHandler(BaseHandler[Update, CCT]):
 
     Attributes:
         callback (:term:`coroutine function`): The callback function for this handler.
-        _chat_ids (:frozenset(:obj:`int`), optional):
-            Which chat ID(s) to allow through.
-
-            .. versionadded:: 20.0
-        _usernames (:frozenset(:obj:`str`), optional):
-            Which username(s) to allow through.
-
-            .. versionadded:: 20.0
         block (:obj:`bool`): Determines whether the callback will run in a blocking way..
 
     """
--- a/telegram/ext/_chatjoinrequesthandler.py
+++ b/telegram/ext/_chatjoinrequesthandler.py
@@ -32,8 +32,10 @@ class ChatJoinRequestHandler(BaseHandler[Update, CCT]):
     :attr:`telegram.Update.chat_join_request`.
 
     Note:
-        If either the :paramref:`username` or the :paramref:`chat_id` corresponds to one
-        in the appropriate, the request will be passed to this handler.
+        If neither of :paramref:`username` and the :paramref:`chat_id` are passed, this handler
+        accepts *any* join request. Otherwise, this handler accepts all requests to join chats
+        for which the chat ID is listed in :paramref:`chat_id` or the username is listed in
+        :paramref:`username`, or both.
 
         .. versionadded:: 20.0
 
--- a/telegram/ext/_chatjoinrequesthandler.py
+++ b/telegram/ext/_chatjoinrequesthandler.py
@@ -18,7 +18,7 @@
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains the ChatJoinRequestHandler class."""
 
-from typing import Optional
+from typing import FrozenSet, Optional
 
 from telegram import Update
 from telegram._utils.defaultvalue import DEFAULT_TRUE
@@ -89,7 +89,7 @@ class ChatJoinRequestHandler(BaseHandler[Update, CCT]):
         self._usernames = self._parse_username(username)
 
     @staticmethod
-    def _parse_chat_id(chat_id: Optional[SCT[int]]) -> frozenset[int]:
+    def _parse_chat_id(chat_id: Optional[SCT[int]]) -> FrozenSet[int]:
         if chat_id is None:
             return frozenset()
         if isinstance(chat_id, int):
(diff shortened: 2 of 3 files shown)


telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -43,17 +44,41 @@ async def callback(update: Update, context: CallbackContext)
 
             The return value of the callback is usually ignored except for the special case of
             :class:`telegram.ext.ConversationHandler`.
+        chat_id(:obj:`int` | Collection[:obj:`int`], optional): Filters requests to allow only
```
Poolitzer, line 47: I think we use internally the SCT typehint here, so SCT[int]

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -43,17 +44,41 @@ async def callback(update: Update, context: CallbackContext)
...
+            those which are from a specified chat ID.
+
+            .. versionadded:: 20.0
```
Poolitzer, line 50: This currently only allows chat IDs. The issue also requested usernames. Can you implement that as well?
The issue said in one argument, I would be fine with two as well, do not really mind.

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -43,17 +44,41 @@ async def callback(update: Update, context: CallbackContext)
...
+    ):
+        super().__init__(callback, block=block)
+
+        if chat_id is not None:
+            if isinstance(chat_id, int):
+                chat_id = frozenset({chat_id})
+            else:
+                chat_id = frozenset(chat_id)
+
+        self.chat_ids = chat_id
```
Poolitzer, line 81: This should be a private attribute. You probably want to implement it with a setter, so you can check if either chat ids or usernames are already set, see https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/filters.py#L627-L641

Poolitzer (changes requested):
Thanks for the PR, good job implementing the tests.

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -43,17 +44,60 @@ async def callback(update: Update, context: CallbackContext)
 
             The return value of the callback is usually ignored except for the special case of
             :class:`telegram.ext.ConversationHandler`.
+        chat_id (:SCT[:obj:`int`], optional): Filters requests to allow only
```
Poolitzer: Ah, sadly using our shortcut does not work with sphinx, sorry for misleading you there.
``` `:obj:`int` | Collection[:obj:`int`], optional``` was right in the written type hint.

File: telegram/ext/_chatjoinrequesthandler.py [hunk 5]
```diff
@@ -43,17 +44,60 @@ async def callback(update: Update, context: CallbackContext)
...
+            those which are from a specified chat ID.
+
+            .. versionadded:: 20.0
+        username (:SCT[:obj:`str`], optional): Filters requests to allow only
+            those which are from a specified username.
```
Poolitzer: same as above

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -43,17 +44,60 @@ async def callback(update: Update, context: CallbackContext)
 
             The return value of the callback is usually ignored except for the special case of
             :class:`telegram.ext.ConversationHandler`.
+        chat_id (:SCT[:obj:`int`], optional): Filters requests to allow only
+            those which are from a specified chat ID.
```
Poolitzer, line 48: This only mentions one chat id, we add usually do (s), so ID(s). And then without the a.

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -43,17 +44,60 @@ async def callback(update: Update, context: CallbackContext)
...
     Attributes:
         callback (:term:`coroutine function`): The callback function for this handler.
+        _chat_ids (:frozenset(:obj:`int`), optional):
+            Which chat ID(s) to allow through.
+
+            .. versionadded:: 20.0
+        _usernames (:frozenset(:obj:`str`), optional):
+            Which username(s) to allow through.
+
+            .. versionadded:: 20.0
```
Poolitzer, line 68: The whole point behind making attributes private would be to hide it from the users, so no need to document them.

miles170:
The CI failure is not related to code changes.

tests/test_chatjoinrequesthandler.py
```diff
@@ -127,6 +127,28 @@ async def callback(self, update, context):
...
+        assert handler.check_update(chat_join_request_update)
+        handler = ChatJoinRequestHandler(self.callback, chat_id=[1])
+        assert handler.check_update(chat_join_request_update)
+
+        handler = ChatJoinRequestHandler(self.callback, chat_id=2)
+        assert not handler.check_update(chat_join_request_update)
+        handler = ChatJoinRequestHandler(self.callback, chat_id=[2])
+        assert not handler.check_update(chat_join_request_update)
+
+    def test_with_username(self, chat_join_request_update):
```
Poolitzer, line 141: that one misses a test when the user does not have a username. It wont break your code but it is important to have it in the tests for future changes.

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -43,17 +44,60 @@ async def callback(update: Update, context: CallbackContext)
 
             The return value of the callback is usually ignored except for the special case of
             :class:`telegram.ext.ConversationHandler`.
+        chat_id (SCT[:obj:`int`], optional): Filters requests to allow only
```
Poolitzer, line 47: I am sorry, I mislead you here. While the type hint for the argument is shortcutable, the sphinx docs don't work this way sadly. So you still have to do ```:obj:`int` | Collection[:obj:`int`]``` here.

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -43,17 +44,60 @@ async def callback(update: Update, context: CallbackContext)
...
+            those which are from a specified chat ID.
+
+            .. versionadded:: 20.0
+        username (SCT[:obj:`str`], optional): Filters requests to allow only
+            those which are from a specified username.
+
+            .. versionadded:: 20.0
```
Poolitzer, line 54: both points from above apply to this as well ofc

Poolitzer (changes requested):
Also, could you please not rebase/force push? It makes the review process much worse. Just make a new commit, its fine, all our PRs have lots of commits.

Poolitzer (approved):
LGTM

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -43,6 +44,14 @@ async def callback(update: Update, context: CallbackContext)
 
             The return value of the callback is usually ignored except for the special case of
             :class:`telegram.ext.ConversationHandler`.
+        chat_id (:obj:`int` | Collection[:obj:`int`], optional): Filters requests to allow only
+            those which are from specified chat ID(s).
```
Bibo-Joshi, line 48:
```suggestion
            those which are asking to join the specified chat ID(s).
```

or similar wording. My intention here is to make it more clear that `chat_join_request.chat.id` is filtered, not `chat_join_request.from_user.id` :)

Same for `username`

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -65,4 +101,12 @@ def check_update(self, update: object) -> bool:
             :obj:`bool`
 
         """
-        return isinstance(update, Update) and bool(update.chat_join_request)
+        if isinstance(update, Update) and update.chat_join_request:
+            chat = update.chat_join_request.chat
+            if self._chat_ids and chat.id not in self._chat_ids:
+                return False
```
Bibo-Joshi, line 107: consider a chat with username `@A` and id 1. if `'A' in `self._usernames`, but `1 not in `self._chat_ids`, the request will be rejected. Is this intended? CC @Poolitzer 

Whatever the decision is, I would ask to explicitly clarify that in the docs and also add a test case for that.

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -53,7 +62,34 @@ async def callback(update: Update, context: CallbackContext)
...
+    ):
+        super().__init__(callback, block=block)
+
+        if chat_id is not None:
+            if isinstance(chat_id, int):
+                chat_id = frozenset({chat_id})
+            else:
+                chat_id = frozenset(chat_id)
+
+        if username is not None:
```
Bibo-Joshi, line 85: usernames should be sanitized to add/drop a leading `@` IMO. Please also add a corresponding unit test that makes sure that both versions are accepted.

Bibo-Joshi (changes requested):
Hi, thanks also from my side for the PR!

telegram/ext/_chatjoinrequesthandler.py
Poolitzer: Dropp the @ please, telegram doesn't send it with the @ either.

telegram/ext/_chatjoinrequesthandler.py
Poolitzer: I would argue to implement it like an or filter. If either the username or the chat id correspond to one in appropriate set, set it to `True`; otherwise `False`. So change from current behavior.

But yes, documenting this clearly would be great.

AUTHORS.rst
```diff
@@ -70,6 +70,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `Loo Zheng Yuan <https://github.com/loozhengyuan>`_
 - `LRezende <https://github.com/lrezende>`_
 - `macrojames <https://github.com/macrojames>`_
+- `miles <https://github.com/miles170>`_
 - `Matheus Lemos <https://github.com/mlemosf>`_
 - `Michael Elovskikh <https://github.com/wronglink>`_
```
harshil21, line 75:
```suggestion
- `Matheus Lemos <https://github.com/mlemosf>`_
- `Michael Elovskikh <https://github.com/wronglink>`_
- `miles <https://github.com/miles170>`_
```
your name should be right after michael's

telegram/ext/_chatjoinrequesthandler.py
harshil21: nice! Important distinction imo

harshil21 (commented):
good work!

telegram/ext/_chatjoinrequesthandler.py
harshil21: there are type aliases in sphinx which would work, but we'd have to make it public first, which isn't sooo

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -18,16 +18,25 @@
...
+from telegram.ext._utils.types import CCT, HandlerCallback
 
 
 class ChatJoinRequestHandler(BaseHandler[Update, CCT]):
     """BaseHandler class to handle Telegram updates that contain
     :attr:`telegram.Update.chat_join_request`.
 
+    Note:
+        If either the :paramref:`username` or the :paramref:`chat_id` corresponds to one
+        in the appropriate, the request will be passed to this handler.
```
Bibo-Joshi, line 36:
```suggestion
        If neither of :paramref:`username` and the :paramref:`chat_id` are passed, this handlers accepts *any* join request. Otherwise, this handler accepts all requests to join chats for which the chat ID is listed in :paramref:`chat_id` or the username is listed in :paramref:`username`, or both.
```

Making this note a bit more verbose should avoid some confusion :)

Bibo-Joshi (approved):
Except the suggested reformulation, this LGTM :)

telegram/ext/_chatjoinrequesthandler.py
```diff
@@ -43,6 +54,15 @@ async def callback(update: Update, context: CallbackContext)
 
             The return value of the callback is usually ignored except for the special case of
             :class:`telegram.ext.ConversationHandler`.
+        chat_id (:obj:`int` | Collection[:obj:`int`], optional): Filters requests to allow only
+            those which are asking to join the specified chat ID(s).
+
+            .. versionadded:: 20.0
+        username (:obj:`str` | Collection[:obj:`str`], optional): Filters requests to allow only
+            those which are asking to join the specified username(s).
+            Leading ``'@'`` in username will be discarded.
```
Poolitzer, line 63: We dont mention this in filters, and I dont think its necessary. Doesn't matter what the user input, the code just handles it. So I would remove this comment.

Poolitzer (changes requested):
One last comment, code is good!

telegram/ext/_chatjoinrequesthandler.py
miles170: Remove `Leading ``'@'`` in username will be discarded.` from comment?

Poolitzer (approved):
LGTM

Bibo-Joshi:
Thank you for your contribution @miles170 and thanks also for your patience during the review process :)

Issue #3109, opened 2022-06-21
[Documentation]  Add valuable information to attribute docstrings of telegram classes


Bibo-Joshi:
The docs of Telegram classes are currently written such that the docstring of the arguments are more informative than the docstring of the respective attributes. E.g. for `Chat.has_private_forwards`, "Returned only in `get_chat()`" is only present in the [argument docstrings](https://docs.python-telegram-bot.org/en/v20.0a1/telegram.chat.html#telegram.Chat.params.has_private_forwards), not in the [attribute docstring](https://docs.python-telegram-bot.org/en/v20.0a1/telegram.chat.html#telegram.Chat.has_private_forwards). For `Message.text`, the length limitation is only documented for the argument, not the attribute.

This logic makes sense for classes that are to be instantiated by the user (like `BotCommand` or `InputMedia*`), but for classes that are rarely instantiated by the user like `Chat` and `Message` this makes it sometimes hard to find valuable information.

I suggest to revisit the documentation of the Telegram classes and either unify the docstrings of args & attributes or at least check which information would be worth adding to the attributes as well.

Contributions in this direction are very welcome! They also don't have to cover all Telegram classes at once - instead this can be done step by step. If you want to cover (parts of) this, please leave a comment below, so that we can avoid duplicate work.

PRs for this should be made against the `doc-fixes` branch. Please also check out the [contribution guide](https://docs.python-telegram-bot.org/en/v20.0a1/contributing.html), especially the part on how to [build the documentation locally](https://docs.python-telegram-bot.org/en/v20.0a1/contributing.html#documenting).

Classes are listed [below](https://github.com/python-telegram-bot/python-telegram-bot/issues/3109#issuecomment-1333572895) for convenience.

Bibo-Joshi:
You're very welcome to contribute! please leave a short comment about which classes you want to cover in a pull request :)

simonfongnt:
is this still available for a beginner?

Bibo-Joshi:
You're very welcome to contribute! There haven't been any PRs regarding this so far. please leave a short comment about which classes you want to cover in a pull request :)

simonfongnt:
i shall go for the `Message` class and unify the docstrings of args & attributes in the PR.

penguinpee:
Collision avoidance: I'll start at the bottom and take care of the _web family of classes.

Happy Hacktober! :beers:

Bibo-Joshi:
[...]

PR #3272, opened 2022-10-01
Fix typos in Filters

rglsk:
Refer to #3109, done:
* Updated method and class docstring to follow the same convention (start with new line)
* Fix couple of typos
* Fix typing override issue with `check_update` method (@Bibo-Joshi : Please let me know if this change should be cherry-picked to other branch to merge across master -> I'm happy to do if if the proposed suggestion fits you 😄  )

### Checklist for PRs

### DIFF
--- a/telegram/ext/filters.py
+++ b/telegram/ext/filters.py
@@ -116,7 +116,8 @@ DataDict = Dict[str, list]
 
 
 class BaseFilter:
-    """Base class for all Filters.
+    """
+    Base class for all Filters.
 
     Filters subclassing from this class can combined using bitwise operators:
 
@@ -234,7 +235,8 @@ class BaseFilter:
 
 
 class MessageFilter(BaseFilter):
-    """Base class for all Message Filters. In contrast to :class:`UpdateFilter`, the object passed
+    """
+    Base class for all Message Filters. In contrast to :class:`UpdateFilter`, the object passed
     to :meth:`filter` is :obj:`telegram.Update.effective_message`.
 
     Please see :class:`BaseFilter` for details on how to create custom filters.
@@ -257,7 +259,8 @@ class MessageFilter(BaseFilter):
 
     @abstractmethod
     def filter(self, message: Message) -> Optional[Union[bool, DataDict]]:
-        """This method must be overwritten.
+        """
+        This method must be overwritten.
 
         Args:
             message (:class:`telegram.Message`): The message that is tested.
@@ -269,7 +272,8 @@ class MessageFilter(BaseFilter):
 
 
 class UpdateFilter(BaseFilter):
-    """Base class for all Update Filters. In contrast to :class:`MessageFilter`, the object
+    """
+    Base class for all Update Filters. In contrast to :class:`MessageFilter`, the object
     passed to :meth:`filter` is an instance of :class:`telegram.Update`, which allows to create
     filters like :attr:`telegram.ext.filters.UpdateType.EDITED_MESSAGE`.
 
@@ -292,7 +296,8 @@ class UpdateFilter(BaseFilter):
 
     @abstractmethod
     def filter(self, update: Update) -> Optional[Union[bool, DataDict]]:
-        """This method must be overwritten.
+        """
+        This method must be overwritten.
 
         Args:
             update (:class:`telegram.Update`): The update that is tested.
@@ -304,7 +309,8 @@ class UpdateFilter(BaseFilter):
 
 
 class _InvertedFilter(UpdateFilter):
-    """Represents a filter that has been inverted.
+    """
+    Represents a filter that has been inverted.
 
     Args:
         f: The filter to invert.
@@ -330,7 +336,8 @@ class _InvertedFilter(UpdateFilter):
 
 
 class _MergedFilter(UpdateFilter):
-    """Represents a filter consisting of two other filters.
+    """
+    Represents a filter consisting of two other filters.
 
     Args:
         base_filter: Filter 1 of the merged filter.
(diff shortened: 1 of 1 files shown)


telegram/ext/filters.py
```diff
@@ -182,8 +183,16 @@ def __init__(self, name: str = None, data_filter: bool = False):
         self._name = self.__class__.__name__ if name is None else name
         self._data_filter = data_filter
 
-    def check_update(self, update: Update) -> Optional[Union[bool, DataDict]]:  # skipcq: PYL-R0201
-        """Checks if the specified update is a message."""
+    def check_update(self, update: Update) -> bool:  # skipcq: PYL-R0201
```
harshil21, line 186: the type hint change should be reverted. The subclasses of `BaseFilter`s `check_update` can return `DataDict` based on `data_filter` property. That's why pylint also complained in `UpdateFilter` and `MessageFilter`

telegram/ext/filters.py
```diff
@@ -153,7 +154,7 @@ class BaseFilter:
     If you want to create your own filters create a class inheriting from either
     :class:`MessageFilter` or :class:`UpdateFilter` and implement a ``filter()``
     method that returns a boolean: :obj:`True` if the message should be
-    handled, :obj:`False` otherwise.
+    handled, otherwise :obj:`False`.
```
harshil21, line 157: I think the previous one made more sense

harshil21 (changes requested):

telegram/ext/filters.py
rglsk: @harshil21 Are you sure? the base method can only return the True of False 🤔 The methods which overrides this one can return the `DataDict` 🤔 
Screenshot 2022-10-01 at 16 21 31

telegram/ext/filters.py
rglsk: ✅ Done

telegram/ext/filters.py
harshil21: imo, base method should have same sig as children (if possible). Let's see what others think

rglsk: @harshil21 I reverted the change for now, we can do it in another PR to get others opinions on this specific case later 👍

telegram/ext/filters.py
```diff
@@ -183,7 +184,15 @@ def __init__(self, name: str = None, data_filter: bool = False):
...
     def check_update(self, update: Update) -> Optional[Union[bool, DataDict]]:  # skipcq: PYL-R0201
-        """Checks if the specified update is a message."""
+        """
+        Checks if the specified update is a message.
+
+        Args:
+            update (:class:`telegram.Update`): Incoming update.
+
+        Returns:
+            :obj:`bool`
```
harshil21, line 194:
```suggestion
            :obj:`bool` | :obj:`Dict[str, list]`
```

harshil21 (commented):
let's also wait for other reviews. And, please avoid force pushing since it makes it harder to review

telegram/ext/filters.py
```diff
@@ -183,7 +184,15 @@ def __init__(self, name: str = None, data_filter: bool = False):
...
     def check_update(self, update: Update) -> Optional[Union[bool, DataDict]]:  # skipcq: PYL-R0201
-        """Checks if the specified update is a message."""
+        """
+        Checks if the specified update is a message.
+
+        Args:
+            update (:class:`telegram.Update`): Incoming update.
+
+        Returns:
+            :obj:`bool` | :obj:`Dict[str, list]`
```
harshil21, line 194:
```suggestion
            :obj:`bool` | Dict[str, list]
```
sorry, forgot that the previous one fails the docs

telegram/ext/filters.py
Bibo-Joshi: I agree with harshil here.

telegram/ext/filters.py
```diff
@@ -183,7 +184,15 @@ def __init__(self, name: str = None, data_filter: bool = False):
...
     def check_update(self, update: Update) -> Optional[Union[bool, DataDict]]:  # skipcq: PYL-R0201
-        """Checks if the specified update is a message."""
+        """
+        Checks if the specified update is a message.
+
+        Args:
+            update (:class:`telegram.Update`): Incoming update.
+
+        Returns:
+            :obj:`bool` | Dict[str, list]
```
Bibo-Joshi, line 194:
```suggestion
            :obj:`bool` | Dict[:obj:`str`, :obj:`list`]
```

telegram/ext/filters.py
```diff
@@ -116,7 +116,8 @@
 
 
 class BaseFilter:
-    """Base class for all Filters.
+    """
+    Base class for all Filters.
```
Bibo-Joshi, line 120: TBH I don't really see any value in these changes - in this PR or even at all, that is.
True, some style guides suggest to start the docstring on a new line, but none of the styling tools that we currently use will report on these things.  __If__ we want to have docstrings start on a new line, I would rather like to make it a standalone issue to apply this to the whole code base.
I'm inclined to ask to revert these changes. Let's first hear what others think, though.

Bibo-Joshi (changes requested):

telegram/ext/filters.py
harshil21: I also agree to revert, they are rendered the same and [PEP 257](https://peps.python.org/pep-0257/) does not comment on this design.

Bibo-Joshi:
PS: you mentioned #3109 in all your PRs so far - I'd like to point out that none of your PRs addressed #3109 in any way 😉

rglsk:
Alright folks, sorry for that, I'll try better next time 🙏

PR #3275, opened 2022-10-01
Point to correct "edit on github" link

harshil21:
Uses the new furo `source_edit_link` to fix the edit button link at the top of every page


harshil21:
As I thought, this doesn't work as we want since it inserts the source name e.g. `telegram.bot.rst` into the `{filename}`. See the [source](https://github.com/pradyunsg/sphinx-basic-ng/blob/main/src/sphinx_basic_ng/theme/basic-ng/components/edit-this-page.html#L5). 

Closing as wont fix

PR #3276, opened 2022-10-01
Add `recursive` parameter to `TO.to_dict()`

### LINKED ISSUE #3146: [FEATURE] `TelegramObject.to_dict(recursion=value)`

Have a dict representation of `TelegramObject` without having the attributes also converted to dicts

Make `TO.to_dict` accept a parameter `recursion` that specifies up to which level attributes should be converted to dict. Logic:

* `recursion=None` (default) recurs infinitely
* `0` gives a dict representation but doesn't convert attributes to dicts
* `1` converts attributes, but not attributes of attributes
* etc

Alternatively, a simple `recursion=True/False` could be a step in the same direction with, having the same effect as `None`/`0` described above

### Additional context

Would e.g. allow to do `ptbcontrib.send_by_kwargs(bot, chat_id=chat_id, **message.to_dict(recursion=0)`. See also https://github.com/python-telegram-bot/ptbcontrib/issues/39 which was implemented only partly

Just a wild idea so far and of no major importance to me :)

harshil21:
Closes #3146 

I went with the simple boolean approach instead of specifying a number for the recursion. 
Also based on the TO-overhaul branch since I had to make sure these changes work with that too. (Merge after #3233)

### Checklist for PRs

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -8205,7 +8205,7 @@ CUSTOM_EMOJI_IDENTIFIER_LIMIT` custom emoji identifiers can be specified.
             api_kwargs=api_kwargs,
         )
 
-    def to_dict(self, recursive: bool = True) -> JSONDict:
+    def to_dict(self, recursive: bool = True) -> JSONDict:  # skipcq: PYL-W0613
         """See :meth:`telegram.TelegramObject.to_dict`."""
         data: JSONDict = {"id": self.id, "username": self.username, "first_name": self.first_name}
 
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -8205,7 +8205,7 @@ CUSTOM_EMOJI_IDENTIFIER_LIMIT` custom emoji identifiers can be specified.
             api_kwargs=api_kwargs,
         )
 
-    def to_dict(self) -> JSONDict:
+    def to_dict(self, recursive: bool = True) -> JSONDict:
         """See :meth:`telegram.TelegramObject.to_dict`."""
         data: JSONDict = {"id": self.id, "username": self.username, "first_name": self.first_name}
 
--- a/telegram/_telegramobject.py
+++ b/telegram/_telegramobject.py
@@ -170,10 +170,6 @@ class TelegramObject:
         """
         data = {}
 
-        # __dict__ has attrs from superclasses, so no need to loop through them
-        if hasattr(self, "__dict__"):
-            data.update(self.__dict__)  # important when class has no __slots__.
-
         # We want to get all attributes for the class, using self.__slots__ only includes the
         # attributes used by that class itself, and not its superclass(es). Hence, we get its MRO
         # and then get their attributes. The `[:-1]` slice excludes the `object` class
--- a/telegram/_telegramobject.py
+++ b/telegram/_telegramobject.py
@@ -170,29 +170,28 @@ class TelegramObject:
         """
         data = {}
 
-        try:
-            # __dict__ has attrs from superclasses, so no need to recompute/duplicate
-            data.update(self.__dict__)
-        except AttributeError:
-            pass
+        # __dict__ has attrs from superclasses, so no need to recompute/duplicate
+        if hasattr(self, "__dict__"):
+            data.update(self.__dict__)  # important when class has no __slots__.
 
+        org_data = tuple(data)  # make a copy of the keys to avoid changing length during iteration
         # We want to get all attributes for the class, using self.__slots__ only includes the
         # attributes used by that class itself, and not its superclass(es). Hence, we get its MRO
         # and then get their attributes. The `[:-1]` slice excludes the `object` class
-        for cls in self.__class__.__mro__[:-1]:
-            # chain the class's slots with the user defined subclass __dict__ (class has no slots)
-            for key in chain(cls.__slots__, data):  # type: ignore[attr-defined]
-                if not include_private and key.startswith("_"):
-                    continue
+        all_slots = (slot for cls in self.__class__.__mro__[:-1] for slot in cls.__slots__)
+        # chain the class's slots with the user defined subclass __dict__ (class has no slots)
+        for key in chain(org_data, all_slots):
+            if not include_private and key.startswith("_"):
+                continue
 
-                value = getattr(self, key, None)
-                if value is not None:
-                    if recursive and hasattr(value, "to_dict"):
-                        data[key] = value.to_dict(recursive=True)  # pylint: disable=no-member
-                    else:
-                        data[key] = value
-                elif not recursive:
+            value = getattr(self, key, None)
+            if value is not None:
+                if recursive and hasattr(value, "to_dict"):
+                    data[key] = value.to_dict(recursive=True)  # pylint: disable=no-member
+                else:
                     data[key] = value
+            elif not recursive:
+                data[key] = value
 
         if recursive and data.get("from_user"):
             data["from"] = data.pop("from_user", None)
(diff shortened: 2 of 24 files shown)


telegram/_telegramobject.py
```diff
@@ -169,17 +169,19 @@ def _get_attrs(
...
+            pass
+
+        tuple_data = tuple(data)
         # We want to get all attributes for the class, using self.__slots__ only includes the
         # attributes used by that class itself, and not its superclass(es). Hence, we get its MRO
         # and then get their attributes. The `[:-1]` slice excludes the `object` class
         for cls in self.__class__.__mro__[:-1]:
-            for key in cls.__slots__:  # type: ignore[attr-defined]
+            # add the class's slots with the user defined subclass __dict__ (class has no slots)
+            for key in tuple(cls.__slots__) + tuple_data:  # type: ignore[attr-defined]
```
Bibo-Joshi, line 184: maybe we can use `set` instead? we don't need immutability or ordering, right?

Also `recursive` should be passed to `to_dict` in line 191 below, right? For the tests to cover that, they probably need one more layer of nesting …

tests/test_telegramobject.py
```diff
@@ -138,6 +138,26 @@ def test_to_dict_api_kwargs(self):
         to = TelegramObject(api_kwargs={"foo": "bar"})
         assert to.to_dict() == {"foo": "bar"}
 
+    def test_to_dict_recursion(self):
+        class Recusive(TelegramObject):
+            def __init__(self):
+                super().__init__()
+                self.recursive = "recursive"
+
+        class SubClass(TelegramObject):
```
Bibo-Joshi, line 147: maybe we can add `__slots__` to one of these two and a comment "doesn't have `__slots__`, so has `__dict__`" to the other? just to emphasize that both should work :)

Bibo-Joshi (commented):
Thanks for the nice PR :)

telegram/_telegramobject.py
harshil21: > we can use set instead? 

I benchmarked this and found tuple to be faster.. (on 3.11)
``` python
>>> timeit.timeit("set(a)|set(b)", setup="a=(k for k in range(50)); b={k:k for k in range(10)}")
0.2501877999984572
>>> timeit.timeit("tuple(a)+tuple(b)", setup="a=(k for k in range(50)); b={k:k for k in range(10)}")
0.12552469999900495
```

> Also recursive should be passed to to_dict in line 191 

yes, but the default to recurse is `True` so wouldn't really matter in this case

Bibo-Joshi: > > we can use set instead? 
> 
> I benchmarked this and found tuple to be faster.. on 3.11: 
> ``` python
> >>> timeit.timeit("set(a)|set(b)", setup="a=(k for k in range(50)); b={k:k for k in range(10)}")
> 0.2501877999984572
> >>> timeit.timeit("tuple(a)+tuple(b)", setup="a=(k for k in range(50)); b={k:k for k in range(10)}")
> 0.12552469999900495
> ```
> 3.10 has a smaller difference - 0.26s for set, 0.18s for tuple

Mh, the more interesting thing IMO would be how fast loops go through tuples and sets. I also just recalled `itertools.chain` - with that we don't even have to built tuples/sets but can just use `chain(__dict__, __slots__)`. Ought to be faster ...

> > Also recursive should be passed to to_dict in line 191 
> 
> yes, but the default to recurse is `True` so wouldn't really matter in this case. So no need for additional tests

A right. Maybe still explicitly pass `recursive=True`? For confused future me? 😄

harshil21: itertools.chain is indeed faster. 
``` python
>>> timeit.timeit("for i in tuple(a)+tuple(b):pass", setup="a=(k for k in range(50)); b={k:k for k in range(10)}")
0.198202909000031
>>> timeit.timeit("for i in itertools.chain(a, b):pass", setup="a=(k for k in range(50)); b={k:k for k in range(10)};")
0.18076100700000097
>>> timeit.timeit("for i in set(a)|set(b):pass", setup="a=(k for k in range(50)); b={k:k for k in range(10)}")
0.3299522259999321
```

telegram/_telegramobject.py
```diff
@@ -169,28 +170,28 @@ def _get_attrs(
         """
         data = {}
 
-        if not recursive:
-            try:
-                # __dict__ has attrs from superclasses, so no need to put in the for loop below
-                data.update(self.__dict__)
-            except AttributeError:
-                pass
+        # __dict__ has attrs from superclasses, so no need to recompute/duplicate
```
Bibo-Joshi, line 173: what exactly do you mean by "recompute/duplicate" here? If the parent has `__slots__`, then those aren't included in `__dict__` …

telegram/_telegramobject.py
```diff
@@ -169,28 +170,28 @@ def _get_attrs(
...
+        if hasattr(self, "__dict__"):
+            data.update(self.__dict__)  # important when class has no __slots__.
+
+        org_data = tuple(data)  # make a copy of the keys to avoid changing length during iteration
```
Bibo-Joshi, line 177: this is only non-empty if `self.__dict__` is present, right? why not just use `self.__dict__` again instead of creating this tuple?

telegram/_telegramobject.py
```diff
@@ -279,16 +280,23 @@ def to_json(self) -> str:
...
 
-    def to_dict(self) -> JSONDict:
+    def to_dict(self, recursive: bool = True) -> JSONDict:
         """Gives representation of object as :obj:`dict`.
 
         .. versionchanged:: 20.0
             Now includes all entries of :attr:`api_kwargs`.
 
+        Args:
+            recursive (:obj:`bool`, optional): If :obj:`True`, will convert any TelegramObjects
```
Bibo-Joshi, line 290:
```suggestion
            recursive (:obj:`bool`, optional): If :obj:`True`, will convert any ``TelegramObjects``
```

tests/test_telegramobject.py
```diff
@@ -138,6 +138,28 @@ def test_to_dict_api_kwargs(self):
...
 
+    def test_to_dict_recursion(self):
+        class Recursive(TelegramObject):
+            __slots__ = ("recursive",)
+
+            def __init__(self):
+                super().__init__()
+                self.recursive = "recursive"
+
+        class SubClass(TelegramObject):  # doesn't have `__slots__`, so has `__dict__` instead.
```
Bibo-Joshi, line 149: comment could be a docstring :)

Bibo-Joshi (commented):
two merge questions :)

telegram/_telegramobject.py
harshil21: hmm I guess I can revert that comment

telegram/_telegramobject.py
```diff
@@ -160,37 +161,36 @@ def _get_attrs(
...
-        if not recursive:
-            try:
-                # __dict__ has attrs from superclasses, so no need to put in the for loop below
-                data.update(self.__dict__)
-            except AttributeError:
-                pass
+        # __dict__ has attrs from superclasses, so no need to loop through them
+        if hasattr(self, "__dict__"):
+            data.update(self.__dict__)  # important when class has no __slots__.
+
```
Bibo-Joshi, line 176: isn't this already covered by the below loop now? sorry for still harping on this point 😬

harshil21: haha yes. Looks like we're optimizing this even more

telegram/_bot.py
```diff
@@ -8205,7 +8205,7 @@ async def create_invoice_link(
             api_kwargs=api_kwargs,
         )
 
-    def to_dict(self) -> JSONDict:
+    def to_dict(self, recursive: bool = True) -> JSONDict:
```
Bibo-Joshi, line 8208:
```suggestion
    def to_dict(self, recursive: bool = True) -> JSONDict:  # skipcq: PYL-W0613
```

to shut up DS …

PR #3267, opened 2022-09-30
Optional 3rd Party dependencies

Bibo-Joshi:
Makes all 3rd party dependencies optional, i.e. addresses https://github.com/python-telegram-bot/python-telegram-bot/projects/7#card-77767364.

Currently based against the `read-only-cdc` branch, b/c that must be merged first and the diff is better readible this way.

### Checklist for PRs

- [x] Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)
- [x] Created new or adapted existing unit tests
- [x] Documented code changes according to the [CSI standard](https://standards.mousepawmedia.com/en/stable/csi.html)
- [ ] ~Added myself alphabetically to `AUTHORS.rst` (optional)~
- [ ] ~Added new classes & modules to the docs and all suitable `__all__` s~

PR #3283, opened 2022-10-04
Run Tests in Parallel

harshil21:
Runs the test suite in parallel by using the `pytest-xdist` plugin. In my case, the test suite went from taking ~14 mins to complete to just 3m 28s!
 That's roughly a **75%** improvement!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,19 +17,8 @@ jobs:
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11.0-rc.2']
         os: [ubuntu-latest, windows-latest, macos-latest]
-        experimental: [false]
-        include:
-          - python-version: 3.11.0-rc.2
-            os: ubuntu-latest
-            experimental: true
-          - python-version: 3.11.0-rc.2
-            os: windows-latest
-            experimental: true
-          - python-version: 3.11.0-rc.2
-            os: macos-latest
-            experimental: true
       fail-fast: False
     steps:
       - uses: actions/checkout@v3
@@ -53,15 +42,16 @@ jobs:
         # 4. Test everything else
         # The first & second one are achieved by mocking the corresponding import
         # See test_helpers.py & test_no_passport.py for details
+        # `--dist loadfile` distributes each test file to a different worker
         run: |
-          pytest -v --cov -k test_no_passport.py
+          pytest -v --cov tests/test_no_passport.py
           no_passport_exit=$?
           export TEST_PASSPORT='true'
-          pytest -v --cov --cov-append -k test_helpers.py
+          pytest -v --cov --cov-append tests/test_helpers.py
           no_pytz_exit=$?
           export TEST_PYTZ='true'
           pip uninstall aiolimiter -y
-          pytest -v --cov --cov-append -k test_ratelimiter.py
+          pytest -v --cov --cov-append tests/test_ratelimiter.py
           no_rate_limiter_exit=$?
           export TEST_RATE_LIMITER='true'
           pip install -r requirements-opts.txt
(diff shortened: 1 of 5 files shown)


requirements-dev.txt
```diff
@@ -3,6 +3,8 @@ pre-commit
 pytest==7.1.3
 pytest-asyncio==0.19.0
 pytest-timeout==2.1.0 # used to timeout tests
+pytest-xdist==2.5.0  # used to run tests in parallel
+pytest-xdist[psutil]  # used for better auto detection of cpu cores to run in parallel
```
Bibo-Joshi, line 7:
```
pytest-xdist[psutil]==2.5.0
```
doesn't that work?

tests/test_datetime.py
```diff
@@ -31,8 +31,8 @@
 from tests.conftest import env_var_2_bool
 
 ABSOLUTE_TIME_SPECS = [
-    dtm.datetime.now(tz=dtm.timezone(dtm.timedelta(hours=-7))),
-    dtm.datetime.utcnow(),
+    dtm.datetime.now(tz=dtm.timezone(dtm.timedelta(hours=-7))).replace(second=0, microsecond=0),
+    dtm.datetime.utcnow().replace(second=0, microsecond=0),
```
Bibo-Joshi, line 35: why?

requirements-dev.txt
harshil21: hm, didn't think chaining these would work, but it seems that it does

tests/test_datetime.py
harshil21: when initializing pytest, each worker got a slightly different value for this, hence it showed an error saying that it got different items in this list.

Bibo-Joshi: ah, interesting!

.github/workflows/test.yml
```diff
@@ -53,6 +41,7 @@ jobs:
         # 4. Test everything else
         # The first & second one are achieved by mocking the corresponding import
         # See test_helpers.py & test_no_passport.py for details
+        # `--dist loadfile` distributes each test file to a different worker
```
Bibo-Joshi, line 44:
```suggestion
```

.github/workflows/test.yml
```diff
@@ -65,7 +54,7 @@ jobs:
           no_rate_limiter_exit=$?
           export TEST_RATE_LIMITER='true'
           pip install -r requirements-opts.txt
-          pytest -v --cov --cov-append
+          pytest -v --cov --cov-append -n auto --dist loadfile
```
Bibo-Joshi, line 57:
```suggestion
          # `-n auto --dist loadfile` uses pytest-xdist to run each test file on a different CPU worker
          pytest -v --cov --cov-append -n auto --dist loadfile
```

requirements-dev.txt
```diff
@@ -3,6 +3,7 @@ pre-commit
 pytest==7.1.3
 pytest-asyncio==0.19.0
 pytest-timeout==2.1.0 # used to timeout tests
+pytest-xdist[psutil]==2.5.0  # xdist runs tests in parallel, psutil for better cpu no. detection
```
Bibo-Joshi, line 6: I re-checked and psutil is not pure-python so that installation may not be available on all setups. Maybe skip the requirement here and instead manually install it in the test workflow? that way we avoid headaches for contribs trying to install the dev reqs

Bibo-Joshi (changes requested):
Sorry for getting back to this late.

The speedup in the CI doesn't seem as big as for you locally, but it's nice nonetheless. Tests seem to run rather stable, too, so let's add it :) Can you also add a note to the contrib guide that tells users how they can use pytest-xdist? Probably also add a warning that this *may* effect the test results in rare edge cases :D

.github/workflows/test.yml
harshil21: can we add comments in that block?

Bibo-Joshi: Yes, see [here](https://github.com/python-telegram-bot/python-telegram-bot/pull/3267/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88)

Bibo-Joshi (approved):
LGTM 👍 if there's nothing left from your side, I can merge once the tests pass :)

PR #3295, opened 2022-10-16
feat(`Bot`): add shortcut to pass caption to media group

### LINKED ISSUE #3185: [FEATURE] Add `caption` parameter to functions that consume a media group

For displaying a caption on a media group, one needs to set the `caption` parameter on the first element of the  list. 

I'd like to propose a _shortcut_ parameter for the methods that consume a media group; e.g. `reply_media_group` which will take a `caption` parameter and apply it on the first element of the list.

- What happens if the first item already has a caption present? Do we overwrite it?
- If `caption` is being passed, then `parse_mode` will need to be as well 

Let me know if this feature is desirable and I'll get started on making a PR.

_No response_

### Additional context

<details>
  <summary>Example of a media group with a caption</summary>
  
image
  
Only the first item in the list needs to have a caption for the whole media group to display it.
</details>

lemontree210:
closes #3185

Adds 3 keyword-only arguments to pass caption to entire media group.

All tests pass (locally) including `test_official.py`.

All pre-commit hooks passed except for `pylint` that found errors in code that I hadn't changed (cyclic import in `telegram/ext/_chatjoinrequesthandler.py`).

I'm not sure if I have to: 
- [ ] Add`.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes) (_I changed doctrings themselves, just not sure about these `version...` notes_)
- [ ] Add more tests, e.g. for shortcuts (`Chat`, `User`, `Message`)

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -2078,7 +2078,6 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         ):
             raise ValueError("You can only supply either group caption or media with captions.")
 
-        changed_media = None
         if caption:
             # Apply group caption to the first media item.
             # This will lead to the group being shown with this caption.
@@ -2089,12 +2088,12 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
             item_to_get_caption.caption_entities = caption_entities
 
             # copy the list (just the references) to avoid mutating the original list
-            changed_media = media[:]
-            changed_media[0] = item_to_get_caption
+            media = media.copy()
+            media[0] = item_to_get_caption
 
         data: JSONDict = {
             "chat_id": chat_id,
-            "media": changed_media if caption else media,
+            "media": media,
             "disable_notification": disable_notification,
             "allow_sending_without_reply": allow_sending_without_reply,
             "protect_content": protect_content,
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -2076,18 +2076,20 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
             raise ValueError("You can only supply either group caption or media with captions.")
 
         if caption:
-            # Apply group caption to the first media item.
+            # Create copies of objects in media and apply group caption to the first item.
             # This will lead to the group being shown with this caption.
-            # Copy first item to avoid mutation of original object
-            item_to_get_caption = copy.copy(media[0])
+
+            # Only copying of first item should be enough, but the behaviour of Defaults is such
+            # that a bot with a default parse mode will automatically set `parse_mode` for
+            # each item to default parse mode of the bot.
+            # This will throw ValueError at subsequent calls with same media items.
+            media = [copy.copy(item) for item in media]
+
+            item_to_get_caption = media[0]
             item_to_get_caption.caption = caption
             if parse_mode is not DEFAULT_NONE:
                 item_to_get_caption.parse_mode = parse_mode
             item_to_get_caption.caption_entities = caption_entities
-
-            # copy the list (just the references) to avoid mutating the original list
-            media = media.copy()
-            media[0] = item_to_get_caption
 
         data: JSONDict = {
             "chat_id": chat_id,
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -1999,20 +1999,18 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         connect_timeout: ODVInput[float] = DEFAULT_NONE,
         pool_timeout: ODVInput[float] = DEFAULT_NONE,
         api_kwargs: JSONDict = None,
-        group_caption: Optional[str] = None,
-        # Notes on group_caption_parse_mode:
-        # 1. To be fully consistent with media items it would be called
-        # `group_parse_mode`, but that is ambiguous.
-        # 2. Typing ODVInput[str] = DEFAULT_NONE will make tests for shortcuts fail
+        caption: Optional[str] = None,
+        # Note on parse_mode:
+        # Typing ODVInput[str] = DEFAULT_NONE will make tests for shortcuts fail
         # (check_defaults_handling) despite it being used for parse mode for individual captions.
-        group_caption_parse_mode: Optional[str] = None,
-        group_caption_entities: Union[List["MessageEntity"], Tuple["MessageEntity", ...]] = None,
+        parse_mode: Optional[str] = None,
+        caption_entities: Union[List["MessageEntity"], Tuple["MessageEntity", ...]] = None,
     ) -> List[Message]:
         """Use this method to send a group of photos or videos as an album.
 
         Note:
-            If you supply a :paramref:`group_caption` (along with either
-            :paramref:`group_caption_parse_mode` or :paramref:`group_caption_entities`),
+            If you supply a :paramref:`caption` (along with either
+            :paramref:`parse_mode` or :paramref:`caption_entities`),
             then items in :paramref:`media` must have no captions, and vice verca.
 
         .. seealso:: :attr:`telegram.Message.reply_media_group`,
@@ -2051,18 +2049,18 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
                 :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
             api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
                 Telegram API.
-            group_caption (:obj:`str`, optional): Caption that will be added to the
+            caption (:obj:`str`, optional): Caption that will be added to the
                 first element of :paramref:`media`, so that it will be used as caption for the
                 whole media group.
                 Defaults to :obj:`None`.
-            group_caption_parse_mode (:obj:`str` | :obj:`None`, optional):
-                Parse mode for :paramref:`group_caption`.
+            parse_mode (:obj:`str` | :obj:`None`, optional):
+                Parse mode for :paramref:`caption`.
                 See the constants in :class:`telegram.constants.ParseMode` for the
                 available modes.
                 Defaults to :obj:`None`.
-            group_caption_entities (List[:class:`telegram.MessageEntity`], optional):
-                List of special entities for :paramref:`group_caption`,
-                which can be specified instead of :paramref:`group_caption_parse_mode`.
+            caption_entities (List[:class:`telegram.MessageEntity`], optional):
+                List of special entities for :paramref:`caption`,
+                which can be specified instead of :paramref:`parse_mode`.
                 Defaults to :obj:`None`.
 
         Returns:
@@ -2071,7 +2069,7 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         Raises:
             :class:`telegram.error.TelegramError`
         """
-        if group_caption and any(
+        if caption and any(
             [
                 any(item.caption for item in media),
                 any(item.caption_entities for item in media),
@@ -2081,14 +2079,14 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
             raise ValueError("You can only supply either group caption or media with captions.")
 
         changed_media = None
-        if group_caption:
+        if caption:
             # Apply group caption to the first media item.
             # This will lead to the group being shown with this caption.
             # Copy first item to avoid mutation of original object
             item_to_get_caption = copy.copy(media[0])
-            item_to_get_caption.caption = group_caption
-            item_to_get_caption.parse_mode = group_caption_parse_mode
-            item_to_get_caption.caption_entities = group_caption_entities
+            item_to_get_caption.caption = caption
+            item_to_get_caption.parse_mode = parse_mode
+            item_to_get_caption.caption_entities = caption_entities
 
             # copy the list (just the references) to avoid mutating the original list
             changed_media = media[:]
@@ -2096,7 +2094,7 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
 
         data: JSONDict = {
             "chat_id": chat_id,
-            "media": changed_media if group_caption else media,
+            "media": changed_media if caption else media,
             "disable_notification": disable_notification,
             "allow_sending_without_reply": allow_sending_without_reply,
             "protect_content": protect_content,
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -2054,12 +2054,16 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
             group_caption (:obj:`str`, optional): Caption that will be added to the
                 first element of :paramref:`media`, so that it will be used as caption for the
                 whole media group.
-                Defaults to :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
+                Defaults to :obj:`None`.
             group_caption_parse_mode (:obj:`str` | :obj:`None`, optional):
                 Parse mode for :paramref:`group_caption`.
-                Defaults to :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
+                See the constants in :class:`telegram.constants.ParseMode` for the
+                available modes.
+                Defaults to :obj:`None`.
             group_caption_entities (List[:class:`telegram.MessageEntity`], optional):
-                List of special entities for :paramref:`group_caption`.
+                List of special entities for :paramref:`group_caption`,
+                which can be specified instead of :paramref:`group_caption_parse_mode`.
+                Defaults to :obj:`None`.
 
         Returns:
             List[:class:`telegram.Message`]: An array of the sent Messages.
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -2051,14 +2051,15 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
                 :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
             api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
                 Telegram API.
-            group_caption (:obj:`str`, optional): Text of the caption
-                that will be shown for the whole media group.
+            group_caption (:obj:`str`, optional): Caption that will be added to the
+                first element of :paramref:`media`, so that it will be used as caption for the
+                whole media group.
                 Defaults to :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
             group_caption_parse_mode (:obj:`str` | :obj:`None`, optional):
-                Parse mode of the caption that will be shown for the whole media group.
+                Parse mode for :paramref:`group_caption`.
                 Defaults to :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
             group_caption_entities (List[:class:`telegram.MessageEntity`], optional):
-                List of special entities that appear in the group caption.
+                List of special entities for :paramref:`group_caption`.
 
         Returns:
             List[:class:`telegram.Message`]: An array of the sent Messages.
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -2073,7 +2073,8 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
             [
                 any(item.caption for item in media),
                 any(item.caption_entities for item in media),
-                any(item.parse_mode for item in media),
+                # if parse_mode was set explicitly, even to None, error must be raised
+                any(item.parse_mode is not DEFAULT_NONE for item in media),
             ]
         ):
             raise ValueError("You can only supply either group caption or media with captions.")
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -2012,7 +2012,7 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
 
         Note:
             If you supply a :paramref:`group_caption` (along with either
-            :paramref:`group_parse_mode` or :paramref:`group_caption_entities`),
+            :paramref:`group_caption_parse_mode` or :paramref:`group_caption_entities`),
             then items in :paramref:`media` must have no captions, and vice verca.
 
         .. seealso:: :attr:`telegram.Message.reply_media_group`,
@@ -2066,8 +2066,14 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         Raises:
             :class:`telegram.error.TelegramError`
         """
-        if group_caption and any(item.caption for item in media):
-            raise ValueError("You can only supply either group caption or images with captions.")
+        if group_caption and any(
+            [
+                any(item.caption for item in media),
+                any(item.caption_entities for item in media),
+                any(item.parse_mode is DEFAULT_NONE for item in media),
+            ]
+        ):
+            raise ValueError("You can only supply either group caption or media with captions.")
 
         # First object could be mutated if group caption has to be applied to it,
         # so it's best to copy media to avoid side effects.
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -2073,7 +2073,7 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
             [
                 any(item.caption for item in media),
                 any(item.caption_entities for item in media),
-                any(item.parse_mode is DEFAULT_NONE for item in media),
+                any(item.parse_mode for item in media),
             ]
         ):
             raise ValueError("You can only supply either group caption or media with captions.")
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -2000,10 +2000,7 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         pool_timeout: ODVInput[float] = DEFAULT_NONE,
         api_kwargs: JSONDict = None,
         caption: Optional[str] = None,
-        # Note on parse_mode:
-        # Typing ODVInput[str] = DEFAULT_NONE will make tests for shortcuts fail
-        # (check_defaults_handling) despite it being used for parse mode for individual captions.
-        parse_mode: Optional[str] = None,
+        parse_mode: ODVInput[str] = DEFAULT_NONE,
         caption_entities: Union[List["MessageEntity"], Tuple["MessageEntity", ...]] = None,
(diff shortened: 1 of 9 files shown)


lemontree210:
Just a side note: only one check failed for penultimate commit 07b175eae7341ccb6ddf0dd12a1cd03ce3b6c3e0 (`test_official.py` failed), so I don't think I introduced something that caused so many tests to fail in my last commit.

telegram/_bot.py
```diff
@@ -1998,9 +1999,22 @@ async def send_media_group(
         connect_timeout: ODVInput[float] = DEFAULT_NONE,
         pool_timeout: ODVInput[float] = DEFAULT_NONE,
         api_kwargs: JSONDict = None,
+        group_caption: Optional[str] = None,
+        # Notes on group_caption_parse_mode:
+        # 1. To be fully consistent with media items it would be called
+        # `group_parse_mode`, but that is ambiguous.
+        # 2. Typing ODVInput[str] = DEFAULT_NONE will make tests for shortcuts fail
+        # (check_defaults_handling) despite it being used for parse mode for individual captions.
```
Bibo-Joshi, line 2007: Regarding naming, I'd also be okay with dropping the `group_` prefix. What ambiguity do you see with `group_parse_mode`?

Regarding defaults: I hadn't thought of that yet, tbh, but we should make sure that defaults are properly applied. IISC line 2008 doesn't need to be changed for that, but we should in any case add tests cases that make sure of it. If you need help with that, please feel free to reach out (unit tests can be tricky in general and for the defaults-stuff in particular :D) Maybe [this unit test](https://github.com/python-telegram-bot/python-telegram-bot/blob/eb818da93f3df2b439f5aef6065add232a5106e8/tests/test_inputmedia.py#L574-L584) can give you an idea how we currently tests the defaults setpu in the unit tests.

telegram/_bot.py
```diff
@@ -2037,16 +2051,40 @@ async def send_media_group(
                 :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
             api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
                 Telegram API.
+            group_caption (:obj:`str`, optional): Text of the caption
+                that will be shown for the whole media group.
+                Defaults to :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
```
Bibo-Joshi, line 2056:
```suggestion
                Defaults to :obj:`None`.
```

Or even skip it completely, as that's already implied by the `, optional)` :)

telegram/_bot.py
```diff
@@ -2037,16 +2051,40 @@ async def send_media_group(
...
+            group_caption_parse_mode (:obj:`str` | :obj:`None`, optional):
+                Parse mode of the caption that will be shown for the whole media group.
+                Defaults to :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
```
Bibo-Joshi, line 2059:
```suggestion
                Defaults to :obj:`None`.
```
So far, we rarely point to the `Defaults` setup in the docs of the bot methods. Instead, we could leave a note in the docs of the `Defaults.parse_mode` parameter stating that it also applies here :)

telegram/_bot.py
```diff
@@ -2037,16 +2051,40 @@ async def send_media_group(
                 :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
             api_kwargs (:obj:`dict`, optional): Arbitrary keyword arguments to be passed to the
                 Telegram API.
+            group_caption (:obj:`str`, optional): Text of the caption
+                that will be shown for the whole media group.
```
Bibo-Joshi, line 2055: Maybe "the caption is added to the first element of :paramref:`media`, such that it will be used as caption for the whole media group"?

telegram/_bot.py
```diff
@@ -2037,16 +2051,40 @@ async def send_media_group(
...
+                Defaults to :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.
+            group_caption_parse_mode (:obj:`str` | :obj:`None`, optional):
+                Parse mode of the caption that will be shown for the whole media group.
```
Bibo-Joshi, line 2058:
```suggestion
                Parse mode for :paramref:`group_caption`.
```
similar for `group_caption_entities`

telegram/ext/_extbot.py
```diff
@@ -2242,7 +2242,20 @@ async def send_media_group(
...
+        """Shortcut for::
+
+             await bot.send_media_group(update.effective_chat.id, *args, **kwargs)
+
+        For the documentation of the arguments, please see :meth:`telegram.Bot.send_media_group`.
+
+        Returns:
+            List[:class:`telegram.Message`]: On success, instance representing the message posted.
+
+        """
```
Bibo-Joshi, line 2258: This docstring is wrong. please remove.

Bibo-Joshi (changes requested):
Thank you very much for your nice PR! I left a number of comments - but most are actually just nit-picking on the documentation :D

telegram/_bot.py
lemontree210: Thank you for the feedback and the encouraging comment! :blush:

I'll start working on this and other comments, but a quick note regarding `parse_mode`. I thought that in a context of a single media item `parse_mode` was sort of more obvious than in a context of a group. Maybe `parse_mode` should also be `caption_parse_mode`, but at least there are not so many arguments for a single media item than there are for a group. Adding `_caption_` makes it clearer that these 3 arguments sort of belong together.

BTW I see that I put `group_parse_mode` (not `group_caption_parse_mode`) into the Note in docstring, noted.

Anyway, if you think that `group_parse_mode` won't be confusing for users, I'll be happy to make the argument name consistent with that of a single media item.

tests/test_inputmedia.py
```diff
@@ -443,6 +443,59 @@ async def test_send_media_group_photo(self, bot, chat_id, media_group):
             mes.caption_entities == [MessageEntity(MessageEntity.BOLD, 0, 5)] for mes in messages
         )
 
+    @flaky(3, 1)
```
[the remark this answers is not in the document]
lemontree210: got it, thanks. Removing.

telegram/_bot.py
```diff
@@ -2037,16 +2051,40 @@ async def send_media_group(
...
+                List of special entities that appear in the group caption.
 
         Returns:
             List[:class:`telegram.Message`]: An array of the sent Messages.
 
         Raises:
             :class:`telegram.error.TelegramError`
         """
+        if group_caption and any(item.caption for item in media):
+            raise ValueError("You can only supply either group caption or images with captions.")
```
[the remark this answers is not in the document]
lemontree210: thank you, fixing

tests/test_inputmedia.py
```diff
@@ -443,6 +443,59 @@ async def test_send_media_group_photo(self, bot, chat_id, media_group):
...
+            group_caption_parse_mode=group_caption_parse_mode,
+            group_caption_entities=group_caption_entities,
+        )
+        assert isinstance(messages, list)
+        assert len(messages) == 3
+        assert all(isinstance(mes, Message) for mes in messages)
+        assert all(mes.media_group_id == messages[0].media_group_id for mes in messages)
+
+        # Make sure first message got the caption, which will lead
+        # to Telegram displaying its caption as group caption
```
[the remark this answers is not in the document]
lemontree210: of course. fixing in commit af61aac0, hope this is the correct way

telegram/_bot.py
lemontree210: yes, fixing

telegram/_bot.py
lemontree210: yep 👍🏻

telegram/ext/_extbot.py
lemontree210: Yes, I saw that that was quite a lonely docstring. Removing. Just for the sake of understanding, though: why is it wrong?

telegram/_bot.py
lemontree210: fixing in commit e4b7971164c4c08aab6920401bae0bd74397dcf7 (I also add some text, not sure it's good). 

Don't know what's best for the docs - to put `Defaults to` or to skip :)

telegram/_bot.py
lemontree210: Fixed in e4b7971164c4c08aab6920401bae0bd74397dcf7, but not sure I *completely* understand the difference between places where it `DEFAULT_NONE` is there and where it isn't. Maybe the difference is that the new arguments don't reflect Telegram API.

telegram/_bot.py
```diff
@@ -2037,16 +2051,40 @@ async def send_media_group(
...
+        # Shallow copy of every media item should be enough.
+        copied_media = [copy.copy(item) for item in media]
+
+        if group_caption:
+            # Apply group caption to the first media item.
+            # This will lead to the group being shown with this caption.
+            first_item = copied_media[0]
+            first_item.caption = group_caption
+            first_item.parse_mode = group_caption_parse_mode
+            first_item.caption_entities = group_caption_entities
```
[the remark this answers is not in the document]
lemontree210: thank you, good point. Fixing in c38801fab32cc0fad40c07ef176fef00d98ca2f8

is this simple `if group_caption` actually a good enough check?

telegram/ext/_extbot.py
Bibo-Joshi: Because `Bot.send_media_group` can not be a shortcut for `Bot.send_media_group` ;)

telegram/_bot.py
Bibo-Joshi: Mh, my preferred rule of naming here would just be that the parameters that this PR introduces to `send_media_group` should not differ from the parameters of `InputMedia*`. that would give us `parse_mode`, `caption` and `caption_entities`, possibly with a `group_` prefix (although my I personally don't see a big advantage of having that prefix). IMO the correspondence to the established parameters of `InputMedia*` should be enough to avoid confusion …

telegram/ext/_extbot.py
lemontree210: I see, I think I understand the logic of the classes a bit better now.

telegram/_bot.py
lemontree210: OK, I see your logic. I also forget that the project has great docs 😆 

Renaming all arguments to match the arguments of functions that handle individual media (commit 61af4f491fe5390b3f9e30ac28553366531af759)

At least I think it was not such a bad idea to name arguments differently in the beginning because renaming in this direction is much easier than in the opposite one.

> We should make sure that defaults are properly applied. <...> we should in any case add tests cases that make sure of it. If you need help with that, please feel free to reach out. Maybe [this unit test](https://github.com/python-telegram-bot/python-telegram-bot/blob/eb818da93f3df2b439f5aef6065add232a5106e8/tests/test_inputmedia.py#L574-L584) can give you an idea how we currently tests the defaults setup in the unit tests.

I must admit I'm having trouble understanding this, but I guess from pedagogical point of view it would be best for me to keep trying for another day before crying for help 😆

telegram/_bot.py
```diff
@@ -2037,16 +2049,52 @@ async def send_media_group(
...
             List[:class:`telegram.Message`]: An array of the sent Messages.
 
         Raises:
             :class:`telegram.error.TelegramError`
         """
+        if caption and any(
+            [
+                any(item.caption for item in media),
+                any(item.caption_entities for item in media),
+                any(item.parse_mode is DEFAULT_NONE for item in media),
```
Bibo-Joshi, line 2076:
```suggestion
                any(item.parse_mode is not DEFAULT_NONE for item in media),
```

`DEFAULT_NONE` is the value that we expect, as it indicates that the `parse_mode` was not manually specified :)

telegram/_bot.py
```diff
@@ -2037,16 +2049,52 @@ async def send_media_group(
...
+            ]
+        ):
+            raise ValueError("You can only supply either group caption or media with captions.")
+
+        changed_media = None
```
Bibo-Joshi, line 2081: I guess one can skip this variable - just do `media = media.copy()` in the `if caption` case

tests/test_inputmedia.py
```diff
@@ -443,6 +443,64 @@ async def test_send_media_group_photo(self, bot, chat_id, media_group):
...
+        assert first_message.caption
+        assert first_message.caption_entities == [MessageEntity(MessageEntity.BOLD, 0, 5)]
+
+        # Check that other messages have no captions
+        assert all(mes.caption is None for mes in other_messages)
+        assert not any(mes.caption_entities for mes in other_messages)
+        # .parse_mode must be completely absent
+        assert not any(getattr(mes, "parse_mode", None) for mes in other_messages)
```
Bibo-Joshi, line 502: the `Message` class doesn't have a `parse_mode` attribute, so this assertion doesn't really help. You already check that the `caption_entities` are empty - that suffices

tests/test_inputmedia.py
```diff
@@ -443,6 +443,64 @@ async def test_send_media_group_photo(self, bot, chat_id, media_group):
...
+    async def test_send_media_group_with_group_caption(
+        self,
+        bot,
+        chat_id,
+        media_group,
+        caption,
+        parse_mode,
+        caption_entities,
+    ):
+        # clear media_group of all_captions
```
Bibo-Joshi, line 473: we should also check that the `media_group` is not altered. to do this please create a copy `media_group` (e.g. `copied_media_group = media_group.copy()`). Then after calling `send_media_group` add assertions for `copied_media_group == media_group` and `media_group[0] is copied_media_group[0]` (which double-ensures that the first item wasn't replaced)

telegram/_bot.py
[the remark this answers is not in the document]
Bibo-Joshi: I think so, yes. The only downside I see is that `(group_){parse_mode, caption_entities}` are ignored if `(group_)caption` is not passed, but that's the same for `InputMedia*` anyways … so fine with me.

Bibo-Joshi (changes requested):
Thank you for the updates! I left a few more comments below. Take your time with the defaults stuff (nice to see that you're eager to figure it out yourself!), but feel free to reach out for help if you want it :)

telegram/_bot.py
lemontree210: OK, I just thought that would be more readable. But yes, my way requires one more line of code and more complex "media" assignment later on.

Fixing in commit 7643dc6633fd82ab96ef9f1bd7490ddb84b7662b.

BTW, correct me if I'm wrong: as far as I understand, `media.copy()` and `media[:]` do the same thing. I'll put `media.copy()`, but just checking.

lemontree210: OK, I just thought that would be more readable. But yes, my way requires one more line of code and more complex "media" assignment later on.

Fixing in commit 7643dc6633fd82ab96ef9f1bd7490ddb84b7662b.

BTW, correct me if I'm wrong: as far as I understand, `media.copy()` and `media[:]` do the same thing. I'll put `media.copy()`, but just checking.

tests/test_inputmedia.py
lemontree210: got it

telegram/_bot.py
lemontree210: Well, you're absolutely right that I forgot `not` there, I was just lucky `ValueError` was thrown because of preceding statements.

But I can't put `is not DEFAULT_NONE` either because then the test will fail, and in the test the test media group has media items with `parse_mode` explicitly set to None.

So I guess the line should say, just like in preceding lines:
```py
any(item.parse_mode for item in media)
```
Because who knows, not only in test can we explicitly pass `None` to `parse_mode`, maybe some user will do that too for some reason.

tests/test_inputmedia.py
lemontree210: Absolutely

lemontree210:
> Take your time with the defaults stuff, but feel free to reach out for help if you want it 

I guess I understand better **how** these tests work; what I don't understand is **why** we need a test for something that defaults to `None` rather than `DEFAULT_NONE`. 

The only two existing unit tests check arguments that default to `DEFAULT_NONE`. So if we decide to leave `parse_mode: Optional[str] = None` (as opposed to `parse_mode: ODVInput[str] = DEFAULT_NONE`) then I guess I can't understand why an additional check is needed for `parse_mode` (and maybe you meant also for `caption` and `caption_entities`).

lemontree210:
> Take your time with the defaults stuff, but feel free to reach out for help if you want it 

I guess I understand better **how** these tests work; what I don't understand is **why** we need a test for something that defaults to `None` rather than `DEFAULT_NONE`. 

The only two existing unit tests check arguments that default to `DEFAULT_NONE`. So if we decide to leave `parse_mode: Optional[str] = None` (as opposed to `parse_mode: ODVInput[str] = DEFAULT_NONE`) then I guess I can't understand why an additional check is needed for `parse_mode` (and maybe you meant also for `caption` and `caption_entities`).

telegram/_bot.py
lemontree210: > Take your time with the defaults stuff, but feel free to reach out for help if you want it 

I guess I understand better **how** these tests work; what I don't understand is **why** we need a test for something that defaults to `None` rather than `DEFAULT_NONE`. 

The only two existing unit tests check arguments that default to `DEFAULT_NONE`. So if we decide to leave `parse_mode: Optional[str] = None` (as opposed to `parse_mode: ODVInput[str] = DEFAULT_NONE`) then I guess I can't understand why an additional check is needed for `parse_mode` (and maybe you meant also for `caption` and `caption_entities`).

telegram/_bot.py
Bibo-Joshi: Then the unit test needs to be adapted ;)

We want to check if the user explicitly set the `parse_mode` for the `InputMedia*` object. That's the case, if `item.parse_mode is not DEFAULT_NONE`. If it's `None`, then the user explicitly did `InputMediaPhoto(…, parse_mode=None)` and that's a case where we don't want to allow passing `send_media_group(…, caption='stuff')`.

tests/test_inputmedia.py
```diff
@@ -443,6 +443,72 @@ async def test_send_media_group_photo(self, bot, chat_id, media_group):
             mes.caption_entities == [MessageEntity(MessageEntity.BOLD, 0, 5)] for mes in messages
         )
 
+    async def test_send_media_group_throws_error_with_group_caption_and_individual_captions(
```
Bibo-Joshi, line 446: the entries in `media_group` all have both a caption and one of parse_mode/caption_entities. Can you also check that the error is raised if one passe media that *only* have a parse_mode/caption_entities?

lemontree210: Yes, I think that was what I was asking about earlier on, whether my test was enough.

telegram/_bot.py
lemontree210: I see. Let me go a bit back then and ask: how is `parse_mode` different from `caption` and `caption_entities`? Why, say, in the definition of `send_audio()` does `parse_mode` have `= DEFAULT_NONE`, while `caption` and `caption_entities` have `= None`? I think I'm especially curious as to how `parse_mode` is different from `caption_entities`

telegram/_bot.py
lemontree210: I think I see what you mean. 
Still no regrets you let me work on this issue? 🤣

Bibo-Joshi: > Still no regrets you let me work on this issue? 🤣

Not at all 🙆‍♂️

telegram/_bot.py
Bibo-Joshi: `parse_mode` differs from `caption` and `caption_entities` by the fact, that you can set a default value for it via `Defaults.parse_mode` (maybe also have a look at the wiki page on [adding defaults](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Adding-defaults-to-your-bot). Setting default values for `caption` or `caption_entities` doesn't make much sense. So only for `parse_mode` it's relevant to differentviate between 

* `send_media_group(…)` - where `Defaults.parse_mode` must be used
* and `send_media_group(…, parse_mode=None)` - where a `parse_mode` musn't be used

lemontree210: Aah, I understand the difference now, thanks! Added commit 4e8007dd766fcdffdf57c6572fcb7f50f9f56633. I hope it's OK to just set `parse_mode` to `DEFAULT_NONE` in the unit test.

telegram/_bot.py
lemontree210: Added commit b0d20c4dfda80095f644c1ba6542db7fc4674f07
I decided to put `DEFAULT_NONE` for `parse_mode` although I remember you said [here](https://github.com/python-telegram-bot/python-telegram-bot/pull/3295#discussion_r997365686) it was not necessary. But it seems the right way.

What possibly worries me:
- to make tests of shortcuts pass, I had to exclude `parse_mode` from [one check](https://github.com/python-telegram-bot/python-telegram-bot/commit/b0d20c4dfda80095f644c1ba6542db7fc4674f07#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128) in `conftest.py`. It kind of makes sense because `parse_mode` for the group is not a "regular argument" but I still worry that I did something wrong just to avoid an exception being thrown
- I tried to combine tests of defaults (right now I added two separate test functions `test_send_media_group_default_parse_mode` and `test_send_media_group_default_parse_mode_overridden`) in one test function along the lines of `test_send_media_group_default_protect_content`. If do one combined test function it makes the first call to `send_media_group()` just fine but then throws `ValueError` at the second call, as if one of the items had a caption (when in fact, and I have checked, the fixture group was not changed between calls). I suspect I'm missing something tricky here so I'll continue to look into that.

tests/conftest.py
```diff
@@ -712,7 +712,7 @@ async def make_assertion(
         data = request_data.parameters
 
         # Check regular arguments that need defaults
-        for arg in (dkw for dkw in kwargs_need_default if dkw != "timeout"):
+        for arg in (dkw for dkw in kwargs_need_default if dkw not in ("timeout", "parse_mode")):
```
Bibo-Joshi, line 715: excluding `parse_mode` is okay, but it should only happen if we're testing `send_media_group`. Please edit this line to `for arg in kwargs_need_default` (the `timeout` thingy actually is a remnant from v13) and instead add

```python
if method.__name__.endswith("_media_group"):
    # the parse_mode is applied to the first media item and we test this elsewhere
    kwargs_need_default.remove("parse_mode")
```

after the definition of `kwargs_need_default` in lines 696-700

tests/test_inputmedia.py
```diff
@@ -583,6 +677,38 @@ async def test_send_media_group_default_protect_content(
         )
         assert not all(msg.has_protected_content for msg in unprotected)
 
+    @flaky(3, 1)
+    @pytest.mark.parametrize("default_bot", [{"parse_mode": ParseMode.HTML}], indirect=True)
+    async def test_send_media_group_default_parse_mode(
```
Bibo-Joshi, line 682: please also test with explicitly passing `parse_mode=None` - in that case the resutling caption should read `"<b>photo</b> 1"` and there should be no entities

telegram/_bot.py
```diff
@@ -2037,13 +2046,51 @@ async def send_media_group(
...
+
+        if caption:
+            # Create copies of objects in media and apply group caption to the first item.
+            # This will lead to the group being shown with this caption.
+
+            # Only copying of first item should be enough, but the behaviour of Defaults is such
+            # that a bot with a default parse mode will automatically set `parse_mode` for
+            # each item to default parse mode of the bot.
+            # This will throw ValueError at subsequent calls with same media items.
+            media = [copy.copy(item) for item in media]
```
Bibo-Joshi, line 2086: Uff, you're right. I think I stumbled about the "defaults-edits-things-in-place" (let's call it DETIP) issue somewhere before as well … Adding a workaround in `send_media_group` doesn't properly fix it though, instead one should edit `(Ext)Bot._insert_defaults` …

TBH, integrating `Defaults` into this PR has become more complicated than I had anticipated. IMO you've done a remarkable job wrapping your head around it and have prepared a very nice PR! Fixing DETIP would be somewhat out of scope for this PR IMO and also I'd feel bad for putting that on your plate as well 😬 

My suggestion would be: We put this PR on hold for a bit. In the meantime, I'll prepare a separate PR for fixing DETIP. Once that's done, we merge the changes into your branch & finish up this PR (which should hopefully be not too much work). Depending on when that happens, we can also see if you're still interested/available or if I finish up instead. Would that be okay with you?

In any case I want to thank you for the patience and work that you've put into the PR! I hope that you had some fun getting to know some of the internals of the library :)

tests/conftest.py
lemontree210: Done in 7b683154be006b2f591960dbf82c71fad40bc219.

tests/test_inputmedia.py
lemontree210: Yep, done in b23acc53531dc4ec3f35bea791393d62bd01525d.

telegram/_bot.py
lemontree210: That is perfectly fine with me. Thanks for your support, and yes, it's been great fun, I've learned a lot already. Moving on to another issue I picked :)

Bibo-Joshi: I just merged #3311. Once you merge the `master` branch into this branch, it should suffice to copy only the item that changes. Let me know if you're intrested in doing that yourself or if you'd rather have me finish up :)

lemontree210: Thanks! I'm absolutely interested in trying to do this myself.

Bibo-Joshi (approved):
Awesome 🥳

Bibo-Joshi:
Thanks again for your work & patience on this feature :)

lemontree210:
Thank *you*! I hope we didn't forget to add `versionadded` to the docstring. I mean, I know we didn't add it but I don't know whether we should have 😆

Bibo-Joshi:
Ah, damn. I'll add that in the doc-fixes branch :)

PR #3296, opened 2022-10-16
Add 2 new constants related to local Bot API server

harshil21:
Adds 2 new constants to `FileSizeLimit` if you're running in local mode.

Didn't go with `math.inf` for `FILESIZE_DOWNLOAD_LOCAL` since that's a float which can't be converted to `int.` So went with `sys.maxsize`

### Checklist for PRs

### DIFF
--- a/telegram/constants.py
+++ b/telegram/constants.py
@@ -340,7 +340,8 @@ class FileSizeLimit(IntEnum):
        API server.
     """
     FILESIZE_DOWNLOAD_LOCAL = sys.maxsize
-    """obj:`int`: Bots can download files without a size limit when using a local bot API server"""
+    """:obj:`int`: Bots can download files without a size limit when using a local bot API server.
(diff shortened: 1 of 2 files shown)


telegram/constants.py
```diff
@@ -334,6 +335,13 @@ class FileSizeLimit(IntEnum):
     """:obj:`int`: Bots can download files of up to 20MB in size."""
     FILESIZE_UPLOAD = int(50e6)  # (50MB)
     """:obj:`int`: Bots can upload non-photo files of up to 50MB in size."""
+    FILESIZE_UPLOAD_LOCAL = int(2e9)  # (2000MB)
```
Bibo-Joshi, line 338: Maybe `_LOCAL_MODE`?

Bibo-Joshi (approved):
LGTM :)

PR #3308, opened 2022-10-25
Add mention_XXX to Chat class

### LINKED ISSUE #3216: [Suggestion] Add mention_XXX to Chat class
XXX means .html and .markdown(_v2). It is supposed to mention the Chat in the picked markup. Use title or first_name + space + Lastname as the text if the input parameter name is not set (otherwise use that), and the inline mention of the chat.id.

The logic will be close to the User one, have a look at it here: https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/_user.py#L222

Also check out our [contrib](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/.github/CONTRIBUTING.rst) document to get started with this project, and don't forget to unit test your changes!

miles170:
Add `mention_markdown` and `mention_markdown_v2` and `mention_html` to Chat class.

Closes #3216.

### Checklist for PRs

### DIFF
--- a/telegram/_chat.py
+++ b/telegram/_chat.py
@@ -19,6 +19,7 @@
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains an object that represents a Telegram Chat."""
 from datetime import datetime
+from html import escape
 from typing import TYPE_CHECKING, ClassVar, List, Optional, Tuple, Union
 
 from telegram import constants
@@ -30,6 +31,9 @@ from telegram._telegramobject import TelegramObject
 from telegram._utils import enum
 from telegram._utils.defaultvalue import DEFAULT_NONE
 from telegram._utils.types import DVInput, FileInput, JSONDict, ODVInput, ReplyMarkup
+from telegram.helpers import escape_markdown
+from telegram.helpers import mention_html as helpers_mention_html
+from telegram.helpers import mention_markdown as helpers_mention_markdown
 
 if TYPE_CHECKING:
     from telegram import (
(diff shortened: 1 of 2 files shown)


telegram/_chat.py
```diff
@@ -355,6 +359,92 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["Chat"]:
 
         return super()._de_json(data=data, bot=bot, api_kwargs=api_kwargs)
 
+    def mention_markdown(self, name: str = None) -> str:
```
Poolitzer, line 362: Just wondering if we still need to support V1 markdown in new shortcuts...

telegram/_chat.py
```diff
@@ -355,6 +359,92 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["Chat"]:
...
+        """
+        full_name = self.full_name
+        if self.type == self.PRIVATE:
+            if name:
+                return helpers_mention_markdown(self.id, name)
+            if full_name:
+                return helpers_mention_markdown(self.id, full_name)
+            raise TypeError("Can not create a link to a private chat without first name")
+        if self.username:
+            tg_link = f"https://t.me/{self.username}"
```
Poolitzer, line 386: This should be replaced with [self.link](https://docs.python-telegram-bot.org/en/v20.0a4/telegram.chat.html#telegram.Chat.link) (you can probably even skip this variable altogether)

telegram/_chat.py
```diff
@@ -355,6 +359,92 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["Chat"]:
...
+            :obj:`str`: The inline mention for the chat as markdown (version 1).
+
+        """
+        full_name = self.full_name
+        if self.type == self.PRIVATE:
+            if name:
+                return helpers_mention_markdown(self.id, name)
+            if full_name:
+                return helpers_mention_markdown(self.id, full_name)
+            raise TypeError("Can not create a link to a private chat without first name")
```
Poolitzer, line 384: Imo we end our errors with a .

Oh and I would call it `create a mention`, not link, since link is already the attribute.

telegram/_chat.py
```diff
@@ -355,6 +359,92 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["Chat"]:
...
+        if self.username:
+            tg_link = f"https://t.me/{self.username}"
+            if name:
+                return f"[{name}]({tg_link})"
+            if full_name:
+                return f"[{full_name}]({tg_link})"
+            raise TypeError("Can not create a link to a public chat without first name")
+        raise TypeError("Can not create a link to a private group chat")
```
Poolitzer, line 392: see above

telegram/_chat.py
```diff
@@ -355,6 +359,92 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["Chat"]:
...
+            instead.
+
+        Args:
+            name (:obj:`str`): The name used as a link for the chat. Defaults to :attr:`full_name`.
+
+        Returns:
+            :obj:`str`: The inline mention for the chat as markdown (version 1).
+
+        """
+        full_name = self.full_name
```
Poolitzer, line 378: do you actually need this shortcut? full_name is an attribute, so no performance gain, and self. doesn't make much of a difference in code length imo.

Poolitzer (changes requested):
Hey there, quick and well made PR, did not spot a logic fault! Just nitpicking at words, as I usually do.

harshil21 (commented):
Nice and swift work. I think we should also add a `Raises:` block to each of the methods describing when the method fails

telegram/_chat.py
```diff
@@ -355,6 +359,86 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["Chat"]:
...
+            if name:
+                return helpers_mention_markdown(self.id, name)
+            if self.full_name:
+                return helpers_mention_markdown(self.id, self.full_name)
+            raise TypeError("Can not create a mention to a private chat without first name")
+        if self.username:
+            if name:
+                return f"[{name}]({self.link})"
+            if self.full_name:
+                return f"[{self.full_name}]({self.link})"
```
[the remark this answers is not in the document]
Bibo-Joshi: Guess that's right, `full_name` is `None` for non-private chats. Instead of username (which also may be `None`), we should use `title`, I guess

telegram/_chat.py
harshil21: forgot to reply to this.. yes, I think we should. In many cases it's less annoying than MDv2

miles170: I think we meet the decision that drop markdown v1 supports.

telegram/_chat.py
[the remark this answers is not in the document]
miles170: > Guess that's right, `full_name` is `None` for non-private chats. Instead of username (which also may be `None`), we should use `title`, I guess

Isn't `self.username` always `not None` if the chat is public chat?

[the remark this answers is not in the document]
harshil21: title is probably a better "name" than the username

telegram/_chat.py
harshil21: I think there is a misunderstanding - I meant to say we should be supporting MarkdownV1 and thus mention_markdown shouldn't be removed imo

Bibo-Joshi: I'm with harshil here

harshil21 (approved):

Poolitzer (approved):
LGTM

Bibo-Joshi:
Thank you very much for your contribution @miles170 :)

PR #3315, opened 2022-10-28
update: added deep link maximum character constant

BAcode-X:
Based on [Telegram API doc](https://core.telegram.org/bots/features#deep-linking), the parameter can only be up to 64 characters long.

### Checklist
    - [ ] Added new constants at `telegram.constants`


Poolitzer:
Hi. You added the constant above the line where it says that "the constants above this line are tested", while not adding a test. I don't think this variable needs one (and it wouldn't be fun to write either), so just move it down.

Poolitzer (approved):
LGTM

Bibo-Joshi:
Thank you for the contribution :)

PR #3316, opened 2022-10-28
Unified args and attributes of dice and loginurl classes

HatimZ:
Refer to https://github.com/python-telegram-bot/python-telegram-bot/issues/3109, Dice and LoginUrl class arguments docstrings and attributes docstrings are unified.

### DIFF
--- a/telegram/_dice.py
+++ b/telegram/_dice.py
@@ -56,13 +56,15 @@ class Dice(TelegramObject):
         However, this behaviour is undocumented and might be changed by Telegram.
 
     Args:
-        value (:obj:`int`): Value of the dice. 1-6 for dice, darts and bowling balls, 1-5 for
-            basketball and football/soccer ball, 1-64 for slot machine.
+        value (:obj:`int`): Value of dice. telegram.constants.DiceValueRange.DICE for dice, darts and
+            bowling balls, telegram.constants.DiceValueRange.BASKETBALL for basketball and football/soccer ball,
+            telegram.constants.DiceValueRange.SLOT_MACHINE for slot machine.
         emoji (:obj:`str`): Emoji on which the dice throw animation is based.
 
     Attributes:
-        value (:obj:`int`): Value of the dice. 1-6 for dice, darts and bowling balls, 1-5 for
-            basketball and football/soccer ball, 1-64 for slot machine.
+        value (:obj:`int`): Value of dice. telegram.constants.DiceValueRange.DICE for dice, darts and
+            bowling balls, telegram.constants.DiceValueRange.BASKETBALL for basketball and football/soccer ball,
+            telegram.constants.DiceValueRange.SLOT_MACHINE for slot machine.
         emoji (:obj:`str`): Emoji on which the dice throw animation is based.
 
     """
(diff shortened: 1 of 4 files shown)


HatimZ:
@Bibo-Joshi

clot27:
LGTM, @Poolitzer is there something wrong with those tests?

harshil21 (changes requested):

telegram/_loginurl.py
```diff
@@ -59,10 +59,22 @@ class LoginUrl(TelegramObject):
...
+            the original URL without information about the user will be opened. The data added is
+            the same as described in
+            `Receiving authorization data
+            <https://core.telegram.org/widgets/login#receiving-authorization-data>`_.
         forward_text (:obj:`str`): Optional. New text of the button in forwarded messages.
         bot_username (:obj:`str`): Optional. Username of a bot, which will be used for user
-            authorization.
+            authorization. See
+            `Setting up a bot <https://core.telegram.org/widgets/login#setting-up-a-bot>`_
+            for more details. If not specified, the current
```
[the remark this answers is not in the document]
HatimZ: Sure. Thanks!

telegram/_dice.py
```diff
@@ -61,7 +61,8 @@ class Dice(TelegramObject):
         emoji (:obj:`str`): Emoji on which the dice throw animation is based.
 
     Attributes:
-        value (:obj:`int`): Value of the dice.
+        value (:obj:`int`): Value of the dice. 1-6 for dice, darts and bowling balls, 1-5 for
+            basketball and football/soccer ball, 1-64 for slot machine.
```
Poolitzer, line 65: We started moving all of those numeric values into the constant file to compare to easier in you code. I think we should do this here as well.

Poolitzer (changes requested):

telegram/_dice.py
clot27: I think no, as those values are range of numbers and can vary so we can't include them in constant file

harshil21: it's a good question. Should the constants for this be a `range()` or just `MIN_*_VALUE` and `MAX_*_VALUE`? I think it depends on the use case, @Poolitzer what use case for "comparsion" do you have in mind?

Poolitzer: Oh yes, use case is an interesting question. I can't really think of one.

harshil21: hmm, the only use case I think of is like: `if update.message.dice.value == constants.DiceLimits.DICE_MAX_VALUE: update.message.reply_text("congrats! You got the highest value!)`

of course the above can be done with subscription (`[-1]`) if it's a range, but having MAX, MIN is more better imo

Bibo-Joshi: We have a few constants where we just have the max as constant b/c the min value is of no particular interest and/or rather obvious. Would be fine with me here as well, I guess. Otherwise, I'd probably also go for MIN_/MAX_* - Keep It Simple, Stupid 🙆 

In any case, that's out of scope for this PR …

clot27:
Hi, the doc changes are fine, also this pr is against the doc-fixes branch so adding constants should be done in a different pr. So revert those changes..

PR #3319, opened 2022-10-30
Document dunder methods of TO

Bibo-Joshi:
As discussed in https://github.com/python-telegram-bot/python-telegram-bot/pull/3234#discussion_r1005856429

### DIFF
--- a/telegram/_telegramobject.py
+++ b/telegram/_telegramobject.py
@@ -35,20 +35,12 @@ Tele_co = TypeVar("Tele_co", bound="TelegramObject", covariant=True)
 class TelegramObject:
     """Base class for most Telegram objects.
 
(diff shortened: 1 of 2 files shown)


telegram/_telegramobject.py
```diff
@@ -359,6 +401,26 @@ def set_bot(self, bot: Optional["Bot"]) -> None:
         self._bot = bot
 
     def __eq__(self, other: object) -> bool:
+        """Compares this object with :paramref:`other` in terms of equality.
+        If this object and :paramref:`object` ar are `not` objects of the same class,
```
harshil21, line 405:
```suggestion
        If this object and :paramref:`object` are `not` objects of the same class,
```

harshil21 (commented):
amazing, I love this change

PR #3322, opened 2022-10-31
Create setup check

Poolitzer:
This PR creates a small check for common errors when trying to install PTB. Ideally, the questions regarding why PTB does not work when installing should subside on our support groups or at lest should be easier to debug.

I have not created tests because I wasn't sure how and what the best way would be. I imagine somehow running` python setup.py install` on this directory if one of the fail conditions is created before?


setup.py
```diff
@@ -116,13 +117,59 @@ def get_setup_kwargs(raw=False):
     return kwargs
 
 
+def avoid_common_setup_errors(raw=False):
```
Bibo-Joshi, line 120: I'd suggest to give this a more specific name, e.g. `check_namespace_clashes`. This makes it more obvious what this method does and if we are to add more checks in a similar direction in the future, we'll probably split that into mutliple functions anyways :)

setup.py
```diff
@@ -116,13 +117,59 @@ def get_setup_kwargs(raw=False):
...
+    # This runs before the setup, trying to catch common setup errors
+    try:
+        get_distribution("telegram")
+        print(
+            """
+            =============================DEBUG ASSISTANCE==========================
+            You can not use python-telegram-bot and telegram. Please run 
+            pip uninstall telegram and try again.
```
Bibo-Joshi, line 128: The first sentence is a bit opaque for people that are not pip/python seniors IMO. How about

```suggestion
            You are trying to install the `python-telegram-bot` library, but the `telegram` library is already installed. Both libraries provide a Python package called `telegram`, which will lead to a namespace clash.
            If you want to use the `python-telegram-bot` library, please first uninstall the `telegram` library before installing the `python-telegram-bot` library again.
```

This also makes no assumptions on how the user (un)installs packages in their environment … Similar below.

setup.py
```diff
@@ -116,13 +117,59 @@ def get_setup_kwargs(raw=False):
...
+            =============================DEBUG ASSISTANCE==========================
```
Bibo-Joshi, line 129: Can probably be a constant that you reuse 6 times?

setup.py
```diff
@@ -116,13 +117,59 @@ def get_setup_kwargs(raw=False):
...
+            """
+        )
+        raise Exception("BuildError")
```
Bibo-Joshi, line 132: Can we provide a more descriptive error message? 🤔

Bibo-Joshi (commented):
Nice change set! Regarding tests, we already have `test_meta`, which essentially runs `python setup(-raw).py bdist_dumb` in the console - and that's actually the reason why the tests currently fail :D extending `test_meta` to

* uninstall PTB(-RAW) after successfull install
* checking that the error messages are prompted when installing with name clashes

should be possible

setup.py
```diff
@@ -116,13 +117,59 @@ def get_setup_kwargs(raw=False):
     return kwargs
 
 
+def avoid_common_setup_errors(raw=False):
+    # This runs before the setup, trying to catch common setup errors
+    try:
+        get_distribution("telegram")
```
harshil21, line 123: why can't I find any docs on what this does though

Bibo-Joshi: https://setuptools.pypa.io/en/latest/pkg_resources.html#getting-or-creating-distributions

Basically checks for anything that was installed via `pip install telegram`

.github/workflows/test.yml
```diff
@@ -34,11 +34,12 @@ jobs:
...
 
       - name: Test with pytest
-        # We run 4 different suites here
+        # We run 5 different suites here
         # 1. Test just utils.datetime.py without pytz being installed
         # 2. Test just test_no_passport.py without passport dependencies being installed
         # 3. Test just test_rate_limiter.py without passport dependencies being installed
-        # 4. Test everything else
+        # 4. Test just test_meta.py, since otherwise race conditions happened
+        # 5. Test everything else
```
Bibo-Joshi, line 42: I would really prefer to get to the root of this instead of leaving a mysterious comment that will gives us headaches at some point down the road :/

tests/test_meta.py
```diff
@@ -34,10 +35,77 @@ def change_test_dir(request, monkeypatch):
 
 
 @skip_disabled
-def test_build():
-    assert os.system("python setup.py bdist_dumb") == 0  # pragma: no cover
+@pytest.mark.skipif(platform.system() != "Windows", reason="Currently only setup for windows")
```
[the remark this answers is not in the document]
Poolitzer: as the reason states, currently only. I will include linux once I am home and can locally check the behaviour, and then I will figure out mac os.

tests/test_meta.py
```diff
@@ -34,10 +35,77 @@ def change_test_dir(request, monkeypatch):
...
+    def test_build(self):
+        if platform.system() == "Windows":
+            assert (
+                os.system(".\\venv_setup_test\\Scripts\\python.exe setup.py bdist_dumb") == 0
+            )  # pragma: no cover
+        # what follows is a neat hack, convincing the local python venv that ptb isn't installed.
+        # There are more files (see teardown_class), but this is the one python apparently
+        # uses to discover local packages
+        if platform.system() == "Windows":
+            os.system("rmdir /s /q python_telegram_bot.egg-info")
```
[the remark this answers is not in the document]
Poolitzer: `pip uninstall` requires using `pip install .`, which is hardcoded to use _setup.py_ (I once saw this in the pip source code from a stackoverflow answer, I just can't find it right now). 

I do not get the autouse fixture comment, do you mean instead of having this setup in a class with construct and teardown function? I think this setup works nicely and self explains the code flow well.

.github/workflows/test.yml
Poolitzer: I did not understand the root problem one bit, as I said, doing the check if telegram package is installed in the test two function calls before the setup.py install does did not yield it being there, and then setup.py thinks it is. I did not check however if using it in the class created venv folders works or not. I still would not know where the actual problem lies then however.

tests/test_meta.py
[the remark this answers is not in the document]
Bibo-Joshi: Ah okay. I had falsely understood your happy message in the dev chat as "I'm all done". my mistake .

tests/test_meta.py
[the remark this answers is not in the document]
Bibo-Joshi: I see now about having to remove this.

my point about using a fixture is that it can automatically clean up the articfacts after every test withouth any need to repeat the code for that. That would be in addition to a class construct/teardown

Bibo-Joshi:
I did some fiddling around as well and want to share some findings that I had. Probably Pool is already aware of at least some of them …

* running `setup.py` in a different cwd doesn't work because the way we use `requirements.txt` inside `setup.py` is not cwd-agnostic.
* removing a directory only if it exists/not failing if it doesn't:
  * Windows: `if exist dirname rmdir dirname /s /q`. Runnig this with `subprocess.check_call` instead of `os.system` always gives me a `FileNotFoundError` on for some reason that I haven't figured out yet :/
  * ubuntu: `rm -rf dirname`
* One can tell setuptools to write the artifacts to different output directories: https://stackoverflow.com/a/69584004/10606962. In combination with pytests `tmp_path` fixture (used e.g. in `TestPicklePersistence`), this might be helpful to avoid having to clean up the artifacts
* Using the flag `-n` (`--dry-run`) one can tell setuptools to just do nothing at all. This is well might be useful for the tests

setup.py
```diff
@@ -117,13 +118,66 @@ def get_setup_kwargs(raw=False):
...
+                f"""You can not use python-telegram-bot and python-telegram-bot-raw.
+            Please uninstall python-telegram-bot{"" if raw else "-raw"} and try again."""
+            )
+        )
+        raise Exception("BuildError")
+    except DistributionNotFound:
+        pass
+    try:
+        # if the telegram import works here, something locally overrides telegram
+        import telegram
```
harshil21, line 159: just realized, do we need this check? A user could `pip install` from any directory.

Poolitzer:
We realized since we provide wheels, pip uses these when installing PTB, so setup.py is not executed. Thus this PR doesn't make sense since only the small minority who run setup.py manually will benefit from it, and those people should already know what they are doing.

Issue #3107, opened 2022-06-20
Constant improvement


Poolitzer:
We currently have quite some limits still hard coded into the docstrings, for example [vcard,](https://docs.python-telegram-bot.org/en/v20.0a1/telegram.bot.html#telegram.Bot.send_contact.params.vcard) [telegram.Bot.send_poll#explanation](https://docs.python-telegram-bot.org/en/v20.0a1/telegram.bot.html#telegram.Bot.send_poll.params.explanation) or [#open_period](https://docs.python-telegram-bot.org/en/v20.0a1/telegram.bot.html#telegram.Bot.send_poll.params.open_period). We also could think about including [switch_pm_parameter](https://docs.python-telegram-bot.org/en/v20.0a1/telegram.bot.html#telegram.Bot.answer_inline_query.params.switch_pm_parameter) somehow. I would like to collect the thoughts of the maintainers here and maybe motivate someone from the community to contribute this since it is quite an "easy" addition to the project.

harshlancer:
Hey, I'm new to open source and want to learn how to contribute, and I can't quite get what needs to be done here. Can you guide me on which doc is required in order to be updated or changed?

Poolitzer:
Hey, nice to hear.

So I already linked to a few examples in my original message. Have a look at the source doc string of [send_poll](https://github.com/python-telegram-bot/python-telegram-bot/blob/v20.0a1/telegram/_bot.py#L6727-L6872). As you can see in `question`, it links the 1-_number_ to _:tg-const:\`telegram.Poll.MAX_QUESTION_LENGTH\`_  instead of writing out the number in the doc string itself (this happens five arguments later in `explanation`, where it says 0-200 characters without the link).

If you have a look at that linking, you can see that it references MAX_QUESTION_LENGTH in the Poll file. In the [source of Poll](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/_poll.py#L294-L306), you can see that you have been bamboozled and the actual source for these variables lies in the constants.py file, MAX_QUESTION_LENGTH is just a link to `constants.PollLimit.QUESTION_LENGTH`.

In the [source of constants](https://github.com/python-telegram-bot/python-telegram-bot/blob/v20.0a1/telegram/constants.py#L741-L757), you can finally see how it is implemented: An IntEnum Class which has the values mapped to their respective key.

----

[...]

PR #3336, opened 2022-11-05
refactor: add enum constants

lemontree210:
addresses #3107

### DIFF
--- a/telegram/constants.py
+++ b/telegram/constants.py
@@ -507,6 +507,15 @@ class LocationLimit(IntEnum):
     in degrees.
     """
 
+    MIN_LIVE_PERIOD = 60
+    """:obj:`int`: Minimum value allowed for period in seconds for which the location
+    will be updated.
+    """
+    MAX_LIVE_PERIOD = 86400
+    """:obj:`int`: Maximum value allowed for period in seconds for which the location
+    will be updated.
+    """
+
     MIN_PROXIMITY_ALERT_RADIUS = 1
     """:obj:`int`: Minimum distance for proximity alerts about approaching another chat member, in
     meters.
--- a/telegram/constants.py
+++ b/telegram/constants.py
@@ -487,8 +487,8 @@ class InlineQueryResultType(StringEnum):
 
 class LocationLimit(IntEnum):
     """This enum contains limitations for :class:`telegram.Location`/
-    :meth:`telegram.Bot.send_location`. The enum members of this enumeration are instances
-    of :class:`int` and can be treated as such.
+    :meth:`telegram.Bot.edit_message_live_location`/:meth:`telegram.Bot.send_location`.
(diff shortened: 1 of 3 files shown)


telegram/constants.py
```diff
@@ -753,6 +763,25 @@ class MessageType(StringEnum):
...
+    """This enum contains limitations for :paramref:`telegram.Bot.get_updates.limit`.
+    The enum members of this enumeration are instances of :class:`int` and can be treated as such.
+
+    .. versionadded:: 20.0
+    """
+
+    __slots__ = ()
+
+    MIN_LIMIT = 1
+    """:obj:`int`: Minimum value allowed for the ``limit`` parameter of
```
harshil21, line 776:
```suggestion
    """:obj:`int`: Minimum value allowed for the :paramref:`~telegram.Bot.get_updates.limit` parameter of
```
this will actually link to the parameter. Make sure to apply to the rest of the changes as well

lemontree210:
Question regarding limits for `Location` and `send_location`: I know that in Telegram API itself most limits are only listed in [`sendLocation`](https://core.telegram.org/bots/api#sendlocation) and not in [`Location`](https://core.telegram.org/bots/api#location), e.g. `live_period` and `heading`. 

Should I list all the limits in `Location` as well or do I stick to what's listed or not listed in Telegram API?

telegram/constants.py
lemontree210: Yes, I saw examples of this in `constants.py` but it was a bit inconsistent: some of docstrings link to params, some don't.

harshil21: that's because `:paramref:` is a new thing we started in v20, we should use it wherever possible.

harshil21:
> Should I list all the limits in `Location` as well or do I stick to what's listed or not listed in Telegram API?

I think we should stick to TG API docs.

telegram/constants.py
lemontree210: Got it. Maybe it could be a separate task to add `:paramref:`s to existing docstrings. I'll make sure I add them in the classes I add/change.

telegram/constants.py
```diff
@@ -753,6 +806,25 @@ class MessageType(StringEnum):
     """:obj:`str`: Messages with :attr:`telegram.Message.video_chat_participants_invited`."""
 
 
+class NumberOfUpdatesLimit(IntEnum):
```
Bibo-Joshi, line 809: spontanious naming idea: `PollingLimits`? that way this could include any limits that are related to (long) polling, specifically `get_updates`

lemontree210: Done in fcf6e57164278bf2c2381d42334a104485e81b08 in #3351

lemontree210:
closing this PR, all work moved to #3351 (with branches within same repo as opposed to fork)

PR #3343, opened 2022-11-06
Feat: Docstring changes + moving limit to constants

Poolitzer:
This adds the two docstring changes, I took the opportunity to move the changed limit to constants. At least description can be added to the limit (e.g. in #3336), maybe even more, so I kept the Class name broad.

### DIFF
--- a/telegram/constants.py
+++ b/telegram/constants.py
@@ -39,6 +39,7 @@ __all__ = [
     "ChatAction",
     "ChatID",
     "ChatInviteLinkLimit",
+    "ChatLimit",
     "ChatMemberStatus",
     "ChatType",
     "CustomEmojiStickerLimit",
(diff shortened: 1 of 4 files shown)


telegram/constants.py
```diff
@@ -901,3 +901,18 @@ class WebhookLimit(IntEnum):
     """:obj:`int`: Minimum length of the secret token."""
     MAX_SECRET_TOKEN_LENGTH = 256
     """:obj:`int`: Maximum length of the secret token."""
+
+
+class ChatLimits(IntEnum):
```
Bibo-Joshi, line 906: IIRC we named verything else in singular, i.e. `class <…>Limit(enum)`

Poolitzer: You appear to be right

Bibo-Joshi (approved):
LGTM, can be merged IMO

Issue #3340, opened 2022-11-05
[Feature] Bot API 6.3 update


Poolitzer:
This issue exists to keep track of the implementation of the 6.3 Update.

- [x] Topics (in groups):
  - [x] can_manage_topics in classes ChatAdministratorRights, ChatMemberAdministrator, ChatPermissions, ChatMemberRestricted 
  - [x] also in Bot function promoteChatMember
- [x] Message class
  - [x] new empty classes ForumTopicReopened, ForumTopicClosed 
  - [x] new class ForumTopicCreated
  - [x] add to all shortcuts thread_message_id, and add new shortcuts if needed (doubtful, but I did not check)
- [x] changed docstring: setChatTitle title attribute and deleteMessage
- [x] new param message_thread_id to too many methods to list
- [x] Chat
  - [x] needs shortcuts to all new methods except getForumTopicIconStickers
- [x] implement new methods
  - [x] new class ForumTopic

lemontree210:
I think I should be able to handle adding param `message_thread_id` to those numerous methods. Any pitfalls a newcomer wouldn't spot there?

Poolitzer:
@lemontree210 Doubt it. Think about adding it before the * in the param list, the docstring, the tests (if no fitting one exists create them, there should always be one mocking telegrams response and check all params), and you are good.

clot27:
Should I do the Topics (in groups) part?

Poolitzer:
@clot27 If you want to, sure :)

Poolitzer:
I will do the docstring changes now and then a couple new methods

PR #3345, opened 2022-11-06
add `message_thread_id` param to methods, `message_thread_id` and `is_topic_message` to `Message`

lemontree210:
addresses #3340 

- [x] Add `message_thread_id` to methods in `Bot` and `ExtBot`
    - [x] add parameter
    - [x] add docstrings with `.. versionadded:: version` 
    - [x] add argument to a call to `send_message` or item to `data`
- [x] Add method to shortcuts (`Chat`, `User`...)
- [ ] Create new or adapted existing unit tests - **will only be possible when class for topics is created**


telegram/_bot.py
```diff
@@ -965,6 +982,8 @@ async def send_photo(
                 receive a notification with no sound.
             protect_content (:obj:`bool`, optional): Protects the contents of the sent message from
                 forwarding and saving.
+            message_thread_id (:obj:`int`, optional): Unique identifier for the target message
+                thread (topic) of the forum; for forum supergroups only.
 
                 .. versionadded:: 13.10
```
Poolitzer, line 988: you sneaked the message_thread_id into the old versionadded

Poolitzer (changes requested):
Looks good!

telegram/_bot.py
lemontree210: D'oh. Thanks for spotting!

lemontree210:
Hey guys, do you maybe think that we should merge this into `api_6.3_update` and then I'd branch off again when it becomes possible to add tests with real methods and attributes?

telegram/_message.py
```diff
@@ -775,6 +795,7 @@ async def reply_text(
         allow_sending_without_reply: ODVInput[bool] = DEFAULT_NONE,
         entities: Union[List["MessageEntity"], Tuple["MessageEntity", ...]] = None,
         protect_content: ODVInput[bool] = DEFAULT_NONE,
+        message_thread_id: int = None,
```
Bibo-Joshi, line 798: If the bot gets a message that `message` that has `message.message_thread_id`, what happens if we do 

```
send_message(
    chat_id=message.chat_id,
    text="hi",
    reply_to_message_id=message.message_id,
)
```

Does that make a reply in the correct topic? Or do we have to do 

```
send_message(
    chat_id=message.chat_id,
    text="hi",
    reply_to_message_id=message.message_id,
    message_thread_id=message.message_thread_id
)
```

? Can you try that out? How does the behavior differ between passing both `message_thread_id` and `reply_to_message_id` and only passing `message_thread_id`?

I'm trying to figure out, if passing `message_thread_id` should be automated through the `quote` argument 🤔  This doesn't have to be done in this PR though, that can be done on the api 6.3 branch also later on …

Bibo-Joshi (approved):
One inquery, but the changes LGTM :) can be merged IMO

telegram/_message.py
Poolitzer: > Does that make a reply in the correct topic?

yes

> if passing message_thread_id should be automated through the quote argument
not needed

PR #3351, opened 2022-11-08
refactor: add enum constants

lemontree210:
Addresses #3107. This PR replaces #3336 that was a PR from my own fork. Instead, now it's a PR within the repository.

- [x] Among others, be sure to add limits for:
  - [x] set_chat_admin_custom_title
  - [x] set_chat_title
  - [x] set_chat_description
  - [x] create_new_sticker_set name, title
- [x] change docstrings in existing classes in `constants.py`: 
  - [x] add `:paramref:`s
- [x] add `.. versionadded:: 20.0` to every added constant in "target" classes (e.g. like it's done for class constants in `InlineQuery`)


lemontree210:
This is not part of the code I'm changing, but I was just creating a class with limits for `ReplyKeyboardMarkup` and noticed this. I don't think that `InlineKeyboardMarkupLimit` [here](https://github.com/python-telegram-bot/python-telegram-bot/pull/3351/files#diff-9dd9ba2fa5677486ab0a281e94af2aa08bc99c769fec8231e9787ab1b34e149cR377) has actually something to do with limits for any params in `Bot.send_message`, contrary to what's stated in the docstring.

lemontree210:
> I don't think that `InlineKeyboardMarkupLimit` [here](https://github.com/python-telegram-bot/python-telegram-bot/pull/3351/files#diff-9dd9ba2fa5677486ab0a281e94af2aa08bc99c769fec8231e9787ab1b34e149cR377) has actually something to do with limits for any params in `Bot.send_message`.

BTW, class `InlineKeyboardMarkupLimit` in `constants.py` doesn't seem to be used anywhere, not even in `InlineKeyboardMarkup` docstring.

lemontree210:
I also have a (maybe) aesthetic question about naming of constants. Right now I name them e.g. `MIN_LIMIT`, `MAX_LIMIT`, `MIN_RATE` and `MAX_RATE`, which is sort of OK, but when Sphinx generates the docs it sorts the constants alphabetically. So in generated docs they will be ordered thus:

- `MAX_LIMIT`
- `MAX_RATE`
- `MIN_LIMIT`
- `MIN_RATE`

Which is suboptimal for 2 reasons: min comes after max (counterintuitive) and constants are not grouped. 

So although I didn't get this comment during a review, shouldn't constants be rather called `LIMIT_MIN` and `LIMIT_MAX`? That wouldn't solve the problem of max being listed before min, but at least the grouping will be better. If I put `LIMIT_FROM` and `LIMIT_TO`, then both problems will be solved.

Any thoughts?

lemontree210 (commented):
I remember that adding class constants to `Bot` is not optimal, but shouldn't I fully implement the approach [laid out](https://github.com/python-telegram-bot/python-telegram-bot/issues/3107#issuecomment-1166251898) in description for #3107 in `Dice`? Which would mean introducing class constants for emoji and limits in `Dice` class directly and linking them to `constants.py`.

telegram/_dice.py
```diff
@@ -93,3 +105,17 @@ def __init__(self, value: int, emoji: str, *, api_kwargs: JSONDict = None):
     """
     ALL_EMOJI: ClassVar[List[str]] = list(constants.DiceEmoji)
     """List[:obj:`str`]: A list of all available dice emoji."""
+
+    MIN_VALUE: ClassVar[int] = constants.DiceLimit.MIN_VALUE
+    """:const:`telegram.constants.DiceLimit.MIN_VALUE`"""
+
+    MAX_VALUE_DICE_DARTS_BOWLING: ClassVar[int] = constants.DiceLimit.MAX_VALUE_DICE_DARTS_BOWLING
```
Bibo-Joshi, line 112: I'd vote to split that up into `max_value_{dice, darts, bowling}`, even though the value is repeated. If a new emoji is to be added in the future with the same max value, we'd have to rename the constant otherwise …
same for baseketball_football and also in the `constants` file

telegram/constants.py
```diff
@@ -486,24 +525,103 @@ class InlineQueryResultType(StringEnum):
...
     """
 
     __slots__ = ()
 
     HORIZONTAL_ACCURACY = 1500
-    """:obj:`int`: Maximum radius of uncertainty for the location, measured in meters."""
+    """:obj:`int`: Maximum value allowed for:
+
+    * :paramref:`~telegram.Location.horizontal_accuracy` parameter of :class:`telegram.Location`
+    * :paramref:`~telegram.InlineQueryResultLocation.horizontal_accuracy` parameter of
```
Bibo-Joshi, line 540: do these bulleted lists also render correctly? somethimes restructered text is a bit picky when it comes to this - at least in my experience …

docs/source/telegram.constants.rst
```diff
@@ -4,3 +4,4 @@ telegram.constants Module
 .. automodule:: telegram.constants
     :members:
     :show-inheritance:
+    :member-order: bysource
```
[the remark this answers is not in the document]
lemontree210: I would suggest either renaming the constants then - or keeping `bysource` in `telegram.constants.rst` but ditch it in classes because there the amount of constants is small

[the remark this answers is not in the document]
lemontree210: OK, I'll ditch `bysource` in classes. @Bibo-Joshi what do you think about keeping it in `constants.py`? At least when I'm adding classes there, I try to maintain the alphabetical order anyway

docs/source/telegram.dice.rst
```diff
@@ -3,4 +3,5 @@ telegram.Dice
 
 .. autoclass:: telegram.Dice
     :members:
-    :show-inheritance:
\ No newline at end of file
+    :show-inheritance:
+    :member-order: bysource
```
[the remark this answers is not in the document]
lemontree210: Done

telegram/_dice.py
```diff
@@ -34,30 +34,42 @@ class Dice(TelegramObject):
     considered equal, if their :attr:`value` and :attr:`emoji` are equal.
 
     Note:
-        If :attr:`emoji` is "🎯", a value of 6 currently represents a bullseye, while a value of 1
-        indicates that the dartboard was missed. However, this behaviour is undocumented and might
-        be changed by Telegram.
+        If :attr:`emoji` is :tg-const:`telegram.Dice.DARTS`, a value of 6 currently
```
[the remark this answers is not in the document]
lemontree210: Yes, I build locally after every change

telegram/constants.py
lemontree210: They do render (at least locally), yes

telegram/_bot.py
```diff
@@ -2158,13 +2158,18 @@ async def send_location(
...
-                :tg-const:`telegram.constants.LocationLimit.HEADING` if specified.
+                moving, in degrees. Must be between
+                :tg-const:`telegram.constants.LocationLimit.MIN_HEADING` and
+                :tg-const:`telegram.constants.LocationLimit.MAX_HEADING` if specified.
             proximity_alert_radius (:obj:`int`, optional): For live locations, a maximum distance
                 for proximity alerts about approaching another chat member, in meters. Must be
-                between 1 and :tg-const:`telegram.constants.LocationLimit.HEADING` if specified.
+                between :tg-const:`telegram.constants.LocationLimit.MIN_PROXIMITY_ALERT_RADIUS`
+                and :tg-const:`telegram.constants.LocationLimit.MAX_PROXIMITY_ALERT_RADIUS`
+                if specified.
```
lemontree210, line 2172: Here I would leave links to `constants.py` as they are now. Theoretically I could link to `Location` for `heading` and `horizontal_accuracy` and to `constants.py` for `live_period` and `proximity_alert_radius`, but I think it would create a mess.

telegram/_bot.py
```diff
@@ -7375,11 +7389,17 @@ async def send_dice(
...
                 :tg-const:`telegram.constants.DiceEmoji.DARTS` and
-                :tg-const:`telegram.constants.DiceEmoji.BOWLING`, values 1-5 for
+                :tg-const:`telegram.constants.DiceEmoji.BOWLING`, values
+                :tg-const:`telegram.constants.DiceLimit.MIN_VALUE`-
+                :tg-const:`telegram.constants.DiceLimit.MAX_VALUE_BASKETBALL_FOOTBALL` for
                 :tg-const:`telegram.constants.DiceEmoji.BASKETBALL` and
-                :tg-const:`telegram.constants.DiceEmoji.FOOTBALL`, and values 1-64
+                :tg-const:`telegram.constants.DiceEmoji.FOOTBALL`, and values
+                :tg-const:`telegram.constants.DiceLimit.MIN_VALUE`-
+                :tg-const:`telegram.constants.DiceLimit.MAX_VALUE_SLOT_MACHINE`
```
lemontree210, line 7402: This will have to be linked to `Dice` class rather than `telegram.constants`

telegram/constants.py
lemontree210: Screenshot from 2022-11-09 20-15-01

lemontree210:
Do you think I should also replace "emoji literals" in [`filters.py`](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/filters.py#L961) with links to constants?

docs/source/telegram.constants.rst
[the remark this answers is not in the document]
Bibo-Joshi: I don't see the benefit of having the `bysource` thingy for the order of the classes. Having the classes sorted alphabetically in the code certainly is good, but if we don't have to remember that, I'd prefer that …

telegram/_dice.py
Bibo-Joshi: > However, this will lead to docstring of Dice

I'd be fine with just linking one constant, e.g. `1-Dice.MAX_VALUE for dice, darts and bowling`. The link will take the people to the place where the other constants are defined, so they'll see what's going on.

> because of line breaks in docstrings I now get an extra whitespace rendered between min and max value

Maybe a backslash `\` at the end of the line? the whitespace wouldn't be a deal breaker for me though …

docs/source/telegram.constants.rst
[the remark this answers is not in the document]
lemontree210: OK. Removing `bysource` from `constants.py`.

telegram/_dice.py
lemontree210: > Maybe a backslash \ at the end of the line? the whitespace wouldn't be a deal breaker for me though …

It's strange but it doesn't work - maybe because of references. I know it does work
in URLs, for example. But yes, never mind.

lemontree210: Done

telegram/_dice.py
```diff
@@ -34,30 +34,43 @@ class Dice(TelegramObject):
...
 
-        If :attr:`emoji` is "🎰", each value corresponds to a unique combination of symbols, which
+        If :attr:`emoji` is :tg-const:`telegram.Dice.SLOT_MACHINE`, each value
+        corresponds to a unique combination of symbols, which
         can be found at our `wiki <https://github.com/python-telegram-bot/python-telegram-bot/wiki\
         /Code-snippets#map-a-slot-machine-dice-value-to-the-corresponding-symbols>`_.
         However, this behaviour is undocumented and might be changed by Telegram.
 
+    ..
+        In args, some links for limits of `value` intentionally point to constants for only
```
harshil21, line 60:
```suggestion
        In args, some links for limits of :paramref:`value` intentionally point to constants for only
```

telegram/constants.py
```diff
@@ -486,24 +536,103 @@ class InlineQueryResultType(StringEnum):
...
     .. versionadded:: 20.0
     """
 
     __slots__ = ()
 
     HORIZONTAL_ACCURACY = 1500
-    """:obj:`int`: Maximum radius of uncertainty for the location, measured in meters."""
+    """:obj:`int`: Maximum value allowed for:
+
+    * :paramref:`~telegram.Location.horizontal_accuracy` parameter of :class:`telegram.Location`
```
harshil21, line 550: this wasn't needed, but it's a welcome change!

telegram/constants.py
```diff
@@ -888,16 +1057,54 @@ class InvoiceLimit(IntEnum):
...
+    MAX_CONNECTIONS_LIMIT = 100
+    """:obj:`int`: Maximum value allowed for the
+    :paramref:`~telegram.Bot.set_webhook.max_connections` parameter of
+    :meth:`telegram.Bot.set_webhook`.
+    """
     MIN_SECRET_TOKEN_LENGTH = 1
-    """:obj:`int`: Minimum length of the secret token."""
+    """:obj:`int`: Minimum length of the secret token
+    (:paramref:`~telegram.Bot.set_webhook.secret_token` parameter of
+    :meth:`telegram.Bot.set_webhook`).
```
harshil21, line 1104:
```suggestion
    """:obj:`int`: Minimum length of the secret token for the
    :paramref:`~telegram.Bot.set_webhook.secret_token` parameter of
    :meth:`telegram.Bot.set_webhook`.
```
consistency reasons

telegram/constants.py
```diff
@@ -888,16 +1057,54 @@ class InvoiceLimit(IntEnum):
...
+    """
     MAX_SECRET_TOKEN_LENGTH = 256
-    """:obj:`int`: Maximum length of the secret token."""
+    """:obj:`int`: Maximum length of the secret token
+    (:paramref:`~telegram.Bot.set_webhook.secret_token` parameter of
+    :meth:`telegram.Bot.set_webhook`).
```
harshil21, line 1109:
```suggestion
    """:obj:`int`: Maximum length of the secret token for the
    :paramref:`~telegram.Bot.set_webhook.secret_token` parameter of
    :meth:`telegram.Bot.set_webhook`.
```

harshil21 (commented):
well done PR!

lemontree210:
don't forget: 

set_chat_admin_custom_title add constant limit
set_chat_title add constant limit
set_chat_description
create_new_sticker_set name, title

docs/source/telegram.dice.rst
```diff
@@ -3,4 +3,4 @@ telegram.Dice
 
 .. autoclass:: telegram.Dice
     :members:
-    :show-inheritance:
\ No newline at end of file
+    :show-inheritance:
```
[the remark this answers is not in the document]
lemontree210: yes. I added sorting by source, then removed it but left a blank line at EOF. I checked out `master`, the same file there has no blank line at EOF. But since most (all?) other files do, then we might as well let this one have one too :)

telegram/_dice.py
```diff
@@ -34,30 +34,43 @@ class Dice(TelegramObject):
...
+        However, this behaviour is undocumented and might be changed by Telegram.
 
-        If :attr:`emoji` is "🎰", each value corresponds to a unique combination of symbols, which
+        If :attr:`emoji` is :tg-const:`telegram.Dice.SLOT_MACHINE`, each value
+        corresponds to a unique combination of symbols, which
         can be found at our `wiki <https://github.com/python-telegram-bot/python-telegram-bot/wiki\
         /Code-snippets#map-a-slot-machine-dice-value-to-the-corresponding-symbols>`_.
         However, this behaviour is undocumented and might be changed by Telegram.
 
+    ..
```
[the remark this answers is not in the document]
lemontree210: They introduce a [comment](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#comments) because the following paragraph ("In args, some links...") is meant to only be in the source file, not in rendered output.

telegram/_dice.py
lemontree210: not sure this is needed, because, as I say [here](https://github.com/python-telegram-bot/python-telegram-bot/pull/3351#discussion_r1020484710), this paragraph is not meant for rendering. 

I just thought that some developer that comes after me would think a mistake was made because upper limit for `Dice` seems to be erroneously taken from `Bowling`.

telegram/constants.py
lemontree210: Done

telegram/constants.py
lemontree210: Done

telegram/_dice.py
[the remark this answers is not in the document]
harshil21: TIL

lemontree210:
Do we also want to add non-limit-style things like dimensions of images in [ChatPhoto](https://core.telegram.org/bots/api#chatphoto)?

lemontree210:
Saving [comments](https://t.me/c/1494805131/30064) by @Bibo-Joshi here:

I vote to keep `ChatIniviteLinkLimit` separate from a potential `ChatLimit` enum, since CILL addresses limits of the `ChatInviteLink` class rather than generic chat-related limits.

Putting title limits into a `ChatLimit` enum is okay IMO.

I also vote to keep the current `Chat*` enums separate from a `ChatLimit` enum:

- `ChatAction` is mostly related to send_chat_action and has little to do with the Chat class
- `ChatID` also has little to do with the `Chat` class
- `ChatMemberStatus` is about the `ChatMember` class and not about limits
- `ChatType` is not about limits

lemontree210:
If I create constants in a class (that reference constants in `constants.py`), should I add `..versionadded::` to each of them? E.g.:

```py
    MIN_SWITCH_PM_TEXT_LENGTH: ClassVar[int] = constants.InlineQueryLimit.MIN_SWITCH_PM_TEXT_LENGTH
    """:const:`telegram.constants.InlineQueryLimit.MIN_SWITCH_PM_TEXT_LENGTH`

    .. versionadded:: 20.0
    """
```

Bibo-Joshi:
@lemontree210 Adding it to the clas suffices IMO, as that already implies that the attributes can&#x27;t have existed before :)

lemontree210:
There are still 2 tasks left (they are listed in the description), but I think I found all numbers that needed to be included in `constants.py`. So while I'm working on adding stuff to existing docstrings, please feel free to remind me of anything I haven't found :)

telegram/_dice.py
```diff
@@ -93,3 +106,45 @@ def __init__(self, value: int, emoji: str, *, api_kwargs: JSONDict = None):
     """
```
lemontree210, line 106: I didn't add any `.. versionadded` to class constants that I didn't create, though I think this needs to be done. I also think that the version marking in the line above this one might be wrong: a bowling emoji was added in 13.4, but the constant itself is being introduced in 20.0. So maybe all constants referring to  `DiceEmoji` should also get the `20.0` marking.

harshil21 (approved):
This PR is very high effort and simply insane. It was excruciating for me to just review over the past few days! Well done!

PR #3249, opened 2022-09-18
Make `TelegramObject` immutable

Bibo-Joshi:
Currently based on #3233 so that I don't have to add the calls to `super().__init__` everywhere again …

A few notes:

* ~Must be adapted once #3233 is merged to make sure that `api_kwargs` are immutable as well~
* ~In a few places (in `telegram`), I call the protecded `_(un)freeze` methods of objects to altern their values. At least in some of those cases, we could discuss if it would be desirable to instead create a copy of the object.~ (largue alliviated by #3311)
* `to_dict` sill converts tuples into lists. IMO that's okay, as the dict is mutable anyway

TODO:

- [x] Make sure that classes that override `TelegramObject.__hash__` (*if* that's still needed, anyway) also include `self.__class_` in the hash to comply with the behavior of `TelegramObject.__hash__`
- [x] Merge after integrating API 6.3 changes
- [x] Remove the transition script

### Checklist for PRs

- [x] Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)
- [x] Created new or adapted existing unit tests
- [x] Documented code changes according to the [CSI standard](https://standards.mousepawmedia.com/en/stable/csi.html)
- [x] ~Added myself alphabetically to `AUTHORS.rst` (optional)~
- [x] ~Added new classes & modules to the docs and all suitable `__all__` s~

PR #3353, opened 2022-11-09
Supercharge `TelegramObject.to_dict`

harshil21:
Moves the logic of `TO.to_dict` in subclasses to `TelegramObject` itself, which was mainly dealing with sequences and converting datetimes to timestamps.

### Checklist for PRs

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -3665,7 +3665,7 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         if inline_message_id:
             data["inline_message_id"] = inline_message_id
         if entities:
-            data["entities"] = [me.to_dict() for me in entities]
+            data["entities"] = [me.to_dict(recursive=True) for me in entities]
 
         return await self._send_message(
             "editMessageText",
@@ -5085,7 +5085,7 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         data: JSONDict = {"shipping_query_id": shipping_query_id, "ok": ok}
 
         if shipping_options is not None:
-            data["shipping_options"] = [option.to_dict() for option in shipping_options]
+            data["shipping_options"] = [option.to_dict(True) for option in shipping_options]
         if error_message is not None:
             data["error_message"] = error_message
 
--- a/telegram/_passport/credentials.py
+++ b/telegram/_passport/credentials.py
@@ -404,15 +404,6 @@ class SecureValue(TelegramObject):
 
         return super().de_json(data=data, bot=bot)
 
-    def to_dict(self, recursive: bool = True) -> JSONDict:
-        """See :meth:`telegram.TelegramObject.to_dict`."""
-        data = super().to_dict(recursive=recursive)
-
-        data["files"] = [p.to_dict() for p in self.files]  # type: ignore[union-attr]
-        data["translation"] = [p.to_dict() for p in self.translation]  # type: ignore[union-attr]
-
-        return data
-
 
 class _CredentialsBase(TelegramObject):
     """Base class for DataCredentials and FileCredentials."""
@@ -450,15 +441,6 @@ class DataCredentials(_CredentialsBase):
     def __init__(self, data_hash: str, secret: str, *, api_kwargs: JSONDict = None):
         super().__init__(hash=data_hash, secret=secret, api_kwargs=api_kwargs)
 
-    def to_dict(self, recursive: bool = True) -> JSONDict:
-        """See :meth:`telegram.TelegramObject.to_dict`."""
-        data = super().to_dict(recursive=recursive)
-
-        del data["file_hash"]
-        del data["hash"]
-
-        return data
-
 
 class FileCredentials(_CredentialsBase):
     """
@@ -478,12 +460,3 @@ class FileCredentials(_CredentialsBase):
 
     def __init__(self, file_hash: str, secret: str, *, api_kwargs: JSONDict = None):
         super().__init__(hash=file_hash, secret=secret, api_kwargs=api_kwargs)
-
-    def to_dict(self, recursive: bool = True) -> JSONDict:
-        """See :meth:`telegram.TelegramObject.to_dict`."""
-        data = super().to_dict(recursive=recursive)
-
-        del data["data_hash"]
-        del data["hash"]
-
-        return data
--- a/telegram/_telegramobject.py
+++ b/telegram/_telegramobject.py
@@ -17,12 +17,26 @@
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """Base class for Telegram Objects."""
+import datetime
 import inspect
 import json
 from copy import deepcopy
 from itertools import chain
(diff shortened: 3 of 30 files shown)


telegram/_bot.py
```diff
@@ -3665,7 +3665,7 @@ async def edit_message_text(
         if inline_message_id:
             data["inline_message_id"] = inline_message_id
         if entities:
-            data["entities"] = [me.to_dict() for me in entities]
+            data["entities"] = [me.to_dict(recursive=True) for me in entities]
```
Bibo-Joshi, line 3668: Are the changes in `_bot` necessary?

telegram/_telegramobject.py
```diff
@@ -327,6 +341,34 @@ def to_dict(self, recursive: bool = True) -> JSONDict:
             :obj:`dict`
         """
         out = self._get_attrs(recursive=recursive)
+
+        # Now we should convert TGObjects to dicts inside objects such as sequences, and convert
+        # datetimes to timestamps. This mostly eliminates the need for subclasses to override
+        # `to_dict`
+        for key, value in out.items():
+            if isinstance(value, Sequence) and not isinstance(value, str):
```
Bibo-Joshi, line 349:
```suggestion
            if isinstance(value, Sequence) and not isinstance(value, (str, bytes)):
```

Strictly speaking the ` not isintance`  is not even necessary, as the `hasattr(item, "to_dict"` check will fail, but it's good to keep it so that we don't unnecessarily loop over things that we know are not TOs :D

telegram/_telegramobject.py
```diff
@@ -327,6 +341,34 @@ def to_dict(self, recursive: bool = True) -> JSONDict:
...
+                val = []  # empty list to append our converted values to
+                for item in value:
+                    if hasattr(item, "to_dict"):
+                        val.append(item.to_dict(recursive=recursive))  # type: ignore[union-attr]
+                    # This branch is useful for e.g. List[List[PhotoSize|KeyboardButton]]
+                    elif isinstance(item, (tuple, list)):
```
Bibo-Joshi, line 355: why not `Sequence` here?

On second thought, we do know that all the attributes of TO are only lists (or tuples after #3249), so checking for that (both here and above) might actually suffice 🤔 I might have been caught in the "ABC" spirit 😄

telegram/_telegramobject.py
```diff
@@ -327,6 +341,34 @@ def to_dict(self, recursive: bool = True) -> JSONDict:
             :obj:`dict`
         """
         out = self._get_attrs(recursive=recursive)
+
+        # Now we should convert TGObjects to dicts inside objects such as sequences, and convert
+        # datetimes to timestamps. This mostly eliminates the need for subclasses to override
+        # `to_dict`
+        for key, value in out.items():
+            if isinstance(value, Sequence) and not isinstance(value, str):
+                val = []  # empty list to append our converted values to
```
Bibo-Joshi, line 350:
```suggestion
                if len(value) == 0:
                    out[key] = value
                     continue
                     
                val = []  # empty list to append our converted values to
```

go, early returns, go! 🤡 That way you can also skip lines 365-366

tests/test_bot.py
```diff
@@ -2998,7 +2998,10 @@ async def test_arbitrary_callback_data_pinned_message_reply_to_message(
         )
 
         message = Message(
-            1, None, None, reply_markup=bot.callback_data_cache.process_keyboard(reply_markup)
+            1,
+            dtm.datetime.utcnow(),
```
Bibo-Joshi, line 3002: can you pinpoint why passing an (admittably invalid) `None` here doesn't work anymore? 🤔

tests/test_telegramobject.py
```diff
@@ -238,9 +238,9 @@ def test_pickle(self, bot):
         with pytest.raises(RuntimeError):
             unpickled.get_bot()  # There should be no bot when we pickle TGObjects
 
-        assert unpickled.chat == chat
+        assert unpickled.chat == chat, f"{unpickled.chat._id_attrs} != {chat._id_attrs}"
         assert unpickled.from_user == user
-        assert unpickled.date == date
+        assert unpickled.date == date, f"{unpickled.date} != {date}"
```
Bibo-Joshi, line 243: these changes are not really related right? just double checking, we can keep them either way

Bibo-Joshi (changes requested):
Nice, and it's not even fridy 🥳 Could you also extend the `to_dict` tests to make sure that the `recursive` setting now properly extends to the (nested) lists?

telegram/_bot.py
harshil21: no, I passed in the parameter cause you like it 😃

telegram/_telegramobject.py
harshil21: > why not `Sequence` here?

it is code from my initial attempt, before I figured I could use Sequence.

but okay, will revert to using tuple and list only

tests/test_bot.py
harshil21: it's because we did `data["date"] = to_timestamp(self.date)` in `Message.to_dict`. Where `to_timestamp` returned `None` if `self.date` was `None`. This means that `Message.de_json` would pass. But now we don't do that, so `Message.de_json` returns `KeyError: data["date"]`

tests/test_telegramobject.py
harshil21: yeah not related. They had failed during testing, so wanted to find out what the problem was.

harshil21:
> Could you also extend the `to_dict` tests to make sure that the `recursive` setting now properly extends to the (nested) lists?

This is already tested in the classes which have those, e.g. [`test_to_dict` in `test_inlinekeyboardmarkup`](https://github.com/python-telegram-bot/python-telegram-bot/blob/rm-extra-to-dict/tests/test_inlinekeyboardmarkup.py#L137).

tests/test_bot.py
Bibo-Joshi: > But now we don't do that

Because `TO._get_attrs` skips `None` values, right? Okay, that sounds reasonable :)

harshil21: > Because TO._get_attrs skips None values, right?

Precisely

Bibo-Joshi (approved):
DS failure is unrelated. When the tests pass, this LGTM :)

telegram/_passport/credentials.py
```diff
@@ -450,15 +441,6 @@ class DataCredentials(_CredentialsBase):
     def __init__(self, data_hash: str, secret: str, *, api_kwargs: JSONDict = None):
         super().__init__(hash=data_hash, secret=secret, api_kwargs=api_kwargs)
 
-    def to_dict(self, recursive: bool = True) -> JSONDict:
-        """See :meth:`telegram.TelegramObject.to_dict`."""
-        data = super().to_dict(recursive=recursive)
-
-        del data["file_hash"]
-        del data["hash"]
```
harshil21, line 458: btw I hope this doesn't cause problems, the tests passed soo

Bibo-Joshi: mh, seems to be related to lines 429++, so should be good 🤔

PR #3354, opened 2022-11-09
Replace `@flaky` with pytest marker

harshil21:
Simply use the pytest marker instead of the flaky decorator. Same behaviour. Just saving an import statement..

### DIFF
--- a/tests/test_animation.py
+++ b/tests/test_animation.py
@@ -20,7 +20,6 @@ import os
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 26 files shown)


Bibo-Joshi (commented):
TIL :)
Just for clarification: that marker is still provided by the `flaky` library, i.e. we can't magically drop that dependency, right?

harshil21:
@Bibo-Joshi yep that&#x27;s right

PR #3356, opened 2022-11-11
New classes for API 6.3: `ForumTopic`, `ForumTopicCreated`, `ForumTopicClosed`, `ForumTopicReopened`

lemontree210:
addresses #3340 
- [x] create new classes `ForumTopic`, `ForumTopicCreated`, empty classes 
`ForumTopicClosed`, `ForumTopicReopened`
- [x] add arguments and attributes to `Message`: `forum_topic_created`, `forum_topic_closed`, `forum_topic_reopened`
- [x] update docs
- [x] add tests
  - [x] `to_dict`, `__eq__` etc. for each class
  - [x]  createForumTopic
  - [x] editForumTopic
  - [x] closeForumTopic
  - [x] reopenForumTopic
  - [x] deleteForumTopic (I test it in the fixture)
  - [x] unpinAllForumTopicMessages (test created but triggers `telegram.error.BadRequest: Chat_not_modified.`)
  - [x] getForumTopicIconStickers
- [x] add filters


lemontree210:
Actually, again, maybe the tests are worth adding when the methods are there, so it might make sense to review this as it is now, then hopefully merge, and then we'll add tests with real methods.

telegram/_forumtopic.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @classmethod
+    def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["ForumTopicCreated"]:
+        """See :meth:`telegram.TelegramObject.de_json`."""
+        data = cls._parse_data(data)
+
+        if not data:
+            return None
+
+        return super().de_json(data=data, bot=bot)
```
lemontree210, line 124: Actually I don't really understand why a test of `de_json()` will fail in `test_forum.py` unless I override `TelegramObject`'s `de_json()` here in `ForumTopicCreated`. The overridden method doesn't seem to do anything new (unlike other classes where a `User` is attached, or a timestamp converted, or whatnot).

telegram/_forumtopic.py
```diff
@@ -0,0 +1,155 @@
...
+    from telegram import Bot
+
+
+class ForumTopic(TelegramObject):
+    """
+    This object represents the content of a service message about a new forum topic created in
+    the chat.
+
+    Objects of this class are comparable in terms of equality. Two objects of this class are
+    considered equal, if their :attr:`message_thread_id` and :attr:`name` are equal.
```
lemontree210, line 36: Should I include icon_color as well to be consistent with ForumTopicCreated? On the one hand we need consistency, on the other hand it is crucial that no topic is created with the same name, even though the color is different.

That was the reason why I asked about `icon_color` beiung used in comparison in ForumTopicCreated.

telegram/_forumtopic.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @classmethod
+    def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["ForumTopic"]:
+        """See :meth:`telegram.TelegramObject.de_json`."""
+        data = cls._parse_data(data)
+
+        if not data:
+            return None
+
+        return super().de_json(data=data, bot=bot)
```
lemontree210, line 82: same question as the one about `ForumTopicCreated`.

telegram/_message.py
```diff
@@ -380,6 +393,18 @@ class Message(TelegramObject):
         message_thread_id (:obj:`int`, optional): Unique identifier of a message thread to which
```
harshil21, line 393:
```suggestion
        message_thread_id (:obj:`int`): Optional. Unique identifier of a message thread to which
```

telegram/_message.py
```diff
@@ -673,6 +707,9 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["Message"]:
             data.get("video_chat_participants_invited"), bot
         )
         data["web_app_data"] = WebAppData.de_json(data.get("web_app_data"), bot)
+        data["forum_topic_created"] = ForumTopicCreated.de_json(
+            data.get("forum_topic_created"), bot
+        )
```
harshil21, line 712: we also need to do de_json of `ForumTopicClosed` and Reopened

telegram/_forumtopic.py
harshil21: here's what's happening: 

1. `assert ForumTopic.de_json({}, bot=bot) is None`
2. We arrive [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/_telegramobject.py#L258)
3. data is actually `{}` (not `None`), so execution continues.
4. test fails.

For now to make the test pass, just replace that empty dict to `None`.
I propose to change that condition in TGObject to `if not data:` which covers the empty dict case. @Bibo-Joshi

docs/source/telegram.rst
```diff
@@ -45,6 +45,10 @@ Available Types
     telegram.document
     telegram.file
     telegram.forcereply
+    telegram.forumtopic
+    telegram.forumtopiccreated
+    telegram.forumtopicclosed
+    telegram.forumtopicreopened
```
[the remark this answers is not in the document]
lemontree210: yes, absolutely. I just thought I'd follow Telegram's order, plus only the first class is not empty. But yes, I'll reorder

telegram/_forumtopic.py
```diff
@@ -0,0 +1,155 @@
...
+        name (:obj:`str`): Name of the topic
+        icon_color (:obj:`int`): Color of the topic icon in RGB format
+        icon_custom_emoji_id (:obj:`str`, optional): Unique identifier of the custom emoji shown
+            as the topic icon.
+
+    Attributes:
+        message_thread_id (:obj:`int`): Unique identifier of the forum topic
+        name (:obj:`str`): Name of the topic
+        icon_color (:obj:`int`): Color of the topic icon in RGB format
+        icon_custom_emoji_id (:obj:`str`, optional): Unique identifier of the custom emoji shown
```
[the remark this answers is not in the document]
lemontree210: Not sure I really should do this. Wherever in Telegram API docs there is a type and "Optional" note, in docstrings in source code I find something like:
```
callback_data (:obj:`str` | :obj:`object`, optional): Data to be sent in a callback query....
```

telegram/_message.py
```diff
@@ -253,6 +254,18 @@ class Message(TelegramObject):
...
+            .. versionadded:: 20.0
+        forum_topic_created (:class:`telegram.TopicForumCreated`, optional): Service message:
+            forum topic created
+
+            .. versionadded:: 20.0
+        forum_topic_closed (:class:`telegram.TopicForumClosed`, optional): Service message:
+            forum topic closed
+
+            .. versionadded:: 20.0
+        forum_topic_reopened (:class:`telegram.TopicForumReopened`, optional): Service message:
```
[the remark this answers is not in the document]
lemontree210: oh man 🤦🏻 Thanks, done

telegram/_message.py
```diff
@@ -380,6 +393,18 @@ class Message(TelegramObject):
...
```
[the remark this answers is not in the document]
lemontree210: Same as [above](https://github.com/python-telegram-bot/python-telegram-bot/pull/3356#discussion_r1020879928)

telegram/_message.py
lemontree210: Same as [above](https://github.com/python-telegram-bot/python-telegram-bot/pull/3356#discussion_r1020879928)

telegram/_forumtopic.py
lemontree210: Oh, right, I see the difference: in the overridden method it says `if not data:`. Thanks.

lemontree210: Done

docs/source/telegram.rst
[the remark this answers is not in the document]
lemontree210: Done

telegram/_forumtopic.py
lemontree210: Right, but will it not create the situation when a topic can be created with the same name but different colors? We would introduce a more complex logic (yes, the objects are not equal, but these attributes (`name`) are equal and so no, you, user, can't create a topic). Then a question would arise why we have such a strange equality check that isn't that helpful.

But maybe my logic is flawed somewhere. Or maybe it's OK to introduce separate checks, and it's fine that equality of objects and equality of the most important attribute is not the same thing.

telegram/_message.py
lemontree210: Done, thank you. I got confused by the classes being empty. But it may change.

telegram/_forumtopic.py
harshil21: So I wasn't sure of the history of how to implement this, so I looked it up: This [PR](https://github.com/python-telegram-bot/python-telegram-bot/pull/1724#issue-554374164) states what I had said earlier, i.e using all required args. But of course this may not be suitable in all cases, see e.g. the [original issue](https://github.com/python-telegram-bot/python-telegram-bot/pull/604#issuecomment-301236909) and then again in the [next version](https://github.com/python-telegram-bot/python-telegram-bot/pull/1724#issuecomment-657589186).

> why we have such a strange equality check that isn't that helpful

in some cases, yeah. I personally don't really use the object for comparison and just use the attributes instead. But here I would vote to add `icon_color` for consistency reasons

@Bibo-Joshi

telegram/_forumtopic.py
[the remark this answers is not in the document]
harshil21: In the Args section we do ``` ... (:obj:`..`, optional) ``` but in Attributes section we do ``` ... (:obj:`...`): Optional. ``` since it renders better imo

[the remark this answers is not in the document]
lemontree210: got it now, thanks. Done.

tests/test_forum.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+class TestForumTopic:
+    def test_slot_behaviour(self, forum_group_id, mro_slots):
+        topic = ForumTopic(
+            message_thread_id=forum_group_id,
+            name=TEST_TOPIC_NAME,
+            icon_color=TEST_TOPIC_ICON_COLOR,
+            icon_custom_emoji_id=TEST_TOPIC_EMOJI_ID,
+        )
```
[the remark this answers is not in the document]
lemontree210: Done for `ForumTopic` and `ForumTopicCreated`. Not sure they are needed for empty classes, it's kind of same amount of code or even more.

telegram/_forumtopic.py
Bibo-Joshi: > I would vote to add icon_color for consistency reasons

+1 from  me.

IMO the point of providing a custom `__eq__` still is to provide a comparison logic that is more meaningful that object identity. If the user disagrees with that logic, the can use their very own comparison logic manually.

lemontree210: Done

lemontree210:
This seems ready to be merged if there are no other comments.

tests/test_forum.py
```diff
@@ -0,0 +1,286 @@
...
+        "createForumTopic",
+        {"chat_id": forum_group_id, "name": TEST_TOPIC_NAME},
+    )
+
+    yield result
+
+    result = await bot._post(
+        "deleteForumTopic",
+        {"chat_id": forum_group_id, "message_thread_id": result["message_thread_id"]},
+    )
```
harshil21, line 42: now that the method is implemented upstream, let's use it here

tests/test_forum.py
```diff
@@ -0,0 +1,286 @@
...
+        assert a != e
+        assert hash(a) != hash(e)
+
+    async def test_get_forum_topic_icon_stickers(self, bot):
+        # TODO this fails
+        # we expect the first to stay as it is. This might change in the future.
+        # If we have to fix this test too often maybe just checking it is set to "something"
+        # is enough.
+        emoji_sticker_list = await bot.get_forum_topic_icon_stickers()
+        print(emoji_sticker_list[0].emoji)
```
harshil21, line 140:
```suggestion
```

tests/test_forum.py
```diff
@@ -0,0 +1,286 @@
...
+        assert emoji_sticker_list[0].type == Sticker.CUSTOM_EMOJI
+        assert emoji_sticker_list[0].custom_emoji_id == "5420143492263320958"
+        assert emoji_sticker_list[0].thumb.width == 128
+        assert emoji_sticker_list[0].thumb.height == 128
+        assert emoji_sticker_list[0].thumb.file_size == 4036
+        assert emoji_sticker_list[0].thumb.file_unique_id == "AQADfh0AAso3OEty"
+        assert emoji_sticker_list[0].file_size == 57126
+        assert emoji_sticker_list[0].file_unique_id == "AgADfh0AAso3OEs"
+
+    # we sadly do not have access to a test group right now so we only test params right now
```
harshil21, line 156: we do now so let's use that

tests/test_forum.py
lemontree210: is it? I'm struggling to find it in the branch `api_6.3_update`

tests/test_forum.py
lemontree210: These two tests (this comment and [previous one](https://github.com/python-telegram-bot/python-telegram-bot/pull/3356#discussion_r1021672564)) were created by @Poolitzer if I'm not mistaken (I had copy-pasted this file from a different branch). I would tweak them, of course, if @Poolitzer didn't change these lines in a different branch.

lemontree210:
I'm having trouble adding a test for `ForumTopicCreated`. My idea is that I should create a topic, then get updates, check the last Message and make sure it contains a service message in `forum_topic_created` attribute with the correct topic name. Somehow it doesn't work though, which maybe depends on how `bot` fixture works. (I do get service messages, but old ones.)

I tried to see how it was done for other classes representing service messages, but tests for classes like `VideoChatScheduled`, `MessageAutoDeleteTimerChanged` and others only seem to test slots, equality etc., but not the fact of the service message actually being issued. So I'm struggling to find any relevant examples to follow here :)

tests/test_forum.py
```diff
@@ -213,15 +214,33 @@ async def test_close_and_reopen_forum_topic(self, bot, forum_group_id, real_topi
...
         # the inability to post to the topic
 
         result = await bot.reopen_forum_topic(
             chat_id=forum_group_id,
             message_thread_id=message_thread_id,
         )
-        assert result is True
+        assert result is True, "Failed to reopen forum topic"
+
+    # async def test_unpin_all_forum_topic_messages(self, bot, forum_group_id, real_topic):
```
Bibo-Joshi, line 227: we can uncomment and use [xfail](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-mark-xfail-ref) instead :)

harshil21 (approved):

PR #3357, opened 2022-11-11
Fix Deepsource issues

### LINKED ISSUE #3285: Address DeepSource issues
For a while, DeepSource has been reporting issues in a number of PRs which

* mostly are not really related to the changes in those PRs
* partly are false positives
* we probably just want to ignore.

It would be nice if we somehow could review at least most of these issues and address them. I haven't yet found an overview where one can see all the failures at once, but via https://deepsource.io/gh/python-telegram-bot/python-telegram-bot/history/runs one can at least rather quickly access all PR runs of DS.

In this context, I have already openend a feature-request-thread over at DS: https://discuss.deepsource.io/t/improve-granular-setting-possibility-to-pin-a-deepsource-version/472.

harshil21:
Closes #3285

### DIFF
--- a/telegram/__init__.py
+++ b/telegram/__init__.py
@@ -175,7 +175,7 @@ __all__ = (  # Keep this alphabetically ordered
 )
 
 
-from . import _version
+from . import _version, constants, error, helpers, request, warnings
 from ._bot import Bot
 from ._botcommand import BotCommand
 from ._botcommandscope import (
@@ -312,6 +312,15 @@ from ._telegramobject import TelegramObject
 from ._update import Update
 from ._user import User
 from ._userprofilephotos import UserProfilePhotos
+from ._videochat import (
+    VideoChatEnded,
+    VideoChatParticipantsInvited,
+    VideoChatScheduled,
+    VideoChatStarted,
+)
+from ._webappdata import WebAppData
+from ._webappinfo import WebAppInfo
+from ._webhookinfo import WebhookInfo
 
 #: :obj:`str`: The version of the `python-telegram-bot` library as string.
 #: To get detailed information about the version number, please use :data:`__version_info__`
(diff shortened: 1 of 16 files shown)


telegram/__init__.py
```diff
@@ -175,7 +175,7 @@
 )
 
 
-from . import _version
+from . import _version, constants, error, helpers, request, warnings
```
harshil21, line 178: I don't remember if I deliberately didn't add this as an import, adding here since it is pep-8 compliant.

before this, importing those modules from `telegram` worked fine, `from telegram import *` also imported them

Bibo-Joshi: and now `from telegram import *` no longer imports them? 🤔

Bibo-Joshi (commented):
rest looks good to me :) Do you know if these are all things that DS is currently complaining about? A subset would also already be great 😄

telegram/__init__.py
harshil21: it does. I'm saying this change made no difference at all (except for shutting up Deepsource)

harshil21:
> Do you know if these are all things that DS is currently complaining about?

Their UI/UX is confusing, seeing all "issues" in the repo has like only 6. But if I go to the history tab, I can find more things it complains about, which this fixes.

telegram/__init__.py
Bibo-Joshi: ah, perfect :)

Bibo-Joshi:
@harshil21 Yeah, the UI is a bitt messed up. And I really do hope that they do something about https://discuss.deepsource.io/t/improve-granular-setting-possibility-to-pin-a-deepsource-version/472 … but okay, if those are all the issues that you could find, I&#x27;m good with closing #3285 though this PR 👌🏻
let&#x27;s try to remember to merge master into the pending PRs to see if any new issues come up 😅

PR #3365, opened 2022-11-14
try updating meta config

Bibo-Joshi:
setuptools deprected the lincense_file option in favor of a license_files (plural) option. the plural form was introduced in v42.0.0 in 11/2019. the singular form was deprecated in v56.0.0 in 04/2021. setuptools is currently on v65.5.1 (04.11.2022).


Bibo-Joshi:
Before merge, I want to manually test running the setup script on linux, b/c on windows the subprocess call to setup-raw.py fails for me …

Bibo-Joshi:
Confirmed by Pool 🎉

PR #3376, opened 2022-11-20
Fix command handler doc

Viicos:
`PrefixHandler` probably has been moved since.


Bibo-Joshi:
Thanks for the catch!

PR #3380, opened 2022-11-22
Fix accidental wrong renaming in #3223

Bibo-Joshi:
See https://t.me/pythontelegrambotgroup/639703

### Checklist for PRs

- ~[x] Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)~
- ~[x] Documented code changes according to the [CSI standard](https://standards.mousepawmedia.com/en/stable/csi.html)~
- ~[x] Added myself alphabetically to `AUTHORS.rst` (optional)~
- ~[x] Added new classes & modules to the docs and all suitable `__all__` s~

### DIFF
--- a/telegram/_files/file.py
+++ b/telegram/_files/file.py
@@ -130,13 +130,13 @@ class File(TelegramObject):
 
         Note:
             If :paramref:`custom_path` isn't provided and :attr:`file_path` is the path of a
-              local file (which is the case when a Bot API Server is running in local mode), this
-              method will just return the path.
+            local file (which is the case when a Bot API Server is running in local mode), this
+            method will just return the path.
 
             The only exception to this are encrypted files (e.g. a passport file). For these, a
-              file with the prefix `decrypted_` will be created in the same directory as the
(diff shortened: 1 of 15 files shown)


telegram/_files/file.py
```diff
@@ -207,7 +207,7 @@ async def download_to_memory(
         filename.write_bytes(buf)
         return filename
 
-    async def download_to_object(
+    async def download_to_memory(
         self,
         out: BinaryIO,
         read_timeout: ODVInput[float] = DEFAULT_NONE,
```
harshil21, line 213: btw shouldn't these args be kw-only? Or should we do it in another pr later

Bibo-Joshi: oh, right. no, I'll fix it. Don't we test for that anywhere? 🤔

harshil21: I think the test is only for bot

PR #3381, opened 2022-11-22
Fix: Fix return value of `create_forum_topic` shortcut

Poolitzer:
Quick fix, just surprised mypy didn't catch it

### DIFF
--- a/telegram/_chat.py
+++ b/telegram/_chat.py
@@ -2640,7 +2640,7 @@ class Chat(TelegramObject):
         .. versionadded:: 20.0
 
         Returns:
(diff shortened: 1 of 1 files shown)


lemontree210 (approved):

lemontree210:
Just in case: I could also have added the following bit to docstring, but I see that it wasn't done anywhere in shortcut methods for respective methods in `Bot`.

```
Raises: 
    :class:`telegram.error.TelegramError`
```

harshil21:
@lemontree210 not needed since it&#x27;s there in the original

Bibo-Joshi (approved):

PR #3385, opened 2022-11-24
add `caption`, `parse_mode`, `caption_entities`, `disable_web_page_preview` parameters to `Bot._send_message`

### LINKED ISSUE #3372: Add `caption`, `caption_entities` as parameters to `Bot._send_message`
This should do 2 things: 
- Add `caption_entities` to `_send_message`, so we can reduce repetition even further.
- Comment in `_send_message` that we do this to reduce repetition (DRY).

Ref https://github.com/python-telegram-bot/python-telegram-bot/pull/3346#discussion_r1026897846

lemontree210:
closes #3372

Not sure I need to add tests (I think in this case I just needed to make sure existing tests don't fail).


telegram/_bot.py
```diff
@@ -438,10 +438,15 @@ async def _send_message(
         data: JSONDict,
         reply_to_message_id: int = None,
         disable_notification: ODVInput[bool] = DEFAULT_NONE,
+        # To reduce repetition (if-clauses) in methods that call this method, following parameters
+        # are included separately instead of just passing them in data parameter:
+        # reply_to_message_id, reply_markup, message_thread_id, caption, caption_entities.
```
Bibo-Joshi, line 443: that's the whole purpose of this method ;) Please either convert to a docstring or remove this comment.

Bibo-Joshi (changes requested):

telegram/_bot.py
lemontree210: Sure, I get it. But in #3372 @harshil21 suggested that a comment be included in `Bot._send_message`. Since this method has no docstring at all, I decided that a comment was meant. If we're talking about a docstring, then maybe what was meant was that I should add something to a docstring of `Bot.send_message`, not `Bot._send_message`.

Bibo-Joshi: Ah okay. Well, then let's do a docstring for `_send_message`, i.e. give the method an explanation about why it's there. Adding comments about `_send_message` to the docstring of `Bot.send_message` would not be appropriate since the latter is a public method and the user should not be concerned with the internals.

lemontree210: I added a full-fledged docstring, not sure it was needed though because Sphinx doesn't render docstrings for protected methods (as far as I can tell).

telegram/_bot.py
```diff
@@ -442,13 +442,51 @@ async def _send_message(
...
+        Returns:
+            :class:`telegram.Message`: On success, depending on a type of operation, the message
+                or :obj:`True` is returned.
+
+        Raises:
+            :class:`telegram.error.TelegramError`
+
+        """
         if reply_to_message_id is not None:
             data["reply_to_message_id"] = reply_to_message_id
```
harshil21, line 491: can you move this a few lines below so its a little more consistent

harshil21 (commented):
`copy_message` and `edit_message_caption` still have those if clauses left to remove.

telegram/_bot.py
harshil21: yep sphinx is not going to render this, this is documentation by devs for devs 😄

telegram/_bot.py
```diff
@@ -442,13 +442,51 @@ async def _send_message(
...
+            caption_entities (List[:class:`telegram.MessageEntity`], optional): |caption_entities|
+
+        Returns:
+            :class:`telegram.Message`: On success, depending on a type of operation, the message
+                or :obj:`True` is returned.
+
+        Raises:
+            :class:`telegram.error.TelegramError`
+
+        """
```
Bibo-Joshi, line 489:
```suggestion
        """Protected method to send or edit messages of any type. 

        It is here to reduce repetition of if-else closes in the different bot methods, i.e. this method takes care of adding its parameters to `data` if appropriate.

        Depending on the bot method, returns either `True` or the message.
        """
```

Sorry, I should have been more explicit in my previous comment. 

> give the method an explanation about why it's there.

What I'm looking for here is mostly a short info about what this method does. Since it's a description for the complete method and not only for parts of its body, a docstring seems more approriate to me than a `#` style comment. Nonetheless, because sphinx doesn't render docstrings of protected methods, it *is* basically a comment, so correct sphinx/rst-syntax is not strictly needed.
Because the parameters are already well-documented all over `_bot.py`, I don't see a need to repeat this here. Finally the `Note` that you added is the most integral part of the docstring :)

telegram/_bot.py
lemontree210: Done

telegram/_bot.py
lemontree210: Done

lemontree210:
> `copy_message` and `edit_message_caption` still have those if clauses left to remove.

Yes, I agree on `edit_message_caption()`, but `copy_message()` calls `Bot._post()`, not `Bot._send_message()`, so not sure this change is applicable for `copy_message()`

telegram/_bot.py
lemontree210: Done

lemontree210:
BTW a (maybe too) philosophical question. Can the fact that the generic `._send_message()` method knows about other methods having parameters such as `.caption` (which pertain not to any message but to a specific type of message) be considered as breaking LoD? Maybe it's OK in this context because all of these params are part of official Telegram API.

Bibo-Joshi:
>  fact that the generic `._send_message()` method knows about other methods having parameters such as `.caption`

IMO this is not the case. `_send_message` just offers that parameter, but it doesn't really know if any other method actually uses it.

lemontree210:
Added `parse_mode` as mentioned in [this comment](https://github.com/python-telegram-bot/python-telegram-bot/issues/3372#issuecomment-1326595278) for #3372

telegram/_bot.py
```diff
@@ -713,7 +730,6 @@ async def send_message(
         data: JSONDict = {
             "chat_id": chat_id,
             "text": text,
-            "parse_mode": parse_mode,
             "disable_web_page_preview": disable_web_page_preview,
```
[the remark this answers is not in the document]
lemontree210: But it _was_ moved to `_send_message`...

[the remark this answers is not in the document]
harshil21: he means `disable_web_page_preview`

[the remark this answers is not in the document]
Bibo-Joshi: my bad. github doesn't let me comment on lines that weren't changed 🤦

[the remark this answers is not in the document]
lemontree210: > github doesn't let me comment on lines that weren't changed

Oh right, yes, that's a bit annoying that it doesn't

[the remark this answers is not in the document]
lemontree210: Done

harshil21 (approved):

telegram/ext/_extbot.py
```diff
@@ -469,6 +469,10 @@ async def _send_message(
         allow_sending_without_reply: ODVInput[bool] = DEFAULT_NONE,
         protect_content: ODVInput[bool] = DEFAULT_NONE,
         message_thread_id: int = None,
+        caption: str = None,
+        parse_mode: ODVInput[str] = DEFAULT_NONE,
+        caption_entities: Union[List["MessageEntity"], Tuple["MessageEntity", ...]] = None,
+        disable_web_page_preview: ODVInput[bool] = DEFAULT_NONE,
```
Bibo-Joshi, line 475: needs to be passed to super as well - that will also fix the failing tests :)

Bibo-Joshi (approved):
LGTM except for the comment below :)

telegram/ext/_extbot.py
lemontree210: I was being sloppy on this occasion, ran only part of the tests locally and left it to CI to crash 🤦🏻

lemontree210: Done

Bibo-Joshi: > I was being sloppy on this occasion, ran only part of the tests locally and left it to CI to crash 🤦🏻 Should at least have checked back sooner to see how many failed

Been there, done that :D

PR #3393, opened 2022-11-25
warning to install optional dependencies

### LINKED ISSUE #3391: Make `CallbackContext.job_queue` raise a warning when `job_queue` is `None`
When a user doesn't have APScheduler installed, doing something like `context.job_queue.run_repeating` will fail since `job_queue` is `None`. So we should:

1. raise a warning inside the `context.job_queue` property telling the user to run `pip install python-telegram-bot[job-queue]`
2. add the same note to each example which uses an optional dependency.

clot27:
closes #3391

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -40,7 +40,7 @@ up a job to send a message to that user after 30 seconds. The user can
 also cancel the timer by sending ``/unset``. To learn more about the
 ``JobQueue``, read `this wiki
 article <https://github.com/python-telegram-bot/python-telegram-bot/wiki/Extensions-%E2%80%93-JobQueue>`__.
-Note: To use JobQueue, you must install ptb via ``pip install python-telegram-bot[job-queue]``
+Note: To use ``JobQueue``, you must install PTB via ``pip install python-telegram-bot[job-queue]``
 
 :any:`examples.conversationbot`
 -------------------------------
@@ -116,7 +116,7 @@ Don’t forget to enable and configure payments with
 `@BotFather <https://telegram.me/BotFather>`_. Check out this
 `guide <https://github.com/python-telegram-bot/python-telegram-bot/wiki/Telegram-Passport>`__
 on Telegram passports in PTB.
-Note: To use telegram passport, you must install ptb via ``pip install python-telegram-bot[passport]``
+Note: To use Telegram Passport, you must install PTB via ``pip install python-telegram-bot[passport]``
 
 :any:`examples.paymentbot`
 --------------------------
@@ -164,7 +164,7 @@ combination with ``telegram.ext.Application``.
 
 This example showcases how PTBs “arbitrary callback data” feature can be
 used.
-Note: To use arbitrary callback data, you must install ptb via ``pip install python-telegram-bot[callback-data]``
+Note: To use arbitrary callback data, you must install PTB via ``pip install python-telegram-bot[callback-data]``
 
 Pure API
 --------
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -40,6 +40,7 @@ up a job to send a message to that user after 30 seconds. The user can
 also cancel the timer by sending ``/unset``. To learn more about the
 ``JobQueue``, read `this wiki
 article <https://github.com/python-telegram-bot/python-telegram-bot/wiki/Extensions-%E2%80%93-JobQueue>`__.
+Note: To use JobQueue, you must install ptb via ``pip install python-telegram-bot[job-queue]``
 
 :any:`examples.conversationbot`
 -------------------------------
@@ -115,6 +116,7 @@ Don’t forget to enable and configure payments with
 `@BotFather <https://telegram.me/BotFather>`_. Check out this
 `guide <https://github.com/python-telegram-bot/python-telegram-bot/wiki/Telegram-Passport>`__
 on Telegram passports in PTB.
+Note: To use telegram passport, you must install ptb via ``pip install python-telegram-bot[passport]``
 
 :any:`examples.paymentbot`
 --------------------------
@@ -162,6 +164,7 @@ combination with ``telegram.ext.Application``.
 
 This example showcases how PTBs “arbitrary callback data” feature can be
 used.
+Note: To use arbitrary callback data, you must install ptb via ``pip install python-telegram-bot[callback-data]``
 
 Pure API
 --------
--- a/examples/arbitrarycallbackdatabot.py
+++ b/examples/arbitrarycallbackdatabot.py
@@ -8,7 +8,8 @@ For detailed info on arbitrary callback data, see the wiki page at
 https://github.com/python-telegram-bot/python-telegram-bot/wiki/Arbitrary-callback_data
 
 Note:
-To use arbitrary callback data, you must install ptb via `pip install python-telegram-bot[callback-data]`
+To use arbitrary callback data, you must install PTB via
+`pip install python-telegram-bot[callback-data]`
 """
 import logging
 from typing import List, Tuple, cast
--- a/examples/arbitrarycallbackdatabot.py
+++ b/examples/arbitrarycallbackdatabot.py
@@ -6,6 +6,9 @@
 
 For detailed info on arbitrary callback data, see the wiki page at
 https://github.com/python-telegram-bot/python-telegram-bot/wiki/Arbitrary-callback_data
+
+Note:
+To use arbitrary callback data, you must install ptb via `pip install python-telegram-bot[callback-data]`
 """
 import logging
 from typing import List, Tuple, cast
--- a/examples/passportbot.py
+++ b/examples/passportbot.py
@@ -10,6 +10,8 @@ See https://telegram.org/blog/passport for info about what telegram passport is.
 See https://github.com/python-telegram-bot/python-telegram-bot/wiki/Telegram-Passport
  for how to use Telegram Passport properly with python-telegram-bot.
 
+Note:
+To use telegram passport, you must install PTB via `pip install python-telegram-bot[passport]`
 """
 import logging
 from pathlib import Path
--- a/examples/passportbot.py
+++ b/examples/passportbot.py
@@ -11,7 +11,8 @@ See https://github.com/python-telegram-bot/python-telegram-bot/wiki/Telegram-Pas
  for how to use Telegram Passport properly with python-telegram-bot.
 
 Note:
-To use telegram passport, you must install PTB via `pip install python-telegram-bot[passport]`
+To use Telegram Passport, you must install PTB via
+`pip install python-telegram-bot[passport]`
 """
 import logging
 from pathlib import Path
--- a/examples/timerbot.py
+++ b/examples/timerbot.py
@@ -16,6 +16,9 @@ Usage:
 Basic Alarm Bot example, sends a message after a set time.
 Press Ctrl-C on the command line or send a signal to the process to stop the
 bot.
+
+Note:
+To use arbitrary callback data you must install ptb via `pip install python-telegram-bot[callback-data]`
 """
 
 import logging
--- a/examples/timerbot.py
+++ b/examples/timerbot.py
@@ -18,7 +18,8 @@ Press Ctrl-C on the command line or send a signal to the process to stop the
 bot.
 
 Note:
-To use arbitrary callback data you must install ptb via `pip install python-telegram-bot[callback-data]`
+To use arbitrary callback data, you must install ptb via
+`pip install python-telegram-bot[callback-data]`
 """
 
 import logging
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -64,7 +64,7 @@ from telegram.ext._utils.types import BD, BT, CCT, CD, JQ, UD, ConversationKey, 
 
 if TYPE_CHECKING:
     from telegram import Message
-    from telegram.ext import ConversationHandler
+    from telegram.ext import ConversationHandler, JobQueue
     from telegram.ext._applicationbuilder import InitApplicationBuilder
     from telegram.ext._jobqueue import Job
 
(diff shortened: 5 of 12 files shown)


Bibo-Joshi (changes requested):
The changes themself look got to me 👍 

* pre-commit is complaining a bit. If you run `pre-commit install` locally, the pre-commit hooks will run before you can commit 
* We'll need a test to ensure that the warning is issued as expected, i.e. the warning is there, the messages is the expected one and the warning points the user to the line where `context.job_queue` is accessed. You can e.g. have a look at `TestJobQueue.test_run_daily` for some inspiration on how that can be done :)

docs/source/examples.rst
```diff
@@ -40,6 +40,7 @@ up a job to send a message to that user after 30 seconds. The user can
 also cancel the timer by sending ``/unset``. To learn more about the
 ``JobQueue``, read `this wiki
 article <https://github.com/python-telegram-bot/python-telegram-bot/wiki/Extensions-%E2%80%93-JobQueue>`__.
+Note: To use JobQueue, you must install ptb via ``pip install python-telegram-bot[job-queue]``
```
harshil21, line 43:
```suggestion
Note: To use `JobQueue`, you must install PTB via ``pip install python-telegram-bot[job-queue]``
```

docs/source/examples.rst
```diff
@@ -115,6 +116,7 @@ Don’t forget to enable and configure payments with
 `@BotFather <https://telegram.me/BotFather>`_. Check out this
 `guide <https://github.com/python-telegram-bot/python-telegram-bot/wiki/Telegram-Passport>`__
 on Telegram passports in PTB.
+Note: To use telegram passport, you must install ptb via ``pip install python-telegram-bot[passport]``
```
harshil21, line 119:
```suggestion
Note: To use Telegram Passport, you must install PTB via ``pip install python-telegram-bot[passport]``
```

docs/source/examples.rst
```diff
@@ -162,6 +164,7 @@ combination with ``telegram.ext.Application``.
 
 This example showcases how PTBs “arbitrary callback data” feature can be
 used.
+Note: To use arbitrary callback data, you must install ptb via ``pip install python-telegram-bot[callback-data]``
```
harshil21, line 167:
```suggestion
Note: To use arbitrary callback data, you must install PTB via ``pip install python-telegram-bot[callback-data]``
```

examples/arbitrarycallbackdatabot.py
```diff
@@ -6,6 +6,9 @@
 
 For detailed info on arbitrary callback data, see the wiki page at
 https://github.com/python-telegram-bot/python-telegram-bot/wiki/Arbitrary-callback_data
+
+Note:
+To use arbitrary callback data, you must install ptb via `pip install python-telegram-bot[callback-data]`
```
harshil21, line 11:
```suggestion
To use arbitrary callback data, you must install PTB via `pip install python-telegram-bot[callback-data]`
```

examples/passportbot.py
```diff
@@ -10,6 +10,8 @@
 See https://github.com/python-telegram-bot/python-telegram-bot/wiki/Telegram-Passport
  for how to use Telegram Passport properly with python-telegram-bot.
 
+Note:
+To use telegram passport, you must install PTB via `pip install python-telegram-bot[passport]`
```
harshil21, line 14:
```suggestion
To use Telegram Passport, you must install PTB via `pip install python-telegram-bot[passport]`
```

examples/timerbot.py
```diff
@@ -16,6 +16,9 @@
 Basic Alarm Bot example, sends a message after a set time.
 Press Ctrl-C on the command line or send a signal to the process to stop the
 bot.
+
+Note:
+To use arbitrary callback data you must install ptb via `pip install python-telegram-bot[callback-data]`
```
harshil21, line 21:
```suggestion
To use arbitrary callback data you must install PTB via `pip install python-telegram-bot[callback-data]`
```

telegram/ext/_callbackcontext.py
```diff
@@ -389,6 +390,11 @@ def job_queue(self) -> Optional["JobQueue"]:
         .. seealso:: `Job Queue <https://github.com/python-telegram-bot/
             python-telegram-bot/wiki/Extensions-%E2%80%93-JobQueue>`_
         """
+        if self._application.job_queue is None:
+            warn(
+                "No `JobQueue` Set up. To use `JobQueue`, you must install PTB via"
```
harshil21, line 395:
```suggestion
                "No `JobQueue` set up. To use `JobQueue`, you must install PTB via"
```

harshil21 (changes requested):
Some stylistic suggestions below. 

Also the more likely scenario is when user does `application.job_queue.run_*` and faces an error. Shouldn't we make that a property to raise this warning as well?

Bibo-Joshi:
> Also the more likely scenario is when user does `application.job_queue.run_*` and faces an error. Shouldn't we make that a property to raise this warning as well?

I guess it's hard to say if `app.job_queue` or `context.job_queue` is used more … but anyway, making `Application.job_queue` a property and having it issue a warning just like `CallbackContext.job_queue` sounds like a nice idea :)

telegram/ext/_conversationhandler.py
```diff
@@ -675,12 +675,15 @@ def _schedule_job(
...
-            )
+            if application.job_queue:
+                j_queue = application.job_queue
+                self.timeout_jobs[conversation_key] = j_queue.run_once(
+                    self._trigger_timeout,
+                    self.conversation_timeout,  # type: ignore[arg-type]
+                    data=_ConversationTimeoutContext(
+                        conversation_key, update, application, context
+                    ),
+                )
```
Bibo-Joshi, line 686: the comment in line 677 essentially means that we don't need an if-clause here :)

telegram/ext/_application.py
```diff
@@ -337,6 +337,22 @@ def concurrent_updates(self) -> int:
         """
         return self._concurrent_updates
 
+    @property
+    def job_queue(self) -> Optional["JobQueue"]:
+        """
+        :class:`telegram.ext.JobQueue`: The :class:`JobQueue` used by the
```
Bibo-Joshi, line 343: The lines on `job_queue` in the `Attrs` block of the class-docstring can be removed, then :)

telegram/ext/_callbackcontext.py
```diff
@@ -389,6 +390,11 @@ def job_queue(self) -> Optional["JobQueue"]:
         .. seealso:: `Job Queue <https://github.com/python-telegram-bot/
             python-telegram-bot/wiki/Extensions-%E2%80%93-JobQueue>`_
         """
+        if self._application.job_queue is None:
+            warn(
+                "No `JobQueue` set up. To use `JobQueue`, you must install PTB via "
+                "`pip install python-telegram-bot[job_queue]`."
+            )
```
Bibo-Joshi, line 397: the `if self._application.job_queue` will already trigger a warning if there is no `JobQueue`. But that warning will point the user to this line in `CallbackContext` instead of the line in the users code where they accessed `context.job_queue` … We could [surpress the first warning](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) and issue a new one (as currently done). but I'm not sure if it's worth the effort … cc @harshil21

harshil21: hmm or we could use the private attribute here instead of the property, so we bypass the warning. That's better isn't it?

telegram/ext/_conversationhandler.py
clot27: Welp, the mypy-ptb pre commit run fails if I dont put if clause here... :thinking:

telegram/ext/_callbackcontext.py
Bibo-Joshi: mh, would be okay with me, I guess. The only other way around that would be a `Application.get_job_queue` method, I guess and I still think that would be overkill

telegram/ext/_conversationhandler.py
```diff
@@ -676,11 +676,14 @@ def _schedule_job(
...
-                data=_ConversationTimeoutContext(conversation_key, update, application, context),
-            )
+            if j_queue:
+                self.timeout_jobs[conversation_key] = j_queue.run_once(
+                    self._trigger_timeout,
+                    self.conversation_timeout,  # type: ignore[arg-type]
+                    data=_ConversationTimeoutContext(
+                        conversation_key, update, application, context
+                    ),
+                )
```
Bibo-Joshi, line 686: interesting that mypy fails for you locally, since it's not failing in the CI. if you add `--no-verify` to `git commit …` you can skip the pre-commit hooks. Let's see what the CI thinks

tests/test_application.py
```diff
@@ -384,23 +394,18 @@ def test_builder(self, app):
         builder_1.token(app.bot.token)
         builder_2.token(app.bot.token)
 
-    @pytest.mark.parametrize("job_queue", (True, False))
-    async def test_start_stop_processing_updates(self, bot, job_queue):
+    # @pytest.mark.parametrize("job_queue", (True, False))
```
Bibo-Joshi, line 397: We'll still want to test both cases. The test probably failed due to the new warnings being issued, right? You can suppress that by adding a decorator

```
@pytest.mark.filterwarnings("ignore::telegram.warnings.PTBUserWarning")
```

Same for `TestApplicationBuilder.test_no_job_queue` btw :)

tests/test_callbackcontext.py
```diff
@@ -58,6 +58,20 @@ def test_from_job(self, app):
...
+            "`pip install python-telegram-bot[job_queue]`."
+        )
+
+        callback_context = CallbackContext(app)
+        assert callback_context.job_queue is app.job_queue
+        app = ApplicationBuilder().job_queue(None).token(bot.token).build()
+        callback_context = CallbackContext(app)
+        assert callback_context.job_queue is None
+        assert len(recwarn) == 1
+        assert str(recwarn[0].message) == expected_warning
```
Bibo-Joshi, line 73: looks good overall :)

```suggestion
        assert str(recwarn[0].message) == expected_warning
        assert recwarn[0].filename == __file__, "wrong stacklevel"
```

The additional assertion makes sure that the warning is issued directly from the line where the user uses `context.job_queue` so that they are pointed to the right place.

Please also add an analogous test directly to `TestApplication` where the warning behavior of `Application.job_queue` should be tested as well

harshil21 (commented):
~okay I'm not sure why Github is not showing me the changes in conversationhandler.py in the files changed tab~, but I think you should replace all instances of `application.job_queue` with `application._job_queue` to avoid unintentionally raising a warning in user code

tests/test_callbackcontext.py
clot27: IISC this is failing the tests :thinking:

tests/test_application.py
```diff
@@ -201,6 +201,18 @@ async def post_shutdown(application: Application) -> None:
...
+        expected_warning = (
+            "No `JobQueue` set up. To use `JobQueue`, you must install PTB via "
+            "`pip install python-telegram-bot[job_queue]`."
+        )
+        assert app.job_queue is app._job_queue
+        application = ApplicationBuilder().token(bot.token).job_queue(None).build()
+        assert application.job_queue is None
+        assert len(recwarn) == 1
+        assert str(recwarn[0].message) == expected_warning
+        assert recwarn[0].filename == __file__, "wrong stacklevel"
```
lemontree210, line 214: I would add some comment here explaining the intention. Maybe it's just me, but without reading @Bibo-Joshi 's comments in the review I would be pretty confused if I saw this line in the test

clot27: specifically which line? the last one?

tests/test_application.py
```diff
@@ -201,6 +201,18 @@ async def post_shutdown(application: Application) -> None:
                 post_shutdown=None,
             )
 
+    def test_job_queue(self, bot, app, recwarn):
+        expected_warning = (
+            "No `JobQueue` set up. To use `JobQueue`, you must install PTB via "
+            "`pip install python-telegram-bot[job_queue]`."
```
lemontree210, line 207: There's a lot of repetition of this warning message, so I usually have the urge to extract it into a constant, but I know there's some skepticism about this, especially in tests.

tests/test_application.py
lemontree210: Yes, with "wrong stacklevel". That's funny because I chose that line when reviewing, somehow GH isn't showing it because I made no suggested changes :)

lemontree210: And then maybe @Bibo-Joshi wouldn't agree that this line deserves a special comment anyway

Bibo-Joshi: I'm very much fine with that if it's of help for someone :)

tests/test_callbackcontext.py
Bibo-Joshi: Then the stacklevel is indeed wrong :) The `stacklevel` parameter of `warn` does the following. When an update is handled, roughly the following methods are called:

`App.process_update`, which calls `BaseHandler.handle_update`, which calls the users `callback` in which the user accesses `context.job_queue`, which accesses the property `Application.job_queue`, which calls `_utils.warn` which issues the warning.

Now by default the warning is issued directly from `tg._utils.warn` and that filename will be shown to the user as the origin of the warning. But this is not helpful for the user at all b/c the user didn't manually call `_utils.warn`. Passing a higher value to `stacklevel` tells python to go up in the "call stack" and we want to pass a value such that the accessing of `context.job_queue` is shown as origin of the warning :)

lemontree210: Wow, thanks for the explanation!

tests/test_application.py
lemontree210: After reading up on this in [Python docs](https://docs.python.org/3/library/warnings.html#warnings.warn), I'm not sure anymore if a comment is warranted. I don't know what level of Python proficiency we're expecting :)

Bibo-Joshi (commented):
I took the liberty to tie up the very last loose ends :) I'm happy now. If harshil is as well, we can merge.

PR #3396, opened 2022-11-27
Drop `if` clauses from each `Bot` method

harshil21:
Simplifies the bot methods by passing all arguments in `data`. Since `None` values are dropped in `Bot._post` anyway, we can do this safely

### Checklist for PRs

- [ ] ~Documented code changes according to the [CSI standard](https://standards.mousepawmedia.com/en/stable/csi.html)~

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -729,13 +729,7 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
             :class:`telegram.error.TelegramError`
 
         """
-        data: JSONDict = {
-            "chat_id": chat_id,
-            "text": text,
-        }
-
-        if entities:
-            data["entities"] = entities
+        data: JSONDict = {"chat_id": chat_id, "text": text, "entities": entities}
 
         return await self._send_message(  # type: ignore[return-value]
             "sendMessage",
@@ -859,14 +853,12 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         Raises:
             :class:`telegram.error.TelegramError`
         """
-        data: JSONDict = {}
+        data: JSONDict = {
+            "chat_id": chat_id,
+            "from_chat_id": from_chat_id,
+            "message_id": message_id,
+        }
 
-        if chat_id:
-            data["chat_id"] = chat_id
-        if from_chat_id:
-            data["from_chat_id"] = from_chat_id
-        if message_id:
-            data["message_id"] = message_id
         return await self._send_message(  # type: ignore[return-value]
             "forwardMessage",
             data,
@@ -1078,17 +1070,11 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         data: JSONDict = {
             "chat_id": chat_id,
             "audio": self._parse_file_input(audio, Audio, filename=filename),
+            "duration": duration,
+            "performer": performer,
+            "title": title,
+            "thumb": self._parse_file_input(thumb, attach=True),  # type: ignore[arg-type]
         }
-
-        if duration:
-            data["duration"] = duration
-        if performer:
-            data["performer"] = performer
-        if title:
-            data["title"] = title
-
-        if thumb:
-            data["thumb"] = self._parse_file_input(thumb, attach=True)
 
         return await self._send_message(  # type: ignore[return-value]
             "sendAudio",
@@ -1205,12 +1191,9 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
         data: JSONDict = {
             "chat_id": chat_id,
             "document": self._parse_file_input(document, Document, filename=filename),
+            "disable_content_type_detection": disable_content_type_detection,
+            "thumb": self._parse_file_input(thumb, attach=True),  # type: ignore[arg-type]
         }
-
-        if disable_content_type_detection is not None:
-            data["disable_content_type_detection"] = disable_content_type_detection
-        if thumb:
-            data["thumb"] = self._parse_file_input(thumb, attach=True)
 
         return await self._send_message(  # type: ignore[return-value]
             "sendDocument",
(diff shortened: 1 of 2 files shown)


telegram/_bot.py
```diff
@@ -4644,32 +4531,22 @@ async def promote_chat_member(
...
+            "can_post_messages": can_post_messages,
+            "can_edit_messages": can_edit_messages,
+            "can_delete_messages": can_delete_messages,
+            "can_invite_users": can_invite_users,
+            "can_restrict_members": can_restrict_members,
+            "can_pin_messages": can_pin_messages,
+            "can_promote_members": can_promote_members,
+            "can_manage_chat": can_manage_chat,
+            "can_manage_video_chats": can_manage_video_chats,
+            "can_manage_topics": can_manage_topics,
```
lemontree210, line 4548: these and similar lines really create a temptation to use `locals()` :)

harshil21: yep, only have to remove "self" from locals()

at least I had copilot to write these for me, which should reduce typos

telegram/_bot.py
```diff
@@ -1078,18 +1070,12 @@ async def send_audio(
         data: JSONDict = {
             "chat_id": chat_id,
             "audio": self._parse_file_input(audio, Audio, filename=filename),
+            "duration": duration,
+            "performer": performer,
+            "title": title,
+            "thumb": self._parse_file_input(thumb, attach=True),  # type: ignore[arg-type]
```
Bibo-Joshi, line 1076: Instead of the `type: ignore` we can probably improve the type annotation of `_parse_file_input` (maybe with `@override`) or just do something like `"thumb" = self._pfi(thumb) if thumb else None` - or in fact keep the if-clause as is :D

telegram/_bot.py
```diff
@@ -2879,12 +2817,9 @@ async def get_file(
             api_kwargs=api_kwargs,
         )
 
-        if result.get("file_path") and not is_local_file(  # type: ignore[union-attr]
-            result["file_path"]  # type: ignore[index]
-        ):
-            result[  # type: ignore[index]
-                "file_path"
-            ] = f"{self._base_file_url}/{result['file_path']}"  # type: ignore[index]
+        file_path = result.get("file_path") if isinstance(result, dict) else None
```
Bibo-Joshi, line 2820: ignore return-type or a `typing.cast` would be fine with me instead of using `isinstance (result, dict)`

tests/test_bot.py
```diff
@@ -1910,7 +1910,7 @@ async def test_set_webhook_delete_webhook_drop_pending_updates(
         self, bot, drop_pending_updates, monkeypatch
     ):
         async def make_assertion(url, request_data: RequestData, *args, **kwargs):
-            data = request_data.json_parameters
+            data = request_data.parameters
```
Bibo-Joshi, line 1913: why? 🤔

telegram/_bot.py
```diff
@@ -3230,16 +3158,13 @@ async def edit_message_text(
...
-        if inline_message_id:
-            data["inline_message_id"] = inline_message_id
-        if entities:
-            data["entities"] = [me.to_dict(recursive=True) for me in entities]
+        data: JSONDict = {
+            "text": text,
+            "chat_id": chat_id,
+            "message_id": message_id,
+            "inline_message_id": inline_message_id,
+            "entities": [e.to_dict(recursive=True) for e in entities] if entities else None,
```
Bibo-Joshi, line 3166: do we need to `to_dict`? IIRC `RequestParameter` takes care of that …

telegram/_bot.py
```diff
@@ -4374,12 +4260,14 @@ async def answer_shipping_query(  # pylint: disable=invalid-name
...
-        data: JSONDict = {"shipping_query_id": shipping_query_id, "ok": ok}
-
-        if shipping_options is not None:
-            data["shipping_options"] = [option.to_dict(True) for option in shipping_options]
-        if error_message is not None:
-            data["error_message"] = error_message
+        data: JSONDict = {
+            "shipping_query_id": shipping_query_id,
+            "ok": ok,
+            "shipping_options": [option.to_dict(True) for option in shipping_options]
```
Bibo-Joshi, line 4266: same

Bibo-Joshi (commented):
Left a few comments :)

BTW small script for double checking for typos:

```python
import re
from pathlib import Path

bot = Path("telegram/_bot.py")
text = bot.read_text(encoding="utf-8")

for match in re.finditer(r"\"([\w_]+)\"\: ([^\},]+)(\}|,)", text):
    if match.group(1) != match.group(2) and not match.group(2).startswith(
        f"self._parse_file_input({match.group(1)}"
    ):
        print(f'"{match.group(1)}": {match.group(2)}')
```

tests/test_bot.py
Bibo-Joshi: Sorry, I don't quite follow. Are we talking about a bug in `_bot.py` or a bug in the test?

harshil21: both actually. point 1. is about the bot method, 2. is this test

Bibo-Joshi: > do we not want to pass False values to Telegram?

I guess we want to and hopefully we had `if value is not None` checks instead in those cases …

so the issue was that `request_data.json_parameters["drop_pending_updates"]` was a string, right? and `request_data.parameters["drop_pending_updates"]` is a bool or `None`? can't we drop the `bool(…)` in line 1914, then?

harshil21: > so the issue was that `request_data.json_parameters["drop_pending_updates"]` was a string, right? 

yes

> and `request_data.parameters["drop_pending_updates"]` is a bool or `None`? can't we drop the `bool(…)` in line 1914, then?

yes

Bibo-Joshi (approved):
Sweet, then I'm happy as well 👍 If there is nothing left from your side, give me a short ping and I'll merge :)

PR #3404, opened 2022-12-02
Unified args and attrs in classes

### LINKED ISSUE #3109: [Documentation]  Add valuable information to attribute docstrings of telegram classes

lemontree210:
closes #3109 
changes for `Dice` and `LoginUrl` reflect those in a closed PR #3316


telegram/_chatinvitelink.py
```diff
@@ -83,6 +84,7 @@ class ChatInviteLink(TelegramObject):
             :tg-const:`telegram.constants.ChatInviteLinkLimit.MIN_MEMBER_LIMIT`-
             :tg-const:`telegram.constants.ChatInviteLinkLimit.MAX_MEMBER_LIMIT`.
         name (:obj:`str`): Optional. Invite link name.
+            0-:tg-const:`telegram.constants.ChatInviteLinkLimit.NAME_LENGTH` characters.
```
lemontree210, line 87: Telegram API contains no explicit limit for `name` arg/attr

telegram/_chatinvitelink.py
```diff
@@ -44,7 +44,8 @@ class ChatInviteLink(TelegramObject):
        * Comparing objects of this class now also takes :attr:`creates_join_request` into account.
 
     Args:
-        invite_link (:obj:`str`): The invite link.
+        invite_link (:obj:`str`): The invite link. If the link was created by another chat
+            administrator, then the second part of the link will be replaced with ``'…'``.
```
lemontree210, line 48: not sure I had to add the 2nd sentence for `invite_link` to args, but it is there in Telegram API

clot27: Seems like this is recently change in telegram's docs, ig we can change this in argument docs as well :eyes:

lemontree210: > we can change this in argument docs as well eyes

Well that's the thing, it is already there in the arguments section. So I thought it might have been deliberately left out of attrs.

lemontree210: > we can change this in argument docs as well eyes

Sorry, my deleted comment was nonsense 😆 . I mean that this second sentence was already there in attrs section, so maybe it was deliberately left out of args section because this note may not make a lot of sense in the args.

telegram/_files/file.py
```diff
@@ -58,18 +58,19 @@ class File(TelegramObject):
...
-        file_id (:obj:`str`): Identifier for this file.
+        file_id (:obj:`str`): Identifier for this file, which can be used to download
+            or reuse the file.
         file_unique_id (:obj:`str`): Unique identifier for this file, which
             is supposed to be the same over time and for different bots.
             Can't be used to download or reuse the file.
-        file_size (:obj:`str`): Optional. File size in bytes.
-        file_path (:obj:`str`): Optional. File path. Use e.g. :meth:`download_to_drive` to get
-            the file.
+        file_size (:obj:`int`): Optional. File size in bytes, if known.
```
lemontree210, line 71: [Telegram API](https://core.telegram.org/bots/api#file) has following addition to `file_size` attr. Since it contains the phrase "some programming languages", we may not need to include it.

> It can be bigger than 2^31 and some programming languages may have difficulty/silent defects in interpreting it. But it has at most 52 significant bits, so a signed 64-bit integer or double-precision float type are safe for storing ths value.

harshil21: yeah I vote not to include it

lemontree210:
I think I covered the classes, instances of which are **returned** from methods. I will continue looking at classes, though.

telegram/_replykeyboardmarkup.py
```diff
@@ -70,13 +70,31 @@ class ReplyKeyboardMarkup(TelegramObject):
             .. versionadded:: 13.7
 
     Attributes:
-        keyboard (List[List[:class:`telegram.KeyboardButton` | :obj:`str`]]): Array of button rows.
-        resize_keyboard (:obj:`bool`): Optional. Requests clients to resize the keyboard.
+        keyboard (List[List[:class:`telegram.KeyboardButton`]]): Array of button rows,
+            each represented by an Array of :class:`telegram.KeyboardButton` objects.
```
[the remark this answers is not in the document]
harshil21: agree

telegram/_inline/inlinequeryresultgif.py
```diff
@@ -71,7 +71,8 @@ class InlineQueryResultGif(InlineQueryResult):
         gif_duration (:obj:`int`): Optional. Duration of the GIF in seconds.
         thumb_url (:obj:`str`): URL of the static (JPEG or GIF) or animated (MPEG4) thumbnail for
             the result.
-        thumb_mime_type (:obj:`str`): Optional. MIME type of the thumbnail.
+        thumb_mime_type (:obj:`str`): Optional. MIME type of the thumbnail, must be one of
+            ``'image/jpeg'``, ``'image/gif'``, or ``'video/mp4'``. Defaults to ``'image/jpeg'``.
```
lemontree210, line 75: should we add this to constants? If so, just as plain strings or are there any object references?

telegram/_inline/inlinequeryresultmpeg4gif.py
```diff
@@ -71,7 +72,8 @@ class InlineQueryResultMpeg4Gif(InlineQueryResult):
         mpeg4_duration (:obj:`int`): Optional. Video duration in seconds.
         thumb_url (:obj:`str`): URL of the static (JPEG or GIF) or animated (MPEG4) thumbnail for
             the result.
-        thumb_mime_type (:obj:`str`): Optional. MIME type of the thumbnail.
+        thumb_mime_type (:obj:`str`): Optional. MIME type of the thumbnail, must be one of
+            ``'image/jpeg'``, ``'image/gif'``, or ``'video/mp4'``. Defaults to ``'image/jpeg'``.
```
lemontree210, line 76: same as [here](https://github.com/python-telegram-bot/python-telegram-bot/pull/3404/files#r1044321066)

telegram/_chatinvitelink.py
Bibo-Joshi: `createChatInviteLink` does though. IMO it's okay to mention it here as well

telegram/_chatinvitelink.py
Bibo-Joshi: Skipping it in args sounds ok to me

telegram/_inline/inlinequeryresultgif.py
Bibo-Joshi: constants mime types could be an option although I'm wondering if there is a list of constants for mime types available somewhere outh there already … I don't quite get the second question.
Can be done in a separate PR though

Bibo-Joshi (approved):
I looked through everything (maybe not every single word :D) and tbh I didn't find anything to complain about 🙆‍♂️ You've been very thorough - also with keeping https://github.com/python-telegram-bot/python-telegram-bot/issues/3109#issuecomment-1333572895 up to date 🤯

telegram/_chatinvitelink.py
lemontree210: Done (removed this phrase from the args again, the way it used to be)

telegram/_inline/inlinequeryresultgif.py
lemontree210: In the second question I meant that these lines just say ``'image/jpeg'``, but in constants.py I find a special reference to a mime type:
```
Bots can send :mimetype:`audio/ogg` files of up to 1MB in size
```

telegram/_inline/inlinequeryresultgif.py
```diff
@@ -77,7 +77,8 @@ class InlineQueryResultGif(InlineQueryResult):
...
```
lemontree210, line 81:
```suggestion
        thumb_mime_type (:obj:`str`): Optional. MIME type of the thumbnail, must be one of
            :mimetype:`image/jpeg`, mimetype:`image/gif`, or mimetype:`video/mp4`. Defaults to mimetype:`image/jpeg`.
```

lemontree210: It seems logical, but in docs it renders as italics, and we want to signal to the user that they should put literals in there. 

Maybe I should use
```
``'image/jpeg'``
```
in the args and 
```
:mimetype:`image/jpeg`
```
in the attrs. 

Or I'm being too meticulous :)

telegram/_inline/inlinequeryresultgif.py
lemontree210: See [this comment](https://github.com/python-telegram-bot/python-telegram-bot/pull/3404#discussion_r1056438950)

telegram/_inline/inlinequeryresultgif.py
Bibo-Joshi: The sphinx-docs list the `:mimetype:` role in the section "Other semantic markup - The following roles don’t do anything special except formatting the text in a different style:" (see [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-mimetype)).

Having the text render as monospaced literals makes sense and that's more important to me than to use the :`mimetype` role just for the sake of it …

lemontree210: OK! No changes here, then.

PR #3412, opened 2022-12-08
Use `Sequence` instead of `Union[List[…], Tuple[…]]`

### LINKED ISSUE #3350: Use `Sequence` instead of `Union[List[…], Tuple[…]]`
Currently we accept only `List` for arguments that Telegram documents as "array". This limitation is rather unnecessary. Passing a tuple will in fact also work out of the box since the `json` module automatically converts tuples to json arrays. 

IMO it's sane to expect every object that is iterable, with finite length and fixed order of elements to work - which brings us to [`Sequence`](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes). Once could argue that we don't strictly need fixed order and instead go for `Collection`, but for most objects, the order of elements does change the behavior of Telegram and using `Sequence` avoids one possible source of confusion for users. The only example I can think of where order doesn't matter are the `(caption_)entities` arguments …

This is also in symmetry with #3249, which accepts `Sequence` for class arguments and converts them to tuple.

The only downside of this approach is that we need one additional check somewhere in `RequestParameter` along the lines

```
if not isinstance(object, (tuple, list)):
    object = list(object)
```

as otherwise `json.dumps` would fail.

Edit: More precisely, the edit would be required here:

https://github.com/python-telegram-bot/python-telegram-bot/blob/5c3b06f015c7ae5e02f23a2a9a2afea37c551dee/telegram/request/_requestparameter.py#L141-L146

clot27:
Working on #3350 with this

closes #3350


Viicos:
As mentioned in the related issue, using `Sequence` can be problematic with `str` (see https://github.com/python/typing/issues/256, I still haven't figured how to handle this case properly, so you might have some ideas!)

clot27:
> As mentioned in the related issue, using `Sequence` can be problematic with `str` (see [python/typing#256](https://github.com/python/typing/issues/256), I still haven't figured how to handle this case properly, so you might have some ideas!)

as mentioned in the [issue](https://github.com/python-telegram-bot/python-telegram-bot/issues/3350#issuecomment-1307236795) itself, using Sequence[something other than str] already excludes str :)

Bibo-Joshi:
> Indeed but I see that there's a few places where `Sequence[str]` is used in this PR, so maybe this would require `isinstance` checks as stated in the issue comment you mentioned :)

The meaning https://github.com/python-telegram-bot/python-telegram-bot/issues/3350#issuecomment-1307236795 is that we have to make sure not to pass strings as list of chars to TG, e.g. for `message.reply_text(text="Hi")`, we must tell TG that the text is `"Hi"`, not `["H", "i"]`.

In the cases of the parameters `allowed_updates`, `custom_emoji_ids` and `options` (the params currently typed as `Sequence[str]`, such a check will have the following effect: If a user passes `allowed_updates="message"`, we will pass the value `"message"` to TG and TG will probably complain because it expects a JSON array and not a string. So while a type checker will not complain, there will be exceptions at runtime, so I doubt that this will lead to hard-to-debug edge cases.

> This typing issue really is a problem in my opinion, as you can see in the `typing` issue mentioned this caused quite some trouble in production environments for some people :/

You are right that providing a precise type annotation for this doesn't seem possible atm. Personally I'd still be okay with using `Sequence[str]` because

* it's a general limitation of Pynthon that we can do little about
* passing a false value to such a parameter will likely be easy to detect (see arguments above)
* so far only 3 parameters are affected …

Viicos:
Yes I agree `Sequence[str]` should still be used as this is a limitation of the language itself. The way I'm doing it currently is using `isinstance` as a runtime check, but even without this check as you said TG will probably complain so this isn't really important.

(Some people suggested the idea of using a `Diff` type annotation (e.g. `Diff[Sequence[str], str]`), imo `Annotated` could also be used in the future when it will be correctly supported).

telegram/_bot.py
```diff
@@ -1828,9 +1855,11 @@ async def send_media_group(
 
         Args:
             chat_id (:obj:`int` | :obj:`str`): |chat_id_channel|
-            media (List[:class:`telegram.InputMediaAudio`, :class:`telegram.InputMediaDocument`, \
-                :class:`telegram.InputMediaPhoto`, :class:`telegram.InputMediaVideo`]): An array
-                describing messages to be sent, must include 2–10 items.
+            media (List[:class:`telegram.InputMediaAudio`, :class:`telegram.InputMediaDocument`,\
```
lemontree210, line 1858:
```suggestion
            media (Sequence[:class:`telegram.InputMediaAudio`, :class:`telegram.InputMediaDocument`,\
```

telegram/_bot.py
```diff
@@ -1889,7 +1923,7 @@ async def send_media_group(
                 item_to_get_caption.caption_entities = parse_sequence_arg(caption_entities)
 
             # copy the list (just the references) to avoid mutating the original list
-            media = media[:]
+            media = list(media)
```
lemontree210, line 1926: If no caption is passed and user passes a tuple as `media`, then it will remain a tuple in `data` dict. Is it OK? I looked at some protected methods (what they do with items in `data` before sending to Telegram) and I'm not sure a tuple will be handled correctly. For instance, `RequestParameter.from_input()` has an `isinstance(value), list` check.

If I'm correct, then maybe `media = list(media)` should be moved up in the code, outside of `if caption` clause.

clot27: As mentioned in the original issue, json module automatically convert tuples to json arrays which will also work. You can see that `RequestParameter.from_input()` has changed its check from `isinstance(value, list)` to  `if not isinstance(value, str) and isinstance(value, Sequence):`

lemontree210: Right, I was looking at the old file in my local branch 🤦🏻

telegram/_bot.py
```diff
@@ -715,9 +715,12 @@ async def send_message(
...
                 entities parsing.
             parse_mode (:obj:`str`): |parse_mode|
-            entities (List[:class:`telegram.MessageEntity`], optional): List of special entities
-                that appear in message text, which can be specified instead of
+            entities (Sequence[:class:`telegram.MessageEntity`], optional): Sequence of special
+                entities that appear in message text, which can be specified instead of
                 :paramref:`parse_mode`.
+
+                .. versionchanged:: 20.0
+                    Accept :obj:`Sequence` instead of :obj:`List`.
```
harshil21, line 723: let's make this an rst substitution (see docs/substituions/global.rst), we could do:

```
.. |sequenceargs| replace:: Accepts any :class:`collections.abc.Sequence` as input instead of just a list.
``` 
then `sequenceclassargs` would become:
```
.. |sequenceclassargs| replace:: |sequenceargs| The input is converted to a tuple.
```

telegram/_bot.py
```diff
@@ -1855,13 +1886,16 @@ async def send_media_group(
...
                 .. versionadded:: 20.0
-            caption_entities (List[:class:`telegram.MessageEntity`], optional):
+            caption_entities (Sequence[:class:`telegram.MessageEntity`], optional):
                 List of special entities for :paramref:`caption`,
                 which can be specified instead of :paramref:`parse_mode`.
                 Defaults to :obj:`None`.
 
                 .. versionadded:: 20.0
 
+                .. versionchanged:: 20.0
```
harshil21, line 1896: no need of versionchanged in 20 if we added it in 20

telegram/_bot.py
```diff
@@ -4293,8 +4346,11 @@ async def answer_shipping_query(  # pylint: disable=invalid-name
             ok (:obj:`bool`): Specify :obj:`True` if delivery to the specified address is possible
                 and :obj:`False` if there are any problems (for example, if delivery to the
                 specified address is not possible).
-            shipping_options (List[:class:`telegram.ShippingOption`]), optional]: Required if
-                :paramref:`ok` is :obj:`True`. A list of available shipping options.
+            shipping_options (Sequence[:class:`telegram.ShippingOption`]), optional]: Required if
```
harshil21, line 4349:
```suggestion
            shipping_options (Sequence[:class:`telegram.ShippingOption`], optional): Required if
```

docs/substitutions/global.rst
```diff
@@ -42,8 +42,10 @@
 
 .. |reply_to_msg_id| replace:: If the message is a reply, ID of the original message.
 
-.. |sequenceclassargs| replace:: Accepts any :class:`collections.abc.Sequence` as input instead of just a list. The input is converted to a tuple.
+.. |sequenceclassargs| replace:: |sequenceargs| The input is converted to a tuple.
```
harshil21, line 45: did you check if the docs substitute this correctly? cause it's sequenceargs is defined after this, so..

telegram/_bot.py
```diff
@@ -3246,7 +3280,11 @@ async def edit_message_caption(
                 0-:tg-const:`telegram.constants.MessageLimit.CAPTION_LENGTH` characters after
                 entities parsing.
             parse_mode (:obj:`str`, optional): |parse_mode|
-            caption_entities (List[:class:`telegram.MessageEntity`], optional): |caption_entities|
+            caption_entities (Sequence[:class:`telegram.MessageEntity`], optional):
+                |caption_entities|
```
Bibo-Joshi, line 3284: Does the text of this substitution need adaption? you edited "List" to "Sequence"e.g. in line 3206, maybe this is needed here as well? If `|caption_enitites|` is used in docstrings of attributes as well, then we'll probably need a second substitution that says "Tuple" …

telegram/request/_requestparameter.py
```diff
@@ -143,7 +143,7 @@ def from_input(cls, key: str, value: object) -> "RequestParameter":
         """Builds an instance of this class for a given key-value pair that represents the raw
         input as passed along from a method of :class:`telegram.Bot`.
         """
-        if isinstance(value, list):
+        if not isinstance(value, (str, bytes)) and isinstance(value, Sequence):
```
Bibo-Joshi, line 146: We'll need new unit tests that ensure that

1. bytes and str are in fact not treated as sequence
2. different types of sequences work, e.g. `list`, `tuple` and also custom subclasses of `collections.abc.Sequence`

Can you add corresponding tests to `test_requestparameter.py`? Maybe `test_from_input_str_and_bytes` and `test_from_input_different_sequences`?

Bibo-Joshi (changes requested):
Nice job! :) Thanks for the PR! Harshil in fact already covered two comments that I had in the pipeline xD

docs/substitutions/global.rst
clot27: Yes, it works

Bibo-Joshi:
Thank you very much for your work on this nice PR @clot27 🙂

Issue #3421, opened 2022-12-12
[DOCS] add references to Code Resources Wiki section?


lemontree210:
### What kind of feature are you missing? Where do you notice a shortcoming of PTB?

in #3398 I added `seealso` references to Wiki pages to close #3110. I think I could add more references - from docs to section with code resources. 

For example, from `ReplyKeyboardMarkup` to "Keyboard Menus" section of [Code Snippets](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Code-snippets).

### Describe the solution you'd like

If you are in favor of this, what do you think should be the link text? In other Wiki references the link text is the name of Wiki page, but here we might want to choose something different. The target of the link should in any case be not the page but a heading within it.

- "Code Snippets"
- "Code Snippets in Wiki"
- "Code Snippets: Keyboard Menus"
- "Keyboard Menus"

### Additional context

I think that as a beginning bot developer, I would like to have links to these snippets from docs.

Bibo-Joshi:
@lemontree210 Tbh I&#x27;m not very convinced by this. The code snipped page is far from being well organized (meaning that links can easily become outdated) and most of the things on the page are in my impression either direct translations of the docs into code or so specific that don&#x27;t fit into the docs.

The only thing in there that could IMO be worth linking to from the docs is the &quot;working with files an media&quot; section and for that I&#x27;d rather convert that section to a standalone wiki page.

lemontree210:
Sounds good. Yes, I thought too that the code snippets may not be that organized. 
Do you see this standalone wiki page in "Code resources" or in "Notable features"?

Bibo-Joshi:
@lemontree210 I&#x27;d say &quot;concepts &amp; import elements&quot;. The section should also be updated to reflect that you can now always pass file paths as arguments

PR #3306, opened 2022-10-23
Add references to wiki

lemontree210:
addresses #3110.

Issue #3110, opened 2022-06-21
[Documentation] Add more cross-references


Bibo-Joshi:
It would be nice to add more cross-references to the documentation. This should help users to better understand "where is this used?"/"what is this relevant for?".

For this, we can use the sphinx directive `.. seealso::` like done e.g. [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/67e7468366098f79cd2aa4f88b40e6e47d93ae5a/telegram/ext/_applicationbuilder.py#L332).

Notable things that can be cross-referenced better include:

* point to shortcuts from the bot methods. e.g. point to `Message.reply_text` from `Bot.send_message`, point to `CallbackQuery.answer` from `Bot.answer_callback_query` and so on.
* point to examples & wiki: All classes/arguments that are show cased in an example or explained in a wiki page should point to those. e.g. `JobQueue` should point to `timerbot` and the wiki page on `JobQueue`. `Bot.arbitrary_callback_data` should point to the wiki page on arbitrary callback data and to the `arbitrarycallbackbot` example.
* for classes in tg.ext (and for `tg.Bot`) point to attributes of that type: e.g. `JobQueue` should point to `{Application, CallbackContext}.job_queue`, `Bot` should point to `{Application, Updater, CallbackContext}.bot`, `CallbackDataCache` should point to `ExtBot.callback_data_cache`

Contributions in this direction are very welcome! They also don't have to cover everything classes at once - instead this can be done step by step. If you want to cover (parts of) this, please leave a comment below, so that we can avoid duplicate work.

PRs for this should be made against the `doc-fixes` branch. Please also check out the [contribution guide](https://docs.python-telegram-bot.org/en/v20.0a1/contributing.html), especially the part on how to [build the documentation locally](https://docs.python-telegram-bot.org/en/v20.0a1/contributing.html#documenting).

aditya-yadav-27:
Can I work on this?

Bibo-Joshi:
Yes, please :) if you can leave a short comment on which places you want to cover, that would be great (you can also edit that comment in case you extend your work)

aditya-yadav-27:
All `bot.send_*` methods, all updating message methods e.g. `answer_inline_query`, all chat moderation and information methods e.g. `ban_chat_member`. Point to examples as you have mentioned. Point classes to attributes as you have mentioned

lemontree210:
[...]

PR #3423, opened 2022-12-14
Add links to "Working with Files and Media" wiki page from docstrings

lemontree210:
after #3421 (that was closed) and in the spirit of #3306 addressing #3110.

After creating a [wiki page on Files and Media](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Working-with-Files-and-Media) I'm adding links to this page to docstrings of some classes/methods.

(Sorry, I forgot the branch name should be a noun)

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -694,8 +694,9 @@ class Bot(TelegramObject, AbstractAsyncContextManager):
     ) -> Message:
         """Use this method to send text messages.
 
-        .. seealso:: :meth:`telegram.Message.reply_text`, :meth:`telegram.Chat.send_message`,
-            :meth:`telegram.User.send_message`
+        .. seealso:: :meth:`telegram.Chat.send_message`, :meth:`telegram.User.send_message`,
+            :meth:`telegram.Message.reply_text`, :meth:`telegram.Message.reply_html`,
+            :meth:`telegram.Message.reply_markdown`, :meth:`telegram.Message.reply_markdown_v2`
 
         Args:
             chat_id (:obj:`int` | :obj:`str`): |chat_id_channel|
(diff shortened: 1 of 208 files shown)


telegram/_bot.py
```diff
@@ -906,7 +906,9 @@ async def send_photo(
         """Use this method to send photos.
 
         .. seealso:: :meth:`telegram.Message.reply_photo`, :meth:`telegram.Chat.send_photo`,
-            :meth:`telegram.User.send_photo`
+            :meth:`telegram.User.send_photo`,
+            `Working with Files and Media <https://github.com/\
+            python-telegram-bot/python-telegram-bot/wiki/Working-with-Files-and-Media>`_
```
Bibo-Joshi, line 911:
```suggestion
            |files_and_media_wiki|
```

Maybe we can use a subsitution for this?

lemontree210: Absolutely, but maybe in a separate PR then? I mean, there are lots of similar links that could be in substitutions.

Bibo-Joshi: I'd prefer to use a substitution here right away instead of changing it later. There are many links/docstrings/notes/… that can benefit of substitutions, so IMO it's more sane to convert those bit by bit rather than trying to change everything at once …

lemontree210: OK! A regular `|substitution|` won't work with hyperlinks, so I tried using `extlinks` in commit 8e5ae0408d914fbb7709f9550386e40fdbf4cd95

I would also love to find a way of replacing hyphens with whitespaces automatically. Then it would say
```
:wiki:`Working-with-Files-and-Media`
```

instead of the custom link text:

```
:wiki:`Working with Files and Media <Working-with-Files-and-Media>`
```

But it doesn't seem possible. I tried putting something like `"%s".replace(...)` [here](https://github.com/python-telegram-bot/python-telegram-bot/pull/3423/commits/8e5ae0408d914fbb7709f9550386e40fdbf4cd95#diff-008dcb3426febd767787b1521f1fe33086313b927ea37eaab86df5fa88a51698R54), but it doesn't do the trick.

Bibo-Joshi (approved):
LGTM! If there's nothing left on this PR from your side, feel free to merge :)

PR #3426, opened 2022-12-15
Refactor and overhaul the test suite

### LINKED ISSUE #3407: [REFACTOR] Split test classes into two: `TestXXXNoRequest` & `TestXXXRequest`
Currently each of our test files has about one TestClass, which tests everything about the class together. This is not optimal because:

1. Some tests in the class make actual requests to Telegram, thus slowing down development if you've made substantial changes to the class, but don't actually want to test the request part yet.
2. It makes the tests slightly more readable, bearable and just a bit more organized.

### Implementation

For example, `test_sticker.py` should now look like:

``` python
class StickerSpace:  # namespace class
     emoji = "💪🏽 "
     ...

@pytest.mark.no_req   # for faster development cycle
class TestStickerNoRequest:
    def test_de_json(...):
        ...
    def test_a_mock_request(...):
        ...

@pytest.mark.req  # don't think we need this, but let me know if you'd like this
@pytest.mark.flaky(3, 1)
class TestStickerRequest:
    def test_custom_emoji(...):
        await bot.get_sticker_set(...)

```

Along with this change, I'll also try to simplify/shorten some tests, if possible.

harshil21:
Closes #3407. I did not include the `req` marker, but if you'd like that can be easily added as well.

Also some new "features":

- `Dict(Ext)Bot` now overrides `get_me` to return a cached `User` object, which is obtained from the `bot_info` fixture. This saves lots of requests to the real `get_me` every time we initialize a bot, app, or updater.
- A test marked as flaky won't re-run if the test failed with a flood, or was marked to xfail anyway.
- Some tests which make a lot of requests are now run concurrently instead.
- The testing without dependencies part in the CI is now done in one go instead of starting separate runs

Many tests were optimized as well, which should decrease the flakyness in the CI.

#### How do I review this?!

It is best to review by commit, the changes are specific and I rebased multiple times to make sure changes fit in a certain category.


harshil21:
These tests don't fail for me locally, will have a look later.

Bibo-Joshi (changes requested):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/fc5aea4307b6d48ea922cc834dc4c7621b2f0730#.

I didn't go through all the files, but I assume that nothing change besides splitting the classes, reordering, dropping `flaky` and switching the fixture scope from `class` to `module`, as the commit description also says

tests/conftest.py
```diff
@@ -148,23 +146,51 @@ def __init__(self, *args, **kwargs):
         # Makes it easier to work with the bot in tests
         self._unfreeze()
 
+    # Here we override get_me for caching because we don't want to call the API repeatedly in tests
+    async def get_me(self, *args, **kwargs):
+        if self._bot_user is None:
+            self._bot_user = make_mock_user(self.token)
+        return self._bot_user
+
```
Bibo-Joshi, line 158: can this be moved to a common base class or at least a separate function that both classes call?

tests/conftest.py
```diff
@@ -148,23 +146,51 @@ def __init__(self, *args, **kwargs):
...
 
 class DictApplication(Application):
     pass
 
 
+def make_mock_user(token: str) -> User:
+    """Used to return a mock user in bot.get_me(). This saves API calls on every init."""
+    bot_info = get_bot()
+    user_id = int(token.split(":")[0] or bot_info["token"].split(":")[0])
```
Bibo-Joshi, line 180: what is the `or` needed for? in which cases do we have `token.split(":")[0] is None`?

tests/conftest.py
```diff
@@ -148,23 +146,51 @@ def __init__(self, *args, **kwargs):
...
+    first_name = bot_info.get("bot_name") or bot_info.get("name")
+    username = (bot_info.get("bot_username") or bot_info.get("username")).strip("@")
```
Bibo-Joshi, line 182: can be simplified by now

tests/test_bot.py
```diff
@@ -95,7 +95,7 @@ def to_camel_case(snake_str):
     return components[0] + "".join(x.title() for x in components[1:])
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="class")
```
Bibo-Joshi: Can give an explanation on which scopes you setteld for in the end and why? :D

tests/test_bot.py
```diff
@@ -295,10 +295,10 @@ async def shutdown():
 
         assert self.test_flag == "stop"
 
-    async def test_log_decorator(self, bot, caplog):
+    async def test_log_decorator(self, bot: DictExtBot, caplog):
         # Second argument makes sure that we ignore logs from e.g. httpx
         with caplog.at_level(logging.DEBUG, logger="telegram"):
-            await bot.get_me()
+            await super(bot.__class__, bot).get_me()  # call original get_me instead of overriden
```
Bibo-Joshi: This might be fragile if the inheritance order of `DictExtBot` changes. Please instead just instaniate an instance of `ExtBot` with `bot.token`.

tests/test_bot.py
```diff
@@ -365,12 +365,17 @@ async def test_get_me_and_properties_not_initialized(self, bot: Bot, attribute):
...
 
-    async def test_get_me_and_properties(self, bot: Bot):
-        get_me_bot = await bot.get_me()
+    async def test_get_me_and_properties(self, bot):
+        mocked = "mocked"
+        with bot._bot_user._unfrozen():
+            bot._bot_user.username = mocked  # set the attribute to a mocked value
+        assert bot.username == mocked  # To test we are using the mocked bot
+        # call the parent class get_me method instead of the overriden get_me method
+        get_me_bot = await super(bot.__class__, bot).get_me()
```
Bibo-Joshi, line 350: see above. that way we also don't have a need for using this mocking setup, which tbh seems rather convoluted to me …

Bibo-Joshi (changes requested):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/77ecda18b6e21caf883234ca6bbfbf2196df0e0d#

tests/test_bot.py
```diff
@@ -1496,6 +1443,66 @@ def localize(dt, tzinfo):
             # fail if this branch is ever reached.
             sys.exit(1)
 
+    test_flag = None
```
Bibo-Joshi: please move up directly beneath the docstring. having this below a method is confusing …

tests/test_bot.py
```diff
@@ -197,94 +197,41 @@ def __init__(
...
-    async def test_multiple_init_cycles(self, bot):
-        # nothing really to assert - this should just not fail
-        test_bot = Bot(bot.token)
-        async with test_bot:
-            await test_bot.get_me()
-        async with test_bot:
-            await test_bot.get_me()
-
     async def test_context_manager(self, monkeypatch, bot):
+        test_flag = []
```
Bibo-Joshi, line 266: If `TestBotNoReq` would inherit from `Space`, you could make `test_flag` a classvar of `Space` and use it in both classes and also move the `reset` function there. Another reason for subclassing.

Bibo-Joshi (changes requested):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/f1207dc26a98bedd73e063e833594dc0b943a3db

tests/test_bot.py
```diff
@@ -2501,14 +2490,15 @@ async def test_create_chat_invite_link_basics(
         )
         assert revoked_link.is_revoked
 
+    @pytest.mark.skipif(not TEST_WITH_OPT_DEPS, reason="requires pytz")
```
Bibo-Joshi, line 2602: mh, the functionality in this test doesn't actually require `pytz`. But that doesn't really matter IMO. Maybe the `reason` can be adapted for that a bit though, something like `reason="This tests implementation currently requires pytz"`

Bibo-Joshi (commented):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/4fe99558baa9afb16fa279c8957a14331c183ae9

tests/test_chatphoto.py
```diff
@@ -185,6 +175,16 @@ async def test_get_and_download(self, bot, chat_photo):
 
         assert jpg_file.is_file()
 
+    async def test_send_all_args(
```
Bibo-Joshi: if tests need to be in a specific order her to pass, please leave comments in the respective tests explaining why they need to be before/after certain other tests

Bibo-Joshi (changes requested):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/8725e4b0602bb35a6375e863014b7f5f231118ed

tests/test_application.py
```diff
@@ -53,7 +53,18 @@
...
+    call_after,
+    make_bot,
+    make_message_update,
+    send_webhook_message,
+)
+
+
+@pytest.fixture(scope="function")
+def test_bot(bot_info):  # was needed since we get HTTPXRequest not initialized after app tests
+    return make_bot(bot_info)
```
Bibo-Joshi, line 67: 1. comment -> docstring
2. I don't understand the comment and don't recall what the problem was 😄 Can you elaborate?
3. `test_bot` is not a very speaking name. I'd like to improve that, but that depends on the explanation for 2 …

tests/test_application.py
```diff
@@ -433,9 +444,10 @@ async def callback(u, c):
             await asyncio.sleep(0.05)
             assert app.update_queue.empty()
             assert self.received == 1
-
-            await app.updater.start_polling()
-            await app.stop()
+            try:  # just in case start_polling times out
```
Bibo-Joshi, line 447: does that tend to happen? If so, please state that in the comment. what happens with the tests when `start_polling` times out? does it pass or fail?

tests/test_application.py
```diff
@@ -1942,11 +1958,10 @@ def signal_handler_test(*args, **kwargs):
         loop = asyncio.get_event_loop()
         monkeypatch.setattr(loop, "add_signal_handler", signal_handler_test)
 
-        async def abort_app():
-            await asyncio.sleep(2)
+        def abort_app():
             raise SystemExit
 
-        loop.create_task(abort_app())
+        loop.call_later(0.5, abort_app)
```
Bibo-Joshi, line 1964: TIL. do you have an explanation on why this works better than `create_task`?

Bibo-Joshi (commented):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/0920f954b7678fb34f68d23c1550613aab735d8a#

tests/test_application.py
```diff
@@ -481,7 +481,7 @@ async def test_one_context_per_update(self, app):
         async def one(update, context):
             self.received = context
 
-        def two(update, context):
+        async def two(update, context):
```
Bibo-Joshi, line 484: Why didn't this test fail so far? Does that mean that we have a bug somewhere?

tests/test_application.py
```diff
@@ -874,7 +874,7 @@ async def error_handler(_, context):
             assert self.received == (CustomContext, float, complex, int)
 
     async def test_custom_context_handler_callback(self, test_bot):
-        def callback(_, context):
+        async def callback(_, context):
```
Bibo-Joshi, line 877: same

tests/test_jobqueue.py
```diff
@@ -566,7 +566,7 @@ async def test_custom_context(self, bot, job_queue):
         )
         job_queue.set_application(application)
 
-        def callback(context):
+        async def callback(context):
```
Bibo-Joshi, line 569: same

Bibo-Joshi (commented):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/e2cc5367c1b8bc1859dc42376390c52b423c7807

Bibo-Joshi (commented):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/40ad364b0c1a402b268a657d5de3b8281b69dbab

I don't really get what this commit is about. What's the benefit of using `DictBot` in these places?

tests/test_bot.py
```diff
@@ -1433,14 +1433,10 @@ def test_camel_case_aliases(self, bot_class, bot_method_name, bot_method):
         assert camel_case_function is bot_method, f"{camel_case_name} is not {bot_method}"
 
     @bot_methods()
-    def test_coroutine_functions(self, bot_class, bot_method_name, bot_method):
+    def test_coroutine_functions(self, bot_class, bot_method_name, bot_method, monkeypatch):
```
Bibo-Joshi: Is monkeypatch needed?

tests/test_bot.py
```diff
@@ -1433,14 +1433,10 @@ def test_camel_case_aliases(self, bot_class, bot_method_name, bot_method):
...
         """Check that all bot methods are defined as async def  ..."""
-        # unfortunately `inspect.iscoroutinefunction` doesn't do the trick directly,
-        # as the @_log decorator interferes
-        source = "".join(inspect.getsourcelines(bot_method)[0])
-        assert (
-            f"async def {bot_method_name}" in source
-        ), f"{bot_method_name} should be a coroutine function"
+        meth = getattr(bot_method, "__wrapped__", bot_method)
```
Bibo-Joshi: please leave a CSI comment on why we need the `__wrapped__`.

Bibo-Joshi (changes requested):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/dea47cf2e77331160bff68097e0a152f6a5a548e

tests/test_chatphoto.py
```diff
@@ -157,23 +158,18 @@ async def test_get_and_download(self, bot, chat_photo):
...
+        tasks = {bot.get_file(chat_photo.small_file_id), bot.get_file(chat_photo.big_file_id)}
 
-        assert new_file.file_unique_id == chat_photo.small_file_unique_id
-        assert new_file.file_path.startswith("https://")
+        for task in asyncio.as_completed(tasks):
+            file = await task
+            assert file.file_unique_id in (
+                chat_photo.small_file_unique_id,
+                chat_photo.big_file_unique_id,
+            )
```
Bibo-Joshi, line 168: I would prefer if we could make the precise assertion

tests/test_constants.py
```diff
@@ -131,23 +130,23 @@ def test_bot_api_version_info(self):
...
-                chat_id=chat_id, text="a" * (constants.MessageLimit.MAX_TEXT_LENGTH + 1)
-            )
+        tasks = asyncio.gather(
+            bot.send_message(chat_id, text="a" * constants.MessageLimit.MAX_TEXT_LENGTH),
+            bot.send_message(chat_id, text="a" * (constants.MessageLimit.MAX_TEXT_LENGTH + 1)),
+            return_exceptions=True,
+        )
+        results = await tasks
+        assert len(results) == 2
+        assert isinstance(results[1], BadRequest) and "Message is too long" in str(results[1])
```
Bibo-Joshi, line 140: please use the `good_msg, bad_msg` thingy as below in `max_caption_length`

tests/test_invoice.py
```diff
@@ -170,57 +172,61 @@ def test_equality(self):
...
+        )
+        create_inv_link_task = asyncio.create_task(
+            bot.create_invoice_link(
+                title=Space.title,
+                description=Space.description,
+                payload=Space.payload,
+                provider_token=provider_token,
+                currency=Space.currency,
+                prices=Space.prices,
+            )
```
Bibo-Joshi, line 206: please revert the changes for this tests, the tasks are not gathered and this decreases readeability IMO

tests/test_invoice.py
```diff
@@ -170,57 +172,61 @@ def test_equality(self):
...
+                    chat_id,
+                    Space.title,
+                    Space.description,
+                    Space.payload,
+                    provider_token,
+                    Space.currency,
+                    Space.prices,
+                    **i,
+                )
+                for i in ({}, {"protect_content": False})
```
Bibo-Joshi, line 237: `i` -> `kwargs`

tests/test_sticker.py
```diff
@@ -804,12 +810,13 @@ async def test_bot_methods_3_png(self, bot, chat_id, sticker_set_thumb_file):
...
         animated_test = f"animated_test_by_{bot.username}"
-        assert await bot.set_sticker_set_thumb(animated_test, chat_id, animated_sticker_file)
         file_id = animated_sticker_set.stickers[-1].file_id
-        # also test with file input and mask
-        assert await bot.set_sticker_set_thumb(animated_test, chat_id, file_id)
+        tasks = asyncio.gather(
+            bot.set_sticker_set_thumb(animated_test, chat_id, animated_sticker_file),
+            bot.set_sticker_set_thumb(animated_test, chat_id, file_id),
+        )
+        assert all(await tasks)
```
Bibo-Joshi, line 846: both requests make changes to the same sticker set. awaiting them simultaniously could lead to strange situations. Please don't gather them

Bibo-Joshi (changes requested):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/857949aa829098271d5e9fa34edd5f45c3c97de0#

tests/test_invoice.py
```diff
@@ -84,33 +84,45 @@ def test_to_dict(self, invoice):
...
-                and kwargs["photo_height"] == "photo_height"
-                and kwargs["need_name"] == "need_name"
-                and kwargs["need_phone_number"] == "need_phone_number"
-                and kwargs["need_email"] == "need_email"
-                and kwargs["need_shipping_address"] == "need_shipping_address"
-                and kwargs["send_phone_number_to_provider"] == "send_phone_number_to_provider"
-                and kwargs["send_email_to_provider"] == "send_email_to_provider"
-                and kwargs["is_flexible"] == "is_flexible"
-            )
+            return all([kwargs[i] == i for i in kwargs])
```
Bibo-Joshi, line 92: `i` -> `key`

Bibo-Joshi (changes requested):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/b330c9732e65e298aa83c7f3bcfa770662979273

tests/test_user.py
```diff
@@ -100,43 +100,6 @@ def test_de_json(self, json_dict, bot):
...
-        assert user.first_name == Space.first_name
-        assert user.last_name is None
-        assert user.username is None
-        assert user.language_code == Space.language_code
-        assert user.can_join_groups == Space.can_join_groups
-        assert user.can_read_all_group_messages == Space.can_read_all_group_messages
-        assert user.supports_inline_queries == Space.supports_inline_queries
-        assert user.is_premium == Space.is_premium
-        assert user.added_to_attachment_menu == Space.added_to_attachment_menu
-
```
Bibo-Joshi: Why don't these tests serve any purpose? Maybe git blame has some insights into why we have them?

Bibo-Joshi (commented):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/cc1e45e28115cc8560620a0ec5344477f2a44449#

requirements-dev.txt
```diff
@@ -1,9 +1,10 @@
-pre-commit
+pre-commit  # needed for pre-commit hooks in the git commit command
 
+# For the test suite
 pytest==7.2.0
-pytest-asyncio==0.20.3
+pytest-asyncio==0.20.3  # needed for async tests
```
Bibo-Joshi, line 5: "needed because `pytest` doesn't come with native support for coroutines as tests"

tests/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+    $ pytest
+
+This will run all the tests, including the ones which make a request to the Telegram servers, which
+may take a long time (about > 13 mins). To run only the tests that don't require a connection, you
+can run the following command:
+
+    $ pytest -m no_req
+
+To further speed up the tests, you can run them in parallel using the `-n` flag:
```
Bibo-Joshi, line 19: please explain that this uses multiple CPU cores. May be a no-go for people on potato machines

tests/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+    $ pytest --lf
+
+
+### Writing tests
+
+PTB has a separate test file for every file in the `telegram.*` namespace. Further, the tests for
+the `telegram` module are split into two classes, based on whether the test methods in them make a
+request or not. When writing tests, make sure to split them into these two classes, and make sure
+to name the test class as: `TestXXXNoReq` for tests that don't make a request, and `TestXXXReq` for
```
Bibo-Joshi, line 34: naming needs updating

tests/README.md
```diff
@@ -0,0 +1,51 @@
...
+are a few conventions that you should follow:
+
+- Each new test class needs a `test_slot_behaviour`, `test_to_dict`, `test_de_json` and
+  `test_equality` (in most cases).
+
+- Make use of pytest's fixtures and parametrize wherever possible. Having knowledge of pytest's
+  tooling can help you as well. You can look at the existing tests for examples.
+
+We also have another marker, `@pytest.mark.dev`, which is used to mark tests that you want to run selectively.
+Use as follows:
```
Bibo-Joshi, line 47: also mention that the marker needs to be added to the tests that you want to run ;)

tests/README.md
```diff
@@ -0,0 +1,51 @@
+## Testing in PTB
+
+PTB uses [pytest](https://docs.pytest.org/en/latest/) for testing. To run the tests, you need to
```
Bibo-Joshi, line 3: /stable ?

Bibo-Joshi (commented):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/6ea3515eb13bf4d01e63cd5a094bee7463b73177

A few more suggestions:

* Link to the test-readme from the contrib-guide. Elaborations on pytest may be removed there in favor of the new readme
* the readme could use a note on the CI bots, i.e.
  * CI bots are provided by PTB
  * there are different ones for testing locally and in the CI
  * maybe point to the wiki page on the CI bots
* notes on the meta-stucture of the test can also be helpful, e.g.
  * fixtures and pytest-stuff is set up in `conftest.py`
  * other auxil-stuff goes into `auxil`.
  I can also add that in my PR though …

tests/test_audio.py
```diff
@@ -227,11 +227,12 @@ async def test_send_all_args(self, bot, chat_id, audio_file, thumb_file):
         assert message.audio.thumb.height == Space.thumb_height
         assert message.has_protected_content
 
-    async def test_get_and_download(self, bot, audio):
+    async def test_get_and_download(self, bot, chat_id, audio):
         path = Path("telegram.mp3")
         if path.is_file():
             path.unlink()
 
+        audio = (await bot.send_audio(chat_id, audio.file_id, read_timeout=50)).audio
```
Bibo-Joshi, line 235: why not add the read_timeout directly in the fixture?

tests/test_document.py
```diff
@@ -196,11 +196,12 @@ async def test_error_send_empty_file_id(self, bot, chat_id):
         with pytest.raises(TelegramError):
             await bot.send_document(chat_id=chat_id, document="")
 
-    async def test_get_and_download(self, bot, document):
+    async def test_get_and_download(self, bot, document, chat_id):
         path = Path("telegram.png")
         if path.is_file():
             path.unlink()
 
+        document = (await bot.send_document(chat_id, document, read_timeout=50)).document
```
Bibo-Joshi, line 204: here as well and also below

Bibo-Joshi (commented):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/4b55ee567ae31bef0f7a0a200daf50351829c8bf

tests/bots.py
```diff
@@ -47,6 +47,8 @@
 
 FALLBACKS = json.loads(base64.b64decode(FALLBACKS).decode("utf-8"))  # type: list[dict[str, str]]
 
+chosen_bot = {}
```
Bibo-Joshi, line 50: please upper-case this

Bibo-Joshi (commented):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/4ea871eba81708e41bcd04d5caf760044f903308

Alternative idea: Write a small class `BotInfoProvider` (or sim.) instead of using a module-level global variable

tests/test_application.py
```diff
@@ -1944,6 +1944,7 @@ async def raise_method(*args, **kwargs):
 
         assert len(recwarn) == 0
 
+    @pytest.mark.flaky(3, 1)  # We could run into a flood error from run_webhook
```
Bibo-Joshi, line 1947: shouldn't flood errors xfail thanks to our custom test-request-class?

Bibo-Joshi (commented):
Review of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/4f69b52ca4b7b5519fdde2c79cf8575de15b4be6

setup.cfg
```diff
@@ -36,7 +36,10 @@ filterwarnings =
 ;    Unfortunately due to https://github.com/pytest-dev/pytest/issues/8343 we can't have this here
 ;    and instead do a trick directly in tests/conftest.py
 ;    ignore::telegram.utils.deprecate.TelegramDeprecationWarning
-markers = dev: If you want to test a specific test, use this
+markers =
+    dev: If you want to test a specific test, use this
+    no_req
+    req
```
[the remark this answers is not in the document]
harshil21: I was torn on whether to add it or not, but now I've decided to add it.

tests/conftest.py
```diff
@@ -148,6 +164,7 @@ class DictApplication(Application):
 async def bot(bot_info):
     """Makes an ExtBot instance with the given bot_info"""
     async with make_bot(bot_info) as _bot:
+        print("making new bot")
```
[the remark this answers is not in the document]
harshil21: was done in next commit. Didn't mean to commit this actually, mistakenly ran `git commit -a` and realized too late.

tests/conftest.py
```diff
@@ -248,7 +280,7 @@ def thumb_file():
     f.close()
 
 
-@pytest.fixture(scope="class")
+@pytest.fixture(scope="module")
```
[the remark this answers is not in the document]
harshil21: so far it's only been used in the fixtures of inputmedia, whose scope is module as well.

tests/test_animation.py
```diff
@@ -49,7 +49,7 @@ async def animation(bot, chat_id):
         ).animation
 
 
-class TestAnimation:
+class Space:
```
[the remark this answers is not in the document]
harshil21: what about `AnimationSpace` like I proposed in the issue?

tests/test_animation.py
```diff
@@ -76,19 +78,131 @@ def test_creation(self, animation):
         assert animation.file_unique_id != ""
 
     def test_expected_values(self, animation):
-        assert animation.mime_type == self.mime_type
-        assert animation.file_name.startswith("game.gif") == self.file_name.startswith("game.gif")
+        assert animation.mime_type == Space.mime_type
```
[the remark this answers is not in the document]
harshil21: that's a great idea, don't know why I didn't think of that.

tests/conftest.py
harshil21: i'll go with a separate function, the base class may cause some issues in tests

tests/test_bot.py
harshil21: settled for `function` in the end since bot tests modify the message returned from this, making it best to return a new message for each test.

tests/test_bot.py
harshil21: was moved to a local variable (inside the test method) instead.

tests/test_bot.py
harshil21: ah okay. The `localize` static method did say it needs pytz to run, and that static method was only referenced in that test method sooo

tests/test_animation.py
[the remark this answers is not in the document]
Bibo-Joshi: "Space" alone is not sufficient to get the meaning across. With "Base" it's clear that it's a base class and my favorite in the context of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426#discussion_r1051648571. "NameSpace" somewhat gets across the meaning although I like it less today than I did yesterday because we're not really defining any (test) functionality in that "namespace". "Constants" could also work.

tests/test_chatphoto.py
harshil21: not a problem after https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/4b55ee567ae31bef0f7a0a200daf50351829c8bf

tests/conftest.py
Bibo-Joshi: okay, I was confused, b/c the argument isn't even optional. But if `get_bot` already contains the token, then we can drop the parameter alltogether, right? In that case I would like to move this function into the `bots.py` file (since it only relies on the info that we load from the encoded data there) and rename it to maybe `get_bot_user`

tests/test_application.py
harshil21: 2. For whatever reason, if the session `bot` is used (from conftest), these app tests would work fine, but the test files executed after `test_application.py` would fail saying that `HTTPXRequest` was not initialized. So I just resorted to making a new bot. 

P.S: I think I figured out why that happened, I think it's cause of:
```
bot = ApplicationBuilder().bot(bot).build()
async with app:
```
done in a few tests.

3. Yes, don't know why I named it starting with `test_` 😞

tests/test_animation.py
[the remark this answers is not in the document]
harshil21: hmm, I like "Constants" more compared to others. Just "Constants" or something like "AnimationConstants"?

tests/conftest.py
harshil21: I remember at least one test (test_eq in test_bot.py) where we did something like: `make_bot(FALLBACKS[0]["token"])` so in that case, we'll need the parameter so we can pass the token explicitly instead of getting one from `get_bot()`. I'll rename the function regardless

tests/test_application.py
harshil21: so far it happened only once, iirc. It does fail the test yes

tests/test_application.py
harshil21: No, app.process_error was eating up that error, by logging it instead. I only discovered this accidentally while running pytest with `-rP` which also shows log outputs. Test didn't fail since `await two()` would end up calling the function first and thus the results were expected

tests/test_bot.py
harshil21: must've added that accidentally. Was removed in later commits

tests/test_bot.py
harshil21: added in https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/b2e292ea6bd24e57d7cdd92dc169843868871411

tests/test_animation.py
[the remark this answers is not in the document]
Bibo-Joshi: a `Test<ClassName>`, `_Test<ClassName>` or at very last `<ClassName>` prefix is a must for me. since you agree on subclassing, I definitely vote for `Base` though. That way we also don't get weird naming if at some point we e.g. start adding auxiliary functions to that class.

tests/conftest.py
Bibo-Joshi: `make_bot` is a different function than `make_mock_user` though, isn't it? The token isn't even used to get the user info of the bot with that token …

tests/test_bot.py
Bibo-Joshi: I don't know what your reaction is supposed to tell me :D My point is: with the subclassing approach, the `test_flag` thing can be used just as it was before your PR ;)

tests/test_application.py
Bibo-Joshi: ah okay, that makes sense, the problem is that `bot.shutdown` was called. How about `disposable_bot` or `one_time_bot` or `function_scope_bot` then to signal that we get a new bot every time we use that fixture?

tests/test_application.py
Bibo-Joshi: Mh, one time doesn't sound much to me … and what difference does it make if we stop the app before failing the test? 🤔

tests/test_application.py
Bibo-Joshi: Ah, 1. is the important thing than probably. Nice improvement!

tests/test_application.py
Bibo-Joshi: ah okay. IIRC we configured pytest to convert warnings into exceptions. Maybe we could add a default error handler to the `app` fixture that issues warnings instead of logging? that might help at least a bit with these kinds of things.

tests/test_bot.py
harshil21: yes, I was just wondering if the reset thing would work if it's defined in the Base class, will have to check

tests/test_application.py
harshil21: sure

tests/test_application.py
harshil21: I will revert this and check if I can reproduce again

tests/test_sticker.py
harshil21: in this test we only we checked if the sticker thumb was set, by checking the boolean result of it. What strange situations do you think would happen?

tests/test_user.py
harshil21: no idea. It was in leandro's first commit for this test all the way back in 2015. Safe to say everything has improved since then.

tests/test_audio.py
harshil21: it is there in the fixture already

tests/bots.py
harshil21: I'll make add it in a class so I guess I won't have to upper case

tests/test_application.py
harshil21: Yes, they do xfail, but loop.call_later will trigger the raise SystemExit, which then fails the test so

tests/test_application.py
Bibo-Joshi: having multiple error handlers is supported by default though unless one of then raises `ApplicationHandlerStop` 🤔 If introducing something like this makes more problems than it adds benefits, then we should ofc not do that, but maybe it's worth a shot :)

tests/test_sticker.py
Bibo-Joshi: Req_1 start
Req_2 start
Req_1 finish
Req_2 finish

What happens in this scenario? Does TG care about the starting of the requests or does it just treat that is "req_1 came in and then req_2 came in"? The chance of something going wrong here is probably very low but so is the advantage of gathering  IMO …

tests/test_user.py
Bibo-Joshi: okay, let's remove then

tests/test_audio.py
Bibo-Joshi: then I don't quite unnderstand what these added lines do …

tests/test_application.py
Bibo-Joshi: mh, okay. Then please update the comment to highlight this connection between flood errors and call_later

tests/test_sticker.py
harshil21: > What happens in this scenario

nothing... TG just sees whatever request came first to them, and would set that as the thumb. The test has never failed for me for that reason... I could still revert it if you'd like

Bibo-Joshi: My gut feeling is not to use `gather` here. But I won't veto.

tests/test_audio.py
harshil21: the reason I added them is essentially the commit message of https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/4b55ee567ae31bef0f7a0a200daf50351829c8bf.

It's also due to this behaviour from Telegram I don't understand, see this code snippet:
``` python
# audio is a tg.Audio, now a "module" scope, hence same audio object used for all tests
audio1 = (await bot.send_audio(chat_id, audio.file_id, read_timeout=50)).audio
audio2 = (await bot.send_audio(chat_id, audio.file_id, read_timeout=50)).audio
new_file = await bot.get_file(audio.file_id)
print(audio.file_id == new_file.file_id)  # False  -- This should be true?!
print(audio1.file_id == new_file.file_id)  # False
print(audio2.file_id == new_file.file_id)  # True  -- Why is this True but the previous one is False?
print(audio1.file_id == audio2.file_id)  # False
```
since other tests send audio using that file_id, it seems to change file_id returned by `get_file`.

Bibo-Joshi: So the problem is that files have multiple file ids (which is documented) and when exactly they change is not obvious and more importantly not documented. I would therefore suggest to remove tests that assume that the `file_id` stays unchanged. i.e. things like `assert (await bot.get_file(file_id)).file_id == file_id`. Even with your updated logic, TG could change how things behave und we run into problems again.

harshil21:
Phew. By now the test suite runs fairly well with the only `test_build_raw` failing occasionally (which is being fixed in #3322). So let me summarize the "Experiments" I've been running:

1. Job queue tests now `Event` based, meaning they will return only after the callback has been run a specified number of times. This has the advantage of making those tests _semi_ time independent, so its less flaky and works on Windows and MacOS. 
2. Tried running pytest-xdist with more workers, but this seems to make tests slightly more flaky. My local machine seems to handle large number of workers just fine, not sure why that's not the case here. 
Anyway switching to `--dist=loadgroup` yields about ~1-4 min time reduction. Having more number of workers could reduce it by another 2 mins at least. 

3. Testing without dependencies is now done in a single pass, since spawning many pytest instances was just wasting time, not sure if it was deliberately not done like this in the first place.

Onto the reviews!

tests/conftest.py
```diff
@@ -52,12 +52,28 @@
...
+        if (
+            "NoReq" not in parent.name
+            and parent.name.endswith("Req")
+            and not parent.get_closest_marker(name="flaky")
+        ):
+            parent.add_marker(pytest.mark.flaky(3, 1))
+        elif parent.name.endswith("NoReq") and not parent.get_closest_marker(name="no_req"):
+            parent.add_marker(pytest.mark.no_req)
+
+
```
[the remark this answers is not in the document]
harshil21: `pytest.mark.req` not added if it exists since we don't want to stack them on a test class again and again..

tests/conftest.py
```diff
@@ -170,20 +187,35 @@ async def raw_bot(bot_info):
...
-    default_bot = make_bot(bot_info, defaults=Defaults(**param))
-    async with default_bot:
-        yield default_bot
+    # If the bot is already created, return it. Else make a new one.
+    default_bot = default_bots.get(defaults, None)
+    if default_bot is None:
+        default_bot = make_bot(bot_info, defaults=defaults)
+        await default_bot.initialize()
+        default_bots[defaults] = default_bot  # Defaults object is hashable
+    return default_bot
```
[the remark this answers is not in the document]
harshil21: This is causing more problem than it solves, doing it on `pytest_sessionfinish` using loop.run_until_complete or even asyncio.run always ends with an Event loop is closed error

tests/conftest.py
harshil21: > make_bot is a different function than make_mock_user though, isn't it

yes, but since it creates a new DictExtBot, `make_mock_user` is used in its `get_me`. 

`make_bot(FALLBACKS[0]["token"]).id != make_bot(FALLBACKS[1]["token"])` only because of the different user id which is fetched from the token. This important for the equality test

tests/test_bot.py
harshil21: not doing this, I think the nonlocal approach is better; a reset function would make sense if all the methods are using it

tests/test_application.py
harshil21: so if `start_polling` times out, then while exiting the `async with`, we end up with the `This application is still running error`. That's why I stop the updater so it can exit cleanly without failing the test.

tests/test_application.py
harshil21: tried a couple of solutions, and it seems to be giving more problems so not adding

harshil21:
> I don't really get what this commit is about. What's the benefit of using `DictBot` in these places?

Saving calls to real `get_me` by using DictBots.

>   * fixtures and pytest-stuff is set up in `conftest.py`
>   * other auxil-stuff goes into `auxil`.
>     I can also add that in my PR though …

Leaving that up to you..

tests/conftest.py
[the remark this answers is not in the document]
Bibo-Joshi: fair enough. A CSI comment about that would be nice though :)

tests/test_bot.py
Bibo-Joshi: If a large number is using it, that's enoguh IMO. avoids repetition

tests/test_application.py
Bibo-Joshi: then please use `except: TelegramError` or something like that. If there's something fundamentally wrong with `updater.start_polling` I wouldn't like to just ignore it here …

Bibo-Joshi:
> > I don't really get what this commit is about. What's the benefit of using `DictBot` in these places?
> 
> Saving calls to real `get_me` by using DictBots.

Then I recommend renamend to `TestBot` or something like that, i.e. a more generic name for "bot class that we use in tests". `DictBot` still implies that there is one specific change in the class …

> 
> > * fixtures and pytest-stuff is set up in `conftest.py`
> > * other auxil-stuff goes into `auxil`.
> >   I can also add that in my PR though …
> 
> Leaving that up to you..

fair enough :)

tests/test_bot.py
```diff
@@ -366,10 +368,26 @@ def test_bot_pickling_error(self, bot):
         with pytest.raises(pickle.PicklingError, match="Bot objects cannot be pickled"):
             pickle.dumps(bot)
 
+    def test_bot_deepcopy_error(self, bot):
+        with pytest.raises(TypeError, match="Bot objects cannot be deepcopied"):
+            copy.deepcopy(bot)
+
+    @bot_methods(ext_bot=False)
+    def test_api_methods_have_log_decorator(self, bot_class, bot_method_name, bot_method):
```
Bibo-Joshi, line 376: Is this addition right? maybe a merge error?

tests/test_bot.py
```diff
@@ -1761,6 +1775,18 @@ async def test_send_contact(self, bot, chat_id):
         assert message.contact.last_name == last_name
         assert message.has_protected_content
 
+    async def test_send_chat_action_all_args(self, bot, chat_id, monkeypatch):
+        async def make_assertion(*args, **_):
+            kwargs = args[1]
+            return (
```
Bibo-Joshi, line 1781: Is this addition right? maybe a merge error?

tests/test_bot.py
```diff
@@ -3119,7 +3153,7 @@ async def test_replace_callback_data_copy_message(self, cdc_bot, chat_id):
             bot.callback_data_cache.clear_callback_data()
             bot.callback_data_cache.clear_callback_queries()
 
-    async def test_get_chat_arbitrary_callback_data(self, super_group_id, cdc_bot):
+    async def test_get_chat_arbitrary_callback_data(self, channel_id, cdc_bot):
```
Bibo-Joshi, line 3156: why?

tests/test_datetime.py
```diff
@@ -95,12 +96,14 @@ def test_to_float_timestamp_absolute_no_reference(self):
         with pytest.raises(ValueError):
             tg_dtm.to_float_timestamp(dtm.datetime(2019, 11, 11), reference_timestamp=123)
 
-    @pytest.mark.parametrize("time_spec", DELTA_TIME_SPECS, ids=str)
-    def test_to_float_timestamp_delta(self, time_spec):
+    def test_to_float_timestamp_delta(self):
```
Bibo-Joshi, line 99: why not parametrize?

tests/test_datetime.py
```diff
@@ -128,22 +131,22 @@ def test_to_float_timestamp_time_of_day_timezone(self, timezone):
             ref_t + (-utc_offset.total_seconds() % (24 * 60 * 60))
         )
 
-    @pytest.mark.parametrize("time_spec", RELATIVE_TIME_SPECS, ids=str)
-    def test_to_float_timestamp_default_reference(self, time_spec):
+    def test_to_float_timestamp_default_reference(self):
```
Bibo-Joshi, line 134: same

tests/test_datetime.py
```diff
@@ -128,22 +131,22 @@ def test_to_float_timestamp_time_of_day_timezone(self, timezone):
...
+                tg_dtm.to_float_timestamp(i, reference_timestamp=now)
+            ), f"Failed for {i}"
 
     def test_to_float_timestamp_error(self):
         with pytest.raises(TypeError, match="Defaults"):
             tg_dtm.to_float_timestamp(Defaults())
 
-    @pytest.mark.parametrize("time_spec", TIME_SPECS, ids=str)
-    def test_to_timestamp(self, time_spec):
+    def test_to_timestamp(self):
```
Bibo-Joshi, line 146: same

tests/test_inputmedia.py
```diff
@@ -612,6 +628,21 @@ async def func():
...
 
+    @pytest.mark.parametrize("sequence_type", [list, tuple, CustomSequence])
+    @pytest.mark.parametrize("bot_class", ["raw_bot", "ext_bot"])
+    async def test_send_media_group_different_sequences(
+        self, bot, chat_id, media_group, sequence_type, bot_class, raw_bot
+    ):
+        """Test that send_media_group accepts different sequence types. This test ensures that
+        Bot._insert_defaults works for arbitrary sequence types."""
+        bot = bot if bot_class == "ext_bot" else raw_bot
+
```
Bibo-Joshi, line 639: merge error?

tests/test_inputmedia.py
```diff
@@ -712,6 +743,21 @@ async def test_send_media_group_all_args(self, bot, raw_bot, chat_id, media_grou
             )
             assert all(mes.has_protected_content for mes in messages)
 
+    async def test_send_media_group_with_spoiler(
+        self, bot, chat_id, photo_file, video_file  # noqa: F811
+    ):
+        # Media groups can't contain Animations, so that is tested in test_animation.py
```
Bibo-Joshi, line 749: merge error?

tests/test_jobqueue.py
```diff
@@ -68,21 +66,43 @@ def test_init_job(self):
...
         self.job_time = 0
         self.received_error = None
+        self.done_with_cb = asyncio.Event()
+        self.done_with_err_cb = asyncio.Event()
+
+    async def wait_for_cbs(self, number: int = 1):
+        async def _wait_for_cbs():
+            for _ in range(number):
+                await self.done_with_cb.wait()
+                self.done_with_cb.clear()
```
Bibo-Joshi, line 89: maybe `asyncio.Barrier` can help here?

tests/test_jobqueue.py
```diff
@@ -138,7 +173,7 @@ def test_application_weakref(self, bot):
 
     async def test_run_once(self, job_queue):
         job_queue.run_once(self.job_run_once, 0.1)
-        await asyncio.sleep(0.2)
+        await self.wait_for_cbs()
```
Bibo-Joshi, line 176: I don't see the sense in these changes. The tests are no longer testing that the jobs run within the specified time schedule. Now the jobs could run whenever they like.
Not testing on Windows & Mac due to instabilities is more derisable to me than not actually testing the functionality

tests/test_jobqueue.py
```diff
@@ -596,8 +641,8 @@ async def callback(_):
         if wait:
             assert not task.done()
             ready_event.set()
-            await asyncio.sleep(0.1)
+            await task  # no CancelledError here (see source code of JobQueue.stop for details)
```
Bibo-Joshi, line 644: similar as for the other tests: The point here is to ensure that the scheduled job is (not) awaited, which is not tested by your changes …

Bibo-Joshi (changes requested):
I reviewd the commits since my last review. In a few places it looks like you duplicated an already exsting tests. Maybe those are merge errors or maybe it only looks that way if you don't view *all* commits …

tests/bots.py
```diff
@@ -47,7 +47,9 @@
 
 FALLBACKS = json.loads(base64.b64decode(FALLBACKS).decode("utf-8"))  # type: list[dict[str, str]]
 
-chosen_bot = {}
+
+class BotInfoProvider:
+    chosen_bot = {}
```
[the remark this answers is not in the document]
harshil21: https://github.com/python-telegram-bot/python-telegram-bot/pull/3426#pullrequestreview-1222083004

The other idea is to just upper case the variable and keep it global

tests/test_application.py
```diff
@@ -985,7 +986,7 @@ async def callback(update, context):
         ), "incorrect stacklevel!"
 
     async def test_non_blocking_no_error_handler(self, app, caplog):
-        app.add_handler(TypeHandler(object, self.callback_raise_error, block=False))
+        app.add_handler(TypeHandler(object, self.callback_raise_error("Test error"), block=False))
```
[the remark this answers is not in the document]
harshil21: technically the callback is supposed to raise an error, so this raises an error, just not the one we expected 😄

tests/test_bot.py
harshil21: it's to avoid race conditions with `test_pin_and_unpin_message`, we can just use another chat for pinning

tests/test_datetime.py
harshil21: https://github.com/python-telegram-bot/python-telegram-bot/pull/3426/commits/3bdd98128a4f678c21f747b088df2ba0a2e9e21a

there's a tiny chance that there is an error while collecting the tests (happens when time goes from HH:59:00 -> HH+1:00:00)

This was happening before this PR as well

tests/test_jobqueue.py
harshil21: I would use that, but it's py 3.11+ only..

tests/test_jobqueue.py
harshil21: Yes, I could add a `timeout` parameter to `wait_for_cbs()` which ensures the test is still run within the time limit +/- 0.x s. The reason I didn't add this by default is because some tests still use `sleep(x)` which does ensure that

tests/test_jobqueue.py
harshil21: I don't understand. The task of shutting down the job queue is being awaited here, and not the job itself

harshil21:
>  In a few places it looks like you duplicated an already exsting tests. Maybe those are merge errors or maybe it only looks that way if you don't view _all_ commits …

I think it looks like that since I had to rearrange those tests into these classes. Nevertheless i'll check again

tests/bots.py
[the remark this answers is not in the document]
Bibo-Joshi: Ah. I guess I rather meant something like

```python
class BotInfoProvider:
    def __init__(self):
        self._cached = {}
    
    def get_info(self):
        if self._cached:
            return self._cached
        self._cached = {…}
        return self._cached
```
and then build one instance of this class in `conf.py`

tests/test_application.py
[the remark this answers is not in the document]
Bibo-Joshi: lol, fair enough :D

tests/test_datetime.py
Bibo-Joshi: Worth a CIS comment :)

tests/test_jobqueue.py
Bibo-Joshi: ups, overlooked that …

tests/test_jobqueue.py
Bibo-Joshi: We want to test that

* `job_queue.stop(wait=True)` waits for the job to finish before shutting down
* `job_queue.stop(wait=False)` cancels the job and shuts down directly

`await (task = job_queue.stop(wait=wait)` just wait for the `job_queue` to shut down, but is indifferent to the running job

tests/test_jobqueue.py
Bibo-Joshi: Adding a timeout parameter would basically do what we were doing up to now - just more complicated ;)

harshil21: humm, okay I'll revert Experiment 1

tests/conftest.py
Bibo-Joshi: okay, so the point is that we need to be able to make a bot with a specific id, but currently have no way (or are not willing to provide a way) of getting the correct user info from a token. So we're okay building a bot with the specified token & id, but username, first name etc not matching these token & id.
If you can add a CSI style comment about that, I'm good :)

Bibo-Joshi:
Awesome work! 💪

PR #3326, opened 2022-11-01
Move examples in docstrings to separate blocks

lemontree210:
closes #3320 

pre-commit hooks don't pass because of cyclic import in `_pollanswerhandler.py`.

PR #3429, opened 2022-12-17
Replace links to PTB wiki with extlinks

lemontree210:
+ replaces https links to PTB Wiki with extlinks following approach in #3423 
+ moves some examples to "Examples:" block (that I must have missed in #3326)

### DIFF
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -136,10 +136,8 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AbstractAsyncContextManager)
     Examples:
         :any:`Echo Bot <examples.echobot>`
 
-    .. seealso:: `Your First Bot <https://github.com/\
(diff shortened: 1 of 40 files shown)


telegram/ext/_application.py
```diff
@@ -338,8 +336,7 @@ def concurrent_updates(self) -> int:
         """:obj:`int`: The number of concurrent updates that will be processed in parallel. A
         value of ``0`` indicates updates are *not* being processed concurrently.
 
-        .. seealso:: `Concurrency <https://github.com/\
-            python-telegram-bot/python-telegram-bot/wiki/Concurrency>`_
+        .. seealso:: :wiki:`Concurrency`
```
lemontree210, line 339: If the link text is the same as the relative address, the syntax of `:wiki:` link becomes even more concise

Bibo-Joshi (approved):
LGTM 👍 You can merge at will :)

PR #3443, opened 2022-12-27
Fix application run webhook method on readme

zeshuaro:
### Checklist for PRs

I think `Application.run_webhook` should be the correct method instead of `Application.start_webhook` as per the [source code](https://github.com/python-telegram-bot/python-telegram-bot/blob/v20.0b0-2-gf3650364/telegram/ext/_application.py#L715-L843).


harshil21 (approved):

Bibo-Joshi:
Thanks for the catch!

PR #3444, opened 2022-12-28
Add Enum constants

lemontree210:
cherry-picked from reverted commits in #3412

### DIFF
--- a/telegram/constants.py
+++ b/telegram/constants.py
@@ -935,7 +935,7 @@ class MediaGroupLimit(IntEnum):
     :meth:`telegram.Bot.send_media_group`.
     """
     MAX_MEDIA_LENGTH = 10
-    """:obj:`int`: Minimum length of a :obj:`list` passed as the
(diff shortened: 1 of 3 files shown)


telegram/constants.py
```diff
@@ -919,6 +920,27 @@ class MaskPosition(StringEnum):
...
+
+    __slots__ = ()
+
+    MIN_MEDIA_LENGTH = 2
+    """:obj:`int`: Minimum length of a :obj:`list` passed as the
+    :paramref:`~telegram.Bot.send_media_group.media` parameter of
+    :meth:`telegram.Bot.send_media_group`.
+    """
+    MAX_MEDIA_LENGTH = 10
+    """:obj:`int`: Minimum length of a :obj:`list` passed as the
```
Bibo-Joshi, line 938: Max lenght

Bibo-Joshi (changes requested):

telegram/constants.py
lemontree210: Arrgh :)
Done, thanks

PR #3453, opened 2022-12-31
Freeze empty classes

### LINKED ISSUE #3452: [BUG] Telegram classes with no custom `__init__` are not frozen by default
### Steps to Reproduce

1. Initialize a class with no overriden `__init__`, e.g. `VideoChatStarted`
2. do `VideoChatStarted().api_kwargs = "something"`

### Expected behaviour

`AttributeError` should be raised

### Actual behaviour

The attribute is set successfully. 

### Operating System

OS-agnostic

```shell
Python version agnostic
PTB - 20.0b0
```

### Relevant log output

_No response_

### Additional Context

_No response_

Bibo-Joshi:
Closes #3452

I decided to just add `__init__` methods to the empty classes that call `self._freeze()`. That requires fewer changes than setting `self._freeze = True` in `TO.__init__` and using `with self._unfrozen()` in *all* the classes …

### Checklist for PRs

- ~[ ] Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)~

### DIFF
--- a/telegram/_videochat.py
+++ b/telegram/_videochat.py
@@ -42,6 +42,11 @@ class VideoChatStarted(TelegramObject):
 
     __slots__ = ()
 
+    def __init__(self, *, api_kwargs: JSONDict = None) -> None:
+        super().__init__(api_kwargs=api_kwargs)
+
+        self._freeze()
+
 
 class VideoChatEnded(TelegramObject):
     """
--- a/telegram/_forumtopic.py
+++ b/telegram/_forumtopic.py
@@ -122,6 +122,11 @@ class ForumTopicClosed(TelegramObject):
 
     __slots__ = ()
 
+    def __init__(self, *, api_kwargs: JSONDict = None) -> None:
+        super().__init__(api_kwargs=api_kwargs)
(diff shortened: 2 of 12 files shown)


telegram/_videochat.py
```diff
@@ -42,6 +42,11 @@ class VideoChatStarted(TelegramObject):
 
     __slots__ = ()
 
+    def __init__(self, *, api_kwargs: JSONDict = None) -> None:
+        super().__init__(api_kwargs=api_kwargs)
+
+        self._freeze()
+
 
 class VideoChatEnded(TelegramObject):
```
lemontree210, line 51: is there a reason for having used a context manager in other modules but leaving `self._freeze()` in this class and the next class in the same module?

Bibo-Joshi: Yes, that's exactly the reason :) `with self._frozen()` has no advantage here …

harshil21 (approved):

PR #3457, opened 2023-01-01
Downgrade `sphinx` to 5.3.0 to fix docs search

harshil21:
The docs search stopped working when we upgraded to sphinx 6.0.0, so let's downgrade back to 5.3.0 for now.

Reverts https://github.com/python-telegram-bot/python-telegram-bot/pull/3450.

Doc build up at: https://docs.python-telegram-bot.org/en/doc-downgrade-sphinx/ you can check as well

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,7 +5,7 @@ on:
       - master
       - doc-fixes
   push:
-    branches: 
+    branches:
       - master
       - doc-fixes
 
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        python-version: [3.9]
+        python-version: [3.7]
```
Bibo-Joshi, line 18: does this have to be reverted? same for line 8

harshil21: no I guess, i'll bump it back up then

PR #3460, opened 2023-01-01
Stabilize tests on closing and hiding the general forum topic

Bibo-Joshi:
Let's hope this works :)

### DIFF
--- a/tests/test_forum.py
+++ b/tests/test_forum.py
@@ -30,6 +30,7 @@ from telegram import (
     GeneralForumTopicUnhidden,
     Sticker,
 )
+from telegram.error import BadRequest
 
 TEST_MSG_TEXT = "Topics are forever"
 TEST_TOPIC_ICON_COLOR = 0x6FB9F0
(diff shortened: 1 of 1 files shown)


tests/test_forum.py
```diff
@@ -262,30 +263,40 @@ async def test_edit_general_forum_topic(self, bot, forum_group_id):
...
-
-    async def test_hide_and_unhide_general_forum_topic(self, bot, forum_group_id):
-
+        # then hide
         result = await bot.hide_general_forum_topic(
             chat_id=forum_group_id,
         )
         assert result is True, "Failed to hide general forum topic"
 
+        # then unhide, bot don't reopen
```
harshil21, line 291:
```suggestion
        # then unhide, but don't reopen
```

tests/test_forum.py
```diff
@@ -262,30 +263,40 @@ async def test_edit_general_forum_topic(self, bot, forum_group_id):
...
+        """Since reopening also unhides and hiding also closes, testing (un)hiding and
+        closing/reopening in different tests would mean that the tests have to be executed in
+        a specific order. For stability, we instead test all of them in one test."""
+
+        # We first ensure that the topic is open and visible
+        # Otherwise the tests below will fail
+        try:
+            await bot.reopen_general_forum_topic(chat_id=forum_group_id)
+        except BadRequest as exc:
+            if "Topic_not_modified" not in exc.message:
```
harshil21, line 276: maybe add a comment here saying that this is for when the topic is already opened

tests/test_forum.py
```diff
@@ -262,30 +263,40 @@ async def test_edit_general_forum_topic(self, bot, forum_group_id):
...
+        # then unhide, bot don't reopen
         result = await bot.unhide_general_forum_topic(
             chat_id=forum_group_id,
         )
         assert result is True, "Failed to unhide general forum topic"
 
-        # hiding the general topic also closes it, so we reopen it
+        # finally, reopen
+        # as this also unhides, this should ensure that the topic is open and visible
+        # for the next test run
```
lemontree210, line 299: If we create a fixture with the initial reopening (above, with catching of the exception) of the general forum topic, then `yield`, and then this final re-opening, then we can ensure that the general topic remains open even if for some reason this test fails halfway through.

Maybe it will too much code for a simple test like this, though. Plus this fixture is unlikely to be reused elsewhere.

Bibo-Joshi: If we were to use the general topic in more tests, I would definitely agree :) While we only have this one test, I think it's more readable this way.

lemontree210 (approved):

harshil21 (approved):
If this still doesn't work, maybe adding an asyncio sleep next time would work?

PR #3468, opened 2023-01-02
Upload GH action built docs as Artifacts

harshil21:
Uploads the built docs from the workflow run so it is easy to see the rendered html pages without building the docs again locally.

Docs on artifacts: https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

How to view the artifact? Click on the "Details" of Test Doc Build -> Summary (left panel) -> scroll down and download the artifact


.github/workflows/docs.yml
```diff
@@ -30,3 +30,12 @@ jobs:
           python -W ignore -m pip install -r requirements-all.txt
       - name: Build docs
         run: sphinx-build docs/source docs/build/html -W --keep-going -j auto
+      - name: Upload docs
+        uses: actions/upload-artifact@v3
+        with:
+          name: built docs
```
Bibo-Joshi, line 36:
```suggestion
          name: HTML Docs
```

.github/workflows/docs.yml
```diff
@@ -30,3 +30,12 @@ jobs:
...
+          retention-days: 7
+          path: |
```
Bibo-Joshi, line 38: Can you add a short comment on what this does?

Bibo-Joshi (commented):
Also a nice idea! TBH I guess the use case is not overly big, but neither is the effort 🙃

.github/workflows/docs.yml
harshil21: the days? or the path?

Bibo-Joshi: The path

Bibo-Joshi (approved):
Thanks for the updates! As said above, even if the advantage may not be huge, I see no downsides of this change. I'll give the others a chance to chime in if they see the need and merge on thursday evening otherwise :)

PR #3469, opened 2023-01-02
Cache GH action dependencies

harshil21:
Attempts to speed up the test suite by caching GH actions dependencies. About ~1 mins is used every run just to install our dependencies. This should save that.

More info about how the cache is used (and also not used) can be found at https://github.com/actions/setup-python#caching-packages-dependencies

Edit: Confirmed the cache is working correctly by checking both https://github.com/python-telegram-bot/python-telegram-bot/actions/caches & the pytest logs in the workers below.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -24,6 +24,8 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+          cache-dependency-path: '**/requirements*.txt'
       - name: Install dependencies
         run: |
           python -W ignore -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
Nice idea 👍 just to clarify: this "only" removes the need to download the wheels each time, but pip still need to install them.
Can you link to a CI run where downloading and installing took 1min? I clicked around a bit and mostly found that step taking 20-30 or even just 10 seconds...

harshil21:
> Nice idea 👍 just to clarify: this "only" removes the need to download the wheels each time, but pip still need to install them. 

precisely.

> Can you link to a CI run where downloading and installing took 1min? I clicked around a bit and mostly found that step taking 20-30 or even just 10 seconds...

https://github.com/python-telegram-bot/python-telegram-bot/actions/runs/3825220803/jobs/6508026698

This takes 54s in the "install dependencies" tab, but it doesn't even include the one happening in pytest (after we test without dependencies). The precise time for that isn't available, but when I was looking at it live, it took about 30s or so

Bibo-Joshi:
Thanks for the link. Even if the advantage is not huge, I see no downsides of this change. I'll give the others a chance to chime in if they see the need and merge on thursday evening otherwise :)

PR #3481, opened 2023-01-05
Propose to add unit test example for echobot example

zeshuaro:
### Checklist for PRs

Recently, I've achieved full test coverage in my project with `python-telegram-bot`. So I figured to share my approach of testing back to the community and to discuss with you all 🙂 

I noticed the unit tests [section](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Writing-Tests#unit-tests) of the Wiki mentioned an unmaintained unit testing framework. I would argue the approach there is closer to being an integration test than a unit test. In my humble opinion, a unit test would simply be testing a method/API by providing some inputs and checking for the expecting outputs. This is also the approach I've taken in my project.

This PR proposes a unit test example for the echobot [example](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/examples/echobot.py) to test for the callbacks. Basically, it mocks out `Update` and `CallbackContext` which are the expected arguments for the callbacks. We then pass the arguments to the callback and assert for the expected behaviour. Other than being able to unit test the code, another benefit is that this approach doesn't rely on another testing library/framework and all we need is `unittest` and `pytest`.

We should also be able to test out if the handlers have been added to the `Application` with a similar approach, but might require a little bit of refactoring of the example code so maybe will have to discuss on whether it's worth unit testing that bit.

Happy to discuss if this unit test example is appropriate. And I can also add more testing examples for the other example code if you think they're relevant.

PS, I'm not sure if you want this example test to be part of the `tests` folder or the `examples` folder.

### DIFF
--- a/tests/examples/test_echobot.py
+++ b/tests/examples/test_echobot.py
@@ -1,0 +1,42 @@
+from unittest.mock import MagicMock
+
+import pytest
+
+from examples.echobot import echo, help_command, start
+from telegram import ForceReply, Message, Update, User
+from telegram.ext import ContextTypes
+
+
+class TestEchoBot:
+    MENTION_HTML = "mention_html"
+    TEXT = "text"
+
+    def setup_method(self) -> None:
+        self.user = MagicMock(spec=User)
+        self.user.mention_html.return_value = self.MENTION_HTML
+
+        self.message = MagicMock(spec=Message)
+        self.message.text = self.TEXT
+
+        self.update = MagicMock(spec=Update)
+        self.update.effective_user = self.user
+        self.update.message = self.message
+
+        self.context = MagicMock(spec=ContextTypes.DEFAULT_TYPE)
+
+    @pytest.mark.asyncio
+    async def test_start(self) -> None:
+        await start(self.update, self.context)
+        self.update.message.reply_html.assert_called_once_with(
+            f"Hi {self.MENTION_HTML}!", reply_markup=ForceReply(selective=True)
+        )
+
+    @pytest.mark.asyncio
+    async def test_help_command(self) -> None:
+        await help_command(self.update, self.context)
+        self.update.message.reply_text.assert_called_once_with("Help!")
+
+    @pytest.mark.asyncio
+    async def test_echo(self) -> None:
+        await echo(self.update, self.context)
+        self.update.message.reply_text.assert_called_once_with(self.TEXT)


Bibo-Joshi:
Hi. Thank you very much for sharing your ideas and proposing this addition!

Let me start by saying that I would *not* like to include a testing setup for the examples. This has two reasons:

1. The examples are IMO built simple enough to not be worth the effort of automatic testing. Changes in the library that are so big that the examples will break are usually also so big that `pylint` and `mypy` will catch the changes in the examples. Moreover, as each example usually showcases one specific functionality of the library, that particular functionality would have to change for the example to break.
2. A bot-testing framework of any sorts is beyond the scope of this library.

Still, automated testing of bots is a topic that comes up every now and then and that's perfectly reasonable. Since you invited us to discuss your proposal, I will do so :D 

I'm not overly familiar with the correct terminology of "integration/system/unit/… tests", so let me just outline what IMO testing a bot should cover:

1. When testing the individual callback functions, one should test that these trigger the expected result, i.e.
    * return the expected return value (which is only relevant for `ConversationHandler` callbacks)
    * send the expected messages/make the appropriate calls to the Bot API
    * store data in the expected places/make calls to other apis/…

   In short, that they change the state on the local and TG server in the way that you expect them to.
   This ensures that the callback works as expected, without making any assumptions on the implementation of the callback.
3. When testing that updates are handled in the correct way, one should test that `Application.process_update` calls the correct callback function(s) in the correct order. This ensures that updates are handled as expected without making any assumptions on the handler setup. Alternatively one could also here test the things mentioned above, in which case a test wouldn't even make assumptions on the structure of the individual callbacks.

In your test setup, you rely rather heavily on the implementation of the specific callback functions. E.g. if instead of `update.message.reply_text` the implementation would switch to `update.effective_user.send_message` or `context.bot.send_message`, your tests would fail even though the callback still sends the desired message.

I would hence suggest to instead mock the `Bot` object or it's methods - which brings us back rather closely to the unmaintained `ptbtest` library that you mentioned. Another approach that became viable in v20 is to provide a mock-implemantation of `BaseRequest` that does not make calls to the bot API. That way one doesn't need to override all the individual methods of `Bot`, but just `BaseRequest.do_request`.

zeshuaro:
Thanks for the very detailed response. Always appreciate your effort and dedication in maintaining this library and responding to all the issues and PRs.

Responding to your first 2 points:

1. `pylint` and `mypy` are static analysis tools at the end. I think having automated tests would catch unexpected changes more reliably. But like you said, the examples are quite simple, so it's fair enough to not have tests for the examples.
2. The proposed test is not a framework or any sort, basically simple tests to cover the methods/callbacks. But maybe the tests are inappropriate, which I'll go into below to also discuss your other points.

Let me also iterate what I think unit testing would be appropriate. I would say that end-users would want to test for the code written by themselves and the APIs they're using. Hence, testing whether the callbacks are returning the expected value or making the expected API calls like you suggested. How these callbacks are being used, e.g. by `Application.process_update` should not be tested by the end-user as this is a functionality provided by the library and the library should cover the testing of it. And I think most end-users don't need to call `process_update` and hence won't even know about this method.

To add to the above point, an assumption will still have to be made here of how `Application.process_update` handles the update. So that users can provide the appropriate parameters/inputs into this method. And this requires a deeper understanding of the internals of how this method works and going beyond testing of the code written by the end-users.

I understand that there is value of testing whether the correct callback is being used. How I would test this is to mock out the `Application` object and test if the callbacks are added in the order we expected. That way, we're testing the API the end-users are using, e.g. `Application.add_handler` but not the internals of how `Application` deal with these handlers.

> if instead of `update.message.reply_text` the implementation would switch to `update.effective_user.send_message` or `context.bot.send_message`, your tests would fail even though the callback still sends the desired message.

Yes the desired message is still being sent but I would also argue that this is the exact behaviour we want with unit testing to ensure the correct API is being called. If the implementation switches to a different API call, then the test should also ensure the expected API is being called.

> I would hence suggest to instead mock the `Bot` object or it's methods - which brings us back rather closely to the unmaintained `ptbtest` library that you mentioned. Another approach that became viable in v20 is to provide a mock-implemantation of `BaseRequest` that does not make calls to the bot API. That way one doesn't need to override all the individual methods of Bot, but just `BaseRequest.do_request`.

Again in my opinion this is drilling down to the details of how this library is written and would require a deeper knowledge of the internals of the library.

I'm also not sure how easy testing would be by providing a mocked `BaseRequest`. Are we then expected to test if `do_request` is calling the expected Telegram endpoint and the expected payloads? Isn't the whole point of this libary is to abstract all those API calls and hence these tests are only relevant for the library but not the end-users?

Bibo-Joshi:
> I'm also not sure how easy testing would be by providing a mocked `BaseRequest`. Are we then expected to test if `do_request` is calling the expected Telegram endpoint and the expected payloads? Isn't the whole point of this libary is to abstract all those API calls and hence these tests are only relevant for the library but not the end-users?

Fair point

> I understand that there is value of testing whether the correct callback is being used. How I would test this is to mock out the `Application` object and test if the callbacks are added in the order we expected. That way, we're testing the API the end-users are using, e.g. `Application.add_handler` but not the internals of how `Application` deal with these handlers.

That makes some sense, yes. What this type of testing does not cover is that expected the order of handlers is actually correct. You might have set up your filters in a wrong way such that an update is still handled by a different handler than you wanted. Especially when using `ConversationHandler`, testing this `Application.process_update` is valuable IMO since the order of handlers changes.

> And I think most end-users don't need to call `process_update` and hence won't even know about this method.

Mind you, you can ofc also run `Application.run_polling` in your test suite and provide the updates by mocking `Bot.get_updates`, that's just way more effort than just calling `Application.process_update` :)

---

I guess we mainly disagree on where the line between "PTB internals" and "PTB user-interface" lies.

As maintainer of PTB, I can only advocate the treat everything documented on https://docs.python-telegram-bot.org as the the "PTB user-interface", which includes that

* `Message.reply_text` & other methods are convenience shortcuts for the respective methods of the `Bot` class
* `Application.process_update` accepts exactly one argument, name the `update`

As general programmer, I can only advocate to write tests such that they don't rely on the specific implementation of the to-be-tested functionality.

zeshuaro:
Thanks again for your reply. The points you've made are very insightful and would definitely help me to think about the tests. I'll also close off this PR in the meantime 🙂

PR #3483, opened 2023-01-05
Drop `telegram(.ext).` prefixes

Bibo-Joshi:
As suggested in https://t.me/c/1494805131/32127. Build will be up soonish on rtd.

URLs seem to be unchanged.

Before a merge, I want to test if this would imply any changes for rulesbot.

### DIFF
--- a/docs/source/telegram.animation.rst
+++ b/docs/source/telegram.animation.rst
@@ -1,5 +1,5 @@
 Animation
-==================
+=========
 
 .. Also lists methods of _BaseThumbedMedium, but not the ones of TelegramObject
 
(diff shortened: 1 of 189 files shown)


Poolitzer:
@Bibo-Joshi And we should check search

Bibo-Joshi:
Both search and rulesbot-search (which are basically the same thing, IG) seem to work fine :)

harshil21 (commented):
Do we want to drop the prefixes even for the modules?

Bibo-Joshi:
> Do we want to drop the prefixes even for the modules?

For `telegram.{error, request, constants, …}` I kept them so that it's clear at first glance how to import them. Is that what you mean?

harshil21:
@Bibo-Joshi yes

Poolitzer (approved):
Yes, thank you, looks perfect.

harshil21 (approved):

PR #3484, opened 2023-01-05
Custom sidebar brand style

Bibo-Joshi:
As suggested in https://t.me/c/1494805131/32133. I kept the version number in a new line though … build should be up soonish on rtd.
Build at https://docs.python-telegram-bot.org/en/doc-logo-adjustments/

Preview:
image


Poolitzer (approved):
Huge improvement on mobile! Thanks!

clot27 (approved):
Finally some more space :)

docs/source/_static/style_sidebar_brand.css
```diff
@@ -0,0 +1,11 @@
+.sidebar-sticky .sidebar-brand {
+      flex-direction: row;
+}
+
+.sidebar-sticky .sidebar-brand .sidebar-logo-container {
+      align-self: center;
+}
+
+.sidebar-sticky .sidebar-brand .sidebar-brand-text {
+      align-self: center;
```
lemontree210, line 10: Maybe it makes a difference on mobile, but on PC I see no difference with or without these directives

docs/source/_static/style_sidebar_brand.css
```diff
@@ -0,0 +1,11 @@
+.sidebar-sticky .sidebar-brand {
+      flex-direction: row;
+}
+
```
lemontree210, line 4:
```suggestion
.sidebar-container {
      width: 24rem;
}
```
It sort of vexes me that many class names are split over 2 lines in the sidebar. Increasing the width from default `15rem` to `24rem` makes sure even the longest class fits on one line. This may have to be checked on mobile, though

docs/source/_static/style_sidebar_brand.css
Bibo-Joshi: this only ensures that the logo and text are vertically centered. The change is very subtle since both are almost of the same size. You can see the difference if you open the dev tools of your browser and toggle the setting :)

docs/source/_static/style_sidebar_brand.css
lemontree210: The first bit probably depends on screen resolution. When I make my browser window smaller, I see that the line breaks return. This width must be some maximum width and the screen resolution overrides that.

The second bit is especially weird because I'm changing the width of the container, not its content. In my browser, the right-hand-side tree simply disappears if the window is not wide enough to display it.

lemontree210: BTW increasing the container width makes the logo bigger again (because there are percentages there) :)

lemontree210: How will the following work with you?

```css
.sidebar-sticky .sidebar-brand {
      flex-direction: row;
}

.sidebar-sticky .sidebar-brand .sidebar-logo-container {
      align-self: center;
      width: 30%;
}

.sidebar-sticky .sidebar-brand .sidebar-brand-text {
      align-self: center;
}

.content {
      width: 60%;
}

.sidebar-drawer {
      width: 30%;
}

.sidebar-container {
      width: 100%;
}
```
(you can adjust the percentages of `content` and `sidebar-drawer` slightly)

Obviously, the `content` setting may need to go into a different file or the file should be renamed

lemontree210: > the right hand side tree gets so wide that it overlaps the actual content …

I see that if you e.g. open the page for `telegram.BotCommandScopeAllChatAdministrators`, the end of the title will be hidden under the right sidebar. But it happens no matter what settings we put in the new CSS file.

I guess we have to add `overflow-wrap: anywhere` to `h1` tag.

Bibo-Joshi: > How will the following work with you? I tried resizing the browser and it looks fine.

this somewhat works on my desktop screen, but the main content column becomes really narrow and unless you expand the tree on the LHS there's a lot of dead space. TBH, I'd rather like to keep these settings as-is.

> I guess we have to add overflow-wrap: anywhere to h1 tag.

That sounds reasonable. Will add.

lemontree210: Yes, tampering with CSS can be really annoying :) It is to me, at least :)

PR #3485, opened 2023-01-05
Add `effective_name` property to `Chat` object

### LINKED ISSUE #3480: [FEATURE] convenience property for displaying name of Chat object

Because `Chat` objects can be private or groups/user/channels, one needs to check the type every time before using the appropriate property (`Chat.full_name` or `Chat.title` respectively). It would be nice to have a conveniente property that does this automatically

I propose a convenience property that return `Chat.full_name` if the chat.type is private or `Chat.title` otherwise.

_No response_

### Additional context

_No response_

Viicos:
Fixes #3480.

### Checklist for PRs

- [x] Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing

### DIFF
--- a/telegram/_chat.py
+++ b/telegram/_chat.py
@@ -389,9 +389,23 @@ class Chat(TelegramObject):
         self._freeze()
 
     @property
+    def effective_name(self) -> Optional[str]:
+        """
+        :obj:`str`: Convenience property. Gives :attr:`title` if not :obj:`None`,
+        else :attr:`full_name` if not :obj:`None`.
+
+        .. versionadded:: 20.1
+        """
+        if self.title is not None:
+            return self.title
+        if self.full_name is not None:
+            return self.full_name
+        return None
+
+    @property
     def full_name(self) -> Optional[str]:
         """
-        :obj:`str`: Convenience property. If :attr:`first_name` is not :obj:`None` gives,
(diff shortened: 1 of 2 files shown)


Viicos:
Let me know which `.. versionadded::` should be used, and if this property name is fine

Poolitzer:
@Viicos I would call it <code>effective_name</code>, and it needs a test

clot27:
> Let me know which `.. versionadded::` should be used, and if this property name is fine

the next release will be v20.1

telegram/_chat.py
```diff
@@ -388,10 +388,24 @@ def __init__(
...
+        :obj:`str`: Convenience property. Gives :attr:`title` if not :obj:`None`,
+        else :attr:`full_name` if not :obj:`None`.
+
+        .. versionadded:: 20.1
+        """
+        if self.title is not None:
+            return self.title
+        if self.full_name is not None:
+            return self.full_name
+        return None
```
clot27, line 403: You are not testing this line, please add on as:
```
chat = Chat(id=1, type=Chat.GROUP)
assert chat.effective_name is None

```

Viicos: Yes saw that, I'm on something else right now but this will be done in ~30mins

clot27 (approved):
LGTM

Bibo-Joshi:

PR #3500, opened 2023-01-12
Fix the chatmemberbot example

lemontree210:
the bot now keeps track of users who started a private chat

### DIFF
--- a/examples/chatmemberbot.py
+++ b/examples/chatmemberbot.py
@@ -144,14 +144,14 @@ async def start(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
     if it's a private chat. Since no automatic update is issued when a user starts a private chat
     with the bot, we have to track it explicitly here, when handling the /start command.
     """
-    cause_name = update.effective_user.full_name
+    user_name = update.effective_user.full_name
     chat = update.effective_chat
     if chat.type == Chat.PRIVATE:
-        logger.info("%s started a private chat with the bot", cause_name)
+        logger.info("%s started a private chat with the bot", user_name)
         context.bot_data.setdefault("user_ids", set()).add(chat.id)
 
     await update.effective_message.reply_text(
-        f"Welcome {cause_name}. Use /show_chats command to see what chats I'm in."
+        f"Welcome {user_name}. Use /show_chats command to see what chats I'm in."
     )
 
 
--- a/examples/chatmemberbot.py
+++ b/examples/chatmemberbot.py
@@ -87,9 +87,10 @@ async def track_chats(update: Update, context: ContextTypes.DEFAULT_TYPE) -> Non
     chat = update.effective_chat
     if chat.type == Chat.PRIVATE:
         if not was_member and is_member:
-            # This is not really needed in this particular case because /start command will run
-            # after the user unblocks the bot, and the start function will add the user to
-            # "user_ids".  We're including this here for the sake of the example.
+            # This may not be really needed in practice because most clients will automatically
+            # send a /start command after the user unblocks the bot, and start_private_chat()
+            # will add the user to "user_ids".
+            # We're including this here for the sake of the example.
             logger.info("%s unblocked the bot", cause_name)
             context.bot_data.setdefault("user_ids", set()).add(chat.id)
         elif was_member and not is_member:
--- a/examples/chatmemberbot.py
+++ b/examples/chatmemberbot.py
@@ -80,7 +80,10 @@ async def track_chats(update: Update, context: ContextTypes.DEFAULT_TYPE) -> Non
     chat = update.effective_chat
     if chat.type == Chat.PRIVATE:
         if not was_member and is_member:
-            logger.info("%s started the bot", cause_name)
+            # This is not really needed in this particular case because /start command will run
+            # after the user unblocks the bot, and the start function will add the user to
+            # "user_ids".  We're including this here for the sake of the example.
+            logger.info("%s unblocked the bot", cause_name)
             context.bot_data.setdefault("user_ids", set()).add(chat.id)
         elif was_member and not is_member:
             logger.info("%s blocked the bot", cause_name)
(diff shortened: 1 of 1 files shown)


examples/chatmemberbot.py
```diff
@@ -136,11 +139,29 @@ async def greet_chat_members(update: Update, context: ContextTypes.DEFAULT_TYPE)
         )
 
 
+async def start(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    """Greets the user and records that they started a chat with the bot
+    if it's a private chat. Since no automatic update is issued when a user starts a private chat
+    with the bot, we have to track it explicitly here, when handling the /start command.
+    """
+    cause_name = update.effective_user.full_name
```
harshil21, line 147: what's `cause_name` supposed to mean?

examples/chatmemberbot.py
```diff
@@ -136,11 +139,29 @@ async def greet_chat_members(update: Update, context: ContextTypes.DEFAULT_TYPE)
...
+    chat = update.effective_chat
+    if chat.type == Chat.PRIVATE:
+        logger.info("%s started a private chat with the bot", cause_name)
+        context.bot_data.setdefault("user_ids", set()).add(chat.id)
+
+    await update.effective_message.reply_text(
+        rf"Welcome {cause_name}. Use /show_chats command to see what chats I'm in."
```
harshil21, line 154:
```suggestion
        rf"Welcome {cause_name}. Use /show_chats to see what chats I'm in."
```
why the raw string though?

examples/chatmemberbot.py
```diff
@@ -80,7 +80,10 @@ async def track_chats(update: Update, context: ContextTypes.DEFAULT_TYPE) -> Non
     chat = update.effective_chat
     if chat.type == Chat.PRIVATE:
         if not was_member and is_member:
-            logger.info("%s started the bot", cause_name)
+            # This is not really needed in this particular case because /start command will run
+            # after the user unblocks the bot, and the start function will add the user to
+            # "user_ids".  We're including this here for the sake of the example.
```
harshil21, line 85:
```suggestion
            # "user_ids". We're including this here for the sake of the example.
```

examples/chatmemberbot.py
lemontree210: I used the variable name from a different function in the same example where it made more sense. Fixed.

examples/chatmemberbot.py
lemontree210: Oh, I just confused the slashes at first, thought I had to put a backslash, and it required a raw string. Then remembered I was wrong but forgot to remove the raw string :)
Done

examples/chatmemberbot.py
lemontree210: PEP8 [states](https://peps.python.org/pep-0008/#comments):

> You should use two spaces after a sentence-ending period in multi- sentence comments, except after the final sentence.

harshil21: wow, that's insane. TIL

Bibo-Joshi: WTF

examples/chatmemberbot.py
```diff
@@ -80,7 +80,10 @@ async def track_chats(update: Update, context: ContextTypes.DEFAULT_TYPE) -> Non
     chat = update.effective_chat
     if chat.type == Chat.PRIVATE:
         if not was_member and is_member:
-            logger.info("%s started the bot", cause_name)
+            # This is not really needed in this particular case because /start command will run
```
Bibo-Joshi, line 83: This comment is somewhat confusing IMO. If the user starts the bot, but then blocks & unblocks it, we do get the `my_chat_member` update about the unblocking and hence need this check, right?

examples/chatmemberbot.py
```diff
@@ -136,11 +139,29 @@ async def greet_chat_members(update: Update, context: ContextTypes.DEFAULT_TYPE)
         )
 
 
+async def start(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    """Greets the user and records that they started a chat with the bot
+    if it's a private chat. Since no automatic update is issued when a user starts a private chat
```
Bibo-Joshi, line 144:
```suggestion
    if it's a private chat. Since no `my_chat_member` update is issued when a user starts a private chat
```

Bibo-Joshi (commented):
`/start` is the default first text recommended by the official clients, but I'm rather  sure that you can bypass that. So maybe just a `MessageHandler(filters.ALL, …)` that checks if  the user id has been recorded yet?

Poolitzer:
@Bibo-Joshi yes you can send a different first message

examples/chatmemberbot.py
lemontree210: When I press "unblock" in my Telegram App, it unblocks the bot and automatically sends the `/start` command. The CommandHandler catches it, adds the user to the list and logs "user started the bot". 

So after unblocking I get two messages in the logger: user unblocked the bot and user started the bot.

examples/chatmemberbot.py
lemontree210: Done

examples/chatmemberbot.py
```diff
@@ -136,6 +147,22 @@ async def greet_chat_members(update: Update, context: ContextTypes.DEFAULT_TYPE)
         )
 
 
+async def start_private_chat(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    """Greets the user and records that they started a chat with the bot if it's a private chat.
+    Since no `my_chat_member` update is issued when a user starts a private chat with the bot,
+    we have to track it explicitly here, when handling the /start command.
```
Bibo-Joshi, line 153:
```suggestion
    Since no `my_chat_member` update is issued when a user starts a private chat with the bot for the first time,
    we have to track it explicitly here.
```

examples/chatmemberbot.py
```diff
@@ -136,6 +147,22 @@ async def greet_chat_members(update: Update, context: ContextTypes.DEFAULT_TYPE)
...
+    """
+    user_name = update.effective_user.full_name
+    chat = update.effective_chat
+    if chat.type == Chat.PRIVATE:
```
Bibo-Joshi, line 157:
```suggestion
    if chat.type != Chat.PRIVATE or `chat_id` not in context.bot_data["user_ids"]:
        return
```

Tracking the user and sending the `Welcome {user_name}` message only makes sense if the user is not already tracked :)

Bibo-Joshi (changes requested):

examples/chatmemberbot.py
lemontree210: Done

examples/chatmemberbot.py
lemontree210: I think you meant `or "chat_id" in context.bot_data["user_ids"]` (not `not`)?

lemontree210: Anyway, if it's the first ever interaction with the bot, `context.bot_data` has no key "user_ids" and it triggers `KeyError`.

lemontree210: This works, but it's kinda ugly :)

```py
    if (
        chat.type != Chat.PRIVATE 
        or ("user_ids" in context.bot_data and chat.id in context.bot_data["user_ids"])
    ):
        return
```

harshil21: just use `chat.id in context.bot_data.get("user_ids", [])`

lemontree210: bingo. I knew that it was `.get()` but somehow didn't figure out how to word that exactly

Issue #3496, opened 2023-01-10
[DOCUMENTATION] Add new admonitions to telegram classes


harshil21:
### What kind of feature are you missing? Where do you notice a shortcoming of PTB?

I think the docs can do with more explicit types of admonitions. I have a few suggestions:

1. "Returned in":  When viewing a documentation of a class, sometimes I forget which bot method or object through which this class is available.

<details>
<summary> Examples </summary>
<br>

``` python
class ChatInviteLink:
    """
    .. admonition:: Returned In
        :class: returned-in

        :meth:`Bot.create_chat_invite_link`, ...
    """
```

</details>

2. "Available in": Similar to the above, but this would include where and how this object can be found. Motivation: Our docs are currently top-down, i.e. Update -> Message -> Chat, so if I landed in "Update" I can figure out how to get the Chat object, but not vice-versa.  Icon suggestions: [Box-fill](https://icons.getbootstrap.com/icons/box-fill/), [Boxes](https://icons.getbootstrap.com/icons/boxes/), [Geo fill](https://icons.getbootstrap.com/icons/geo-fill/) (liked more by me)

<details>
<summary> Examples </summary>
<br>

``` python
class Chat:
    """
    .. admonition:: Available in
        :class: available-in

        :attr:`telegram.Message.chat`, ...
    """

class Message:
    """
    .. admonition:: Available in
        :class: available-in

        :attr:`telegram.Update.message`, ...
    """
```

</details>

3. "Use in": This would be used in classes which need to be instantiated and then passed into bot methods parameters.
Icon suggestions: [Arrow left](https://icons.getbootstrap.com/icons/arrow-left/), [Paperclip](https://icons.getbootstrap.com/icons/paperclip/), [Funnel](https://icons.getbootstrap.com/icons/funnel/) (liked more by me), [Stack](https://icons.getbootstrap.com/icons/stack/)

<details>
<summary> Examples </summary>
<br>

``` python
class ChatPermissions:
    """
    .. admonition:: Used in
        :class: used-in

        :meth:`Bot.set_chat_permissions`, ...
    """
```

</details>

4. "Shortcuts": This would be used in the bot methods instead of the See Also currently used.

### Describe the solution you'd like

These admonitions are better than "see also" (in most cases) since it tells the user what and why it was linked in the first place.

### Describe alternatives you've considered

see alsos

### Additional context

How to make custom admonitions: https://pradyunsg.me/furo/reference/admonitions/

[...]

PR #3501, opened 2023-01-13
Add admonitions

lemontree210:
addresses #3496

1. "Returned in" admonitions are generated automatically
2. "Available in" admonitions are generated automatically
3. "Use in" admonitions are generated automatically
4. `..seealso` notes were separated and included manually because they contained both shortcuts and links to Wiki


harshil21 (commented):
Nice job. The docs as artifact is already helping.

Can we also make a new icon for this? I have two in mind:

1. Speedometer - https://icons.getbootstrap.com/icons/speedometer/
2. Signpost - https://icons.getbootstrap.com/icons/signpost/

lemontree210:
>     2. Signpost - https://icons.getbootstrap.com/icons/signpost/

Done (signpost)

docs/source/_static/style_admonitions.css
```diff
@@ -0,0 +1,29 @@
...
+}
+.admonition.returned-in > .admonition-title {
+  background-color: rgba(177, 108, 51, 0.1);
+  border-color: rgb(230, 109, 15);
+}
+.admonition.returned-in > .admonition-title::before {
+  background-color: rgb(230, 109, 15);
+  -webkit-mask-image: var(--icon--returned-in);
+  mask-image: var(--icon--returned-in);
+}
```
Bibo-Joshi, line 29: missing new line at end of file :)

docs/source/conf.py
```diff
@@ -441,13 +452,67 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+    return admonition_for_class_name
+
+
+RETURN_ADMONITION_FOR_CLASS_NAME = create_return_admonitions()
+
+
 def autodoc_process_docstring(
     app: Sphinx, what, name: str, obj: object, options, lines: List[str]
 ):
     """We do two things:
```
Bibo-Joshi, line 509:
```suggestion
    """We do three things:
```

;P

docs/source/conf.py
```diff
@@ -441,13 +452,67 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
     )
 
 
+def create_return_admonitions() -> dict[str, str]:
+    """Creates 'Returned in' admonitions for classes that are returned in Bot's methods."""
+
+    methods_for_class = defaultdict(list)
+
+    # checking for islower() to avoid camelCase methods
+    for method in [m for m in dir(telegram.Bot) if not m.startswith("_") and m.islower()]:
```
Bibo-Joshi, line 461: maybe `inspect.getmembers(telegram.Bot, inspect.is_method)` is a bit better? 🤔 Should avoid the try-except below …

docs/source/conf.py
```diff
@@ -441,13 +452,67 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+        try:
+            sig = inspect.Signature.from_callable(getattr(telegram.Bot, method))
```
Bibo-Joshi, line 463:
```suggestion
            sig = inspect.signature(getattr(telegram.Bot, method))
```

that's actually just [a shortcut](https://github.com/python/cpython/blob/ba88628808af14ec904a9afe3b72dca9159d7ce8/Lib/inspect.py#L3277), but at least it's shorter :D

docs/source/conf.py
```diff
@@ -441,13 +452,67 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+
+        class_name = str(sig.return_annotation)
+        if "telegram" not in class_name:
+            continue
+
+        class_name = (
+            class_name.replace("typing.Union[", "")
+            .replace("typing.Tuple[", "")
+            .replace("]", "")
+            .split(", ")[0]  # Union and Tuple give relevant class before the comma
```
Bibo-Joshi, line 475: Union *may* give several relevant classes though, right? e.g. `Union[ChatMemberAdmin, ChatMemberUser]`. I guess we currently only have `Union[something, None]`, but it may still be good if the code can already handle that. just in case …

docs/source/conf.py
```diff
@@ -478,7 +543,29 @@ def autodoc_process_docstring(
...
+        class_name = name.split(".")[-2]
+    except IndexError:
+        class_name = ""
+
+    # check for "__init__" ensures we generate "Returned in" admonition only once per class
+    if (
+        class_name in RETURN_ADMONITION_FOR_CLASS_NAME
+        and name.endswith("__init__")
+        and "slot wrapper" not in str(obj)  # filter out some side effects of __init__ check
+    ):
```
Bibo-Joshi, line 557: can't we just check if `.. admonition:: Returned in` is already somewhere in `lines` 🤔

Bibo-Joshi (changes requested):
Nice work indeed! 🚀 I left a few smaller comments below.

I would like it very much if we could make the bullet point list collapsed by default if there's more than, say, 5 items. for `Message`, the list is really rather long … I'd be okay if that's done in a follow-up PR.

docs/source/_static/style_admonitions.css
lemontree210: I know but all of our CSS files are like that

docs/source/conf.py
lemontree210: Done

docs/source/conf.py
lemontree210: Done

docs/source/_static/style_admonitions.css
Bibo-Joshi: doesn't mean it's good 😆

docs/source/conf.py
lemontree210: Then it tries to insert the admonition e.g. in a docstring of `de_json()` and fails

docs/source/conf.py
lemontree210: ... or `Union[Something, bool]`. But I see your point

docs/source/conf.py
Bibo-Joshi: I see. the `what` argument probably also doesn't help?

docs/source/_static/style_admonitions.css
lemontree210: Done

docs/source/conf.py
lemontree210: Done

docs/source/conf.py
lemontree210: Done

lemontree210:
> I would like it very much if we could make the bullet point list collapsed by default if there's more than, say, 5 items. 

using [this extension](https://sphinx-toolbox.readthedocs.io/en/stable/extensions/collapse.html)?

docs/source/conf.py
lemontree210: no, the `what` is "method" for `de_json()` as well, so I can't see a way to tell them apart

harshil21:
> using [this extension](https://sphinx-toolbox.readthedocs.io/en/stable/extensions/collapse.html)?

I would prefer doing:
``` css
.admonition.returned-in {
    border-color: rgb(230, 109, 15);
    max-height: 200px;
    overflow-y: scroll;
}
```
looks much more aesthetic imo

lemontree210:
> ```css
> .admonition.returned-in {
>     border-color: rgb(230, 109, 15);
>     max-height: 200px;
>     overflow-y: scroll;
> }
> ```

It does look great, but the problem is that the user doesn't expect they are expected to scroll, and the scrolling bar is almost invisible. If I'm new to the docs, I'll just see 6 items in `Message` and think that's it. I could add a note like ("scroll down to see all methods") to the "Returned in" heading if there are more than 6 items (that's what 200px are enough for at 100% scale).

harshil21:
image

with a dropdown I feel they'll just scroll past it, and would still take the whole page when opened. Plus from my end, the overflow of the last item in the image and scrollbar sticking out on the side is enough imo

docs/source/conf.py
```diff
@@ -441,13 +452,68 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+        )
+        for class_name in class_names:
+            if "telegram" not in class_name:
+                continue
+            class_name = class_name.split(".")[-1]  # telegram._botcommand.BotCommand -> BotCommand
+            methods_for_class[class_name].append(method)
+
+    admonition_for_class_name = dict()
+
+    for cls in methods_for_class:
```
harshil21, line 486:
```suggestion
    for cls, methods in methods_for_class.items():
```

lemontree210:
I added a change in cursor when hovering over the list, so that should provide additional clue.

docs/source/conf.py
```diff
@@ -441,13 +452,68 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+            .removesuffix("'>")
+            .removesuffix("]")
+            .split(", ")  # for the cases like Union[telegram.Something, telegram.SomethingElse]
+        )
+        for class_name in class_names:
+            if "telegram" not in class_name:
+                continue
+            class_name = class_name.split(".")[-1]  # telegram._botcommand.BotCommand -> BotCommand
+            methods_for_class[class_name].append(method)
+
```
harshil21, line 483:
```suggestion
    # Now, let's use this mapping to start generating a new mapping of class names to admonitions, i.e. {Message: ".. admonition: Returned in ..."}
```

docs/source/conf.py
```diff
@@ -441,13 +452,68 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+    admonition_for_class_name = dict()
+
+    for cls in methods_for_class:
+        admonition = """
+
+.. admonition:: Returned in
+    :class: returned-in
+"""
```
harshil21, line 491:
```suggestion
"""
    # Add the method names as points
```

docs/source/conf.py
harshil21: > no, the `what` is "method" for `de_json()` as well, so I can't see a way to tell them apart

but shouldn't we be checking for "class"?

harshil21 (commented):
Amazing change, the docs somehow just keep getting better.

lemontree210:
> but shouldn't we be checking for "class"?

@harshil21 I don't think I ever get the "class" in this function, so I hooked up on `__init__`s. Maybe there was something I was missing. It works this way, but I know it isn't very pretty

docs/source/conf.py
```diff
@@ -441,13 +452,68 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+            continue
+
+        class_names = (
+            return_type_string.removeprefix("typing.Union[")
+            .removeprefix("typing.Tuple[")
+            .removeprefix("<class '")
+            .removesuffix("'>")
+            .removesuffix("]")
+            .split(", ")  # for the cases like Union[telegram.Something, telegram.SomethingElse]
+        )
```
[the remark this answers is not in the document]
harshil21: actually let's make it an exact match instead of getting the `__name__` (i.e. `str(telegram_obj)`), and then in `autodoc_process_docstring`, we can do `class_name = str(obj)` (if `what == "class"`). 

We'll get full matches this way: 
``` python
>>> a = typing.get_args(typing.Union[telegram.BotCommand, True])
>>> str(a[0])
"<class 'telegram._botcommand.BotCommand'>"
>>> str(telegram.BotCommand)
"<class 'telegram._botcommand.BotCommand'>"
```

docs/source/conf.py
```diff
@@ -441,13 +452,68 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+def create_return_admonitions() -> dict[str, str]:
+    """Creates 'Returned in' admonitions for classes that are returned in Bot's methods."""
+
+    methods_for_class = defaultdict(list)
+
+    for method in [
+        m[0]
+        for m in inspect.getmembers(telegram.Bot, inspect.isfunction)  # not .ismethod
+        if not m[0].startswith("_") and m[0].islower()  # islower() to avoid camelCase methods
+    ]:
```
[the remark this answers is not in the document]
Bibo-Joshi: I'd prefer not to mix docs & tests. The duplication should be minimal …

docs/source/_static/style_admonitions.css
```diff
@@ -0,0 +1,34 @@
...
+  border-color: rgb(230, 109, 15);
+  max-height: 200px;
+  overflow-y: scroll;
+}
+.admonition.returned-in:hover {
+  cursor: move;
+}
+.admonition.returned-in > .admonition-title {
+  background-color: rgba(177, 108, 51, 0.1);
+  border-color: rgb(230, 109, 15);
```
Bibo-Joshi, line 28:
```suggestion
  border-color: rgb(230, 109, 15);
  position: sticky;
  top: 0;
  backdrop-filter: blur(10px);
```

Another minor suggestion: That way the header doesn't scroll away. The text gets blurred when passing underneath it. Would like it even better if it would just vanish, but with the transparency of the title color and my limited CSS knowledge, I couldn't manage that xD

docs/source/conf.py
lemontree210: Done

docs/source/conf.py
```diff
@@ -441,13 +452,68 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
     )
 
 
+def create_return_admonitions() -> dict[str, str]:
+    """Creates 'Returned in' admonitions for classes that are returned in Bot's methods."""
+
```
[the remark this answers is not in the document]
lemontree210: Done

docs/source/conf.py
```diff
@@ -441,13 +452,68 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+            .removesuffix("]")
+            .split(", ")  # for the cases like Union[telegram.Something, telegram.SomethingElse]
+        )
+        for class_name in class_names:
+            if "telegram" not in class_name:
+                continue
+            class_name = class_name.split(".")[-1]  # telegram._botcommand.BotCommand -> BotCommand
+            methods_for_class[class_name].append(method)
+
+    admonition_for_class_name = dict()
```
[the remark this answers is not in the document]
lemontree210: Done

docs/source/conf.py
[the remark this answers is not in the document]
lemontree210: Very good, wouldn't have figured it out myself.
Done

docs/source/conf.py
lemontree210: Done

docs/source/conf.py
lemontree210: of course
Done

docs/source/_static/style_admonitions.css
lemontree210: The blurring works in Chrome but doesn't work in Firefox on my Ubuntu. We could make the heading opaque, but then its colors must depend on whether the theme is light or dark

Bibo-Joshi: https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter#browser_compatibility

okay, firefox 103 was released in [2022-07](https://en.wikipedia.org/wiki/Firefox_version_history#Firefox_102_through_114), so fairly recent … All other browser support it since ~3 more years though … I guess in FF the blurring is just not applied and the text slides through the background unchanged?

lemontree210: I think I figured out another way :) See 7e30f91297a2bfc3a9bb6e1a9c17173510b794a8

docs/source/conf.py
```diff
@@ -441,13 +452,61 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+    methods_for_class_name = defaultdict(list)
+
+    for method in [
+        m[0]
+        for m in inspect.getmembers(telegram.Bot, inspect.isfunction)  # not .ismethod
+        if not m[0].startswith("_") and m[0].islower()  # islower() to avoid camelCase methods
+    ]:
+        sig = inspect.signature(getattr(telegram.Bot, method))
+        classes = get_args(sig.return_annotation)
+
```
harshil21, line 469:
```suggestion
        sig = inspect.signature(getattr(telegram.Bot, method))
        ret_anot = sig.return_annotation
        classes = get_args(ret_anot)

        if not classes and isinstance(ret_anot, type) and "telegram" in str(ret_anot):
            methods_for_class_name[str(ret_anot)].append(method)
            continue
```

Found something I didn't expect with get_args. If you do `get_args(telegram.Message)`, it returns `()`. So we can do this workaround to fix it.

harshil21 (commented):
Found a tiny bug

docs/source/conf.py
lemontree210: Done

lemontree210:
I noticed that `CallbackContext.bot` is not shown in "Available in" admonition for `Bot` because it has `-> BT` instead of `-> Telegram.Bot`:

```py
    @property
    def bot(self) -> BT:
        """:class:`telegram.Bot`: The bot associated with this context."""
        return self._application.bot
```

lemontree210:
> I noticed that `CallbackContext.bot` is not shown in "Available in" admonition for `Bot` because it has `-> BT` instead of `-> Telegram.Bot`

Nevermind, that's not the reason. The reason is that `CallbackContext.bot` is a property, not an attribute, and it is hence not documented as attribute in the docstring.

I guess I have to inspect properties of classes as well.

docs/source/conf.py
```diff
@@ -441,15 +453,212 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+    # to instances of a given class
+    # i.e. {"telegram.Sticker": ["telegram.Message.sticker", ...]}
+    attrs_for_class_name = defaultdict(list)
+    admonition_for_class_name = {}
+
+    # The colon after the closing parenthesis is important because it makes sure
+    # that the class is mentioned in the attribute description, not in free text.
+    # So far there seem to be no attribute docstrings in which the list of classes
+    # takes up more than one line.
+    attr_docstr_pattern = re.compile(r"^\s*(?P<attr_name>[a-z_]+)\s?\(.*:class:`.+`.*\):\s.*$")
```
Bibo-Joshi, line 478: can you use the `re.VERBOSE` & explain the pattern a bit more? 😬

docs/source/conf.py
```diff
@@ -441,15 +453,212 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+    ):
+        # We need to make "<class 'telegram._files.sticker.StickerSet'>" into "telegram.StickerSet"
+        # because that's the way the classes are mentioned in docstrings...
+        name_of_inspected_class_in_docstr = (
+            str(inspected_class).removeprefix("<class '").removesuffix("'>")
+        )
+        name_of_inspected_class_in_docstr = ".".join(
+            n
+            for n in name_of_inspected_class_in_docstr.split(".")
+            if n in ("telegram", "ext") or not n.islower()  # removing things like "_files.sticker"
```
Bibo-Joshi, line 494: maybe put the comment above line 491?

telegram/_chatadministratorrights.py
```diff
@@ -32,13 +32,12 @@ class ChatAdministratorRights(TelegramObject):
     :attr:`can_post_messages`, :attr:`can_edit_messages`, :attr:`can_pin_messages`,
     :attr:`can_manage_topics` are equal.
 
+    .. seealso:: :meth:`Bot.set_my_default_administrator_rights`
```
Bibo-Joshi, line 35: why did you drop `get_my_…`?

telegram/_keyboardbutton.py
```diff
@@ -58,10 +58,10 @@ class KeyboardButton(TelegramObject):
             sent as a contact when the button is pressed. Available in private chats only.
         request_location (:obj:`bool`, optional): If :obj:`True`, the user's current location will
             be sent when the button is pressed. Available in private chats only.
-        request_poll (:class:`KeyboardButtonPollType`, optional): If specified, the user will be
-            asked to create a poll and send it to the bot when the button is pressed. Available in
-            private chats only.
-        web_app (:class:`WebAppInfo`, optional): If specified, the described `Web App
+        request_poll (:class:`~telegram.KeyboardButtonPollType`, optional): If specified, the user
```
Bibo-Joshi, line 61: is this necessary for the automatic detection to work? If so: I dare say that we'll at some point forget about this and it would be nice if the code could handle it 😬 I guess `getattr(telegram, "KeyboardButtonPollType")` should work in this case?

telegram/_passport/credentials.py
```diff
@@ -114,14 +114,14 @@ class EncryptedCredentials(TelegramObject):
...
-        data (:class:`telegram.Credentials` or :obj:`str`): Decrypted data with unique user's
+        data (:class:`telegram.Credentials` | :obj:`str`): Decrypted data with unique user's
             nonce, data hashes and secrets used for EncryptedPassportElement decryption and
             authentication or base64 encrypted data.
         hash (:obj:`str`): Base64-encoded data hash for data authentication.
         secret (:obj:`str`): Decrypted or encrypted secret used for decryption.
 
     Attributes:
-        data (:class:`telegram.Credentials` or :obj:`str`): Decrypted data with unique user's
+        data (:class:`telegram.Credentials` | :obj:`str`): Decrypted data with unique user's
```
Bibo-Joshi, line 124: similar to my above comment

Bibo-Joshi (commented):
Really nice code I'm seeing here 👌

telegram/_chatadministratorrights.py
lemontree210: because it's in the autogenerated "Returned in" admonition

docs/source/conf.py
```diff
@@ -632,6 +632,78 @@ def create_return_admonitions() -> dict[str, str]:
...
+
+    # First, generate a mapping of class names to the methods which accept them as arguments,
+    # i.e. {InlineQueryResult: [answer_inline_query, ...]}
+    methods_for_class_name = defaultdict(set)  # using set here because there can be repetitions
+
+    forward_ref_pattern = re.compile(r"ForwardRef\('(?P<class_name>.+?)'\)")  # non-greedy
+    skip_pattern = re.compile(
+        r"^typing.Union\[ForwardRef\('DefaultValue\[DVType]'\), ((float)|(bool)|(int)|(str))"
+        r"(, NoneType)?]$"
+    )
```
Bibo-Joshi, line 648: more verbose explanations of the patterns would again be nice - I fear that in two weeks nobody will know what these do 😄

docs/source/conf.py
```diff
@@ -632,6 +632,78 @@ def create_return_admonitions() -> dict[str, str]:
...
+
+    for method in [
+        m[0]
+        for m in inspect.getmembers(telegram.Bot, inspect.isfunction)  # not .ismethod
+        if not m[0].startswith("_")
+        and m[0].islower()  # islower() to avoid camelCase methods
+        and m[0] in telegram.Bot.__dict__  # method is not inherited from TelegramObject
+    ]:
```
Bibo-Joshi, line 656: This is the same as in the "returned in" admonitions right? Can we compute that once and use it both functions?

docs/source/conf.py
```diff
@@ -632,6 +632,78 @@ def create_return_admonitions() -> dict[str, str]:
...
+
+        sig = inspect.signature(getattr(telegram.Bot, method))
```
Bibo-Joshi, line 658: maybe even including the signatures?

docs/source/conf.py
```diff
@@ -632,6 +632,78 @@ def create_return_admonitions() -> dict[str, str]:
...
+            if isinstance(annotation, type) and "telegram" in annotation_text:
+                methods_for_class_name[str(annotation)].add(method)
+
+            # this is a _UnionGenericAlias
+            elif "telegram" in annotation_text or "ForwardRef" in annotation_text:
+                for match in forward_ref_pattern.finditer(annotation_text):
+                    # resolving class
+                    try:
+                        klass = eval(f"telegram.{match.group('class_name')}")
+                    except AttributeError:
```
Bibo-Joshi, line 677: Can you add a CSI comment on what the try-except does?

docs/source/conf.py
```diff
@@ -632,6 +632,78 @@ def create_return_admonitions() -> dict[str, str]:
...
+        if len(methods) > 1:
+            for method_name in sorted(list(methods)):
+                admonition += "\n    * " + f":meth:`telegram.Bot.{method_name}`"
+        else:
+            admonition += f"\n    :meth:`telegram.Bot.{list(methods)[0]}`"
+
+        admonition += "\n    "  # otherwise an unexpected unindent warning will be issued
+        admonition_for_class_name[cls] = admonition
+
+    return admonition_for_class_name
```
Bibo-Joshi, line 701: this is pretty much the same in "returned in", "used in", and "available in". Can you extract that to a new function?

Bibo-Joshi (commented):
One general suggestion: It would be cool, if `Used in` would also consider subclasses - e.g. `BotCommandScope*` should also point to `set_my_commands`, not only the `BotCommandScope` base class. For that, it should suffice to compute the parent classes of each class (excluding `TelegramObject`) and accessing the mapping entries for all parent classes

docs/source/conf.py
```diff
@@ -441,15 +453,289 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+                name_of_class_in_attr = match.group("class_name")
+                if "telegram" not in name_of_class_in_attr:
+                    continue
+
+                # Writing to dictionary: matching the class found in the docstring to the
+                # attribute of the class being inspected.
+                # The class in the attribute docstring is the key, and the attribute of the
+                # class currently being inspected is the value.
+                attrs_for_class_name[build_full_class_name(name_of_class_in_attr)].append(
+                    f":attr:`{name_of_inspected_class_in_docstr}.{target_attr}`"
```
harshil21, line 521: what happens if the entire name is linked, i.e. `` :attr:`telegram._files.sticker.StickerSet` ``, does it link in the docs? Does it show as that or as `telegram.StickerSet`?

docs/source/conf.py
```diff
@@ -441,15 +453,289 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+                matches = [m for m in forward_ref_pattern.finditer(annotation_text)] + [
+                    m for m in telegram_pattern.finditer(annotation_text)
+                ]
+
+                for match in matches:
+                    # resolving class
+                    try:
+                        klass = eval(f"telegram.{match.group('class_name')}")
+                    except AttributeError:
+                        klass = eval(f"telegram.ext.{match.group('class_name')}")
```
harshil21, line 683: does this handle cases like `telegram.ext.filters.Sticker` as well?

harshil21 (commented):
Great work. Dare I say we might need tests for these now 😬

lemontree210:
I think I'll start by reworking the "Use in" but to parse annotations, then refactor to avoid code repetition, then address what will be left of the comments.

Would you mind typing in here (or in the issue) a couple of examples of different types that must work for all 3 types of admonitions (with regard to our more recent discussions about subclasses and the bugs you had noticed that something wasn't shown)? So that I don't write here that everything's ready, then you go and check and find something is missing. I obviously don't mean a big list and it won't save me from all potential bugs, just a couple of examples of different kinds.

docs/source/conf.py
```diff
@@ -478,7 +639,13 @@ def autodoc_process_docstring(
...
             )
 
-    # 2) Get the file names & line numbers
+    # 2-3) Insert "Returned in" and "Available in" admonitions into classes (where applicable)
+    if what == "class":
+        class_name = str(obj)
+        for dict_ in (AVAILABLE_IN_ADMONITION_FOR_CLASS_NAME, RETURN_ADMONITION_FOR_CLASS_NAME):
+            insert_admonition(dict_)
+
+    # 3) Get the file names & line numbers
```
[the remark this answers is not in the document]
lemontree210: actually 5) :) 
Done

harshil21:
Here are some items which can be improved still:

#### Copy-pasting admonitions in subclasses:

Here's 2 examples of this (but there are more):
- [ ] `telegram.ChatMember`'s Available in and Returned in should be present in all `ChatMember` subclasses.
- [ ] All `telegram.InlineQueryResult` subclasses should have the same use in's as the parent class.

#### a) Available in:

- [ ] [`telegram.PersonalDetails`, `telegram.IdDocumentData`, `telegram.ResidentialAddress`] should show it's available in `telegram.EncryptedPassportElement.data`.
- [ ] `telegram.PassportElementErrorUnspecified` should show its available in `telegram.EncryptedPassportElement.hash`
 
#### b) Use in:

- [ ] `InputMedia` should have Use in linking to `Bot.edit_message_media`. Its subclasses also must have the same use in to `Bot.edit_message_media`.
- [ ] `telegram.InlineQueryResult` should have another use in linking to `Bot.answer_web_app_query`
- [ ] The link to the bot methods in `telegram.Location`, `telegram.MaskPosition`, `telegram.BotCommandScope` (there are more) is broken (it only shows the text)

If anyone find's more, feel free to update this list

docs/source/conf.py
lemontree210: Done

docs/source/conf.py
lemontree210: Done

docs/source/conf.py
lemontree210: The second one is irrelevant (I removed it). I'll add a small note on the first one

docs/source/conf.py
lemontree210: Done

docs/source/conf.py
lemontree210: Done

telegram/_passport/credentials.py
lemontree210: I didn't change it because my pattern doesn't capture it (it does). It was just inconsistent and maybe even wrong, so I corrected it :)

telegram/_keyboardbutton.py
lemontree210: I changed the procedure slightly to make sure the "telegram" word is not needed. I'll probably leave the docstrings as I changed them, though, because this way they're consistent with args.

docs/source/conf.py
lemontree210: I don't think that'll work in the current version because I need both the method name and its signature:

```py
            for method in relevant_methods_for_class:
                method_link = self._generate_link_to_method(method, klass)
                sig = inspect.signature(getattr(klass, method))
```

docs/source/conf.py
lemontree210: Done

docs/source/conf.py
```diff
@@ -441,15 +453,289 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+    ):
+        # We need to make "<class 'telegram._files.sticker.StickerSet'>" into "telegram.StickerSet"
+        # because that's the way the classes are mentioned in docstrings...
+        name_of_inspected_class_in_docstr = (
+            str(inspected_class).removeprefix("<class '").removesuffix("'>")
+        )
+        name_of_inspected_class_in_docstr = ".".join(
+            n
+            for n in name_of_inspected_class_in_docstr.split(".")
+            if n in ("telegram", "ext") or not n.islower()  # removing things like "_files.sticker"
```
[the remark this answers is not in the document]
lemontree210: I'm not sure. A regex will look more concise, but here the intent and the steps seem to be very clear: remove suffix&prefix, then remove the middle.

docs/source/conf.py
lemontree210: The method for resolving the full class name (in this older version it was a function called `build_full_class_name()`) first tries to evaluate the entire name:
```py 
for option in (name, f"telegram.{name}", f"telegram.ext.{name}"):
    try:
        return str(eval(option))
```
So it does work if the entire name is linked.

As for how it shows - I guess it is Sphinx's behavior: all full class names that I return from the aforementioned method look like this:
```
<class 'telegram._keyboardbuttonpolltype.KeyboardButtonPollType'>
```
And when I generate the links, Sphinx shows then in rendered docs as `telegram.KeyboardButtonPollType`

docs/source/conf.py
lemontree210: Well it didn't, I just added this functionality. In practice, though, docstrings with `filters` use full class name. I doubt you will put `:class: Sticker` if you mean `filters.Sticker`. But this addition is reasonable anyway (I could imagine someone putting `:class: BaseFilter` into the docstring).

lemontree210:
In my current state of code it generates "Use In" and "Returned In" admonitions for `ApplicationBuilder` in `ApplicationBuilder` itself, because its method return itself. I kind of think that it's good this way because someone who is not familiar with it may not be prepared to this method chaining mechanism. But if you think that's no good, I'll exclude `ApplicationBuilder` itself from classes that should be linked to `ApplicationBuilder`.

lemontree210:
>  `telegram.PassportElementErrorUnspecified` should show its available in `telegram.EncryptedPassportElement.hash`

I disagree with this one because `.hash` is: 

```
(:obj:`str`): Base64-encoded element hash for using in   :class:`telegram.PassportElementErrorUnspecified`.
```

so it's a string. We can't say an instance of the class is used in this attribute.

harshil21:
@lemontree210 ah, my bad

lemontree210:
> * `telegram.ChatMember`'s Available in and Returned in should be present in all `ChatMember` subclasses.

I think it gets tricky. Do all subclasses of `telegram.ChatMember` must contain same "Returned In" admonitions? For me it doesn't make sense that `telegram.ChatMemberBanned` would contain "Returned in: Bot.get_chat_administrators()"

harshil21:
> For me it doesn't make sense that `telegram.ChatMemberBanned` would contain "Returned in: Bot.get_chat_administrators()"

hmm, I would be okay with an exclude list/dict, but I want to see what others think as well, cause Telegram could change that overnight and our docs would be wrong then and we wouldn't really find out

lemontree210:
>     * [x]  [`telegram.PersonalDetails`, `telegram.IdDocumentData`, `telegram.ResidentialAddress`] should show it's available in `telegram.EncryptedPassportElement.data`.

this had nothing to do with subclasses, it was a tricky regex thing with a multi-line attribute typing string. Fixed, thanks for the catch.

Bibo-Joshi:
@harshil21 I&#x27;d also prefer false positives over false negatives here :)

telegram/_poll.py
```diff
@@ -146,7 +146,7 @@ class Poll(TelegramObject):
         id (:obj:`str`): Unique poll identifier.
         question (:obj:`str`): Poll question, :tg-const:`telegram.Poll.MIN_QUESTION_LENGTH`-
             :tg-const:`telegram.Poll.MAX_QUESTION_LENGTH` characters.
-        options (Sequence[:class:`PollOption`]): List of poll options.
+        options (Sequence[:class:`~telegram.PollOption`]): List of poll options.
```
Bibo-Joshi, line 149: just double checking: These changes are not strictly needed anymore, right? I'm okay with keeping them to avoid having to revert though :)

docs/source/conf.py
```diff
@@ -441,15 +446,552 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+
+    # In some cases it is useful that some admonitions are displayed not only for base classes
+    # but also for their subclasses.
+    CLASSES_WHOSE_SUBCLASSES_NEED_ADMONITIONS = (
+        telegram.BotCommandScope,
+        telegram.InlineQueryResult,
+        telegram.InputMedia,
+        telegram.MenuButton,
+        telegram.request.BaseRequest,
+    )
```
Bibo-Joshi, line 466: Maybe I missed part of the discussion … Can't this be automated or were there arguments against doing that? Keeping this list updated is one more thing to remember …

harshil21 (changes requested):
The code is much easier to read and well formatted. 👍🏽

Some newer issues I found:

- [x] `telegram.Credentials`'s admonition is not inserted in the right place.
- [ ] `telegram.InputMedia{Video, Photo, Document}` should have another use in to send_media_group (related to regex not handling new lines?)
- [ ] `telegram.MenuButton` subclasses doesn't have the parent class' Returned in. Was this intentional?
- [ ] `telegram.ext.Application`'s Use in shows for the `AppBuilder.post_*` methods. That's not supposed to happen?
- [ ] `telegram.Bot`, `telegram.ext.CallbackContext` has a Returned in linking to `AppBuilder.build()`.
- [x] Missing Use in for `JobQueue` -> `AppBuilder.job_queue()`?

> But if you think that's no good, I'll exclude `ApplicationBuilder` itself from classes that should be linked to `ApplicationBuilder`

not a fan of it tbh.

docs/source/conf.py
```diff
@@ -441,15 +446,552 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+        "available_in": {<class 'telegram._chatinvitelink.ChatInviteLink'>: 
+        <"Available in" admonition">, ...},
+        "returned_in": {...}
+        }
+        ```
+        """
+
+    def insert_admonitions_for_class(
+        self,
+        klass: type,
```
Bibo-Joshi, line 515:
```suggestion
        cls: type,
```

that's the widely used name for variables of type `type` :D more an FYI, I'm fine with keeping as is

Bibo-Joshi (commented):
The class-refactor makes it really nice to read 👍 Looks well organized and clean.
TBH I skipped the implementation details this time and only left a bit of nitpicking. Harshil apparently provides you with the edge cases to cover, so I think the two of you got it from here 🙂

telegram/_poll.py
lemontree210: That is correct

lemontree210:
>     `telegram.MenuButton` subclasses doesn't have the parent class' Returned in. Was this intentional?

Yes, it was. I introduced admonitions for subclasses for "Use in" admonitions only. Maybe I don't understand the logic of the Bot method correctly: can it really return a subclass of the `telegram.MenuButton`? As far as I can tell, it can't:

```py
return MenuButton.de_json(result, bot=self)
```

docs/source/conf.py
```diff
@@ -441,15 +446,552 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+
+                    elif isinstance(annotation, typing.TypeVar):
+                        # gets access to the "bound=..." parameter
+                        resolve_and_add(arg=annotation.__bound__)
+
+                    elif typing.get_origin(annotation) in (
+                        dict,
+                        type,
+                        collections.Sequence,
+                        collections.Callable,
```
Bibo-Joshi, line 732:
```suggestion
                        collections.abc.Sequence,
                        collections.abc.Callable,
```

those have been moved for quite a while and `collections.{Sequence, Callbale}` have been removed in 3.9 or so IIRC

lemontree210:
> `telegram.InputMedia{Video, Photo, Document}` should have another use in to `send_media_group` (related to regex not handling new lines?)

Subclasses of `InputMedia` do have this in their "Use in" admonitions. `InputMedia` does not (because I process subclasses, not parents).

I can also add processing of parents, this will make the logic even more complex, I guess.

docs/source/conf.py
Bibo-Joshi: Clarification: I'd like to have the "link all subclasses wherever the base class is mentioned" logic applied to all classes. Unless there are specific reasons not to do that.

lemontree210:
>  `telegram.Credentials`'s admonition is not inserted in the right place.

Fixed. It has no args, didn't expect that.

docs/source/conf.py
lemontree210: I can see that for "Use in" admonitions. Maybe for "Available in", not sure. But for "Return in"? See [this comment](https://github.com/python-telegram-bot/python-telegram-bot/pull/3501#issuecomment-1398898882).

docs/source/conf.py
```diff
@@ -478,7 +1020,15 @@ def autodoc_process_docstring(
                 ),
             )
 
-    # 2) Get the file names & line numbers
+    # 2-4) Insert "Returned in", "Available in", "Use in" admonitions into classes
+    # (where applicable)
+    if what == "class":
+        AdmonitionInserter().insert_admonitions_for_class(
```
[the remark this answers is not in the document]
lemontree210: yes, a very good idea. Done.

docs/source/conf.py
```diff
@@ -441,15 +446,552 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+            # The ^ colon above along with parenthesis is important because it makes sure that
+            # the class is mentioned in the attribute description, not in free text.
+            r".*$",  # Any number of characters, end of string (end of line)
+            re.VERBOSE,
+        )
+
+        # for properties: there is no attr name in docstring.  Just check if there's a class name.
+        prop_docst_pattern = re.compile(r":class:`.+`.*:")
+
+        single_class_name_pattern = re.compile(r":class:`~?(?P<class_name>[\w.]*)`")
```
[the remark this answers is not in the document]
lemontree210: I thought the name was self-explanatory :) will add

docs/source/conf.py
```diff
@@ -441,15 +446,552 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+                for n in name_of_inspected_class_in_docstr.split(".")
+                if n in ("telegram", "ext") or not n.islower()
+            )
+
+            # Parsing part of the docstring with attributes (parsing of properties follows later)
+            docstring_lines = inspect.getdoc(inspected_class).splitlines()
+            lines_with_attrs = []
+            for idx, line in enumerate(docstring_lines):
+                if line.strip() == "Attributes:":
+                    lines_with_attrs = docstring_lines[idx + 1 :]
```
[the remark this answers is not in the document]
lemontree210: yes, of course. Done.

docs/source/conf.py
```diff
@@ -441,15 +446,552 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+            raise NotImplementedError(f"Base class {base_class} not supported")
+
+    def _resolve_arg_and_add_link_to_method(
+        self,
+        arg: Any,
+        dict_of_methods_for_class: defaultdict,
+        link_to_method: str,
+        include_subclasses_if_applicable: bool = False,
+    ) -> None:
+        """A helper method for "Returned in" and "Use in admonition.
```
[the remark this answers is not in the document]
lemontree210: Done

docs/source/conf.py
lemontree210: no you're right. I used `klass` because they were all `str`. Then I figured out I could use `type` and now `cls` makes sense.
Done.

docs/source/conf.py
lemontree210: I guess `Coroutine` as well then. Done.
(It works on Python 3.9 but still)

docs/source/conf.py
```diff
@@ -441,15 +446,552 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+
+                    elif typing.get_origin(annotation) in (
+                        dict,
+                        type,
+                        collections.Sequence,
+                        collections.Callable,
+                        Union,
+                    ):
+                        for arg in typing.get_args(annotation):
+                            resolve_and_add(arg=arg)
```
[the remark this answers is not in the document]
lemontree210: Thanks a lot for spotting this. That was before I added some recursion.
Done.

lemontree210:
>  `telegram.ext.Application`'s Use in shows for the `AppBuilder.post_*` methods. That's not supposed to happen?

Does it? When I render the docs, `Application` has only one Use In:
`telegram.ext.ApplicationBuilder.application_class()`

harshil21:
> Does it? When I render the docs, `Application` has only one Use In: `telegram.ext.ApplicationBuilder.application_class()`

yeah, I downloaded docs as artifact (that's python 3.9) and built docs locally (on python 3.11) with same result. Maybe you forgot to push a commit or something?

lemontree210:
>  `telegram.Bot`, `telegram.ext.CallbackContext` has a Returned in linking to `AppBuilder.build()`.

That's an interesting one. Here's the code:
```py
    def build(
        self: "ApplicationBuilder[BT, CCT, UD, CD, BD, JQ]",
    ) -> Application[BT, CCT, UD, CD, BD, JQ]:
```
My methods try to resolve the args and they do so recursively, so it thinks that `Application` contains more args and resolves each of them, adding "Returned in" to each of them. Why does this return annotation contain all these classes anyway?

docs/source/conf.py
```diff
@@ -441,15 +446,552 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+
+    # We cannot move this code to a separate module within docs because there is no way
+    # to import it.  This class is simply a way of encapsulating the code and providing a better
+    # overview of conf.py.
+
+    ADMONITION_TYPES = ("use_in", "available_in", "returned_in")
+
+    # In some cases it is useful that some admonitions are displayed not only for base classes
+    # but also for their subclasses.
+    CLASSES_WHOSE_SUBCLASSES_NEED_ADMONITIONS = (
```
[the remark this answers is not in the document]
harshil21: with hinrich in favour of including the `ChatMember` case, we can drop this entirely, i.e. add admonitions for all subclasses

docs/source/conf.py
harshil21: >See [this comment](https://github.com/python-telegram-bot/python-telegram-bot/pull/3501#issuecomment-1398898882).

check dev chat

lemontree210:
> Missing Use in for `JobQueue` -> `AppBuilder.job_queue()`?

Thank you, helped me uncover a bug. That recursion within loops can be very confusing...

lemontree210:
> > Does it? When I render the docs, `Application` has only one Use In: `telegram.ext.ApplicationBuilder.application_class()`
> 
> yeah, I downloaded docs as artifact (that's python 3.9) and built docs locally (on python 3.11) with same result. Maybe you forgot to push a commit or something?

This is so strange! I fetched and merged `origin/doc-fixes` to be sure, re-built the docs and still this is the only thing I can see for `Application`:

Screenshot from 2023-01-21 13-19-23

docs/source/conf.py
```diff
@@ -899,8 +899,8 @@ def _resolve_arg(self, arg: Any) -> Union[type, None]:
             dict,
             tuple,
             Union,
-            collections.Coroutine,
-            collections.Sequence,
+            collections.abc.Coroutine,
+            collections.abc.Sequence,
             typing.Sequence,
```
harshil21, line 882: let's use collections.abc.Sequence here as well since this is deprecated in py 3.9 (which is what the docs build out of)

docs/source/conf.py
```diff
@@ -441,15 +446,552 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+            # docstrings
+            name_of_inspected_class_in_docstr = (
+                str(inspected_class).removeprefix("<class '").removesuffix("'>")
+            )
+            # removing things like "_files.sticker"
+            name_of_inspected_class_in_docstr = ".".join(
+                n
+                for n in name_of_inspected_class_in_docstr.split(".")
+                if n in ("telegram", "ext") or not n.islower()
+            )
```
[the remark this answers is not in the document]
lemontree210: Thanks! Done

docs/source/conf.py
lemontree210: Sorry, I didn't get that :)

I now just `import collections.abc` and then use things like `collections.abc.Coroutine` in code.

docs/source/conf.py
lemontree210: Done

docs/source/conf.py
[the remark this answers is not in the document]
lemontree210: Done

docs/source/conf.py
```diff
@@ -441,15 +446,552 @@ def check_timeout_and_api_kwargs_presence(obj: object) -> int:
...
+    ADMONITION_TYPES = ("use_in", "available_in", "returned_in")
+
+    # In some cases it is useful that some admonitions are displayed not only for base classes
+    # but also for their subclasses.
+    CLASSES_WHOSE_SUBCLASSES_NEED_ADMONITIONS = (
+        telegram.BotCommandScope,
+        telegram.InlineQueryResult,
+        telegram.InputMedia,
+        telegram.MenuButton,
+        telegram.request.BaseRequest,
```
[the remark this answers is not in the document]
lemontree210: Done (removed include list)

docs/source/conf.py
harshil21: my bad, I meant that there is `typing.Sequence` and `collections.abc.Sequence` in that list. Do we need both of them?

lemontree210: Well in practice probably not because we don't have any args that are `typing.Sequence`. But if there were any, then `typing.Sequence` and `collections.abc.Sequence` wouldn't be the same.

docs/auxil/admonition_inserter.py
```diff
@@ -0,0 +1,507 @@
...
+        elif isinstance(arg, type):
+            if "telegram" in str(arg):
+                yield arg
+
+        # for custom generics like telegram.ext._application.Application[~BT, ~CCT, ~UD...]
+        elif str(type(arg) == "<class 'typing._GenericAlias'>"):
+            if "telegram" in str(arg):
+                # get_origin() of telegram.ext._application.Application[~BT, ~CCT, ~UD...]
+                # will produce <class 'telegram.ext._application.Application'>
+                yield typing.get_origin(arg)
```
harshil21, line 455:
```suggestion
                yield origin
```
?

harshil21:
Found some more missing links, I think this is the last of it:

Use-in:

- [ ]  `telegram.ext.{Defaults, ContextTypes, {Base, Pickle, Dict}Persistence, {Base, AIO}RateLimiter`-> `telegram.ext.ApplicationBuilder`

Available-in:

- [ ] `ChatMember` subclasses do not have the Available in that `ChatMember` has, Returned in is copied correctly.

docs/auxil/admonition_inserter.py
```diff
@@ -0,0 +1,507 @@
...
+            # class denoted by `name` actually belongs to `telegram.ext`:
+            # "module 'telegram' has no attribute 'ApplicationBuilder'".
+            # If neither option works, this is not a PTB class.
+            except (NameError, AttributeError):
+                continue
+
+
+if __name__ == "__main__":
+    # just try instantiating for debugging purposes
+    AdmonitionInserter()
```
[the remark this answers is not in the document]
lemontree210: It's very convenient for me to run, though, without wasting time rebuilding the docs

docs/auxil/admonition_inserter.py
lemontree210: Done

docs/auxil/admonition_inserter.py
```diff
@@ -0,0 +1,507 @@
...
+            origin in (collections.abc.Callable, typing.IO)
+            or arg is None
+            # no other check available (by type or origin) for these:
+            or str(type(arg)) in ("<class 'typing._SpecialForm'>", "<class 'ellipsis'>")
+        ):
+            pass
+
+        # RECURSIVE CALLS
+        # for cases like Union[Sequence....
+        elif typing.get_origin(arg) in (
```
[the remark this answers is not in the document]
lemontree210: Done

docs/auxil/admonition_inserter.py
```diff
@@ -0,0 +1,507 @@
...
+            type,
+            Union,
+            collections.abc.Coroutine,
+            collections.abc.Sequence,
+            typing.Sequence,
+        ):
+            for sub_arg in typing.get_args(arg):
+                yield from self._resolve_arg(sub_arg)
+
+        elif isinstance(arg, list):
```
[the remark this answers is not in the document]
lemontree210: for some reason `get_origin()` of a list returns `None`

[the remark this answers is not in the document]
harshil21: > for some reason `get_origin()` of a list returns `None`

hmm, well I found the same behaviour on tuples and dicts then. on py3.11:

``` python
import typing
a = typing.get_origin([1, 2, 3])  # None
b = typing.get_origin((1, 2, 3))  # None
c = typing.get_origin({1: 2})  # None
```

docs/auxil/admonition_inserter.py
```diff
@@ -194,13 +196,16 @@ def _create_available_in(self) -> dict[type, str]:
...
-                    # The class in the property docstring is the key, and the property of the
-                    # class currently being inspected is the value.
-                    attrs_for_class[self._resolve_class(name_of_class_in_prop)].add(
-                        f":attr:`{name_of_inspected_class_in_docstr}.{prop_name}`"
-                    )
+                    # Writing to dictionary: matching the class found in the docstring and its
+                    # subclasses to the property of the class being inspected.
+                    # The class in the property docstring (or its subclass) is the key,
+                    # and the property of the class currently being inspected is the value.
+                    cls = self._resolve_class(name_of_class_in_prop)
```
harshil21: shouldn't we check for `None` here? That's what was done in the attribute section (`if resolved_class is None ..`)

docs/auxil/admonition_inserter.py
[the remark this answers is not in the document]
lemontree210: I think I'll just do the opposite thing: I'll remove the types and origins that are practically unused and create a more informative exception message so that if someone creates a new argument and Sphinx fails, the user would know exactly what happened.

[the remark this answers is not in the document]
harshil21: okay, so the docs passing means those conditions weren't actually being hit before, and was intended as a failsafe.

> more informative exception message

nice idea

harshil21 (approved):
Phew, I think I'm happy with all the changes. This feature makes it almost impossible to get lost in the docs!

PR #3503, opened 2023-01-13
Issue warning in `run_daily` only if `days` parameter was passed

harshil21:
As discussed in private chat, this was done to reduce users getting confused on why they got this warning if they never used the days parameter in the first place.

### Checklist for PRs

### DIFF
--- a/telegram/ext/_jobqueue.py
+++ b/telegram/ext/_jobqueue.py
@@ -461,10 +461,6 @@ class JobQueue:
                 .. versionchanged:: 20.0
                     Changed day of the week mapping of 0-6 from monday-sunday to sunday-saturday.
 
-                .. versionchanged:: 20.1
-                    The warning about the change of day of week mapping is now only shown if this
-                    parameter was passed.
-
             data (:obj:`object`, optional): Additional data needed for the callback function.
                 Can be accessed through :attr:`Job.data` in the callback. Defaults to
                 :obj:`None`.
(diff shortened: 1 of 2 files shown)


telegram/ext/_jobqueue.py
```diff
@@ -460,6 +460,11 @@ async def callback(context: CallbackContext)
 
                 .. versionchanged:: 20.0
                     Changed day of the week mapping of 0-6 from monday-sunday to sunday-saturday.
+
+                .. versionchanged:: 20.1
+                    The warning about the change of day of week mapping is now only shown if this
+                    parameter was passed.
+
```
Bibo-Joshi, line 467: Is that worth a versionchanged? The warning isn't documented behavior …

telegram/ext/_jobqueue.py
```diff
@@ -460,6 +460,11 @@ async def callback(context: CallbackContext)
 
                 .. versionchanged:: 20.0
                     Changed day of the week mapping of 0-6 from monday-sunday to sunday-saturday.
+
+                .. versionchanged:: 20.1
+                    The warning about the change of day of week mapping is now only shown if this
+                    parameter was passed.
```
Bibo-Joshi, line 466:
```suggestion
                    the job does not run every day.
```

We don't actually check if you manually pass `tuple(range(7))` :)

Bibo-Joshi (commented):
Just commenting on the docstring change, code LGTM :)

telegram/ext/_jobqueue.py
harshil21: don't know.. guess I'll remove

Issue #3367, opened 2022-11-16
[FEATURE] Typings: switch to explicit optional types


Viicos:
### What kind of feature are you missing? Where do you notice a shortcoming of PTB?

Since [mypy 0.990](https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html), implicit optional types are disabled by default. Maybe it is a good time to switch to explicit optional types, if so I can work on this.

Future reference: https://github.com/hauntsaninja/no_implicit_optional

### Describe the solution you'd like

Switch to explicit optional types (e.g. `def foo(s: str | None = None)`)

Bibo-Joshi:
Thanks for reaching out. Are there any advantages of switching to explicit optional types other than complying with PEP 484?

harshil21:
> Thanks for reaching out. Are there any advantages of switching to explicit optional types other than complying with PEP 484?

It would work slightly better with pylance/pyright for one...

Bibo-Joshi:
I checked with harshil again and it seems that changing this in our library does not make a difference when __using__ PTB with pylance. I doubt that there will be any changes for users using other type checkers.
While the "explicit is better than implicit" argument is ofc somewhat right, I don't see a compelling benefit of this change. We can just re-enable the implicit optional mode once needed.

I vote -1 on this.

Viicos:
Alright that's understandable, you might want to update the pre-commit config to use `no_implicit_optional = False`, as iirc the default value have changed (not an issue for now as mypy is still on 0.982).

Bibo-Joshi:
our pre-commit-ci integration will inform us about that once the dependency is updated :)

I'll close this issue. Still, thanks again for taking the time and making the suggestion, @Viicos !

PR #3508, opened 2023-01-19
More explicit type annotations

Bibo-Joshi:
Basically a long-time result of https://github.com/microsoft/pylance-release/issues/1983: Explicity adds type annotations to attributes of classes.
While I still don't really like that there is a (partial) need for this, it's mostly a one-time effort.
Almost all additions where ofc automated. It's "only" type hints, so if I got something wrong here, it doesn't have runtime effect, but a "slow-scrolling-quick-read" review would still be good at some point :D

Todo:

- [x] check with `pyright --verifytypes telegram` what that tool thinks about the type completeness of our library and check what we can improve about it
- [x] think about adding a step to the CI that ensures that type completeness only ever increases
- [x] check in vscode if the types are better now


Bibo-Joshi:
Small update: Apparently pyright still thinks that PTb is not very type complete:

```json
{
        "exportedSymbolCounts": {
            "withKnownType": 415,
            "withAmbiguousType": 30,
            "withUnknownType": 294
        },
        "otherSymbolCounts": {
            "withKnownType": 1815,
            "withAmbiguousType": 702,
            "withUnknownType": 566
        },
        "missingFunctionDocStringCount": 25,
        "missingClassDocStringCount": 0,
        "missingDefaultParamCount": 107,
        "completenessScore": 0.5615696887686062,
}
```
[Full json output](https://github.com/python-telegram-bot/python-telegram-bot/files/10461188/pyright.zip).

Bibo-Joshi:
Update:

```json
{
        "exportedSymbolCounts": {
            "withKnownType": 426,
            "withAmbiguousType": 277,
            "withUnknownType": 36
        },
        "otherSymbolCounts": {
            "withKnownType": 2292,
            "withAmbiguousType": 671,
            "withUnknownType": 120
        },
        "missingFunctionDocStringCount": 25,
        "missingClassDocStringCount": 0,
        "missingDefaultParamCount": 107,
        "completenessScore": 0.5764546684709067,
}
```

So at least a rather large jump from "unknown" to "ambiguous - even though the completeness increased only ~1.5 points 😄
I found out that as soon as *anything* in a class A is unknown/ambiguous (annotation of param/return value/attr, base class), that whole class A is of unknown/ambiguous type and using this class A in some other class B as type annotation will hence also make B unknown/ambiguous.

Also I apparently messed up the tests big time

Bibo-Joshi:
```json
{
        "exportedSymbolCounts": {
            "withKnownType": 705,
            "withAmbiguousType": 0,
            "withUnknownType": 34
        },
        "otherSymbolCounts": {
            "withKnownType": 3180,
            "withAmbiguousType": 0,
            "withUnknownType": 102
        },
        "missingFunctionDocStringCount": 23,
        "missingClassDocStringCount": 0,
        "missingDefaultParamCount": 147,
        "completenessScore": 0.9539918809201624,
}
```

😎

Bibo-Joshi:
Last update for today messe up the tests, but got to 96%:

```
Symbols exported by "telegram": 734
  With known type: 707
  With ambiguous type: 0
  With unknown type: 27
    (Ignoring unknown types imported from other packages)
  Functions without docstring: 13
  Functions without default param: 0
  Classes without docstring: 0

Other symbols referenced but not exported by "telegram": 2496
  With known type: 2408
  With ambiguous type: 1
  With unknown type: 87

Type completeness score: 96.3%
```

IISC most of what's left boils down to the generic nature of `Application` and `CallbackContext`. I'll see how far I'll get with that …

Bibo-Joshi:
The GH Actions thingy doesn't quite work yet and sphinx is acting up while documenting `ContextTypes.DEFAULT_TYPES`. but type completeness is at 100% now 🥳 
If anyone's using VSCode + pylance, I'd be happy to know if this improves autocompletion etc noticable :)

harshil21:
Okay so I tested it and it's just as good as before, except now the autocompletion of `context.{user, chat,bot}_data` works better. Also sadly it still shows None for something like `update.message.poll.id`, though I guess we can't fix that without a TypeGuard(?). 

Anyway running pyright on my machine still found some unknown types, sending a pastebin in dev chat

Bibo-Joshi:
@harshil21 Nice :)
I doubt that <code>TypeGuard</code> can help with that. To know in which handlers <code>update.message.poll</code> is <code>None</code> and in which it&#x27;s a <code>Message</code>, we&#x27;d have to make <code>Update</code> and <code>Message</code> generic classes <i>and</i> couple that with the handlers &amp; filters - (nearly) impossible

which Python version are did you run pyright on? did you use the <code>--ignoreexternal</code> flag?

telegram/ext/filters.py
```diff
@@ -172,10 +173,6 @@ class variable.
             return a dict with lists. The dict will be merged with
             :class:`telegram.ext.CallbackContext`'s internal dict in most cases
             (depends on the handler).
-
-    Attributes:
-        name (:obj:`str`): Name for this filter.
-        data_filter (:obj:`bool`): Whether this filter is a data filter.
```
harshil21, line 178: can you revert this change (and the same in `{Message, Update}Filter`) since these attributes are now being repeated in classes like `filters.Command`.

The `:inherited-members:` option [was added ](https://github.com/python-telegram-bot/python-telegram-bot/pull/2759#discussion_r744147138) since I wanted to document the members of `_ChatUserBaseFilter` such as `(add,remove)_usernames` in its subclasses (`filters.{Chat, User, etc}`) without duplicating them.

Alternatively, it maybe possible to modify `autodoc_skip_member` to exclude this in other classes I think

harshil21 (commented):
Is it worth to also modify bot method parameters to `Optional[...]` as well? Even the newer mypy version complains of implicit optionals now

Bibo-Joshi:
> Is it worth to also modify bot method parameters to Optional[...] as well? Even the newer mypy version complains of implicit optionals now

You mean e.g. `parse_mode: Optional[str] = None` instead of `parse_mode: str = None`? Does mypy copmlain about that when you _use_ PTB or only when developing PTB?

telegram/_bot.py
```diff
@@ -114,7 +114,7 @@
 BT = TypeVar("BT", bound="Bot")
 
 
-class Bot(TelegramObject, AbstractAsyncContextManager):
+class Bot(TelegramObject, AsyncContextManager["Bot"]):
```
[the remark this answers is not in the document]
Bibo-Joshi: it's not generic in py3.7 & py3.8, same as `list` and `dict` etc. [PEP 585](https://peps.python.org/pep-0585/) is new in py3.9.

harshil21:
> You mean e.g. `parse_mode: Optional[str] = None` instead of `parse_mode: str = None`?

yes 

> Does mypy copmlain about that when you _use_ PTB or only when developing PTB?

pylance complains about that while developing PTB. I mentioned mypy since they made a similar change with implicit_optional

telegram/ext/filters.py
Bibo-Joshi: I excluded `{Base, Massage, Update}Filter` from the `:inherited-members:`, that does the trick :)

Bibo-Joshi:
> pylance complains about that while developing PTB.

So this changed? https://github.com/python-telegram-bot/python-telegram-bot/issues/3367#issuecomment-1318837905

If it has: Fair enough, let's do it. But I vote to do that in a different PR - we can reopen #3367 for that.

telegram/ext/filters.py
harshil21: though now those 2 properties aren't documented for `Message, UpdateFilter`.. is that okay with you?

Bibo-Joshi: Yes. We list the base classes, that should be enough. We don't normalle repeat all the parents attributes/methods/… :)

harshil21 (approved):

PR #3514, opened 2023-01-21
add check if keyboard has more than 2 dimensions

### LINKED ISSUE #3511: Don't allow more than 2 dimensional input for `Inline/ReplyKeyboardMarkup`

When a user passes their list to `Inline/ReplyKeyboardMarkup`, we currently check if their input is valid (exactly 2 dimensional input is valid), i.e. 

* the input is a sequence, not just a single button
* it's not just a simple sequence (sequence of buttons)

However it doesn't check if it's more than 2D, i.e. `[[[KeyboardButton(...)], [...], ]]` is invalid.

Modify `tg._utils.markup.check_keyboard_type` to return `False` if we find another sequence.

_No response_

### Additional context

_No response_

clot27:
Closes #3511

### DIFF
--- a/telegram/_utils/markup.py
+++ b/telegram/_utils/markup.py
@@ -31,7 +31,7 @@ from collections.abc import Sequence
 
 
 def check_keyboard_type(keyboard: object) -> bool:
-    """Checks if the keyboard provided is of the correct type - A list of lists.
+    """Checks if the keyboard provided is of the correct type - A sequence of sequences.
     Implicitly tested in the init-tests of `{Inline, Reply}KeyboardMarkup`
     """
     # string and bytes may actually be used for ReplyKeyboardMarkup in which case each button
@@ -39,6 +39,11 @@ def check_keyboard_type(keyboard: object) -> bool:
     # allow it here.
     if not isinstance(keyboard, Sequence) or isinstance(keyboard, (str, bytes)):
         return False
+
+    # check whether the keyboard has more than two dimensions.
+    if isinstance(keyboard[0][0], Sequence) and not isinstance(keyboard[0][0], (str, bytes)):
+        return False
+
     for row in keyboard:
         if not isinstance(row, Sequence) or isinstance(row, (str, bytes)):
             return False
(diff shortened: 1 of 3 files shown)


telegram/_utils/markup.py
```diff
@@ -31,14 +31,19 @@ class ``telegram.ReplyMarkup``.
...
     Implicitly tested in the init-tests of `{Inline, Reply}KeyboardMarkup`
     """
     # string and bytes may actually be used for ReplyKeyboardMarkup in which case each button
     # would contain a single character. But that use case should be discouraged and we don't
     # allow it here.
     if not isinstance(keyboard, Sequence) or isinstance(keyboard, (str, bytes)):
         return False
+
+    # check whether the keyboard has more than two dimensions.
+    if isinstance(keyboard[0][0], Sequence) and not isinstance(keyboard[0][0], (str, bytes)):
```
Bibo-Joshi, line 44: At this point we haven't yet established that the rows are sequences as well. `check_keyboard_type([1])` would raise an exception with this addition IISC …

telegram/_utils/markup.py
```diff
@@ -31,14 +31,19 @@ class ``telegram.ReplyMarkup``.
...
 
 def check_keyboard_type(keyboard: object) -> bool:
-    """Checks if the keyboard provided is of the correct type - A list of lists.
+    """Checks if the keyboard provided is of the correct type - A sequence of sequences.
     Implicitly tested in the init-tests of `{Inline, Reply}KeyboardMarkup`
     """
     # string and bytes may actually be used for ReplyKeyboardMarkup in which case each button
     # would contain a single character. But that use case should be discouraged and we don't
     # allow it here.
     if not isinstance(keyboard, Sequence) or isinstance(keyboard, (str, bytes)):
```
harshil21, line 40: make this particular condition a new function, because we will be using it three times

harshil21 (approved):

Bibo-Joshi:
Thanks for the contribution :)

PR #3515, opened 2023-01-22
Clean up `docs/source/conf.py`

Bibo-Joshi:
Moves the different functions used for custom scripting to the new `docs/auxil` directly. This

* doesn't clutter `conf.py` as much
* groups the code by functionality, making it easier to maintain


docs/auxil/admonition_inserter.py
```diff
@@ -1,3 +1,20 @@
...
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser Public License
+#  along with this program.  If not, see [http://www.gnu.org/licenses/].
 import collections.abc
 import inspect
 import re
```
lemontree210, line 20: Somewhere in this file I mention `autodoc_process_docstring`. I guess that comment can be changed so that the user knows where to look for it.

lemontree210 (commented):
My small general concern is that all modules are in one dir, while actually only half of them are imported in `conf.py` while the rest is used by those three. Maybe they should me moved to a subdir.

docs/source/conf.py
```diff
@@ -291,344 +283,16 @@
...
-                parts = parts[:index] + parts[-1:]
-                break
-
-        # Replace private base classes with their respective parent
-        parts = [PRIVATE_BASE_CLASSES.get(part, part) for part in parts]
-
-        base = ".".join(parts)
 
-        bases[idx] = f":class:`{base}`"
+# Not used but must be imported for the linkcode extension to find it
```
harshil21, line 288: maybe also add a comment that the sphinx logger was imported in the other files in docs.auxil so not needed to reimport

docs/auxil/sphinx_logging.py
```diff
@@ -0,0 +1,22 @@
...
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser Public License
+#  along with this program.  If not, see [http://www.gnu.org/licenses/].
+from sphinx.util import logging
+
+# get the sphinx(!) logger
+# Makes sure logs render in red and also plays nicely with e.g. the `nitpicky` option.
+sphinx_logger = logging.getLogger(__name__)
```
harshil21, line 22: do we really need a new file just to import the logger?

tests/test_applicationbuilder.py
```diff
@@ -305,7 +306,7 @@ def test_all_bot_args_custom(self, builder, bot, monkeypatch):
         # other means that the parameters are passed correctly
 
         assert built_bot.token == bot.token
-        assert built_bot.base_url == "base_url" + bot.token
+        assert docs.auxil.link_code.base_url == "base_url" + bot.token
```
harshil21, line 309: why not just use `built_bot`? how is this different?

.pre-commit-config.yaml
```diff
@@ -68,7 +68,7 @@ repos:
     rev: v3.3.1
     hooks:
     -   id: pyupgrade
-        files: ^(telegram|examples|tests)/.*\.py$
+        files: ^(telegram|examples|tests|docs)/.*\.py$
         args:
           - --py37-plus
```
harshil21, line 73: our docs should be python 3.9+ compatible. So the args passed here is wrong?

docs/source/conf.py
Bibo-Joshi: What's benefit in mentioning that here? If we don't need a logger here anyway, why comment on it ... :)

docs/auxil/sphinx_logging.py
Bibo-Joshi: Yeah, maybe it's Overkill and I can just get the logger where needed ...

tests/test_applicationbuilder.py
Bibo-Joshi: Irg, probably a search&replace mistake ...

.pre-commit-config.yaml
Bibo-Joshi: 3.7 code is still compatible with 3.9 though - just not voce versa

harshil21: ah right

PR #3523, opened 2023-01-30
Add tests for admonitions

lemontree210:
adds tests for functionality introduced in #3501 that resolved #3496 

the tests are meant to only run in GH actions before building docs. They are not included in the main test suite, one of the reasons being that Python 3.9+ is required to run the admonition generator.

PR #3529, opened 2023-02-03
Autogenerate "Shortcuts" admonitions

### LINKED ISSUE #3496: [DOCUMENTATION] Add new admonitions to telegram classes

lemontree210:
closes #3496 (last part of it; see also #3501 and #3523 that address the same issue): automatically generates "Shortcuts" admonitions for `Bot` methods and removes manual ones.

### DIFF
--- a/docs/auxil/admonition_inserter.py
+++ b/docs/auxil/admonition_inserter.py
@@ -86,54 +86,27 @@ class AdmonitionInserter:
         ```
         """
 
-    def insert_admonitions_for_class(
+    def insert_admonitions(
         self,
-        cls: type,
+        obj: Union[type, collections.abc.Callable],
         docstring_lines: list[str],
     ):
-        """Inserts admonitions into docstring lines for a given class.
+        """Inserts admonitions into docstring lines for a given class or method.
 
         **Modifies lines in place**.
         """
         # A better way would be to copy the lines and return them, but that will not work with
         # docs.auxil.sphinx_hooks.autodoc_process_docstring()
 
-        for admonition_type in self.CLASS_ADMONITION_TYPES:
+        for admonition_type in self.ALL_ADMONITION_TYPES:
 
-            # If there is no admonition of the given type for the given class,
-            # continue to the next admonition type, maybe the class is listed there.
-            if cls not in self.admonitions[admonition_type]:
+            # If there is no admonition of the given type for the given class or method,
+            # continue to the next admonition type, maybe the class/method is listed there.
+            if obj not in self.admonitions[admonition_type]:
                 continue
 
-            # TODO repetition?
             insert_idx = self._find_insert_pos_for_admonition(docstring_lines)
-            admonition_lines = self.admonitions[admonition_type][cls].splitlines()
-
-            for idx in range(insert_idx, insert_idx + len(admonition_lines)):
-                docstring_lines.insert(idx, admonition_lines[idx - insert_idx])
-
-    def insert_admonitions_for_bot_method(
-        self,
-        method: collections.abc.Callable,
-        docstring_lines: list[str],
-    ):
-        """Inserts admonitions into docstring lines for a given Bot method.
-
-        **Modifies lines in place**.
-        """
-        # A better way would be to copy the lines and return them, but that will not work with
-        # docs.auxil.sphinx_hooks.autodoc_process_docstring()
-
-        for admonition_type in self.METHOD_ADMONITION_TYPES:
-
-            # If there is no admonition of the given type for the given class,
-            # continue to the next admonition type, maybe the class is listed there.
-            if method not in self.admonitions[admonition_type]:
-                continue
-
-            # TODO repetition?
-            insert_idx = self._find_insert_pos_for_admonition(docstring_lines)
-            admonition_lines = self.admonitions[admonition_type][method].splitlines()
+            admonition_lines = self.admonitions[admonition_type][obj].splitlines()
 
             for idx in range(insert_idx, insert_idx + len(admonition_lines)):
                 docstring_lines.insert(idx, admonition_lines[idx - insert_idx])
@@ -184,9 +157,7 @@ class AdmonitionInserter:
             # We need to make "<class 'telegram._files.sticker.StickerSet'>" into
             # "telegram.StickerSet" because that's the way the classes are mentioned in
             # docstrings.
-            # Check for potential presence of ".ext.", we will need to keep it.
-            ext = ".ext" if ".ext." in str(inspected_class) else ""  # TODO repetition?
-            name_of_inspected_class_in_docstr = f"telegram{ext}.{class_name}"
+            name_of_inspected_class_in_docstr = self._generate_class_name_for_link(inspected_class)
 
             # Parsing part of the docstring with attributes (parsing of properties follows later)
             docstring_lines = inspect.getdoc(inspected_class).splitlines()
@@ -279,12 +250,12 @@ class AdmonitionInserter:
         methods_for_class = defaultdict(set)
 
         for cls, methods in self.METHODS_FOR_BOT_AND_APPBUILDER.items():
-            for method in methods:
+            for method_name in methods:
 
-                sig = inspect.signature(getattr(cls, method))
+                sig = inspect.signature(getattr(cls, method_name))
                 ret_annot = sig.return_annotation
 
-                method_link = self._generate_link_to_method(method, cls)
+                method_link = self._generate_link_to_method(method_name, cls)
 
                 try:
                     self._resolve_arg_and_add_link(
@@ -295,7 +266,7 @@ class AdmonitionInserter:
                 except NotImplementedError as e:
                     raise NotImplementedError(
                         f"Error generating Sphinx 'Returned in' admonition "
-                        f"(admonition_inserter.py). {cls}, method {method}. "
+                        f"(admonition_inserter.py). {cls}, method {method_name}. "
                         f"Couldn't resolve type hint in return annotation {ret_annot}. {str(e)}"
                     )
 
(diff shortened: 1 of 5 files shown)


docs/auxil/admonition_inserter.py
```diff
@@ -265,12 +266,54 @@ def _create_returned_in(self) -> dict[type, str]:
...
+        """Creates a dictionary with 'Shortcuts' admonitions for Bot methods that
+        have shortcuts in other classes.
+        """
+
+        # pattern for looking for calls to Bot methods only
+        bot_method_pattern = re.compile(
+            r"""\s*  # any number of whitespaces
+            (?<=return\sawait\sself\.get_bot\(\)\.)  # lookbehind
+            .+  # the method name we are looking for
+            (?=\() # lookahead: closing bracket
```
Bibo-Joshi, line 285: I'm not overly familiar with lookahead, so just double checking: If I were to type `return await self.get_bot().one().two()`, would this match `one` ore `one().two`?

docs/auxil/admonition_inserter.py
```diff
@@ -265,12 +266,54 @@ def _create_returned_in(self) -> dict[type, str]:
...
+            re.VERBOSE,
+        )
+
+        # Generate a mapping of methods of classes to links to Bot methods which they are shortcuts
+        # for, i.e. {<function Bot.send_voice at ...>: {:meth:`telegram.User.send_voice`, ...}
+        shortcuts_for_bot_method = defaultdict(set)
+
+        # inspect methods of all telegram classes for return statements that indicate
+        # that this given method is a shortcut for a Bot method
+        for class_name, cls in inspect.getmembers(telegram, predicate=inspect.isclass):
```
Bibo-Joshi, line 296: if the class is `Bot`, we can `continue`

docs/auxil/admonition_inserter.py
```diff
@@ -265,12 +266,54 @@ def _create_returned_in(self) -> dict[type, str]:
...
+
+            for method_name, method in inspect.getmembers(cls, predicate=inspect.isfunction):
```
Bibo-Joshi, line 298: we can exclude protected & private methods

docs/auxil/admonition_inserter.py
```diff
@@ -265,12 +266,54 @@ def _create_returned_in(self) -> dict[type, str]:
...
+
+                # .getsourcelines() returns a tuple. Item [1] is an int (length of line?)
+                relevant_return_lines = [
+                    line
+                    for line in inspect.getsourcelines(method)[0]
+                    if bot_method_pattern.search(line)
+                ]
+
+                for line in relevant_return_lines:
+                    bot_method_name = bot_method_pattern.search(line).group()
```
Bibo-Joshi, line 308: why do we need to `search(line)` twice? and what happens with things like

```python
return await (
    selt.get_bot()
    .some_long_method_name(…)
)
```
Does the regex catch these multiline things?

docs/auxil/admonition_inserter.py
```diff
@@ -265,12 +266,54 @@ def _create_returned_in(self) -> dict[type, str]:
...
+                    bot_method = eval(f"telegram.Bot.{bot_method_name}")
```
Bibo-Joshi, line 309: does `getattr(telegram.Bot, bot_method_name)` do the trick? I like to avoid `eval` when possible …

Bibo-Joshi (commented):
Thanks for the PR! Looks good overall :)

docs/auxil/admonition_inserter.py
lemontree210: First of all, thanks for helping me spot a typo in the last line: it's actually an **opening** bracket. 

This regex will match `one`.

lemontree210: The usual addition of `?` makes it non-greedy. This will match `one`:

```py
bot_method_pattern = re.compile(
    r"""\s*  # any number of whitespaces
    (?<=return\sawait\sself\.get_bot\(\)\.)  # lookbehind
    .+?  # NON-GREEDY
    (?=\() # lookahead: opening bracket before the args of the method start
    """,
    re.VERBOSE,
)
```

lemontree210: Alternatively, I could be looking for only letters and underscores:

```py
bot_method_pattern = re.compile(
    r"""\s*  # any number of whitespaces
    (?<=return\sawait\sself\.get_bot\(\)\.)  # lookbehind
    \w+  # the method name we are looking for
    (?=\() # lookahead: opening bracket before the args of the method start
    """,
    re.VERBOSE,
)
```

lemontree210: None of this is really relevant in our case because it's always a hanging open bracket after the `Bot` method, but maybe I should replace `.+` with `\w+` to make it match only letters and underscores.

docs/auxil/admonition_inserter.py
lemontree210: No it doesn't catch them but we don't have such things in the code :)

Maybe even our pre-commit hooks wouldn't allow this :)

docs/auxil/admonition_inserter.py
lemontree210: Absolutely. I wonder why I hadn't thought about it.
Done

docs/auxil/admonition_inserter.py
Bibo-Joshi: I'm rather sure that black would enforce that 😉 I guess if you `"".join(lines)` and regex-search on that, that would both cover the multiline-case as well as remove the need to loop twice

docs/auxil/admonition_inserter.py
Bibo-Joshi: should be `[\w_]+`, then, right? Or are underscores included in `\w`?
Anyway, I'm fine with both :) the "non-greedy" part was what I wanted to ensure.

lemontree210: Yes, funnily enough, underscores are included in `\w`.

lemontree210: Done

docs/auxil/admonition_inserter.py
lemontree210: Done

docs/auxil/admonition_inserter.py
lemontree210: Done

docs/auxil/admonition_inserter.py
lemontree210: I refactored in 659faa19e2157a0354b199f4dc8d83e9e14a9f08 to avoid using the pattern twice. Still not sure it's worth it to join the lines because we don't have any Bot names that are that long. Even this method name would fit into one line:
```py
return await self.get_bot().send_message_and_forget_all_about_it_the_very_next_morning(
```

Bibo-Joshi: mh okay. let's keep it, then :D

docs/auxil/admonition_inserter.py
```diff
@@ -295,18 +302,20 @@ def _create_shortcuts(self) -> dict[collections.abc.Callable, str]:
         # that this given method is a shortcut for a Bot method
         for class_name, cls in inspect.getmembers(telegram, predicate=inspect.isclass):
 
-            for method_name, method in inspect.getmembers(cls, predicate=inspect.isfunction):
+            # no need to inspect Bot's own methods, as Bot can't have shortcuts in But
```
Bibo-Joshi, line 305:
```suggestion
            # no need to inspect Bot's own methods, as Bot can't have shortcuts in Bot
```

docs/auxil/admonition_inserter.py
```diff
@@ -295,18 +302,20 @@ def _create_shortcuts(self) -> dict[collections.abc.Callable, str]:
...
+            if cls is telegram.Bot:
+                continue
+
+            for method_name, method in _iter_own_public_methods(cls):
+
+                # .getsourcelines() returns a tuple. Item [1] is an int
+                for line in inspect.getsourcelines(method)[0]:
+                    bot_method_match = bot_method_pattern.search(line)
```
Bibo-Joshi, line 313:
```suggestion
                    if not bot_method_match := bot_method_pattern.search(line):
                        continue
```

since running this requires py3.9+ now anyway :)

docs/auxil/admonition_inserter.py
lemontree210: I'm glad at least I didn't add another "t"

docs/auxil/admonition_inserter.py
lemontree210: so you think an occasional regex might be too complex but the walrus operator that no-one ever learned is not confusing 😆 

because of some precedence issues this had to be wrapped in another set of parentheses

Bibo-Joshi: walrus is way less confusing than regex ;P

Bibo-Joshi (approved):

harshil21 (approved):
LGTM!

PR #3531, opened 2023-02-03
Increase verbosity of type completeness CI Job

Bibo-Joshi:
Follow-up for #3508

### DIFF
--- a/.github/workflows/type_completeness.yml
+++ b/.github/workflows/type_completeness.yml
@@ -48,7 +48,7 @@ jobs:
             pr = float(
               json.load(open("pr.json", "rb"))["typeCompleteness"]["completenessScore"]
             )
-            base_text = "After this PR, type completeness will be {round(pr, 3)}.".format(base)
+            base_text = f"After this PR, type completeness will be {round(pr, 3)}."
             if pr < (base - 0.1):
                 text = f"This PR decreases type completeness by {round(base - pr, 3)} ❌"
                 set_summary(text)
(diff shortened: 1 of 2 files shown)


.github/workflows/type_completeness.yml
```diff
@@ -33,29 +33,34 @@ jobs:
...
             )
             pr = float(
               json.load(open("pr.json", "rb"))["typeCompleteness"]["completenessScore"]
             )
+            base_text = f"After this PR, type completeness will be {round(pr, 3)}."
             if pr < (base - 0.1):
                 text = f"This PR decreases type completeness by {round(base - pr, 3)} ❌"
                 set_summary(text)
-                error(text)
+                error(f"{base_text} {text}")
```
harshil21, line 55: shouldn't this be the opposite, i.e `error(f"{text} {base_text}")` so it reads as 
```
This PR decreases type completeness by 2% ❌
After this PR, type completeness will be 98%
``` 
something like this would be shown?

Bibo-Joshi: mh, I don't really care for the order … Those are two independent true statements and switching them around makes them no less true. If you prefer it as `{text} {base_text}`, I can change it.

harshil21: for me I'd like to see the total amount after I've seen how much it has been reduced (mostly cause of the wording "This PR", "After this PR")

Poolitzer (approved):
Looks good!

harshil21 (approved):

PR #3538, opened 2023-02-07
Resolve frame paths in `was_called_by` (#3536)

### LINKED ISSUE #3536: [BUG] telegram.ext._utils.stack.was_called_by gives incorrect result on 64-bit machine
### Steps to Reproduce

1. Set up virtualenv using `python3 -m venv` on a 64-bit machine.
2. Initialize an `application` object using `telegram.ext.ApplicationBuilder`:
``` python
from telegram.ext import ApplicationBuilder
def main() -> None:
    application = ApplicationBuilder().token("TOKEN").build()

if __name__ == "__main__":
    main()
```
4. Run the bot in virtualenv and it will give a warning messgae like ```PTBUserWarning: `Application` instances should be built via the `ApplicationBuilder`.```

### Expected behaviour

The warning message shouldn't be given since `ApplicationBuilder` is being used.

### Actual behaviour

``` bash
$ python test.py 
../venv/lib64/python3.11/site-packages/telegram/ext/_applicationbuilder.py:292: 
PTBUserWarning: `Application` instances should be built via the `ApplicationBuilder`.
  ] = DefaultValue.get_value(  # pylint: disable=not-callable
```

### Operating System

Fedora Linux 37 (Server Edition)

```shell
python-telegram-bot 20.0
Bot API 6.4
Python 3.11.1 (main, Dec  7 2022, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)]
```

### Relevant log output

_No response_

### Additional Context

I believe this is caused by comparing a resolved path with an unresolved path [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/_application.py#L273). 

In my case, it finds `../venv/lib/python3.11/site-packages/telegram/ext/_applicationbuilder.py` not equal to `../venv/lib64/python3.11/site-packages/telegram/ext/_applicationbuilder.py`, the directory `lib64` being a symlink to `lib`.

A quick (maybe not final) fix is to modify [stack.py](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/_utils/stack.py) so that `was_called_by` always resolves paths from frame:
``` python
    while frame.f_back:
        frame = frame.f_back
        if Path(frame.f_code.co_filename).resolve() == caller:
            return True
```
I have tested it and the warning no longer appears.

hyfc:
Fixes #3536 which causes false warning message during bot initialization. Now `telegram.ext._utils.stack.was_called_by` will always resolve paths in frame before comparing them with caller path.

### Checklist for PRs

- [ ] ~~Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)~~
- [ ] ~~Added myself alphabetically to `AUTHORS.rst` (optional)~~
- [ ] ~~Added new classes & modules to the docs and all suitable `__all__` s~~


Bibo-Joshi (commented):
Thanks for the quick PR! Nice that you also managed to write a test case for it 👍 I have a few comments about that:

* Could you replace the `os` calls by the corresponding `pathlib` equivalents?
* I'm wondering if pytest could help with creating temporary directories: https://docs.pytest.org/en/6.2.x/tmpdir.html. that could make it possible to remove the custom cleanup logic. Plus, if the temporary directories/files are not created in the test directory, there no risk of producing left over artifacts ...
* IISC your test case currently cover only one of the changes lines. Maybe it's possible to cover also the second one?

tests/test_stack.py
```diff
@@ -31,5 +31,45 @@ def test_called_by_current_file(self):
...
+        # Here we expect `was_called_by` to recognize
+        # "`tmp_path`/foo_link.py" as same as "`tmp_path`/foo.py".
+        temp_file = tmp_path / "foo.py"
+        foo_content = "import inspect\ndef foo_func():\
+            \n    return inspect.currentframe() "
+        with temp_file.open('w') as f:
+            f.write(foo_content)
+        symlink_file = tmp_path / "foo_link.py"
+        symlink_file.symlink_to(temp_file)
+        import sys
```
Bibo-Joshi, line 46: Is it necessary to import sys here or can that be moved to the top of the file?

tests/test_stack.py
```diff
@@ -31,5 +31,45 @@ def test_called_by_current_file(self):
...
+        with temp_file2.open('w') as f:
+            f.write(bar_content)
+        symlink_file2 = tmp_path / "bar_link.py"
+        symlink_file2.symlink_to(temp_file2)
+        import sys
+        sys.path.append(tmp_path.as_posix())
+        from bar_link import bar_func
+        frame = bar_func()
+        assert was_called_by(frame, temp_file2)
+
```
Bibo-Joshi, line 77: The below comment is basically an addition to `test_called_by_current_file`. Can you move it into that test similar to the comments that you adde to your new test cases?

tests/test_stack.py
```diff
@@ -31,5 +31,45 @@ def test_called_by_current_file(self):
         file = Path(__file__)
         assert was_called_by(frame, file)
 
+    def test_called_by_symlink_file(self, tmp_path):
+        # Set up a call from a linked file in a temp directory,
+        # then test it with its resolved path.
+        # Here we expect `was_called_by` to recognize
+        # "`tmp_path`/foo_link.py" as same as "`tmp_path`/foo.py".
```
Bibo-Joshi, line 38:
```suggestion
        # "`tmp_path`/caller_link.py" as same as "`tmp_path`/caller.py".
```

A bit more speaking names will help future readors of the test files to graps the test more quickly :)

tests/test_stack.py
```diff
@@ -31,5 +31,45 @@ def test_called_by_current_file(self):
...
+        symlink_file.symlink_to(temp_file)
+        import sys
+        sys.path.append(tmp_path.as_posix())
+        from foo_link import foo_func
+        frame = foo_func()
+        assert was_called_by(frame, temp_file)
+
+    def test_called_by_symlink_file_nested(self, tmp_path):
+        # Same as test_called_by_symlink_file except 
+        # foo_func is nested inside bar_func to test
```
Bibo-Joshi, line 56:
```suggestion
        # inner_func is nested inside outer_func to test
```

Bibo-Joshi (commented):
Thank you for the updates! I have some more nitpicking, but nitpicking is really all it is :)

Bibo-Joshi:
btw, don't worry about the faild `Check Type Completeness` check, that apparently is a problem of the workflow …

hyfc:
Looks like adding `resolve` may cause trouble on Windows: [CI log](https://github.com/python-telegram-bot/python-telegram-bot/actions/runs/4128408660/jobs/7132814454#step:5:12668`)
`OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: '<string>'`

Haven't tested it on Windows, but if it breaks, I'm ok to drop this PR. We can live with a minor warning message:sweat_smile:   .

Bibo-Joshi:
That's unfortunate :/ I'll try to have a closer look at it soon™, but I agree that the priority on this may not be too high :D

hyfc:
Not able to reproduce the failed `TestApplication` unit tests on `Windows 11` with `Python 3.9` and `Python 3.11`. 
I suspect it has something to do with the built-in `nt._getfinalpathname` on some other versions of Windows.

Bibo-Joshi:
FYI: I tried around some more in #3552 and arrived at a solution that I'm happy with.

hyfc:
Good work! It looks good to me.

PR #3542, opened 2023-02-08
Dummy PR

Bibo-Joshi:
Testing if something in #3537 broke tests on py3.11 & windows


Bibo-Joshi:
My findings:

The part 

```python
self._client_kwargs = {
            "timeout": timeout,
            "proxies": proxy_url,
            "limits": limits,
            "http1": http1,
            "http2": not http1,
        }
```

in `HTTPXRequest` is what breaks the CI. Changing to 

```python
self._client_kwargs = {}
self._client_kwargs["timeout"] = timeout
...
```

does not work either, but 

```python
self._client_kwargs = dict()
self._client_kwargs["timeout"] = timeout
...
```

does, so it's definitely the curly braces.

* I still can't reproduce this locally, even on the same python version and with the same packages installed
* It's still not clear to me what exactly the problem even is. The log files read a bit like the recursion depth error happens while a traceback is being formatting, but I don't know which

---

Since this is hard to debug and it's only this one place that acts up, I'll just go back to the `dict(…)` approach. Maybe we find a solution in the future …

PR #3543, opened 2023-02-09
Fix Application.create_task type hinting

### LINKED ISSUE #3540: [BUG] Incorrect type hinting for Application.create_task
### Steps to Reproduce

In the wiki you have provided an example for [Avoiding flood limits](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Avoiding-flood-limits) :

```python
async def broadcast(update: Update, context: ContextTypes.DEFAULT_TYPE):
    await update.message.reply_text(text="Starting the broadcast...")
    context.application.create_task(
        asyncio.gather(
            *(
                context.bot.send_message(
                    chat_id=user_id, text="Hello World!", rate_limit_args={'priority': -1}
                )
                for user_id in user_ids
            )
        )
    )

```
which has a typing issue because `asyncio.gather` returns a `asyncio.Future` and `Application.create_task` expects an `asyncio.Coroutine`

### Expected behaviour

`Application.create_task` should also expect a `Future` besides `Coroutine`.

### Actual behaviour

```python
def create_task(self, coroutine: Coroutine, update: object = None) -> asyncio.Task:
```

### Operating System

Manjaro Linux 22.0.0

```shell
python-telegram-bot 20.0
Bot API 6.4
Python 3.10.6 (main, Aug 23 2022, 08:36:38) [GCC 10.2.1 20210110]
```

### Relevant log output

_No response_

### Additional Context

_No response_

sam-mosleh:
Closes #3540 

---
### Checklist for PRs

### DIFF
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -953,8 +953,8 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
         Args:
             coroutine (:term:`awaitable`): The awaitable to run as task.
 
-            .. versionchanged:: 20.2
-                Accepts :class:`asyncio.Future` and generator-based coroutine functions.
+                .. versionchanged:: 20.2
+                    Accepts :class:`asyncio.Future` and generator-based coroutine functions.
             update (:obj:`object`, optional): If set, will be passed to :meth:`process_error`
                 as additional information for the error handlers. Moreover, the corresponding
                 :attr:`chat_data` and :attr:`user_data` entries will be updated in the next run of
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -36,6 +36,7 @@ from typing import (
     Coroutine,
     DefaultDict,
     Dict,
+    Generator,
     Generic,
     List,
     Mapping,
@@ -933,7 +934,11 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
                 if close_loop:
                     loop.close()
 
-    def create_task(self, coroutine: Awaitable[RT], update: object = None) -> "asyncio.Task[RT]":
+    def create_task(
+        self,
+        coroutine: Union[Generator[Optional["asyncio.Future[object]"], None, RT], Awaitable[RT]],
+        update: object = None,
+    ) -> "asyncio.Task[RT]":
         """Thin wrapper around :func:`asyncio.create_task` that handles exceptions raised by
         the :paramref:`coroutine` with :meth:`process_error`.
 
@@ -958,7 +963,10 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
         return self.__create_task(coroutine=coroutine, update=update)
 
     def __create_task(
-        self, coroutine: Awaitable[RT], update: object = None, is_error_handler: bool = False
+        self,
+        coroutine: Union[Generator[Optional["asyncio.Future[object]"], None, RT], Awaitable[RT]],
+        update: object = None,
+        is_error_handler: bool = False,
     ) -> "asyncio.Task[RT]":
         # Unfortunately, we can't know if `coroutine` runs one of the error handler functions
         # but by passing `is_error_handler=True` from `process_error`, we can make sure that we
@@ -993,11 +1001,13 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
 
     async def __create_task_callback(
         self,
-        coroutine: Awaitable[RT],
+        coroutine: Union[Generator[Optional["asyncio.Future[object]"], None, RT], Awaitable[RT]],
         update: object = None,
         is_error_handler: bool = False,
     ) -> RT:
         try:
+            if isinstance(coroutine, Generator):
+                return await asyncio.create_task(coroutine)
             return await coroutine
         except asyncio.CancelledError as cancel:
             # TODO: in py3.8+, CancelledError is a subclass of BaseException, so we can drop this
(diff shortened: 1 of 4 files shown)


Bibo-Joshi:
`Awaitable` was explicitly removed from the type hints in python/typeshed#6779, so I don't really think that we should re-add it. Why `Future` is not listed in the annotations is not quite clear to me either. IMO we should first investigate if that's a shortcoming or concious decision in typeshed before making any moves here.

sam-mosleh:
`asyncio.create_task` only accepts `Coroutine`s and `Future`s are not acceptable:
```python
# Not gonna work
asyncio.create_task(
    asyncio.gather(
        ...
    )
)
```
So in order to make a `Task` out of it you have to await inside a coroutine which is happening in [#L994-L1031](https://github.com/sam-mosleh/python-telegram-bot/blob/227f72ab482f08a374aec457fc9c04b960f7a16f/telegram/ext/_application.py#L994-L1031)

Bibo-Joshi:
Aha! Thanks for the finding :) Them I'm good with changing to `Awaitable`. I'd like to have a new unit test though that explicitly check that `Future` is supported by `Application.create_task`. Could you add that to `test_application.py`? Please also add to the `Note` in the docs of `Application.create_task` a bullet points that clarifies that support for `Future` is offered __in addition__ to the functionality of `asyncio.create_task`.

Moreover, I realize that `Application.create_task` does in fact not support generators - in contrast to `asyncio.create_task`. This can be fixed by changing `return await coroutine` to `return await asyncio.create_task(coroutine)`  in `__create_task_callback`. That way, `Application.create_task` would fully cover the functionality of `asyncio.gather` __plus__ also supporting `Future`.

Would you like to add this to your PR? this would include

* A unit test that makes sure that generators work
* Updating the type hints & docs of `Application.create_task` accordingly

If you'd rather like to keep your PR as-is, that's perfectly fine. It can go in a follow-up PR, then.

PS: the type completeness check can't cope with PRs from external repos yet. before merging we should check the type completeness manually

sam-mosleh:
Is supporting an edge-case (generators) necessary for the `Application.create_task` because the method is already different than `asyncio.create_task` by supporting `Future`s?

Bibo-Joshi:
I think it would be benefitial, yes. The idea behind `Application.create_task` was to provide a drop-in replacement for `asyncio.create_task` that plays will with the PTB framework. If it can do a bit more than `asyncio.create_task` that's okay, but it should be able to do everything that `asyncio.create_task` can do IMO.

sam-mosleh:
@Bibo-Joshi Since I'm not familiar with the workflows, I'll leave the versioning and documentation of `Application.create_task` to you.

sam-mosleh:
I'm not sure why, but it appears that the Codecov upload token is missing, and because of that, tests are currently failing.

tests/test_application.py
```diff
@@ -1287,6 +1287,24 @@ async def callback_2():
             await asyncio.sleep(0.05)
             assert stop_task.done()
 
+    async def test_create_task_awaiting_future(self, app):
+        async def callback():
+            await asyncio.sleep(0.01)
+            return 42
+
+        out = await app.create_task(asyncio.gather(callback()))
```
Bibo-Joshi, line 1295:
```suggestion
        # `asyncio.gather` returns an `asyncio.Future` and not an
        # `asyncio.Task`
        out = await app.create_task(asyncio.gather(callback()))
```

A reminder for future devs maybe is useful :)

tests/test_application.py
```diff
@@ -1287,6 +1287,24 @@ async def callback_2():
...
+        assert out == [42]
+
+    async def test_create_task_awaiting_generator(self, app):
+        event = asyncio.Event()
+
+        def gen():
+            yield
+            event.set()
+
+        await app.create_task(gen())
```
Bibo-Joshi, line 1305: can we also check a return value here?

telegram/ext/_application.py
```diff
@@ -948,7 +951,7 @@ def create_task(
         .. seealso:: :wiki:`Concurrency`
 
         Args:
-            coroutine (:term:`coroutine function`): The coroutine to run as task.
+            coroutine (:term:`awaitable`): The awaitable to run as task.
```
Bibo-Joshi, line 954:
```suggestion
            coroutine (:term:`awaitable`): The awaitable to run as task.
            
                .. versionchanged:: 20.2
                    Accepts generators.
```

Bibo-Joshi (changes requested):
Thanks for the updates! I left some small comments :) 
Codecove sometime has a hiccup - rerunning the tests fixed it …

tests/test_application.py
sam-mosleh: We can do that if we use `asyncio.coroutine` decorator but it prints out deprecated warnings like hell for python>=3.8
Generator-based coroutine functions in asyncio has been kept  only for backward compatibility and you should drop it after 4 months (Python 3.7 EOL)

Bibo-Joshi: All right, let's skip it, then

telegram/ext/_application.py
```diff
@@ -948,7 +951,10 @@ def create_task(
...
+
+            .. versionchanged:: 20.2
+                Accepts :class:`asyncio.Future` and generator-based coroutine functions.
```
Bibo-Joshi, line 957:
```suggestion
                .. versionchanged:: 20.2
                    Accepts :class:`asyncio.Future` and generator-based coroutine functions.
```

That fixes the failing doc test :)

Bibo-Joshi (approved):

sam-mosleh:
@Bibo-Joshi I apologize for the misclick in my request for a review.

Poolitzer (approved):
Changes LGTM

Bibo-Joshi:
Thank you for your contribution @sam-mosleh ! :)

PR #3546, opened 2023-02-09
Fixed example in timerbot.py

PBartrina:
parameter was passed as `name` but used as `chat_id`

### Checklist for PRs


harshil21:
Thanks for the spot, but I'm going to have to reject this since even though that would work, we want to showcase using `job.chat_id` here as a parameter to `bot.send_message`

PBartrina:
> Thanks for the spot, but I'm going to have to reject this since even though that would work, we want to showcase using `job.chat_id` here as a parameter to `bot.send_message`

Sure, I'll PR again with the correct one

PR #3552, opened 2023-02-11
Handle symlinks in `was_called_by`

### LINKED ISSUE #3536: [BUG] telegram.ext._utils.stack.was_called_by gives incorrect result on 64-bit machine

Bibo-Joshi:
Supersedes #3538. Closes #3536

### Checklist for PRs

- ~[x] Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)~
- [x] ~Added new classes & modules to the docs and all suitable `__all__` s~


Bibo-Joshi:
After a bit of trial and error, I figured out that the windows test fail due to the usage of `pytest-xdist`. I did not investigate, why exactly that breaks it, though. Instead I chose a programatic apporach: `was_called_by` does a `try-except` and on exceptions, it logs the error and returns `False`.

This may give false warnings in rare cases where `ApplicationBuilder` was used correctly but the `_was_called_by` raises an exception, but I reckon that these cases are even more rare than symlinks.

The benefit of this change is that `was_called_by` now can never cause the code to fail - IMO this *is* an upside, as PTB figuring out if it should raise a warning should not be a cause for the program not running.

hyfc:
Good work! It looks good to me.

PR #3420, opened 2022-12-12
Extend tests/auxil in order to tidy up conftest.py

Bibo-Joshi:
As announced in #3419

to be merged only after harshils great test-overhaul :D

- [ ] https://github.com/python-telegram-bot/python-telegram-bot/pull/3426#issuecomment-1382532833

PR #3553, opened 2023-02-11
Tidy up tests directory

Bibo-Joshi:
Replacement for #3420


tests/auxil/asyncio.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python
```
harshil21, line 1: is it a good idea to name this file as `asyncio`? Maybe `asyncio_extension`/`async_helpers`

tests/auxil/ci_bots.py
```diff
@@ -1,4 +1,23 @@
 #!/usr/bin/env python
+
+#
+#  A library that provides a Python interface to the Telegram Bot API
```
harshil21, line 4: this copyright has been duplicated

tests/auxil/networking.py
```diff
@@ -0,0 +1,117 @@
...
+from httpx import AsyncClient, Response
+
+from telegram._utils.defaultvalue import DEFAULT_NONE
+from telegram._utils.types import ODVInput
+from telegram.error import BadRequest, RetryAfter, TimedOut
+from telegram.request import HTTPXRequest, RequestData
+
+
+class NonchalantHttpxRequest(HTTPXRequest):
+    """This Request class is used in the tests to supress errors that we don't to care about
```
harshil21, line 31:
```suggestion
    """This Request class is used in the tests to suppress errors that we don't want to care about
```

harshil21 (commented):
Nice work cleaning up the tests structure. Maybe worth doing these as well:

- Adding the `ext`, `inline`, `passport` folders in the tests so we can shift related test files in there. I believe they can have their own `conftest`s as well if that helps in better organizing it

tests/auxil/ci_bots.py
Bibo-Joshi: damn, thought I got all …

Bibo-Joshi:
> * Adding the `ext`, `inline`, `passport` folders in the tests so we can shift related test files in there. I believe they can have their own `conftest`s as well if that helps in better organizing it

I guess then it would make sense to mimic the whole directory structure of `telegram`. I'll have to look more closely into how pytest discovers tests. should have time on the weekend

harshil21 (approved):
LGTM

PR #3576, opened 2023-03-06
Feat: Downgrade to 1.1

Poolitzer:
I decided to go for warning when a custom http://, not https:// URL is set. I also decided to call the next version 20.1.1, I think this issue warrants a release since users face errors because of us which are hard to spot/track down in production.

I also decided to change away the [http2] dependency, I hope this doesn't break anything.

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -250,16 +250,22 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
 
         error_string = ""
 
-        if isinstance(self._request[0], HTTPXRequest):
-            if self._request[0].http_version == "2" and not base_url.startswith("https"):
-                error_string = "get_updates_request"
+        if (
+            isinstance(self._request[0], HTTPXRequest)
+            and self._request[0].http_version == "2"
+            and not base_url.startswith("https")
+        ):
+            error_string = "get_updates_request"
 
-        if isinstance(self._request[1], HTTPXRequest):
-            if self._request[1].http_version == "2" and not base_url.startswith("https"):
-                if error_string:
-                    error_string += " and request"
-                else:
-                    error_string = "request"
+        if (
+            isinstance(self._request[1], HTTPXRequest)
+            and self._request[1].http_version == "2"
+            and not base_url.startswith("https")
+        ):
+            if error_string:
+                error_string += " and request"
+            else:
+                error_string = "request"
 
         if error_string:
             warn(
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -243,9 +243,9 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
             HTTPXRequest() if request is None else request,
         )
 
-        # this section is about issuing a warning when using HTTP/2 and connect to a self hosted bot api
-        # instance, which currently only supports HTTP/1.1. Checking if a custom base url is set
-        # is the best way to do that.
+        # this section is about issuing a warning when using HTTP/2 and connect to a self hosted
+        # bot api instance, which currently only supports HTTP/1.1. Checking if a custom base url
+        # is set is the best way to do that.
 
         warning_string = ""
 
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -243,35 +243,34 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
             HTTPXRequest() if request is None else request,
         )
 
-        # this section is about raising an error when using HTTP/2 and connect to a local bot api
-        # instance. Checking if a custom url starts with http is the best way to do that since
-        # this ensures a) a local bot api instance is used and b) probably no proxy runs in front
-        # of it.
+        # this section is about issuing a warning when using HTTP/2 and connect to a local bot api
+        # instance, which currently only supports HTTP/1.1. Checking if a custom base url is set
+        # is the best way to do that.
 
-        error_string = ""
+        warning_string = ""
 
         if (
             isinstance(self._request[0], HTTPXRequest)
             and self._request[0].http_version == "2"
-            and not base_url.startswith("https")
+            and not base_url.startswith("https://api.telegram.org/bot")
         ):
-            error_string = "get_updates_request"
+            warning_string = "get_updates_request"
 
         if (
             isinstance(self._request[1], HTTPXRequest)
             and self._request[1].http_version == "2"
-            and not base_url.startswith("https")
+            and not base_url.startswith("https://api.telegram.org/bot")
         ):
-            if error_string:
-                error_string += " and request"
+            if warning_string:
+                warning_string += " and request"
             else:
-                error_string = "request"
+                warning_string = "request"
 
-        if error_string:
+        if warning_string:
             warn(
-                f"You set the HTTP version for the {error_string} HTTPX instance to HTTP/2. The "
-                "local bot api instances do not natively support this. You need to either run a "
-                "http proxy in front of it or change the version back.",
+                f"You set the HTTP version for the {warning_string} HTTPXRequest instance to "
+                f"HTTP/2. The self hosted bot api instances only support HTTP/1.1. You should "
+                f"either run a HTTP proxy in front of it which supports HTTP/2 or use HTTP/1.1.",
                 PTBUserWarning,
                 stacklevel=2,
             )
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -243,7 +243,7 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
             HTTPXRequest() if request is None else request,
         )
 
-        # this section is about issuing a warning when using HTTP/2 and connect to a local bot api
+        # this section is about issuing a warning when using HTTP/2 and connect to a self hosted bot api
         # instance, which currently only supports HTTP/1.1. Checking if a custom base url is set
         # is the best way to do that.
 
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -92,12 +92,14 @@ from telegram._utils.argumentparsing import parse_sequence_arg
 from telegram._utils.defaultvalue import DEFAULT_NONE, DefaultValue
 from telegram._utils.files import is_local_file, parse_file_input
 from telegram._utils.types import DVInput, FileInput, JSONDict, ODVInput, ReplyMarkup
+from telegram._utils.warnings import warn
 from telegram._webhookinfo import WebhookInfo
 from telegram.constants import InlineQueryLimit
 from telegram.error import InvalidToken
 from telegram.request import BaseRequest, RequestData
 from telegram.request._httpxrequest import HTTPXRequest
 from telegram.request._requestparameter import RequestParameter
+from telegram.warnings import PTBUserWarning
 
 if TYPE_CHECKING:
     from telegram import (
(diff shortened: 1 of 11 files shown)


telegram/request/_httpxrequest.py
```diff
@@ -99,8 +101,9 @@ def __init__(
         write_timeout: Optional[float] = 5.0,
         connect_timeout: Optional[float] = 5.0,
         pool_timeout: Optional[float] = 1.0,
-        http_version: str = "2",
+        http_version: str = "1.1",
     ):
+        self.http_version = http_version
```
Bibo-Joshi, line 106:
```suggestion
        self._http_version = http_version
```

Please make this a property, since this value will not be changable by the user at runtime

telegram/ext/_applicationbuilder.py
```diff
@@ -705,15 +719,28 @@ def get_updates_pool_timeout(
...
         .. seealso:: :meth:`http_version`
 
         .. versionadded:: 20.1
+        .. versionchanged:: 20.1.1
+            Reset the default version to 1.1.
+
+        Note:
+            HTTPX errors out when Telegram cancels the keepalive connection. Be mindful of those
+            when you want to change the version to 2. See `#3556 <https://github.com/
+            python-telegram-bot/python-telegram-bot/issues/3556>`_ for more details.
```
Bibo-Joshi, line 733:
```suggestion
            Users have obsered stability issues with HTTP/2, which are due to how the h2 <link here> library handles cancellations of keepalive connections. See `#3556 <https://github.com/
            python-telegram-bot/python-telegram-bot/issues/3556>`_ for a discussion.
```

telegram/_bot.py
```diff
@@ -241,6 +243,39 @@ def __init__(
             HTTPXRequest() if request is None else request,
         )
 
+        # this section is about raising an error when using HTTP/2 and connect to a local bot api
```
Bibo-Joshi, line 246:
```suggestion
        # this section is about issuing a warning when using HTTP/2 and connect to a local bot api
```

Please replace "error" with "warning" in this whole dif block :)

telegram/_bot.py
```diff
@@ -241,6 +243,39 @@ def __init__(
...
+            and not base_url.startswith("https")
+        ):
+            if error_string:
+                error_string += " and request"
+            else:
+                error_string = "request"
+
+        if error_string:
+            warn(
+                f"You set the HTTP version for the {error_string} HTTPX instance to HTTP/2. The "
```
Bibo-Joshi, line 272:
```suggestion
                f"You set the HTTP version for the {error_string} HTTPXRequest instance to HTTP/2. The "
```

telegram/_bot.py
```diff
@@ -241,6 +243,39 @@ def __init__(
...
+                "local bot api instances do not natively support this. You need to either run a "
+                "http proxy in front of it or change the version back.",
```
Bibo-Joshi, line 274:
```suggestion
                "local bot api instances do not natively support this. You should to either run a "
                "http proxy in front of it or use HTTP/1.1.",
```

tests/request/test_request.py
```diff
@@ -413,16 +414,6 @@ async def test_http_1_response(self):
...
-    async def test_http_2_response(self):
-        httpx_request = HTTPXRequest()
-        async with httpx_request:
-            resp = await httpx_request._client.request(
-                url="https://python-telegram-bot.org",
-                method="GET",
-                headers={"User-Agent": httpx_request.USER_AGENT},
-            )
-            assert resp.http_version == "HTTP/2"
-
```
Bibo-Joshi, line 425: Please re-add the test and pass `http_version="2"` to `HTTPXRequest`

telegram/_bot.py
```diff
@@ -241,6 +243,39 @@ def __init__(
...
+        ):
+            if error_string:
+                error_string += " and request"
+            else:
+                error_string = "request"
+
+        if error_string:
+            warn(
+                f"You set the HTTP version for the {error_string} HTTPX instance to HTTP/2. The "
+                "local bot api instances do not natively support this. You need to either run a "
```
Bibo-Joshi, line 273:
```suggestion
                "self hosted bot api instances do not natively support this. You need to either run a "
```

Bibo-Joshi (changes requested):
Thanks for the PR!

> I decided to go for warning when a custom http://, not https:// URL is set.

IISC the logic for the warnings-checking is not yet covering everything. After reviewing your changes, the different scenarios (all considering the user passed `http_version="2"`) that I see are:

* User is using cloud api -> base url starts with https://, `local_mode is False`, everything should work
* User is using a self-hosted, non-local API server, with or without HTTP/1.1 proxy -> base_url can be either http://… *or* https://…, `local_mode is False`, will not work due to HTTP/2
* User is using a self-hosted, non-local API server with a HTTP/2 proxy -> base_url can be either http://… *or* https://…, `local_mode is False`, will work due to proxy
* User is using a self-hosted local API server with or without a HTTP/1.1 proxy -> base_url is http://…, `local_mode is True`, will not work due to HTTP/2
* User is using a self-hosted local API server with a HTTP/2 proxy -> base_url can be http://… *or* https://…, `local_mode is True`, will not work due to proxy

In summary all combinations of http(s) & `local_mode=True/False` can either work or not work. IMO we should just issue a warning as soon as we know that the user is using a self hosted API server. This will give a few false positives, but it's only a warning after all …
 I guess the only reliable way to check if the user is using a self-hosted API server is to check if `"://api.telegram.org/" in base_(file_)url` …

> I also decided to call the next version 20.1.1, I think this issue warrants a release since users face errors because of us which are hard to spot/track down in production.

Fine with me 👍 

> I also decided to change away the [http2] dependency, I hope this doesn't break anything.

Please make it a proper optional dependency via `requirements-opts.txt` and also add unit test for that :)

Poolitzer:
@Bibo-Joshi You mention <code>base_file_url</code>, I only check <code>base_url</code>. Do you really think it makes sense to (or) check <code>base_file_url</code> as well? You cant really use it without also changing <code>base_url</code>...

Bibo-Joshi:
@Poolitzer good point. <code>base_url</code> should be enough, then :)

telegram/request/_httpxrequest.py
```diff
@@ -130,14 +133,23 @@ def __init__(
...
+                ) from exc
             raise RuntimeError(
-                "To use Socks5 proxies, PTB must be installed via `pip install "
-                "python-telegram-bot[socks]`."
+                "To use HTTP/2, PTB must be installed via `pip install "
+                "python-telegram-bot[http2]`."
             ) from exc
 
+    @property
+    def http_version(self) -> str:
```
Bibo-Joshi, line 150: Docstring is missing

tests/test_bot.py
```diff
@@ -1648,6 +1649,40 @@ async def post(*args, **kwargs):
             bot.callback_data_cache.clear_callback_data()
             bot.callback_data_cache.clear_callback_queries()
 
+    @pytest.mark.skipif(not TEST_WITH_OPT_DEPS, reason="Optional dependencies not installed")
+    async def test_http2_warning(self):
+        with pytest.raises(
```
Bibo-Joshi, line 1654: Using caplog would be more appropriate than pytest.raises. The warning being converted to an exception is just a side effect of your pytest settings, plus using caplog allows us to also test the stacklevel

tests/test_bot.py
```diff
@@ -1648,6 +1649,40 @@ async def post(*args, **kwargs):
...
+                get_updates_request=HTTPXRequest(http_version="2"),
+            )
+
+    @pytest.mark.skipif(
+        TEST_WITH_OPT_DEPS, reason="Only relevant if the optional dependency is not installed"
+    )
+    async def test_http2_runtime_error(self, bot):
+        with pytest.raises(RuntimeError, match=r"python-telegram-bot\[http2\]"):
+            HTTPXRequest(http_version="2")
+
```
Bibo-Joshi, line 1685:
```suggestion
    @pytest.mark.skipif(
        not TEST_WITH_OPT_DEPS, reason="Only relevant if the optional dependency is not installed"
    )
    async def test_http2_runtime_error(self, bot):
        with pytest.raises(RuntimeError, match=r"python-telegram-bot\[http2\]"):
            HTTPXRequest(http_version="2")
```

?

Please double check that this is actually run correctly. IIRC in the workflow yaml we list all test files that contain tests that should run without optional components. To not run all bot tests in that case it would also make sense to extract this to a standalone test class - cf `test_jobqueue.py`

telegram/ext/_applicationbuilder.py
```diff
@@ -564,15 +564,31 @@ def pool_timeout(self: BuilderType, pool_timeout: Optional[float]) -> BuilderTyp
...
+        :attr:`telegram.Bot.request`. By default, HTTP/1.1 is used.
 
         .. seealso:: :meth:`get_updates_http_version`
 
         .. versionadded:: 20.1
+        .. versionchanged:: 20.1.1
+            Reset the default version to 1.1.
+
+        Note:
+            HTTPX currently errors out with HTTP/2 when Telegram cancels the keepalive connection,
```
Bibo-Joshi, line 576: The phrasing is somewhat awkward IMHO. I made a suggestion at https://github.com/python-telegram-bot/python-telegram-bot/pull/3576#discussion_r1126930937

telegram/ext/_applicationbuilder.py
```diff
@@ -564,15 +564,31 @@ def pool_timeout(self: BuilderType, pool_timeout: Optional[float]) -> BuilderTyp
...
+            see `#3556 <https://github.com/python-telegram-bot/python-telegram-bot/issues/3556>`_
+            for more details.
+
+            You will also need to install the http2 dependency or use your own request backend.
```
Bibo-Joshi, line 580: "using you own Backend" is not relevant here. The method applies a seeing for HTTPXRequest. If the user doesn't want to use that class, they must use the request() method

Please also update the note on the optional dependency to align with those on Apscheduler, Cachetools and aiolimiter

telegram/_bot.py
```diff
@@ -241,6 +243,38 @@ def __init__(
             HTTPXRequest() if request is None else request,
         )
 
+        # this section is about issuing a warning when using HTTP/2 and connect to a local bot api
```
Bibo-Joshi, line 246:
```suggestion
        # this section is about issuing a warning when using HTTP/2 and connect to a self hosted bot api
```

Bibo-Joshi (changes requested):
Thanks for the updates! The warning logic now LGTM :)

tests/test_bot.py
Poolitzer: no its negated, skip if opt installed

Bibo-Joshi: my bad :)

tests/test_bot.py
```diff
@@ -3196,3 +3202,38 @@ async def test_arbitrary_callback_data_get_chat_no_pinned_message(
         finally:
             bot.callback_data_cache.clear_callback_data()
             bot.callback_data_cache.clear_callback_queries()
+
+
+@pytest.mark.skipif(not TEST_WITH_OPT_DEPS, reason="Optional dependencies not installed")
+class TestBotWithOptDeps:
```
Bibo-Joshi, line 3208: Sorry to be pendantic here - but … The main part of the test suite should run with the optional dependencies as one run IMO, i.e. `pytest -v`. So far only the logic for the optional dependencies *not* being installed was extracted to special classes and is tested in the special call to pytest in the workflow.

tests/test_bot.py
```diff
@@ -3196,3 +3202,38 @@ async def test_arbitrary_callback_data_get_chat_no_pinned_message(
...
+    """
+    This class exists to test the behaivour of
+    """
+
+    @pytest.mark.dev
```
Bibo-Joshi, line 3213:
```suggestion
```

tests/request/test_request.py
```diff
@@ -413,8 +414,9 @@ async def test_http_1_response(self):
             )
             assert resp.http_version == "HTTP/1.1"
 
+    @pytest.mark.skipif(not TEST_WITH_OPT_DEPS, reason="Optional dependencies not installed")
```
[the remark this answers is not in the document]
Poolitzer: I will just put them in the same class, watch :O

Bibo-Joshi (approved):
LGTM! If there's nothing left from your side, we can merge IMO :)

Poolitzer:
@Bibo-Joshi Nothing from me

tests/request/test_request.py
```diff
@@ -74,12 +74,30 @@ async def httpx_request():
 @pytest.mark.skipif(
     TEST_WITH_OPT_DEPS, reason="Only relevant if the optional dependency is not installed"
 )
-class TestNoSocks:
+class TestNoSocksHTTP2:
```
harshil21, line 77:
```suggestion
class TestNoSocksHTTP2WithoutRequest:
```

tests/request/test_request.py
```diff
@@ -74,12 +74,30 @@ async def httpx_request():
...
     async def test_init(self, bot):
         with pytest.raises(RuntimeError, match=r"python-telegram-bot\[socks\]"):
             HTTPXRequest(proxy_url="socks5://foo")
+        with pytest.raises(RuntimeError, match=r"python-telegram-bot\[http2\]"):
+            HTTPXRequest(http_version="2")
 
 
+@pytest.mark.skipif(not TEST_WITH_OPT_DEPS, reason="Optional dependencies not installed")
+class TestHTTP2:
```
harshil21, line 86:
```suggestion
class TestHTTP2WithRequest:
```

harshil21 (commented):
looks good, just 2 test class renaming suggestions..

PR #3594, opened 2023-03-12
Extend ruff to use more checkers

harshil21:
Makes ruff use more checkers, so we can now remove pyupgrade and isort from pre-commit hooks.

New checkers used:
- isort
- pylint
- pyupgrade
- ruff's checkers
- flake8-use-pathlib
- flake8-bugbear
- flake8-comprehensions
- flake8-pie
- flake8-pytest-style
- flake8-return
- flake8-simplify
- flake8-raise
- flake8-logging-format
- flake8-implicit-str-concat

- [x] TODO: Remove the ruff noqa line from test_official after api 6.6 is merged


pyproject.toml
```diff
@@ -9,3 +9,10 @@ line_length = 99
 [tool.ruff]
 line-length = 99
 target-version = "py37"
+select = ["E", "F", "I", "PL", "UP", "RUF", "PTH", "C4", "B", "PIE"]
+
+[tool.ruff.pylint]
+max-args = 100
+max-statements = 100
+max-branches = 40
+max-returns = 15
```
Bibo-Joshi, line 18: have these values been used by use before somewhere? Or how did they come to be?

telegram/_bot.py
```diff
@@ -147,7 +147,7 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
     Examples:
         :any:`Raw API Bot <examples.rawapibot>`
 
-    .. seealso:: :wiki:`Your First Bot <Extensions-–-Your-first-Bot>`,
+    .. seealso:: :wiki:`Your First Bot <Extensions---Your-first-Bot>`,
```
Bibo-Joshi, line 150: Does that link still work?

telegram/_games/game.py
```diff
@@ -159,7 +159,8 @@ def parse_text_entity(self, entity: MessageEntity) -> str:
             raise RuntimeError("This Game has no 'text'.")
 
         # Is it a narrow build, if so we don't need to convert
-        if sys.maxunicode == 0xFFFF:
+        max_value = 0xFFFF
+        if sys.maxunicode == max_value:
```
Bibo-Joshi, line 163: what's the benefit of these changes? If the argument is that 0xFFFF should not be hardcoded into the comparison, then I'd vote to move that constant to `tg._utils` instead of introduceding a `max_value` variable everywhere …

telegram/_message.py
```diff
@@ -3303,31 +3308,29 @@ def _parse_html(
                             )
                             + insert
                         )
+                elif sys.maxunicode == max_value:
+                    html_text += message_text[last_offset : entity.offset - offset] + insert
```
Bibo-Joshi, line 3312: Just FYI: I don't think that we ever covered this in the CI

telegram/_message.py
```diff
@@ -3440,17 +3445,16 @@ def _parse_markdown(
                 }
                 parsed_entities.extend(list(nested_entities.keys()))
 
-                orig_text = text
-                text = escape_markdown(text, version=version)
+                escaped_text = escape_markdown(text, version=version)
 
                 if nested_entities:
-                    if version < 2:
+                    if version < md_version_2:
```
Bibo-Joshi, line 3451: what's the benefit of this change?

telegram/_update.py
```diff
@@ -46,7 +46,7 @@ class Update(TelegramObject):
     Note:
         At most one of the optional parameters can be present in any given update.
 
-    .. seealso:: :wiki:`Your First Bot <Extensions-–-Your-first-Bot>`
+    .. seealso:: :wiki:`Your First Bot <Extensions---Your-first-Bot>`
```
Bibo-Joshi, line 49: .

telegram/ext/_aioratelimiter.py
```diff
@@ -167,7 +167,8 @@ def _get_group_limiter(self, group_id: Union[str, int, bool]) -> "AsyncLimiter":
         # We only do that if we have a lot of limiters lying around to avoid looping on every call
         # This is a minimal effort approach - a full-fledged cache could use a TTL approach
         # or at least adapt the threshold dynamically depending on the number of active limiters
-        if len(self._group_limiters) > 512:
+        max_limiters = 512
```
Bibo-Joshi, line 170: similar question as above

telegram/ext/_application.py
```diff
@@ -499,7 +500,8 @@ async def _initialize_persistence(self) -> None:
         ):
             persistent_data = await self.persistence.get_callback_data()
             if persistent_data is not None:
-                if not isinstance(persistent_data, tuple) or len(persistent_data) != 2:
+                data_length = 2
```
Bibo-Joshi, line 503: .

telegram/ext/_applicationbuilder.py
```diff
@@ -114,7 +114,7 @@ class ApplicationBuilder(Generic[BT, CCT, UD, CD, BD, JQ]):
         * Unless a custom :class:`telegram.Bot` instance is set via :meth:`bot`, :meth:`build` will
           use :class:`telegram.ext.ExtBot` for the bot.
 
-    .. seealso:: :wiki:`Your First Bot <Extensions-–-Your-first-Bot>`,
+    .. seealso:: :wiki:`Your First Bot <Extensions---Your-first-Bot>`,
```
Bibo-Joshi, line 117: .

telegram/ext/_jobqueue.py
```diff
@@ -41,6 +41,9 @@
     from telegram.ext import Application
 
 
+_day_tuple = tuple(range(7))
```
Bibo-Joshi, line 44:
```suggestion
_ALL_DAYS = tuple(range(7))
```

telegram/ext/filters.py
```diff
@@ -244,7 +244,7 @@ class MessageFilter(BaseFilter):
 
     Please see :class:`BaseFilter` for details on how to create custom filters.
 
-    .. seealso:: :wiki:`Advanced Filters <Extensions-–-Advanced-Filters>`
+    .. seealso:: :wiki:`Advanced Filters <Extensions---Advanced-Filters>`
```
Bibo-Joshi, line 247: .

telegram/ext/filters.py
```diff
@@ -387,7 +387,7 @@ def _merge(base_output: Union[bool, Dict], comp_output: Union[bool, Dict]) -> Fi
                 if isinstance(base[k], list):
                     base[k] += comp_value
                 else:
-                    base[k] = [base[k]] + comp_value
+                    base[k] = [base[k], *comp_value]
```
Bibo-Joshi, line 390: this line is unchecked. Surely not a result of this PR, but it would be cool if we could increase the coverage :)

telegram/helpers.py
```diff
@@ -55,9 +55,10 @@ def escape_markdown(text: str, version: int = 1, entity_type: str = None) -> str
             See the official API documentation for details. Only valid in combination with
             ``version=2``, will be ignored else.
     """
+    md_version_2 = 2
     if int(version) == 1:
         escape_chars = r"_*`["
-    elif int(version) == 2:
+    elif int(version) == md_version_2:
```
Bibo-Joshi, line 61: see above. If ruff is complaining about hard-coded comparisons, why isn't it complaining about `== 1`?

telegram/helpers.py
```diff
@@ -163,14 +164,16 @@ def create_deep_linked_url(bot_username: str, payload: str = None, group: bool =
     Returns:
         :obj:`str`: An URL to start the bot with specific parameters.
     """
-    if bot_username is None or len(bot_username) <= 3:
+    min_bot_username_length = 3
+    max_payload_length = 64
```
Bibo-Joshi, line 168: .

telegram/request/_baserequest.py
```diff
@@ -289,7 +289,8 @@ async def _request_wrapper(
         except Exception as exc:
             raise NetworkError(f"Unknown error in HTTP implementation: {repr(exc)}") from exc
 
-        if HTTPStatus.OK <= code <= 299:
+        last_http_success_code = 299
```
Bibo-Joshi, line 292: .

.pre-commit-config.yaml
```diff
@@ -64,23 +64,8 @@ repos:
...
-        args:
-          - --py37-plus
--   repo: https://github.com/pycqa/isort
-    rev: 5.12.0
-    hooks:
-    -   id: isort
-        name: isort
-        args:
-        - --diff
-        - --check
```
Bibo-Joshi, line 81: What is the compatibility status here? IIRC @lemontree210 reported that isort detects some things that ruff doesn't. moreover, ruff is still far away from being stable. I'm not quite comfortable with dropping pyupgrade and isort completely yet …

Bibo-Joshi (changes requested):
Thanks for the PR!

Regarding the several changed comparisons: I guess that ruff doesn't like hardcoded values there. However, I see no value of something like

```python
const = 5
if something == const:
    ...
```

especially if the comparison is part of a function. The definition of `const` should at the very least be moved outside of the function definition to avoid defining it on every run.

.pre-commit-config.yaml
lemontree210: I just said that isort was more informative: it shows exactly what the problem is, whereas ruff just states that there is a sorting problem in a given file.

I decided to keep isort in my own project too.

pyproject.toml
harshil21: We just ignored them in pylint, could do the same here if you'd like..
https://github.com/python-telegram-bot/python-telegram-bot/blob/d5cdb94c6f7b35c54b754781e4d99168d572df15/setup.cfg#L19-L22

telegram/_bot.py
harshil21: no :( Maybe let's just change the wiki heading itself then?

harshil21:
> Thanks for the PR!
> 
> Regarding the several changed comparisons: I guess that ruff doesn't like hardcoded values there. However, I see no value of something like
> 
> ```python
> const = 5
> if something == const:
>     ...
> ```
> 
> especially if the comparison is part of a function. The definition of `const` should at the very least be moved outside of the function definition to avoid defining it on every run.

I don't really like it either.. do you want to ignore it?

telegram/_message.py
harshil21: do you want me to cover it? I think we'll have to monkeypatch it then.. I think we can just drop it since `sys.maxunicode` was only different for python 2 back in the day (https://docs.python.org/3/library/sys.html#sys.maxunicode). This code has been there since we had python 2 support, so do you want to just drop it?

telegram/_message.py
harshil21: no benefit at all imo

telegram/ext/_jobqueue.py
harshil21: I think it didn't allow me to make it capital for some reason

telegram/helpers.py
harshil21: no idea, maybe because it's very common to check for hard-coded 0 and 1?

.pre-commit-config.yaml
harshil21: ruff's readme does say it replace pyupgrade and isort.. And also I've never had or seen pyupgrade fail once btw

harshil21: Update: https://twitter.com/charliermarsh/status/1637512274716893187?s=20 (and the next tweet)

Based on the issue linked in the FAQ for isort (https://github.com/charliermarsh/ruff/issues/1381, https://github.com/charliermarsh/ruff/issues/2104), there are slight differences how ruff handles it, which I'm okay with.

The pyupgrade one I believe we can definitely drop, based on the tweet.

telegram/_message.py
Bibo-Joshi: uhm, that might be possible, yes. Can you open a new issue for that?

.pre-commit-config.yaml
Bibo-Joshi: Thanks for the links! Then from compatibility POV I'd be okay.

from stability pov I'm still cautious though. I just had a look at the contributiors page https://github.com/charliermarsh/ruff/graphs/contributors and that makes me more reserved for two reasons:

* the project is barely half a year old
* there is one main contributor

meaning that if the main contributor can no longer advance the development for whichever reason, the project is basically stuck at a way-before-stable status … I'd very much like the project to get more mature before relying too much on it.

Finally, in my experience isort & pyupgrade usually don't take much time - that's mostly pylint.

tests/test_videochat.py
```diff
@@ -116,7 +116,7 @@ def test_de_json(self, user1, user2, bot):
         assert video_chat_participants.users[0].id == user1.id
         assert video_chat_participants.users[1].id == user2.id
 
-    @pytest.mark.parametrize("use_users", (True, False))
+    @pytest.mark.parametrize("use_users", [True, False])
```
Bibo-Joshi, line 119: why is that better? 🤔

tests/ext/test_picklepersistence.py
```diff
@@ -137,10 +137,10 @@ def bad_pickle_files():
         "pickletest",
     ]:
         Path(name).write_text("(())")
-    yield True
+    return True
```
Bibo-Joshi, line 140: mh, what's the difference between yeild and return in this context?

tests/ext/test_jobqueue.py
```diff
@@ -71,7 +71,7 @@ def test_init_job(self):
     not TEST_WITH_OPT_DEPS, reason="Only relevant if the optional dependency is installed"
 )
 @pytest.mark.skipif(
-    os.getenv("GITHUB_ACTIONS", False) and platform.system() in ["Windows", "Darwin"],
+    bool(os.getenv("GITHUB_ACTIONS", "") and platform.system() in ["Windows", "Darwin"]),
```
Bibo-Joshi, line 74: isn't there a helper function for that?

tests/ext/test_dictpersistence.py
```diff
@@ -25,33 +25,33 @@
...
     return {12345: {"test1": "test2", "test3": {"test4": "test5"}}, 67890: {3: "test4"}}
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture()
 def callback_data():
     return [("test1", 1000, {"button1": "test0", "button2": "test1"})], {"test1": "test2"}
 
 
-@pytest.fixture(scope="function")
```
Bibo-Joshi, line 54: why all these changes?

Bibo-Joshi (commented):
Soooo many changes in the tests 😮

tests/ext/test_picklepersistence.py
harshil21: I think it did that cause there's nothing after the `yield`, hence no point in `yield`ing when we can just `return` it

tests/ext/test_jobqueue.py
harshil21: yes, will use

tests/ext/test_dictpersistence.py
harshil21: the default scope is actually function, so I think that tool just prefers it to omit defining it explicitly 🤷🏽

tests/test_videochat.py
Bibo-Joshi: ouch … but yes, fair enough reasoning :)

telegram/_bot.py
harshil21: I think it is better to do this just before release, as if we do it now users will have to click again if they visit the wiki from the docs

harshil21: forgot that we had that..

harshil21: done in live-wiki.. by mistake, little damage anyway

PR #3623, opened 2023-03-24
Overhaul loggers. No tests or docs yet.

### LINKED ISSUE #3591: Overhaul logger names
We currently use `logging.getLogger(__name__)` to get the logger that we use in a module. While this is common practice, this explicitly uses the naming of our files, leading to output like

```
DEBUG:telegram.ext._updater:Updater started (polling)
```

This exposes the protected modules that are protected for the reason that users should not use them directly.

I suggest to revisit every call to `logging.getLogger` in our code base and

* evaluate whether we need a logger on class-instance, class, or module level
* evaluate which name that logger should carry. The name should
  * reflect the public API of PTB
  * make it clear to the user which component is doing the logging

If some auxiliary functionality can be used to automate the derivation of the name, this is a welcome addition and can be implemented in a new `telegram._utils.logging` moduile.

IMO it would be desirable to extend existing logging-related unit tests to check the set name, especially if some automation functionality is used for the name derivation.

Note that so far we have made no promisies about the names of the loggers in the documentation, so that I don't consider this a breaking change. We might want to give a promise in the future.

Bibo-Joshi:
When done, closes #3591

### Checklist for PRs

### DIFF
--- a/telegram/_utils/logging.py
+++ b/telegram/_utils/logging.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+#
+# A library that provides a Python interface to the Telegram Bot API
+# Copyright (C) 2015-2023
+# Leandro Toledo de Souza <devs@python-telegram-bot.org>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser Public License for more details.
+#
+# You should have received a copy of the GNU Lesser Public License
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+"""This module contains helper functions related to logging.
+
+Warning:
+    Contents of this module are intended to be used internally by the library and *not* by the
+    user. Changes to this module are not considered breaking changes and may not be documented in
+    the changelog.
+"""
+import logging
+
+
+def get_logger(file_name: str, class_name: str = None) -> logging.Logger:
+    """Returns a logger with an appropriate name. Use as follows::
+
+        logger = get_logger(__name__)
+
+    If for example `__name__` is `telegram.ext._updater`, the logger will be named
+    `telegram.ext.Updater`. If `class_name` is passed, this will result in
+    `telegram.ext.<class_name>`. Useful e.g. for CamelCase class names.
+
+    If the file name points to a utils module, the logger name will simply be `telegram(.ext)`.
+
+    Returns:
+        :class:`logging.Logger`: The logger.
+    """
+    parts = file_name.split("_")
+    if parts[1].startswith("_utils"):
+        name = parts[0]
+    else:
+        name = f"{parts[0]}{class_name or parts[1].capitalize()}"
+    return logging.getLogger(name)
--- a/telegram/ext/_aioratelimiter.py
+++ b/telegram/ext/_aioratelimiter.py
@@ -21,7 +21,6 @@ library.
 """
 import asyncio
 import contextlib
-import logging
 import sys
 from typing import Any, AsyncIterator, Callable, Coroutine, Dict, List, Optional, Union
 
@@ -32,6 +31,7 @@ try:
 except ImportError:
     AIO_LIMITER_AVAILABLE = False
 
+from telegram._utils.logging import get_logger
 from telegram._utils.types import JSONDict
 from telegram.error import RetryAfter
 from telegram.ext._baseratelimiter import BaseRateLimiter
@@ -46,6 +46,9 @@ else:
     @contextlib.asynccontextmanager
     async def null_context() -> AsyncIterator[None]:
         yield None
+
+
+_logger = get_logger(__name__, class_name="AIORateLimiter")
 
 
 class AIORateLimiter(BaseRateLimiter[int]):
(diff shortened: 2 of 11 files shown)


harshil21 (commented):
just a single comment

telegram/ext/_aioratelimiter.py
```diff
@@ -48,6 +48,9 @@ async def null_context() -> AsyncIterator[None]:
         yield None
 
 
+_logger = get_logger(__name__, class_name="AIORateLimiter")
```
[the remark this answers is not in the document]
Bibo-Joshi: For Bot I made it a class variable because we need different names for Bot and ExtBot. This is not needed for other classes so far

telegram/_utils/logging.py
```diff
@@ -0,0 +1,48 @@
...
+Warning:
+    Contents of this module are intended to be used internally by the library and *not* by the
+    user. Changes to this module are not considered breaking changes and may not be documented in
+    the changelog.
+"""
+import logging
+
+
+def get_logger(file_name: str, class_name: str = None) -> logging.Logger:
+    """Returns a logger with an appropriate name. Use as follows::
```
lemontree210, line 30:
```suggestion
    """Returns a logger with an appropriate name. 
    
    Use as follows:
```
The first line is a summary line, would be strange for it to end with a colon

telegram/_utils/logging.py
```diff
@@ -0,0 +1,48 @@
...
+    If the file name points to a utils module, the logger name will simply be `telegram(.ext)`.
+
+    Returns:
+        :class:`logging.Logger`: The logger.
+    """
+    parts = file_name.split("_")
+    if parts[1].startswith("utils") and class_name is None:
+        name = parts[0].rstrip(".")
+    else:
+        name = f"{parts[0]}{class_name or parts[1].capitalize()}"
```
lemontree210, line 47: I remember doing something with class names when we were creating admonitions. Just in case :)

telegram/ext/_aioratelimiter.py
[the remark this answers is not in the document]
lemontree210: worth a CSI in `Bot`?

telegram/ext/_updater.py
```diff
@@ -51,6 +51,7 @@
 
 
 _UpdaterType = TypeVar("_UpdaterType", bound="Updater")  # pylint: disable=invalid-name
+_logger = get_logger(__name__)
```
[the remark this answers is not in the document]
lemontree210: and should be capitalized anyway, no?

telegram/ext/_updater.py
```diff
@@ -345,7 +344,7 @@ async def polling_action_cb() -> bool:
 
             if updates:
                 if not self.running:
-                    self._logger.critical(
+                    _logger.critical(
                         "Updater stopped unexpectedly. Pulled updates will be ignored and again "
```
lemontree210, line 348: This is unrelated to the changes but something seems to be wrong with the phrase on the next line ("Pulled updates will be ignored and again..."). Maybe a verb is missing?

lemontree210 (commented):
just some minor comments (how could I do more :))

telegram/ext/_aioratelimiter.py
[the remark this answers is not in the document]
harshil21: > worth a CSI in `Bot`?

I think so, yes

telegram/ext/_updater.py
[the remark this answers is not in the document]
Bibo-Joshi: I *guess* one could see this as constant in the sense of https://peps.python.org/pep-0008/#constants though personally I don't have a strong opinion on that. Anyways, if it's important to you, I'll just rename :)

telegram/_utils/logging.py
Bibo-Joshi: There you have access to `cls.__name__` though, right? here the issue is that we call `get_logger` before the class is defined so that my best guess was the file name …

lemontree210: OK :)

telegram/ext/_updater.py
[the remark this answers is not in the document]
Bibo-Joshi: too late 🙆‍♂️

telegram/ext/_application.py
```diff
@@ -1519,13 +1519,13 @@ async def __update_persistence(self) -> None:
                 # *all* tasks will be done.
                 if not new_state.done():
                     if self.running:
-                        _logger.debug(
+                        _LOGGER.debug(
```
harshil21, line 1522: not covered

telegram/_utils/logging.py
```diff
@@ -0,0 +1,49 @@
...
+    Contents of this module are intended to be used internally by the library and *not* by the
+    user. Changes to this module are not considered breaking changes and may not be documented in
+    the changelog.
+"""
+import logging
+
+
+def get_logger(file_name: str, class_name: str = None) -> logging.Logger:
+    """Returns a logger with an appropriate name.
+    Use as follows::
```
lemontree210, line 31:
```suggestion

    Use as follows:
```
additional blank line after summary line + I doubt that the double colon is intentional :)

telegram/ext/_utils/webhookhandler.py
```diff
@@ -31,11 +30,15 @@
 from tornado.httpserver import HTTPServer
 
 from telegram import Update
+from telegram._utils.logging import get_logger
 from telegram.ext._extbot import ExtBot
 
 if TYPE_CHECKING:
     from telegram import Bot
 
+# This module is not visible to users, so we log as Updater
```
lemontree210, line 39: Just a quick understanding check: why as Updater?

lemontree210 (approved):

telegram/ext/_utils/webhookhandler.py
harshil21: because Updater.run_webhook will use this class to start it

telegram/_utils/logging.py
harshil21: colon is intentional so the block below renders as python code, but this is not going to be in the docs anyway so 🤷🏽

PR #3631, opened 2023-03-25
Use pytest's `--lf` and `--lfnf` to stabilize tests

### LINKED ISSUE #3626: Use pytests `--lf` feature to stabalize CI
It rather frequently happens that a handful of tests fail in the CI due to flakyness (even though Harshil always puts admirable effort in trying to stabalize that away :D). Rerunning the failing jobs usually fixes that, but it requires manual work.

Pytest comes with the [`--lf`](https://docs.pytest.org/en/7.2.x/how-to/cache.html) feature, i.e. it allows to re-run only the failing tests. I suggest to make use of featurer in our workflow definition by running `pytest --lf` one (or maybe even twice). For tests failing due to flakyness, this hopefully reduces the number of failing jobs. For tests that fail due to systematic problems, this will not affect the outcome.

harshil21:
Closes #3626


Bibo-Joshi (commented):
Thanks for the PR!

Just for future reference: We can see the effect of the changes in the job runs

* https://github.com/python-telegram-bot/python-telegram-bot/actions/runs/4521712977/jobs/7963608524?pr=3631
* https://github.com/python-telegram-bot/python-telegram-bot/actions/runs/4521712977/jobs/7963608648?pr=3631

Would you mind also applying this to the "test without opt. deps" section? 🤔

.github/workflows/test.yml
```diff
@@ -54,7 +54,10 @@ jobs:
           # - without http2 support
           TO_TEST="test_no_passport.py or test_datetime.py or test_defaults.py or test_jobqueue.py or test_applicationbuilder.py or test_ratelimiter.py or test_updater.py or test_callbackdatacache.py or test_request.py"
           pytest -v --cov -k "${TO_TEST}"
-          status=$?
+          # Rerun only failed tests (--lf), and don't run any tests if none failed (--lfnf=none)
+          pytest -v --cov -k "${TO_TEST}" --lf --lfnf=none
```
Bibo-Joshi, line 58:
```suggestion
          pytest -v --cov --cov-append -k "${TO_TEST}" --lf --lfnf=none
```

Bibo-Joshi (changes requested):

.github/workflows/test.yml
harshil21: sneaky, that's why coverage was failing

PR #3632, opened 2023-03-27
Localize datetimes to defaults timezone

### LINKED ISSUE #3595: [FEATURE] Message.date is UTC-localized
### Steps to Reproduce

`Message.date` is a datetime object created by PTB converting the original unix timestamp sent by telegram. It IS timezone aware, but uses UTC.

### Expected behaviour

I fully expected that - because I've setup my local timezone using `Default.tzinfo` and passing it to `ApplicationBuilder.defaults` - the received `Message.date` would be localized in my local timezone.

### Actual behaviour

This is not the case. AFAIK `Default.tzinfo` is only used to localize the datetime objects used for job_queue's jobs.

### Operating System

Raspberry OS

```shell
python-telegram-bot 20.1
Bot API 6.5
Python 3.11.0 (main, Nov 23 2022, 10:42:54) [GCC 10.2.1 20210110]
```

### Relevant log output

_No response_

### Additional Context

_No response_

Trifase:
When done, will close #3595

### Checklist for PRs


Bibo-Joshi (changes requested):
Thanks for the PR! The changes look good so far. We'll need a few newunit tests though :)

* tests for `extract_tzinfo_from_defaults` in `test_datetime.py`, where you pass a `tg.Bot`, a `tg.ext.ExtBot` without defaults and a `tg.ext.ExtBot` with defaults to ensure that everything works as expected
* For all the changed classes we need ensure that the datetime objects are localized as expected in the three cases described above. For that, please add a `test_de_json_localization` (or sim) after the respective `de_json` tests. The test fixtures linked in #3595 should be used for that.

Please do reach out if you need help with the unit tests - I know that those are not always simple :)

telegram/_utils/datetime.py
```diff
@@ -179,9 +186,17 @@ def from_timestamp(unixtime: Optional[int], tzinfo: dtm.tzinfo = UTC) -> Optiona
...
-    if tzinfo is not None:
-        return dtm.datetime.fromtimestamp(unixtime, tz=tzinfo)
-    return dtm.datetime.utcfromtimestamp(unixtime)
+    return dtm.datetime.fromtimestamp(unixtime, tz=UTC if tzinfo is None else tzinfo)
+
+
+def extract_tzinfo_from_defaults(bot: "Bot") -> Union[dtm.tzinfo, None]:
+    """
+    Extracts the timezone info from the default values of the bot.
+    If the bot has no default values, None is returned.
```
lemontree210, line 195:
```suggestion
    If the bot has no default values, :obj:`None` is returned.
```

tests/test_chatinvitelink.py
```diff
@@ -107,6 +107,33 @@ def test_de_json_all_args(self, bot, creator):
...
+            "name": self.name,
+            "pending_join_request_count": str(self.pending_join_request_count),
+        }
+
+        invite_link_raw = ChatInviteLink.de_json(json_dict, raw_bot)
+        invite_link_bot = ChatInviteLink.de_json(json_dict, bot)
+        invite_link_tz = ChatInviteLink.de_json(json_dict, tz_bot)
+
+        # comparing utcoffsets because comparing timezones is unpredicatable
+        invite_offset = datetime.datetime.utcoffset(invite_link_tz.expire_date)
```
Bibo-Joshi, line 128:
```suggestion
        invite_offset = invite_link_tz.expire_date.utcoffset()
```

`datetime.datetime.utfcoffset` is a bound method rather than a class method, so the argument that you pass to it is actually the `self`. That works, but feels strange to me … The SO link that I pointed you to also does that "wrongly".

Bibo-Joshi (approved):
Thank you for the updates - LGTM! What a nice PR :) If we get another approval (maybe lemontree or pool), we can merge

telegram/_chatmemberupdated.py
Poolitzer, line 1: this is missing the docstring about the change

telegram/_message.py
Poolitzer, line 1: this also misses changestring for forward_date and edit_date

telegram/_webhookinfo.py
Poolitzer, line 1: also misses the docstring

tests/conftest.py
Poolitzer, line 1: this feels unfitting here, but if hinrich doesn't complain I am not opposed

Poolitzer (changes requested):
Missed a few changestrings.

Btw, is the docstring important for the args documentation? It only applies to the attributes of the classes, no...

tests/conftest.py
Bibo-Joshi: I committed myself ;P

telegram/_webhookinfo.py
Trifase: I forgot that i had modified this. I also forgot the tests.

Trifase: I was going to fix this, but I should point out that the documentation of `_webhookinfo.py` states that the two parameter I was going to localize where unix time (it's wrong: they were datetime). Should I just fix it and add my localization versionchanged or put two versionchanged and put the version that was changed from unix to datetime and then add my versionchanged?

Trifase: In the interest of time I simply fixed the docstring and added my localization versiochanged

Bibo-Joshi:
@Poolitzer do you have any comments left? @Trifase I guess there is notheng to be added from your side?

Trifase:
> @Poolitzer do you have any comments left? @Trifase I guess there is notheng to be added from your side?

I think I fullfilled everything that was requested and everything else that was missing!

Poolitzer (approved):
LGTM

Bibo-Joshi:
Thank you very much for this nice contribution @Trifase ! I hope that you also had some fun with the PR :)

PR #3640, opened 2023-03-30
Improved Message.link

### LINKED ISSUE #3475: [FEATURE] Improve `Message.link` property to reflect threads

`Message.link` currently does not reflect if a message is a comment on another message in a supergroup/channel and also does not reflect if it's a message in a forum topic. By now we have a official documentation on how message links work: https://core.telegram.org/api/links#message-links and it would be nice to extend the support for this.

I'd like for `Message.link` to reflect the "threads" that the message was sent in. IISC that can achieved by

* include the `reply_to_message_id` as `thread` parameter in groups
* include the `reply_to_message_id` as `comment` parameter in channels
* for topics, I'm no entirely sure, but IISC the forum id and message_id should suffice, maybe the `forum_id` can also be used as `thread`

Admittedly, this needs some double checking on how things works out.

We could also add a full-fledged `build_message_link` function to `helpers` (or sim naming) that accepts all parameters described in https://core.telegram.org/api/links#message-links and use that function in `Message.link`.

We could also provide a method `Message.build_link` that accepts boolean flags that indicate if the link should represent the message in the thread (if available) or in the whole chat etc, but I think that's over the top.

### Additional context

Originated from https://t.me/pythontelegrambotgroup/648879.

This has no priority to me. Just nice to have.

Trifase:
When done will close #3475

### Checklist for PRs


Trifase:
In the end I just added two lines to detect if the message is a topic message, in which case append the `thread=` parameter to the link. 

I think that's all that is needed because it reflects what official clients do.

Poolitzer (approved):
LGTM thanks trifase

Bibo-Joshi (changes requested):
Thanks for the PR! The changes themself look good - I guess you double checked that this is working on at least the official clients? However I'm missing

* a  `versionchanged::` telling the user that the property now also reflects the reply-thingy
* support for `reply_to_message_id` in non-topic groups

Could you add those as well?

telegram/_message.py
```diff
@@ -826,11 +826,19 @@ def id(self) -> int:  # pylint: disable=invalid-name
...
+                Added support for topics.
         """
         if self.chat.type not in [Chat.PRIVATE, Chat.GROUP]:
             # the else block gets rid of leading -100 for supergroups:
             to_link = self.chat.username if self.chat.username else f"c/{str(self.chat.id)[4:]}"
-            return f"https://t.me/{to_link}/{self.message_id}"
+            baselink = f"https://t.me/{to_link}/{self.message_id}"
+
+            if self.is_topic_message and self.message_thread_id:
+                baselink += f"?thread={self.message_thread_id}"
```
lemontree210, line 839:
```suggestion
                baselink = f"{baselink}?thread={self.message_thread_id}"
```
I heard f-strings are faster than addition, and we have an f-string here anyway

telegram/_message.py
```diff
@@ -826,11 +826,19 @@ def id(self) -> int:  # pylint: disable=invalid-name
...
         """
         if self.chat.type not in [Chat.PRIVATE, Chat.GROUP]:
             # the else block gets rid of leading -100 for supergroups:
             to_link = self.chat.username if self.chat.username else f"c/{str(self.chat.id)[4:]}"
-            return f"https://t.me/{to_link}/{self.message_id}"
+            baselink = f"https://t.me/{to_link}/{self.message_id}"
+
+            # adds the thread for topics and replies
+            if (self.is_topic_message and self.message_thread_id) or self.reply_to_message:
+                baselink = f"{baselink}?thread={self.message_thread_id}"
```
Bibo-Joshi, line 840: `self.reply_to_message` can be `None` without `self.message_thread_id` being present though, right? 🤔

telegram/_message.py
```diff
@@ -826,11 +826,19 @@ def id(self) -> int:  # pylint: disable=invalid-name
     def link(self) -> Optional[str]:
         """:obj:`str`: Convenience property. If the chat of the message is not
         a private chat or normal group, returns a t.me link of the message.
+
+            .. versionchanged:: NEXT.VERSION
+                Added support for topics.
```
Bibo-Joshi, line 831:
```suggestion
                For messages that are replies or part of a forum topic, the link now points to the corresponding thread view.
```

Bibo-Joshi (changes requested):

telegram/_message.py
Trifase: I'm too dumb to parse your rethorical question..
If it's a topic reply (with a `message_thread_id`), or it's a reply to something, we enhance the link with `thread`.
If it's not any of those things, we don't.

Is my logic wrong?

Bibo-Joshi: My point is that a message can be a reply even if it's not in a forum group. To my understanding, `message_thread_id` is set only in froum groups. But maybe I'm mistaken …

Trifase: I think that `message_thread_id` is there even for replies to normal messages in a supergroup, pointing to the same `reply_to_message.id`.

Bibo-Joshi (approved):
Thanks for the Updates! LGTM now :) @Poolitzer do you have any comments left?

Poolitzer:
nope Im good feel free to merge

PR #3647, opened 2023-04-04
Introduce Typing context manager, utility class

yeralin:
### Checklist for PRs

    - [ ] Updated the Bot API version number in all places: `README.rst` and `README_RAW.rst` (including the badge), as well as `telegram.constants.BOT_API_VERSION_INFO`

---

### Notice:
Before I add doc strings and unit tests, I'd like to share this draft PR to be reviewed. I wanted to understand if this feature would be met well by the maintainers.

### Description: 
This PR introduces `Typing` class with additional `bot.typing()` method which is a context manager used for sending "typing..." action every 5 seconds until the context manager exits.

### Usage:
```python
async def message(update: Update, context: CallbackContext) -> None:
    async with context.bot.typing(chat_id=update.effective_chat.id):
        # some expensive operation
    await context.bot.send_message(chat_id=update.effective_chat.id,
                                   text=response)
```

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -100,6 +100,7 @@ from telegram._utils.warnings_transition import warn_about_thumb_return_thumbnai
 from telegram._webhookinfo import WebhookInfo
 from telegram.constants import InlineQueryLimit
 from telegram.error import InvalidToken
+from telegram.ext._typing import Typing
 from telegram.request import BaseRequest, RequestData
 from telegram.request._httpxrequest import HTTPXRequest
 from telegram.request._requestparameter import RequestParameter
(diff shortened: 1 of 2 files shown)


Poolitzer:
@yeralin Hi.

Thanks for sharing this draft with us. We would consider this out of the scope for PTB itself, but would very much welcome this as a contribution over at ptbcontrib!

Two things to note: I would make the specific ChatAction customizable, and I would probably expect a decorator as a user instead of a context manager.

harshil21:
I really like this idea, thanks! As Pool said this would better suited for ptbcontrib than here.

I would also prefer having both context manager and decorator, depending on the use case we can use on over the other.

PR #3654, opened 2023-04-09
Customized concurrent handling

### LINKED ISSUE #3509: [FEATURE] Customizing concurrent handling of updates

Setting `ApplicationBuilder.concurrent_updates(True)` allows to process all updates concurrently (up to a specified number at a time). What you can't do is e.g. processing updates from all users & chats concurrently while processing the updates for each user/chat sequentially. Having the possibility for such a mechanism would e.g. allow to

* handle updates concurrently except those that are revelant for stateful handlers such as `ConversationHandler`
* apply a rate-limiting mechanism for users by delaying processing of their updates
* prioritizing certain types of updates for handling, e.g. inline/callback queries over regular messages

1. Introduce an interface class `BaseThrottler` (other naming suggestions are welcome!). This class should
    1. accept an integer that specifies the maximum overall number of updates that may be processed concurrently
    2. have a single abstract method `async def process_update(update, coroutine)` whose task it is to `await` at an appropriate time, that may be determined depending on the `update`
    3. have a single non-abstract method `asyne def do_process_update` that calls `process_update` after applying a global semaphore, which enforces the limit in a.
2. Introduce a class `SimpleThrottler(BaseThrottler)` (other naming suggestions are welcome) for which `process_update` immediately awaits the coroutine, i.e. does not apply any additional throttling
3. The behavior of `Application.concurrent_updates(…)` should be as follows:
    1. passing an int will make use of the `SimpleThrottler` with the respective max overall number of concurrent updates
    2. passing `True` will behave as a. with the current default number specified for that case
    3. passing an instace of `BaseThrottler` will use that instance for handling concurrency
4. `Application.concurrent_updates` should return the max overall number of concurrent updates as specified for the throttler in use. The throttler itself may be made available via an additional attribute/property

* Instead of an interface class, one could directly provide the `SimpleThrottler` and encourage users to override `process_update`
* One could make `SimpleThrottler` a bit more elaborate by e.g. allowing to sequencify the processing of updates per chat id. In this case, I would vote to go for the interface-class approach as the implementation would surely require more elaborate internals than that of `SimpleThrottler`.

### Additional context

This idea was already discussed a bit in the early stages of the v20 architecture design but was put on hold, since we first wanted to get the basic structure ready before introducing additional functionality.

Before an attempt on an implementation for this is made, I'd like to get feedback on the idea :)

clot27:
Closes #3509 when done.
### Checklist for PRs


docs/source/telegram.ext.BaseProcessor.rst
```diff
@@ -0,0 +1,6 @@
+BaseProcessor
+======================
```
Bibo-Joshi, line 2:
```suggestion
BaseProcessor
=============
```

telegram/ext/_application.py
```diff
@@ -297,14 +299,16 @@ def __init__(
...
-        self._concurrent_updates: int = concurrent_updates or 0
-
+        if isinstance(concurrent_updates, int):
+            if concurrent_updates < 0:
+                raise ValueError("`concurrent_updates` must be a non-negative integer!")
+            concurrent_updates = SimpleProcessor(concurrent_updates)
+        self._concurrent_updates_sem = asyncio.BoundedSemaphore(
+            concurrent_updates.max_concurrent_updates or 1
+        )
+        self._concurrent_updates: BaseProcessor = concurrent_updates
```
Bibo-Joshi, line 311:
```suggestion
        self._update_processor: BaseProcessor = concurrent_updates
```

May be a more descriptive name now :)

Actually, what do you think about `{Base, Simple}UpdateProcessor` as naming instead of just `*Processor`? 🤔

telegram/ext/_application.py
```diff
@@ -361,7 +365,7 @@ def concurrent_updates(self) -> int:
 
         .. seealso:: :wiki:`Concurrency`
         """
-        return self._concurrent_updates
+        return self._concurrent_updates.max_concurrent_updates
```
Bibo-Joshi, line 368: I think it would be benefitial to add a note to the docstring stating that this is now just a shortcut to `processor.max_concurrent_updates`

telegram/ext/_application.py
```diff
@@ -1672,3 +1688,64 @@ async def process_error(
 
         _logger.exception("No error handlers are registered, logging exception.", exc_info=error)
         return False
+
+
+class BaseProcessor:
```
Bibo-Joshi, line 1693: let's move this to a standalone file :)
`SimpleProcessor` could maybe be put in the same file … I guess Pool would like that :D

the two classes also need docstrings

telegram/ext/_application.py
```diff
@@ -1672,3 +1688,64 @@ async def process_error(
...
+    def __init__(self, max_concurrent_updates: int):
+        self.max_concurrent_updates = max_concurrent_updates
```
Bibo-Joshi, line 1695: I suggest to make this a property. That way it's clear that this value can not be changed at runtime (which it can't be anyways since `self.semaphore` wouldn't reflect those changes)

telegram/ext/_application.py
```diff
@@ -1672,3 +1688,64 @@ async def process_error(
...
+        self.semaphore = BoundedSemaphore(self.max_concurrent_updates or 1)
```
Bibo-Joshi, line 1696:
```suggestion
        self._semaphore = BoundedSemaphore(self.max_concurrent_updates or 1)
```

Do you also want to check for negative values here?

telegram/ext/_application.py
```diff
@@ -1672,3 +1688,64 @@ async def process_error(
...
+    async def shutdown(self) -> None:
+        pass
+
+    async def do_process_update(
+        self,
+        update: object,
+        coroutine: "Awaitable[Any]",
+    ) -> None:
+        async with self.semaphore:
+            await self.process_update(update, coroutine)
```
Bibo-Joshi, line 1720: These methods need docstrings :)

Bibo-Joshi (commented):
Thanks for the nice PR @clot27 !
IMO it covers the contents of #3509 pretty well and the overvall logic looks good :) I left some comments already though I'll ofc do a proper review again once it's out of draft status .

telegram/ext/_application.py
clot27: Yeah, Gonna add that

telegram/ext/_application.py
clot27: yup

telegram/ext/_application.py
clot27: Seems a better option to me, lets see what others think. cc @harshil21

telegram/ext/_application.py
```diff
@@ -1061,7 +1072,12 @@ async def _update_fetcher(self) -> None:
 
             if self._concurrent_updates:
                 # We don't await the below because it has to be run concurrently
-                self.create_task(self.__process_update_wrapper(update), update=update)
```
harshil21, line 1064: `__process_update_wrapper` had also marked the task as done in the update queue (`self.update_queue.task_done()`), shouldn't we do that too in `do_process_update`?

telegram/ext/_applicationbuilder.py
```diff
@@ -198,7 +198,7 @@ def __init__(self: "InitApplicationBuilder"):
         self._context_types: DVType[ContextTypes] = DefaultValue(ContextTypes())
         self._application_class: DVType[Type[Application]] = DefaultValue(Application)
         self._application_kwargs: Dict[str, object] = {}
-        self._concurrent_updates: Union[int, DefaultValue[bool]] = DEFAULT_FALSE
+        self._concurrent_updates: Union[int, DefaultValue[bool], "BaseProcessor"] = DEFAULT_FALSE
```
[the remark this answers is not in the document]
harshil21: > I think that's the builders job

You mean https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/ext/_application.py#L301-L305 should be in the builder? I would agree.

telegram/ext/_application.py
harshil21: > more descriptive name now :)

yep good change

> what do you think about {Base, Simple}UpdateProcessor as naming instead of just *Processor?

yes would prefer that too.

telegram/ext/_applicationbuilder.py
```diff
@@ -916,9 +918,11 @@ def concurrent_updates(self: BuilderType, concurrent_updates: Union[bool, int])
         .. seealso:: :attr:`telegram.ext.Application.concurrent_updates`
 
         Args:
-            concurrent_updates (:obj:`bool` | :obj:`int`): Passing :obj:`True` will allow for
-                ``256`` updates to be processed concurrently. Pass an integer to specify a
-                different number of updates that may be processed concurrently.
+            concurrent_updates (:obj:`bool` | :obj:`int` | :class:`BaseProcessor`): Passing
+            :obj:`True` will allow for `256`` updates to be processed concurrently. Pass an
+            integer to specify a different number of updates that may be processed
+            concurrently. Pass an instance of `BaseProcessor` to use that instance
```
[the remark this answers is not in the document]
harshil21: should also specify that passing an `int` will make use of the `SimpleThrottler` class.

harshil21 (changes requested):
initial review. Will review again after tests are added, implementation looks ok to me

telegram/ext/__init__.py
```diff
@@ -25,6 +25,7 @@
     "ApplicationHandlerStop",
     "BaseHandler",
     "BasePersistence",
+    "BaseProcessor",
```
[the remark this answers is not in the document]
clot27: Umm, why? users not gonna use SimpleProcessor directly...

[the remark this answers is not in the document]
harshil21: True, but shouldn't they know it exists, like `RequestData`? If not making it public, then the class name should begin with `_`. @Bibo-Joshi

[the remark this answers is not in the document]
Bibo-Joshi: I see no reason to hide the default implementation from the users ...

telegram/ext/_application.py
clot27: done

telegram/ext/_applicationbuilder.py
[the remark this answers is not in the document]
clot27: done

telegram/ext/_application.py
```diff
@@ -359,9 +356,12 @@ def concurrent_updates(self) -> int:
         """:obj:`int`: The number of concurrent updates that will be processed in parallel. A
         value of ``0`` indicates updates are *not* being processed concurrently.
 
+        Note:
+            This is now just a shortcut to `processor.max_concurrent_updates`.
```
Bibo-Joshi, line 360:
```suggestion
        .. verisonchanged:: NEXT.VERSION
            This is now just a shortcut to ``update_processor.max_concurrent_updates``.
```

telegram/ext/_application.py
```diff
@@ -379,6 +379,14 @@ def job_queue(self) -> Optional["JobQueue[CCT]"]:
             )
         return self._job_queue
 
+    @property
+    def processor(self) -> Optional["BaseUpdateProcessor"]:
+        """:class:`telegram.ext.BaseProcessor`: The processor used by this application.
```
Bibo-Joshi, line 384:
```suggestion
    def update_processor(self) -> Optional["BaseUpdateProcessor"]:
        """:class:`telegram.ext.BaseUpdateProcessor`: The update processor used by this application.
```

telegram/ext/_application.py
```diff
@@ -297,14 +298,10 @@ def __init__(
...
-        if isinstance(concurrent_updates, int) and concurrent_updates < 0:
-            raise ValueError("`concurrent_updates` must be a non-negative integer!")
-        if concurrent_updates is True:
-            concurrent_updates = 256
-        self._concurrent_updates_sem = asyncio.BoundedSemaphore(concurrent_updates or 1)
-        self._concurrent_updates: int = concurrent_updates or 0
-
+        self._concurrent_updates_sem = asyncio.BoundedSemaphore(
+            update_processor.max_concurrent_updates or 1
+        )
```
Bibo-Joshi, line 303: We no longer need this semaphore, as `BaseUpdateProcossor.do_process_updates` now has a semaphore :)

telegram/ext/_applicationbuilder.py
```diff
@@ -198,7 +199,9 @@ def __init__(self: "InitApplicationBuilder"):
         self._context_types: DVType[ContextTypes] = DefaultValue(ContextTypes())
         self._application_class: DVType[Type[Application]] = DefaultValue(Application)
         self._application_kwargs: Dict[str, object] = {}
-        self._concurrent_updates: Union[int, DefaultValue[bool]] = DEFAULT_FALSE
+        self._concurrent_updates: Union[
+            int, DefaultValue[bool], "BaseUpdateProcessor"
+        ] = DEFAULT_FALSE
```
Bibo-Joshi, line 204:
```suggestion
        self._update_processor: "BaseUpdateProcessor" = SimpleUpdateProcessor(max_concurrent_updates=1)
```

If `concurrent_updates` in not called, then this should be the default value that we'd like to user, right? :)

telegram/ext/_applicationbuilder.py
```diff
@@ -306,7 +309,7 @@ def build(
             bot=bot,
             update_queue=update_queue,
             updater=updater,
-            concurrent_updates=DefaultValue.get_value(self._concurrent_updates),
+            update_processor=DefaultValue.get_value(self._concurrent_updates),
```
Bibo-Joshi, line 312:
```suggestion
            update_processor=self._update_processor,
```

telegram/ext/_applicationbuilder.py
```diff
@@ -916,14 +921,33 @@ def concurrent_updates(self: BuilderType, concurrent_updates: Union[bool, int])
...
         Args:
-            concurrent_updates (:obj:`bool` | :obj:`int`): Passing :obj:`True` will allow for
-                ``256`` updates to be processed concurrently. Pass an integer to specify a
-                different number of updates that may be processed concurrently.
+            concurrent_updates (:obj:`bool` | :obj:`int` | :class:`BaseUpdateProcessor`): Passing
+            :obj:`True` will allow for ``256`` updates to be processed concurrently. Pass an
+            integer to specify a different number of updates that may be processed concurrently,
+            in that case :class:`telegram.ext.SimpleUpdateProcessor` is used to handle updates.
+            Pass an instance of :class:`telegram.ext.BaseUpdateProcessor` to use that instance
+            for handling updates concurrently.
```
Bibo-Joshi, line 929:
```suggestion
            concurrent_updates (:obj:`bool` | :obj:`int` | :class:`BaseUpdateProcessor`): Passing
            :obj:`True` will allow for ``256`` updates to be processed concurrently using :class:`telegram.ext.SimpleUpdateProcessor`. Pass an
            integer to specify a different number of updates that may be processed concurrently.
            Pass an instance of :class:`telegram.ext.BaseUpdateProcessor` to use that instance
            for handling updates concurrently.
```

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,140 @@
...
+    your own update processor.
+
+    .. seealso:: :wiki:`Concurrency`
+
+    .. versionadded:: NEXT.VERSION
+
+    Args:
+        max_concurrent_updates (:obj:`int`): The maximum number of concurrent updates to be
+            processed. If this number is exceeded, new updates will be queued until the number of
+            currently processed updates decreases.
```
Bibo-Joshi, line 39:
```suggestion
        max_concurrent_updates (:obj:`int`): The maximum number of updates to be
            processed concurrently. If this number is exceeded, new updates will be queued until the number of
            currently processed updates decreases.
```

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Raises:
+        :exc:`ValueError`: If `max_concurrent_updates` is a negative integer.
```
Bibo-Joshi, line 42:
```suggestion
        :exc:`ValueError`: If :paramref:`max_concurrent_updates` is a negative integer.
```

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,140 @@
...
+        """:obj:`int`: The maximum number of updates that can be processed concurrently."""
+        return self._max_concurrent_updates
+
+    @abstractmethod
+    async def process_update(
+        self,
+        update: object,
+        coroutine: "Awaitable[Any]",
+    ) -> None:
+        """Custom implementation of how to process an update.
```
Bibo-Joshi, line 62:
```suggestion
        """Custom implementation of how to process an update. Must be implemented by a sublass.
        
        Warning:
            This method will be called by :meth:`do_process_update`. It should *not* be called manually.
```

Also we should switch the namings of `process_update` and `do_process_update` to align with `BasePersistence.do_request`. I think I messed that up in the issue description :/

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,140 @@
...
+
+        Args:
+            update (:obj:`object`): The update to be processed.
+            coroutine (:obj:`Awaitable`): The coroutine that will be awaited to process the update.
+        """
+
+    @abstractmethod
+    async def initialize(self) -> None:
+        """Initializes the Processor so resources can be allocated.
```
Bibo-Joshi, line 71:
```suggestion
        """Initializes the processor so resources can be allocated. Must be implemented by a subclass.
```

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,140 @@
...
+
+        .. seealso::
+            :meth:`shutdown`
+        """
+
+    @abstractmethod
+    async def shutdown(self) -> None:
+        """Shutdown the Processor so resources can be freed.
```
Bibo-Joshi, line 79: .

Bibo-Joshi (changes requested):
Thanks for the updates! This will come out great :) I left a bunch of new comments, but nothing major IMO.

telegram/ext/_applicationbuilder.py
```diff
@@ -916,14 +921,33 @@ def concurrent_updates(self: BuilderType, concurrent_updates: Union[bool, int])
...
+            concurrent_updates = 256
+        elif concurrent_updates is False:
+            concurrent_updates = 0
+
+        # If `concurrent_updates` is an integer, create a `SimpleUpdateProcessor`
+        # instance with that integer value; otherwise, raise an error if the value
+        # is negative
+        if isinstance(concurrent_updates, int):
+            if concurrent_updates < 0:
+                raise ValueError("`concurrent_updates` must be a non-negative integer!")
```
Bibo-Joshi, line 945:
```suggestion
            if concurrent_updates < 1:
                raise ValueError("`concurrent_updates` must be a positive integer!")
```

Same in `BaseUpdateProcessor`. 0 concurrent updates would mean that we don't process updates at all :D

telegram/ext/_baseupdateprocessor.py
Bibo-Joshi: Sorry, I somehow missed your comment. I mean this one here: https://docs.python-telegram-bot.org/en/stable/telegram.request.baserequest.html#telegram.request.BaseRequest.do_request

telegram/ext/_baseupdateprocessor.py
```diff
@@ -34,32 +34,42 @@ class BaseUpdateProcessor:
...
     def max_concurrent_updates(self) -> int:
         """:obj:`int`: The maximum number of updates that can be processed concurrently."""
         return self._max_concurrent_updates
 
+    @property
+    def semaphore(self) -> BoundedSemaphore:
+        """:class:`asyncio.BoundedSemaphore`: The semaphore used to limit the number of
+        concurrent updates."""
+        return self._semaphore
+
```
Bibo-Joshi, line 61: Is there as specific reason why you added this property? I don't think that we want the user to fiddle with that … 🤔

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,151 @@
...
+    your own update processor.
+
+    .. seealso:: :wiki:`Concurrency`
+
+    .. versionadded:: NEXT.VERSION
+
+    Args:
+        max_concurrent_updates (:obj:`int`): The maximum number of updates to be processed
+        concurrently. If this number is exceeded, new updates will be queued until the number
+        of currently processed updates decreases.
```
harshil21, line 39: speaking of.. do you want to provide a way to inspect current number of processed updates? (https://github.com/python-telegram-bot/python-telegram-bot/issues/3509#issuecomment-1400636217) can be done by inspecting a property in semaphore itself iirc

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,151 @@
...
+
+    Raises:
+        :exc:`ValueError`: If :paramref:`max_concurrent_updates` is a negative integer.
+    """
+
+    def __init__(self, max_concurrent_updates: int):
+        self._max_concurrent_updates = max_concurrent_updates
+        if self.max_concurrent_updates < 0:
```
harshil21, line 47:
```suggestion
        if self.max_concurrent_updates < 1:
```

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,151 @@
...
+            raise ValueError("`max_concurrent_updates` must be a positive integer!")
+        self._semaphore = BoundedSemaphore(self.max_concurrent_updates or 1)
```
harshil21, line 49:
```suggestion
        self._semaphore = BoundedSemaphore(self.max_concurrent_updates)
```

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,151 @@
...
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
+        """Shutdown the Processor from the context manager."""
+        await self.shutdown()
+
+
+class SimpleUpdateProcessor(BaseUpdateProcessor):
```
harshil21, line 133: this class needs a docstring

telegram/ext/_baseupdateprocessor.py
clot27: uff did that locally for some purpose, forgot to remove...

tests/ext/test_application.py
```diff
@@ -210,6 +210,21 @@ def test_job_queue(self, one_time_bot, app, recwarn):
         assert str(recwarn[0].message) == expected_warning
         assert recwarn[0].filename == __file__, "wrong stacklevel"
 
+    def test_concurrent_updates(self, one_time_bot):
+        application = ApplicationBuilder().bot(one_time_bot).concurrent_updates(2).build()
+        assert application.concurrent_updates == (
+            application.update_processor.max_concurrent_updates
+        )
+
+    def test_active_updates(self, one_time_bot):
```
harshil21, line 219: it would be nice to mock process_update to check if number of active updates is actually correct

tests/ext/test_baseupdateprocessor.py
```diff
@@ -0,0 +1,96 @@
...
+        processor = MockProcessor(5)
+        application = ApplicationBuilder().concurrent_updates(processor).bot(one_time_bot).build()
+        update = Update(1)
+
+        async def coroutine():
+            pass
+
+        await application.update_queue.put(1)
+        await processor.process_update(update, coroutine, application)
+        assert not application.update_queue.empty()
```
Bibo-Joshi, line 49: I'm not sure what this is supposed to test … I guess what we want to make sure is that `process_update` actually calls `do_process` update 

What must also be tested is that not more than `max_concurrent_updates` are being processed at the same time. This can be done in a similar fashion as the `TestApplication.test_update_processor`  test

tests/ext/test_baseupdateprocessor.py
```diff
@@ -0,0 +1,96 @@
...
+
+class TestSimpleUpdateProcessor:
+    async def test_do_process_update(self):
+        processor = SimpleUpdateProcessor(1)
+        update = Update(1)
+
+        async def coroutine():
+            pass
+
+        await processor.do_process_update(update, coroutine())
```
Bibo-Joshi, line 90: this doesn't really test anything :/ We want to make sure that this actually awaits the `coroutine` - for that you could set a test flag in that function.

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,167 @@
...
+        """Does nothing."""
+
+    def __eq__(self, other: object) -> bool:
+        return (
+            isinstance(other, self.__class__)
+            and other.max_concurrent_updates == self.max_concurrent_updates
+        )
+
+    def __hash__(self) -> int:
+        return hash(self.max_concurrent_updates)
```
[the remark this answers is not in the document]
clot27: Well, I did that just for tests most probably here https://github.com/python-telegram-bot/python-telegram-bot/pull/3654/files#diff-8708d60b1d1959712bde5982c7f5d73ea53e7d6c522be904062924f574d3a03aR417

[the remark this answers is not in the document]
Bibo-Joshi: I see. Adding functionality "just" to make testing easier would be an antipattern IMO :D In the tests you can instead use `is` instead of `==`. That even makes sure that `app.update_processor` is actually the very same object as `expected` and not a different object that just happens to have the same `max_concurrent_updates` setting.

telegram/ext/_application.py
```diff
@@ -359,9 +352,20 @@ def concurrent_updates(self) -> int:
...
         """
-        return self._concurrent_updates
+        return self._update_processor.max_concurrent_updates
+
+    @property
+    def active_updates(self) -> int:
+        """:obj:`int`: The number of active updates that are currently being processed by the
+        update processor.
+        """
+        return self._update_processor._semaphore._value  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
harshil21: Okay, I'm fine with the counter approach, but I would definitely like having it so the user is assured that things are running concurrently with an easy way to introspect it.

telegram/ext/_baseupdateprocessor.py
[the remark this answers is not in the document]
clot27: Well I tried "is" prior to "==" and it didnt worked, so I decided to check if it is the instance of SUP and max concurrent updates is same, cant find a way to get the same attribute passed tho...

tests/ext/test_application.py
```diff
@@ -227,6 +213,33 @@ def test_job_queue(self, one_time_bot, app, recwarn):
...
+        processor = MockProcessor(2)
+
+        async def coroutine():
+            pass
+
+        application = ApplicationBuilder().bot(one_time_bot).concurrent_updates(processor).build()
+        assert application.active_updates == processor._semaphore._value
+        await application.update_queue.put(1)
+        await processor.process_update(Update(1), coroutine, application)
+        assert application.active_updates == processor._semaphore._value
```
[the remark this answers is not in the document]
clot27: Will keep this in mind, ref: https://github.com/python-telegram-bot/python-telegram-bot/pull/3654#discussion_r1194253447

telegram/ext/_application.py
```diff
@@ -397,6 +405,11 @@ async def initialize(self) -> None:
             return
 
         await self.bot.initialize()
+        if isinstance(self._update_processor, BaseUpdateProcessor) and not isinstance(
+            self._update_processor, SimpleUpdateProcessor
+        ):
+            await self._update_processor.initialize()
```
Bibo-Joshi, line 411: `self._update_processor` always is an instance of `BUP`, isn't it? And we should just always call `initialize`, no matter which exact implementation is used :) If the implementation decides that `initialize` doesn't do anything that's fine, but that's not the concern of the `Application` class.

Same for `shutdown` :)

tests/ext/test_application.py
```diff
@@ -241,53 +227,114 @@ def test_custom_context_init(self, one_time_bot):
...
+        async def after_initialize_update_processor(*args, **kwargs):
+            self.test_flag.add("update_processor")
+
         async def after_initialize_updater(*args, **kwargs):
             self.test_flag.add("updater")
 
         monkeypatch.setattr(Bot, "initialize", call_after(Bot.initialize, after_initialize_bot))
+        if isinstance(update_processor, BaseUpdateProcessor) and not isinstance(
+            update_processor, SimpleUpdateProcessor
+        ):
```
Bibo-Joshi, line 251: This clause should be removed as well - see my comment above

tests/ext/test_application.py
```diff
@@ -241,53 +227,114 @@ def test_custom_context_init(self, one_time_bot):
...
+        def after_shutdown_update_processor(*args, **kwargs):
+            self.test_flag.add("update_processor")
+
         def after_updater_shutdown(*args, **kwargs):
             self.test_flag.add("updater")
 
         monkeypatch.setattr(Bot, "shutdown", call_after(Bot.shutdown, after_bot_shutdown))
+        if isinstance(update_processor, BaseUpdateProcessor) and not isinstance(
+            update_processor, SimpleUpdateProcessor
+        ):
```
Bibo-Joshi, line 311: .

tests/ext/test_applicationbuilder.py
```diff
@@ -403,7 +416,9 @@ async def post_stop(app: Application) -> None:
         assert app.updater.update_queue is update_queue
         assert app.updater.bot is app.bot
         assert app.context_types is context_types
-        assert app.concurrent_updates == concurrent_updates
+        assert isinstance(app.update_processor, SimpleUpdateProcessor)
+        assert app.update_processor.max_concurrent_updates == expected.max_concurrent_updates
+        assert app.concurrent_updates == app.update_processor.max_concurrent_updates
```
Bibo-Joshi, line 421: I see, using `is` here won't work because you don't pass the `expected` object itself to `concurrent_updates` but rather an integer. Then this is ofc fine. 👍 Please do however also add a test that ensure that passing a `SUP` object to `concurrent_updates` actually works with `is`

Bibo-Joshi (changes requested):
Thanks for the new updates! I left smaller comments :) Also the docs build is failing …

Regarding the unit tests on the update processors: TBH I don't see them going in a right direction and I'm not sure how I can explain better what should be covered. I will try and write some tests myself, but please do feel free to ask anything about that if you have questions :)

telegram/ext/_application.py
clot27: Well, its always an instance of `BUP` but when nothing is passed it uses `SUP` so we need to check it with both `SUP` and `BUP`. The `call_after` only checks upto one stacklevel, i.e. it will be call `SUP.initialize` when `SUP` object is passed and `BUP.initialize` when `BUP` object is passed (same for shutdown ofc), so we have to write seperate tests for both cases thats why I added this implementation. :thinking:

clot27: @Bibo-Joshi

Bibo-Joshi: Let me point you to https://github.com/python-telegram-bot/python-telegram-bot/pull/3654#discussion_r1186681941 ;P If the tests don't work as disered, let's adapt the tests, not the code ;) TBH I don't quite understand where the issue in the tests is, though …

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,149 @@
...
+# You should have received a copy of the GNU Lesser Public License
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+"""This module contains the BaseProcessor class."""
+from abc import abstractmethod
+from asyncio import BoundedSemaphore
+from types import TracebackType
+from typing import Any, Awaitable, Optional, Type
+
+
+class BaseUpdateProcessor:
```
Bibo-Joshi, line 26:
```suggestion
class BaseUpdateProcessor(ABC):
```
TBH I'm astonished that this even worked without … see my question at https://stackoverflow.com/questions/76292008

Bibo-Joshi (approved):
LGTM :) @harshil21 do you have anything to add?

After merge, the dev-wiki should be updated to

* elaborate on the new processor on the page on concurrency
* mention the `BaseUpdateProcessor` on the architecture page. This would be a good chance to convert the graphic to a mermaid diagram

telegram/ext/_application.py
```diff
@@ -379,12 +376,25 @@ def job_queue(self) -> Optional["JobQueue[CCT]"]:
...
+        return self._update_processor
+
     async def initialize(self) -> None:
         """Initializes the Application by initializing:
 
         * The :attr:`bot`, by calling :meth:`telegram.Bot.initialize`.
         * The :attr:`updater`, by calling :meth:`telegram.ext.Updater.initialize`.
         * The :attr:`persistence`, by loading persistent conversations and data.
+        * The :attr:`update_processor` by calling
+            :meth:`telegram.ext.BaseUpdateProcessor.initialize`.
```
harshil21, line 397:
```suggestion
        :meth:`telegram.ext.BaseUpdateProcessor.initialize`.
```

telegram/ext/_applicationbuilder.py
```diff
@@ -917,14 +922,36 @@ def concurrent_updates(self: BuilderType, concurrent_updates: Union[bool, int])
...
+            concurrent_updates = 256
+        elif concurrent_updates is False:
+            concurrent_updates = 1
+
+        # If `concurrent_updates` is an integer, create a `SimpleUpdateProcessor`
+        # instance with that integer value; otherwise, raise an error if the value
+        # is negative
+        if isinstance(concurrent_updates, int):
+            if concurrent_updates < 1:
+                raise ValueError("`concurrent_updates` must be a positive integer!")
```
harshil21, line 949: if we check for negative values in the `__init__` of `BaseUpdateProcessor`, we need not check the same again here right?

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,151 @@
...
+    ) -> None:
+        """Custom implementation of how to process an update. Must be implemented by a subclass.
+
+        Warning:
+            This method will be called by :meth:`process_update`. It should *not* be called
+            manually.
+
+        Args:
+            update (:obj:`object`): The update to be processed.
+            coroutine (:obj:`Awaitable`): The coroutine that will be awaited to process the update.
```
harshil21, line 70: `Awaitable` doesn't render right, I don't remember the correct reference

telegram/ext/_baseupdateprocessor.py
```diff
@@ -0,0 +1,151 @@
...
+    ) -> None:
+        """Shutdown the Processor from the context manager."""
+        await self.shutdown()
+
+
+class SimpleUpdateProcessor(BaseUpdateProcessor):
+    """Instance of :class:`telegram.ext.BaseUpdateProcessor` that immediately awaits the
+    coroutine, i.e. does not apply any additional processing. This is used by default when
+    :attr:`telegram.ext.ApplicationBuilder.concurrent_updates` is :obj:`int`.
+
```
harshil21, line 128: Don't we want to add `Args` to this docstring as well? Or maybe change the .rst file to `:inherit-members:` or something like that

tests/ext/test_application.py
```diff
@@ -277,22 +278,35 @@ async def test_shutdown(self, one_time_bot, monkeypatch, updater):
...
+
         def after_updater_shutdown(*args, **kwargs):
             self.test_flag.add("updater")
 
+        update_processor = SimpleUpdateProcessor(1)
         monkeypatch.setattr(Bot, "shutdown", call_after(Bot.shutdown, after_bot_shutdown))
+        monkeypatch.setattr(
+            SimpleUpdateProcessor,
+            "initialize",
+            call_after(SimpleUpdateProcessor.initialize, after_shutdown_update_processor),
```
harshil21, line 292:
```suggestion
            "shutdown",
            call_after(SimpleUpdateProcessor.shutdown, after_shutdown_update_processor),
```
? how did the tests pass

harshil21 (commented):
final review. Looks much more refined and ready. Good job @clot27. Just some tiny errors caught:

telegram/ext/_application.py
clot27: Why? sphinx gives warning if I do so.

harshil21: it's not rendered properly, just align the `:meth:` with the `The`

tests/ext/test_application.py
clot27: seems scary 😬

telegram/ext/_baseupdateprocessor.py
harshil21: Ok nvm, that didn't fix it, maybe just write add a docstring here saying to refer `BaseUpdateProcessor` for args to pass? @clot27

Bibo-Joshi: We don't list arguments of the base classes anywhere else either, do we? The base classes are listed in any case, which suffices IMO

harshil21: alright then

docs/source/telegram.ext.simpleupdateprocessor.rst
```diff
@@ -0,0 +1,7 @@
+SimpleUpdateProcessor
+=====================
+
+.. autoclass:: telegram.ext.SimpleUpdateProcessor
+    :inherited-members: telegram.ext.BaseUpdateProcessor
```
harshil21, line 5:
```suggestion
```
let's remove this and then I'm done.

harshil21 (approved):
nice upgrade!

Bibo-Joshi:
Big thanks to @clot27 for your work and also your patience! 👏

PR #3657, opened 2023-04-10
shield-update-fetcher

Bibo-Joshi:
See https://t.me/pythontelegrambotgroup/670240?thread=670177 for a reference.
I'll have to check if I can write a unit test for this special case …

### Checklist for PRs

- ~[ ] Added `.. versionadded:: version`, `.. versionchanged:: version` or `.. deprecated:: version` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)~

### DIFF
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -1046,24 +1046,30 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
     async def _update_fetcher(self) -> None:
         # Continuously fetch updates from the queue. Exit only once the signal object is found.
         while True:
-            update = await self.update_queue.get()
+            try:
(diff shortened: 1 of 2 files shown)


harshil21 (commented):
Nice. Though can't we just use asyncio.shield() in `app.start()`?

Bibo-Joshi:
> Though can't we just use asyncio.shield() in `app.start()`?

That is actually way better! Don't know why I didn't think of it :D It works in a manual test. Let's see if the CI is happy as well - I'm not on a linux machine rn …

Bibo-Joshi:
So apparently `asyncio.shield` does not work on linux. TBH I'm not sure why exactly is, but I'm also not overly enthusiastic to dig into it too much when we have a working solution … So I reverted.

PR #3673, opened 2023-04-21
API 6.7

Bibo-Joshi:
TODO:

- [x] Mock the `set/get_my_name` test
- [x] Update the warnings tests to check with bot `Bot` and `ExtBot` in the manner of #3674

### Checklist for PRs

* New classes:
    - [x] `__init__` accepts `api_kwargs` as kw-only

* Added new shortcuts:
    - ~[ ] In `Chat` & `User` for all methods that accept `chat/user_id`~
    - ~[ ] In `Message` for all methods that accept `chat_id` and `message_id`~
    - ~[ ] For new `Message` shortcuts: Added `quote` argument if methods accepts `reply_to_message_id`~
    - ~[ ] In `CallbackQuery` for all methods that accept either `chat_id` and `message_id` or `inline_message_id`~

* If relevant:

    - [x] Link new and existing constants in docstrings instead of hard coded number and strings
    - ~[ ] Add new message types to `Message.effective_attachment`~
    - ~[ ] Added new handlers for new update types~
      - ~[ ] Add the handlers to the warning loop in the `ConversationHandler`~
    - ~[ ] Added new filters for new message (sub)types~
    - [x] Added the new method(s) to `_extbot.py`
    - [x] Added or updated `bot_methods.rst`
    - [x] Updated the Bot API version number in all places: `README.rst` and `README_RAW.rst` (including the badge), as well as `telegram.constants.BOT_API_VERSION_INFO`
    - ~[ ] Added logic for arbitrary callback data in `tg.ext.Bot` for new methods that either accept a `reply_markup` in some form or have a return type that is/contains `telegram.Message`~

### DIFF
--- a/telegram/_inline/inlinekeyboardbutton.py
+++ b/telegram/_inline/inlinekeyboardbutton.py
@@ -23,6 +23,7 @@ from typing import TYPE_CHECKING, ClassVar, Optional, Union
 from telegram import constants
 from telegram._games.callbackgame import CallbackGame
 from telegram._loginurl import LoginUrl
+from telegram._switchinlinequerychosenchat import SwitchInlineQueryChosenChat
 from telegram._telegramobject import TelegramObject
 from telegram._utils.types import JSONDict
 from telegram._webappinfo import WebAppInfo
@@ -111,6 +112,10 @@ class InlineKeyboardButton(TelegramObject):
             in inline mode when they are currently in a private chat with it. Especially useful
             when combined with ``switch_pm*`` actions - in this case the user will be automatically
             returned to the chat they switched from, skipping the chat selection screen.
+
+            Tip:
+                This does the same as :paramref:`switch_inline_query_current_chat`, but for some
+                reason, Telegram has not deprecated this field.
         switch_inline_query_current_chat (:obj:`str`, optional): If set, pressing the button will
             insert the bot's username and the specified inline query in the current chat's input
             field. Can be empty, in which case only the bot's username will be inserted. This
@@ -122,6 +127,20 @@ class InlineKeyboardButton(TelegramObject):
         pay (:obj:`bool`, optional): Specify :obj:`True`, to send a Pay button. This type of button
             **must** always be the **first** button in the first row and can only be used in
             invoice messages.
+        switch_inline_query_chosen_chat (:obj:`telegram.SwitchInlineQueryChosenChat`, optional):
+            If set, pressing the button will prompt the user to select one of their chats of the
+            specified type, open that chat and insert the bot's username and the specified inline
+            query in the input field.
+
+            .. versionadded:: NEXT.VERSION
+
+            Tip:
+                This does the same as :paramref:`switch_inline_query`, but gives more control on
+                which chats can be selected.
+
+            Caution:
+                The PTB team has discovered that this field works correctly only if your Telegram
+                client is released after April 20th 2023.
 
     Attributes:
         text (:obj:`str`): Label text on the button.
@@ -154,6 +173,10 @@ class InlineKeyboardButton(TelegramObject):
             in inline mode when they are currently in a private chat with it. Especially useful
             when combined with ``switch_pm*`` actions - in this case the user will be automatically
             returned to the chat they switched from, skipping the chat selection screen.
+
+            Tip:
+                This does the same as :attr:`switch_inline_query_current_chat`, but for some
+                reason, Telegram has not deprecated this field.
         switch_inline_query_current_chat (:obj:`str`): Optional. If set, pressing the button will
             insert the bot's username and the specified inline query in the current chat's input
             field. Can be empty, in which case only the bot's username will be inserted. This
@@ -165,7 +188,20 @@ class InlineKeyboardButton(TelegramObject):
         pay (:obj:`bool`): Optional. Specify :obj:`True`, to send a Pay button. This type of button
             **must** always be the **first** button in the first row and can only be used in
             invoice messages.
+        switch_inline_query_chosen_chat (:obj:`telegram.SwitchInlineQueryChosenChat`): Optional.
+            If set, pressing the button will prompt the user to select one of their chats of the
+            specified type, open that chat and insert the bot's username and the specified inline
+            query in the input field.
 
+            .. versionadded:: NEXT.VERSION
+
+            Tip:
+                This does the same as :attr:`switch_inline_query`, but gives more control on
+                which chats can be selected.
+
+            Caution:
+                The PTB team has discovered that this field works correctly only if your Telegram
+                client is released after April 20th 2023.
     """
 
     __slots__ = (
@@ -178,6 +214,7 @@ class InlineKeyboardButton(TelegramObject):
         "text",
         "login_url",
         "web_app",
+        "switch_inline_query_chosen_chat",
     )
 
     def __init__(
@@ -191,6 +228,7 @@ class InlineKeyboardButton(TelegramObject):
         pay: bool = None,
         login_url: LoginUrl = None,
         web_app: WebAppInfo = None,
+        switch_inline_query_chosen_chat: SwitchInlineQueryChosenChat = None,
         *,
         api_kwargs: JSONDict = None,
     ):
@@ -207,6 +245,9 @@ class InlineKeyboardButton(TelegramObject):
         self.callback_game: Optional[CallbackGame] = callback_game
         self.pay: Optional[bool] = pay
         self.web_app: Optional[WebAppInfo] = web_app
+        self.switch_inline_query_chosen_chat: Optional[
+            SwitchInlineQueryChosenChat
+        ] = switch_inline_query_chosen_chat
         self._id_attrs = ()
         self._set_id_attrs()
 
@@ -236,6 +277,9 @@ class InlineKeyboardButton(TelegramObject):
         data["login_url"] = LoginUrl.de_json(data.get("login_url"), bot)
         data["web_app"] = WebAppInfo.de_json(data.get("web_app"), bot)
         data["callback_game"] = CallbackGame.de_json(data.get("callback_game"), bot)
+        data["switch_inline_query_chosen_chat"] = SwitchInlineQueryChosenChat.de_json(
+            data.get("switch_inline_query_chosen_chat"), bot
+        )
 
         return super().de_json(data=data, bot=bot)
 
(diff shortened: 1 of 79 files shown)


telegram/_inline/inlinekeyboardbutton.py
```diff
@@ -111,6 +112,10 @@ class InlineKeyboardButton(TelegramObject):
             in inline mode when they are currently in a private chat with it. Especially useful
             when combined with ``switch_pm*`` actions - in this case the user will be automatically
             returned to the chat they switched from, skipping the chat selection screen.
+
+            Tip:
+                This does the same as :paramref:`switch_inline_query_current_chat`, but for some
+                reason, Telegram has not deprecated this field.
```
Bibo-Joshi, line 118: it does not. current chat explicitly switches to the current chat, `switch_inline_query` allows the user to choose *any* chat

telegram/_inline/inlinekeyboardbutton.py
```diff
@@ -122,6 +127,20 @@ class InlineKeyboardButton(TelegramObject):
...
+        switch_inline_query_chosen_chat (:obj:`telegram.SwitchInlineQueryChosenChat`, optional):
+            If set, pressing the button will prompt the user to select one of their chats of the
+            specified type, open that chat and insert the bot's username and the specified inline
+            query in the input field.
+
+            .. versionadded:: NEXT.VERSION
+
+            Tip:
+                This does the same as :paramref:`switch_inline_query`, but gives more control on
+                which chats can be selected.
```
Bibo-Joshi, line 139: "is similar to"

telegram/_inline/inlinekeyboardbutton.py
harshil21: ~darn copilot~, I meant the new parameter hehe

telegram/_inline/inlinekeyboardbutton.py
```diff
@@ -111,6 +112,10 @@ class InlineKeyboardButton(TelegramObject):
             in inline mode when they are currently in a private chat with it. Especially useful
             when combined with ``switch_pm*`` actions - in this case the user will be automatically
             returned to the chat they switched from, skipping the chat selection screen.
+
+            Tip:
+                This is similar to the new parameter :paramref:`switch_inline_query_chosen_chat`,
+                but for some reason, Telegram has not deprecated this parameter.
```
Bibo-Joshi, line 118: But allows the user to select an arbitrary chat.

I see no need to comment in TGs choices if API design...

harshil21: Ok, how about this then: "This is similar to the new parameter :paramref:`switch_inline_query_chosen_chat`, but gives no control over which chats can be selected."

Need to distinguish between both these parameters which essentially do the same thing

telegram/_message.py
```diff
@@ -3522,6 +3526,19 @@ def _parse_markdown(
                             "Spoiler entities are not supported for Markdown version 1"
                         )
                     insert = f"||{escaped_text}||"
+                elif entity.type == MessageEntity.CUSTOM_EMOJI:
+                    if version == 1:
+                        # this ensures compatibility to previous PTB versions
+                        insert = escaped_text
+                    else:
+                        # This should never be needed because ids are numeric but the documentation
+                        # specifically mentions it so here we are
```
Bibo-Joshi, line 3535: Which documentation exactly are we talking about? 🤔

telegram/_message.py
```diff
@@ -3522,6 +3526,19 @@ def _parse_markdown(
                             "Spoiler entities are not supported for Markdown version 1"
                         )
                     insert = f"||{escaped_text}||"
+                elif entity.type == MessageEntity.CUSTOM_EMOJI:
+                    if version == 1:
+                        # this ensures compatibility to previous PTB versions
+                        insert = escaped_text
```
Bibo-Joshi, line 3532: I don't quite understand the comment. Do you mean that before the introduction of this entity type, the alternative emoji would was always being used? By that logic we shouldn't raise ValueError for strikethrough etiher, should we? My point is: I had expected to raise an exception here, too :D

telegram/helpers.py
```diff
@@ -44,23 +44,28 @@
...
         entity_type (:obj:`str`, optional): For the entity types
             :tg-const:`telegram.MessageEntity.PRE`, :tg-const:`telegram.MessageEntity.CODE` and
-            the link part of :tg-const:`telegram.MessageEntity.TEXT_LINK`, only certain characters
+            the link part of :tg-const:`telegram.MessageEntity.TEXT_LINK` and
+            :tg-const:`telegram.MessageEntity.CUSTOM_EMOJI`, only certain characters
             need to be escaped in :tg-const:`telegram.constants.ParseMode.MARKDOWN_V2`.
-            See the official API documentation for details. Only valid in combination with
-            ``version=2``, will be ignored else.
+            See the `official API documentation <https://core.telegram.org/bots/api
+            #formatting-options>`__ for details. Only valid in
```
Bibo-Joshi, line 60:
```suggestion
            #formatting-options>`_ for details. Only valid in
```

telegram/helpers.py
```diff
@@ -44,23 +44,28 @@
...
+            combination with ``version=2``, will be ignored else.
     """
     if int(version) == 1:
         escape_chars = r"_*`["
     elif int(version) == 2:
         if entity_type in ["pre", "code"]:
             escape_chars = r"\`"
-        elif entity_type == "text_link":
+        elif entity_type in ["text_link", "custom_emoji"]:
```
Bibo-Joshi, line 68: I don't see this in the documentation 🤔  Can a `)` even be part of the `tg://emoji?id=<numeric>` thingy?

tests/test_message.py
```diff
@@ -634,17 +638,65 @@ def test_text_markdown_emoji(self):
...
+        assert expected == message[type_]
+
+    @pytest.mark.parametrize(
+        "type_",
+        argvalues=[
+            "text_html",
+            "text_html_urled",
+        ],
+    )
     def test_text_custom_emoji(self, type_):
```
Bibo-Joshi, line 697:
```suggestion
    def test_text_custom_emoji_html(self, type_):
```

telegram/_message.py
Poolitzer: Well we didn't before, because previously it just wasn't supported, so we used the replacement emoji instead. And I would stick to that, it makes more sense to me then raising an error.

telegram/_message.py
Poolitzer: Telegrams ofc. About replacing chars in the () part of inline links and now custom emojis

telegram/helpers.py
Poolitzer: _Inside the (...) part of the inline link and custom emoji definition, all ')' and '\' must be escaped with a preceding '\' character._

:Shrug:

telegram/_message.py
Bibo-Joshi: My quesiton is: If instead of including a custom emoji markup, we fall back to the static emoji, shouldn't we then also fallback to the ordinary text for strikethrough text? i.e. just leave out the markup? I feel like raising for strikethrough but falling back to the static emoji is an inconsistency …

telegram/helpers.py
Bibo-Joshi: Hah. I'm blind. Thanks :)

telegram/_message.py
Bibo-Joshi: Would you mind adding a link to https://core.telegram.org/bots/api#formatting-options?

telegram/_inline/inlinekeyboardbutton.py
Bibo-Joshi: sounds good 👍

telegram/_message.py
Poolitzer: I think it is clear where we get what to escape from, no?

telegram/_message.py
Poolitzer: It is inconsistent, yes. But unlike strikethrough, where we introduced this ValueError when it got added to the MessageEntities right away, we decided to fallback to static emojis when custom emojis got added. No matter the markdown flavor. So I think it would be weird to break this just to keep it consistent.

Maybe we do our fancy new deprecation move here?

telegram/_message.py
Bibo-Joshi: At least I got confused by the "are numeric" part, I expected a documentation about that … but nevermind

telegram/_message.py
Bibo-Joshi: Fair point, makes sense to me, then. But yes, let's deprecate that for md v1 and raise ValueError on the next api update 👍

telegram/_message.py
```diff
@@ -3570,6 +3601,9 @@ def text_markdown(self) -> str:
 
             * |custom_emoji_formatting_note|
 
+        .. deprecated:: NEXT.VERSION
+            |custom_emoji_md1_deprecation|
```
harshil21, line 3605: this tells me the whole method is deprecated, I think we should just modify `custom_emoji_formatting_note` for this or put this in a Caution block instead.

telegram/_writeaccessallowed.py
```diff
@@ -17,21 +17,37 @@
...
 class WriteAccessAllowed(TelegramObject):
     """
-    This object represents a service message about a user allowing a bot added to the attachment
-    menu to write messages. Currently holds no information.
+    This object represents a service message about a user allowing a bot to write messages after
+    adding the bot to the attachment menu or launching a Web App from a link.
 
     .. versionadded:: 20.0
+    .. versionchanged:: NEXT.VERSION
+        Added the optional :attr:`web_app_name` attribute.
```
harshil21, line 33: unnecessary imo, we already mention we added the attribute

harshil21 (commented):
a question or two:
👇🏽

telegram/_message.py
```diff
@@ -578,8 +580,13 @@ class Message(TelegramObject):
 
             .. versionadded:: 20.1
 
-    .. |custom_emoji_formatting_note| replace:: Custom emoji entities will currently be ignored
-        by this function. Instead, the supplied replacement for the emoji will be used.
+    .. |custom_emoji_formatting_note| replace:: Custom emoji entities will be ignored by this
+        function. Instead, the supplied replacement for the emoji will be used.
```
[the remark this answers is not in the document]
Bibo-Joshi: it's not a note, it defines a replacement-thingy that's valid only within the `Message` class :)

telegram/_message.py
Bibo-Joshi: image

I'd read that as "the following is behavior is deprecated since NEXT.VERSION: …" …

telegram/_message.py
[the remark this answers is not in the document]
Bibo-Joshi: To clarify: Once we raise an exception instead, it can ofc be removed

PR #3674, opened 2023-04-21
Improve Warning Categories & Stacklevels

Bibo-Joshi:
* Fixed a few warning categories that were missed in API 6.6.
* Unifies warning handling in `Bot` and `ExtBot`, ensuring that the stacklevel is correct for both
* adds missing tests & modifies existing ones

TODO:

- [x] Update all warning tests to also check the category
- [x] Update tests to double check stacklevel in ExtBot vs Bot

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -273,7 +273,7 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
                 warning_string = "request"
 
         if warning_string:
-            warn(
+            self._warn(
                 f"You set the HTTP version for the {warning_string} HTTPXRequest instance to "
                 f"HTTP/2. The self hosted bot api instances only support HTTP/1.1. You should "
                 f"either run a HTTP proxy in front of it which supports HTTP/2 or use HTTP/1.1.",
@@ -337,6 +337,14 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
         .. versionadded:: 20.0
         """
         return self._private_key
+
+    @classmethod
+    def _warn(
+        cls, message: str, category: Type[Warning] = PTBUserWarning, stacklevel: int = 0
+    ) -> None:
+        """Convencience method to issue a warning. This method is here mostly to make it easier
+        for ExtBot to add 1 level to all warning calls."""
+        warn(message=message, category=category, stacklevel=stacklevel + 1)
 
     def __reduce__(self) -> NoReturn:
         """Customizes how :func:`copy.deepcopy` processes objects of this type. Bots can not
@@ -1157,7 +1165,7 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
 
         """
         thumbnail_or_thumb: FileInput = warn_about_thumb_return_thumbnail(
-            deprecated_arg=thumb, new_arg=thumbnail
+            deprecated_arg=thumb, new_arg=thumbnail, warn_callback=self._warn
         )
         data: JSONDict = {
             "chat_id": chat_id,
@@ -1294,7 +1302,7 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
 
         """
         thumbnail_or_thumb: FileInput = warn_about_thumb_return_thumbnail(
-            deprecated_arg=thumb, new_arg=thumbnail
+            deprecated_arg=thumb, new_arg=thumbnail, warn_callback=self._warn
         )
 
         data: JSONDict = {
@@ -1531,7 +1539,7 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
 
         """
         thumbnail_or_thumb: FileInput = warn_about_thumb_return_thumbnail(
-            deprecated_arg=thumb, new_arg=thumbnail
+            deprecated_arg=thumb, new_arg=thumbnail, warn_callback=self._warn
         )
         data: JSONDict = {
             "chat_id": chat_id,
@@ -1664,7 +1672,7 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
 
         """
         thumbnail_or_thumb: FileInput = warn_about_thumb_return_thumbnail(
-            deprecated_arg=thumb, new_arg=thumbnail
+            deprecated_arg=thumb, new_arg=thumbnail, warn_callback=self._warn
         )
         data: JSONDict = {
             "chat_id": chat_id,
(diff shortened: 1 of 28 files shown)


tests/test_bot.py
```diff
@@ -1676,6 +1678,13 @@ async def test_http2_runtime_error(self, recwarn):
             "You set the HTTP version for the get_updates_request and request HTTPXRequest "
             "instance" in str(recwarn[2].message)
         )
+        for warning in recwarn:
+            print()
+            print(warning)
```
harshil21, line 1683:
```suggestion
```

tests/ext/test_picklepersistence.py
```diff
@@ -891,7 +892,12 @@ async def test_custom_pickler_unpickler_simple(
         await pickle_persistence.update_chat_data(12345, data_with_bot)
         assert len(recwarn) == 1
         assert recwarn[-1].category is PTBUserWarning
+        assert recwarn[-1].category is PTBUserWarning
```
harshil21, line 895:
```suggestion
        assert recwarn[-1].category is PTBUserWarning
```

tests/ext/test_conversationhandler.py
```diff
@@ -1042,12 +1050,27 @@ async def test_no_running_job_queue_warning(self, app, bot, user1, recwarn, jq):
...
             else:
                 assert len(recwarn) == 2
+                assert (
+                    Path(recwarn[0].filename)
+                    == PROJECT_ROOT_PATH / "telegram" / "ext" / "_conversationhandler.py"
+                ), "wrong stacklevel!"
+                assert (
+                    Path(recwarn[1].filename)
+                    == PROJECT_ROOT_PATH / "telegram" / "ext" / "_conversationhandler.py"
+                ), "wrong stacklevel!"
```
harshil21, line 1066:
```suggestion
                for i in (0, 1):
                    assert (
                        Path(recwarn[i].filename)
                        == PROJECT_ROOT_PATH / "telegram" / "ext" / "_conversationhandler.py"
                    ), "wrong stacklevel!"
```

harshil21 (commented):
nice catches

tests/test_bot.py
Bibo-Joshi: I'm still working on that ;)

telegram/_bot.py
```diff
@@ -1157,7 +1165,10 @@ async def send_audio(
 
         """
         thumbnail_or_thumb: FileInput = warn_about_thumb_return_thumbnail(
-            deprecated_arg=thumb, new_arg=thumbnail
+            deprecated_arg=thumb,
+            new_arg=thumbnail,
+            warn_callback=self._warn,
+            stacklevel=4,
```
lemontree210, line 1171: I just thought of one thing: could this indication of stacklevel be some sort of violation of Demeter law or a similar rule? It's like a class whose method is being called knows something about what is calling it and through how many "layers".

tests/auxil/deprecations.py
```diff
@@ -72,3 +72,16 @@ def check_thumb_deprecation_warnings_for_args_and_attrs(
         )
 
     return True
+
+
+def check_thumb_deprececation_warning_for_method_args(
```
lemontree210, line 77:
```suggestion
def check_thumb_deprecation_warning_for_method_args(
```
+ all usages.
I know that's a pun on @Poolitzer 's depreciation 😉

lemontree210 (changes requested):
I'm obviously out of my depth here, but one typo I did catch! :)

Bibo-Joshi:
Just a related note: stackleveling will get way more fun soon :D https://github.com/python-telegram-bot/python-telegram-bot/projects/8#card-88970053

telegram/_bot.py
Bibo-Joshi: mh, the only thing this method knows is how many layers *inside* of the `Bot` class there are between this line and the actual interface of the class. IMO that's no forbidden info …

lemontree210: so a `Bot` being an attribute of `Application` doesn't affect `stacklevel`?

Bibo-Joshi: No. we just want to know how many layers there are between `await bot.send_photo` and `warnings.warn`. In this case the stack is:

1. the call to `bot.send_photo`
2.  the `@_log` decorator
3. the `warn_about_thumb_return_thumbnail` function
4. the `warn_about_deprecated_arg_return_new_arg` function
5. * the `warn_callback` argument of that function, here `self._warn`
6. * the function `telegram._utils.warnings.warn`
7. the actual call to `warnings.warn`

To issue the warning form the `bot.send_photo` line, we need a stacklevel of 6 - that is 6 layers above `warnings.warn`. The two calls marked with * basically already have this "Demeter law" in mind - they add + 1 to the stacklevel so that you don't have to care about them from where you call them.
One could indeed argue that this should also be the case for the two `warn_about_*` functions - you're right here. If we do that, then we'd have to set a stacklevel of 2 in line 1171, which then only "knows about" the `@_log` decorator. That's really a `Bot`-class internal thingy though and whoever issues a warning from somewhere within `Bot` can see if they need to consider the decorator or not.

lemontree210: Great, thanks for the explanation!

tests/_files/test_animation.py
```diff
@@ -191,6 +194,19 @@ async def make_assertion(url, request_data: RequestData, *args, **kwargs):
...
         assert await bot.send_animation(animation=animation, chat_id=chat_id)
 
+    @pytest.mark.parametrize("bot_class", ["Bot", "ExtBot"])
+    async def test_send_animation_thumb_deprecation_warning(
+        self, recwarn, monkeypatch, bot_class, bot, raw_bot, chat_id, animation
+    ):
+        async def make_assertion(url, request_data: RequestData, *args, **kwargs):
+            return True
+
+        bot = raw_bot if bot_class == "Bot" else bot
```
harshil21, line 204:
```suggestion
        bot = request.getfixturevalue("raw_bot/bot")
```
this would be way to dynamically load a fixture, just for reference. I'm okay with that approach also

harshil21 (commented):
much more extensive testing, I like it

harshil21:
I wonder if there is a way to dynamically inspect the stacklevel and pass the correct value for cases like those...

Bibo-Joshi:
@harshil21 I guess there is <a href="https://docs.python.org/3/library/inspect.html?highlight=inspect#inspect.stack">inspect.stack</a>. I&#x27;m not so sure if I want to get the inspect module involved for this though ...

Bibo-Joshi:
Test failures are unrelated - see #3685 and #3673. Merging.

PR #3676, opened 2023-04-23
Recover 100% Type Completeness

Bibo-Joshi:
Noticed by chance that type completeness has dropped. Trying to fix both that and the corresponding worflow. Workflow first.

Based on the API 6.7 branch because that's where I noticed stuff going wrong. Master may have been better, but now it's too late :D 

To be merged after #3673 

- [x] Reactivate unit tests
- [x] remove `api-6.7` branch from workflow file


Bibo-Joshi:
FYI: apparently it's still not possible to display the type completeness status in the checks overview (similar to what codecov does). I openend https://github.com/orgs/community/discussions/53528 about that.

PR #3692, opened 2023-05-05
Use explicit optional instead of implicit optional in whole project

MiguelX413:
### Checklist for PRs

* New classes:
    - [ ] `__init__` accepts `api_kwargs` as kw-only

* Added new shortcuts:

* If relevant:

    - [ ] Link new and existing constants in docstrings instead of hard coded number and strings
    - [ ] Add new message types to `Message.effective_attachment`
      - [ ] Add the handlers to the warning loop in the `ConversationHandler`
    - [ ] Added the new method(s) to `_extbot.py`
    - [ ] Updated the Bot API version number in all places: `README.rst` and `README_RAW.rst` (including the badge), as well as `telegram.constants.BOT_API_VERSION_INFO`

This was causing mypy to have problems verifying the integrity of my own codebase which depended on this project.


Bibo-Joshi:
Hi. Thanks for the PR. Does the current implicit typing have any negative effect on your usage of the python-telegram-bot library? If it does not, then I would like to reject this PR for the reasons discussed in #3367.

MiguelX413:
> Hi. Thanks for the PR. Does the current implicit typing have any negative effect on your usage of the python-telegram-bot library? If it does not, then I would like to reject this PR for the reasons discussed in #3367.
> 
> Edit: Ah, I didnt see your comment at the very end. Can you give an example of how exactly the current status affects the type checking of your code base?

@Bibo-Joshi Mypy says the following about https://github.com/MiguelX413/IgTgBot/blob/4586c01fb89283eada0b2c1567e2ee67a02d1bd2/formatted_text.py#L35-L81 only without this PR:
```formatted_text.py:53: error: Argument "url" to "MessageEntity" has incompatible type "Optional[str]"; expected "str"  [arg-type]
formatted_text.py:54: error: Argument "user" to "MessageEntity" has incompatible type "Optional[User]"; expected "User"  [arg-type]
formatted_text.py:55: error: Argument "language" to "MessageEntity" has incompatible type "Optional[str]"; expected "str"  [arg-type]
formatted_text.py:56: error: Argument "custom_emoji_id" to "MessageEntity" has incompatible type "Optional[str]"; expected "str"  [arg-type]
formatted_text.py:75: error: Argument "url" to "MessageEntity" has incompatible type "Optional[str]"; expected "str"  [arg-type]
formatted_text.py:76: error: Argument "user" to "MessageEntity" has incompatible type "Optional[User]"; expected "User"  [arg-type]
formatted_text.py:77: error: Argument "language" to "MessageEntity" has incompatible type "Optional[str]"; expected "str"  [arg-type]
formatted_text.py:78: error: Argument "custom_emoji_id" to "MessageEntity" has incompatible type "Optional[str]"; expected "str"  [arg-type]```

**Comment by Bibo-Joshi** [comment]
I see. Since there is a user impact then, we should indeed completely drop the implicit optionals in the whole library. For this, the mypy config file needs to be updated and all the function/method signatures need to be updated. I would guess that some regex-replacing + mypy checking will make this a rather straightforward task. @MiguelX413 would you maybe like to extend your PR?

**Comment by MiguelX413** [comment]
> I see. Since there is a user impact then, we should indeed completely drop the implicit optionals in the whole library. For this, the mypy config file needs to be updated and all the function/method signatures need to be updated. I would guess that some regex-replacing + mypy checking will make this a rather straightforward task. @MiguelX413 would you maybe like to extend your PR?

Sure!

**Comment by MiguelX413** [comment]
> I see. Since there is a user impact then, we should indeed completely drop the implicit optionals in the whole library. For this, the mypy config file needs to be updated and all the function/method signatures need to be updated. I would guess that some regex-replacing + mypy checking will make this a rather straightforward task. @MiguelX413 would you maybe like to extend your PR?

Sure! @Bibo-Joshi

**Comment by Bibo-Joshi** [comment]
Thanks for the updates! A number of tests fail because the type hints of shortcut messages don't match the type hint in the `Bot` class (e.g. `Message.reply_text` doesn't match `Bot.send_message`). IISC that's due to some classes being importing only `if TYPE_CHECKING` and the tests can't resolve the forward references. If you need help with addressing this, please do reach out.

On a side note: please avoid force-pushing (at least once you got review comments), since that makes it hard to keep track of what one has already reviewed :)

**Comment by MiguelX413** [comment]
> Thanks for the updates! A number of tests fail because the type hints of shortcut messages don't match the type hint in the `Bot` class (e.g. `Message.reply_text` doesn't match `Bot.send_message`). IISC that's due to some classes being importing only `if TYPE_CHECKING` and the tests can't resolve the forward references. If you need help with addressing this, please do reach out.
> 
> On a side note: please avoid force-pushing (at least once you got review comments), since that makes it hard to keep track of what one has already reviewed :)

Noted!

**Comment by MiguelX413** [comment]
> Thanks for the updates! A number of tests fail because the type hints of shortcut messages don't match the type hint in the `Bot` class (e.g. `Message.reply_text` doesn't match `Bot.send_message`). IISC that's due to some classes being importing only `if TYPE_CHECKING` and the tests can't resolve the forward references. If you need help with addressing this, please do reach out.
> 
> On a side note: please avoid force-pushing (at least once you got review comments), since that makes it hard to keep track of what one has already reviewed :)

I'm not really sure where to start with the testing system tbh @Bibo-Joshi

**Comment by lemontree210** [comment]
> I'm not really sure where to start with the testing system tbh @Bibo-Joshi 

I tried to fix the testing issues. Also, I merged `master` with some conflict resolution, but some new methods that were added to our `master` and were merged without conflicts contain the implicit optionals and you might want to run your script again :)

File: tests/auxil/bot_method_checks.py [hunk 1]
```diff
@@ -62,6 +72,33 @@ def check_shortcut_signature(
...
+            try:
+                return eval(option)
+            # NameError will be raised if trying to eval just name and it doesn't work, e.g.
+            # "Name 'ApplicationBuilder' is not defined".
+            # AttributeError will be raised if trying to e.g. eval f"telegram.{name}" when the
+            # class denoted by `name` actually belongs to `telegram.ext`:
+            # "module 'telegram' has no attribute 'ApplicationBuilder'".
+            # If neither option works, this is not a PTB class.
+            except (NameError, AttributeError):
+                continue
```
Bibo-Joshi, line 97:
```suggestion
            attr = getattr(module, name, None)
            if attr is not None:
                return None
```

I'd like to not rely on `eval` if not strictly necessary :)

tests/auxil/bot_method_checks.py
```diff
@@ -92,10 +129,31 @@ def check_shortcut_signature(
...
                     )
+            elif FORWARD_REF_PATTERN.search(str(shortcut_sig.parameters[kwarg])):
+                # If a shortcut signature contains a ForwardRef, the simple comparison of
+                # annotations can fail. Try and resolve the .__args__, then compare them.
+
+                for shortcut_arg, bot_arg in zip(
+                    shortcut_sig.parameters[kwarg].annotation.__args__,
+                    bot_sig.parameters[kwarg].annotation.__args__,
+                ):
+                    shortcut_arg_to_check = shortcut_arg  # for ruff
```
Bibo-Joshi, line 130: why?

tests/auxil/bot_method_checks.py
```diff
@@ -92,10 +129,31 @@ def check_shortcut_signature(
...
+                    m = FORWARD_REF_PATTERN.search(str(shortcut_arg))
```
Bibo-Joshi, line 131:
```suggestion
                    match = FORWARD_REF_PATTERN.search(str(shortcut_arg))
```

Bibo-Joshi (changes requested):

tests/auxil/bot_method_checks.py
lemontree210: hmm... why are we not doing this in the admonitions?

tests/auxil/bot_method_checks.py
lemontree210: ruff says I'm assigning something to a variable that is a part of the sequence I'm looping over, and that's bad apparently :)

tests/auxil/bot_method_checks.py
lemontree210: I don't understand the `return None` bit. We want to return a class.

tests/auxil/bot_method_checks.py
lemontree210: Fixed

tests/auxil/bot_method_checks.py
Bibo-Joshi: > then maybe return module.name?

yes ofc. my bad 🤦

tests/auxil/bot_method_checks.py
Bibo-Joshi: mh, kind of makes sense

tests/auxil/bot_method_checks.py
harshil21, line 1: can this file get a docstring?

tests/auxil/bot_method_checks.py
```diff
@@ -62,6 +70,33 @@ def check_shortcut_signature(
...
+        """Attempts to resolve a PTB class from a ForwardRef.
+
+        E.g. resolves <class 'telegram._files.sticker.StickerSet'> from "StickerSet".
+
+        Returns a class on success, :obj:`None` if nothing could be resolved.
+        """
+
+        for option in (
+            name,
+            f"telegram.{name}",
```
harshil21, line 84:
```suggestion
            f"telegram.{name}",
            f"telegram.request.{name}",
```
Do we need to include the other modules? This function only checks bot methods isn't it? So we don't need to check ext.filters either correct?

tests/auxil/bot_method_checks.py
lemontree210: Done

tests/auxil/bot_method_checks.py
lemontree210: OK :)

Bibo-Joshi (approved):
LGTM 👍 @harshil21 do you have anything to add?

harshil21:
@Bibo-Joshi nope LGTM ✅

Bibo-Joshi:
Thank you very much for your contribution @MiguelX413 !

PR #3694, opened 2023-05-06
Update filters.py

ibragimovgeorge:
updated use case in docs for proper using filters.Entity


Bibo-Joshi:
Thanks for the contribution!

PR #3695, opened 2023-05-07
Make JobQueues APS-Interface more explicit

Bibo-Joshi:
Inspired by python-telegram-bot/ptbcontrib#72 and python-telegram-bot/ptbcontrib#73

### Checklist for PRs


telegram/ext/_jobqueue.py
```diff
@@ -626,7 +650,7 @@ def jobs(self) -> Tuple["Job[CCT]", ...]:
             Tuple[:class:`Job`]: Tuple of all *scheduled* jobs.
         """
         return tuple(
-            Job._from_aps_job(job)  # pylint: disable=protected-access
+            Job.from_aps_job(job)  # pylint: disable=protected-access
```
lemontree210, line 653:
```suggestion
            Job.from_aps_job(job)
```
We're not accessing a protected method anymore?

telegram/ext/_jobqueue.py
```diff
@@ -158,6 +158,30 @@ def application(self) -> "Application[Any, CCT, Any, Any, Any, JobQueue[CCT]]":
             return application
         raise RuntimeError("The application instance is no longer alive.")
 
+    @staticmethod
+    async def job_callback(job_queue: "JobQueue[CCT]", job: "Job[CCT]") -> None:
+        """This method is used as a callback for the APScheduler jobs.
+
+        More precisely, the ``func`` argument of :class:`apscheduler.job.Job` is set to this method
+        and the ``arg`` argument is set to a tuple containing the :class:`JobQueue`
+        itself and the :class:`~telegram.ext.Job` instance.
```
lemontree210, line 167:
```suggestion
        and the ``arg`` argument (representing positional arguments to ``func``) is set to a tuple
        containing the :class:`JobQueue` itself and the :class:`~telegram.ext.Job` instance.
```
Just for clarity

telegram/ext/_jobqueue.py
```diff
@@ -230,11 +254,11 @@ async def callback(context: CallbackContext)
         date_time = self._parse_time_input(when, shift_day=True)
 
         j = self.scheduler.add_job(
-            job.run,
+            self.job_callback,
             name=name,
             trigger="date",
             run_date=date_time,
-            args=(self.application,),
+            args=(self, job),
```
lemontree210, line 261: Wow :) `args` are passed as args to a callback (self.job_callback), which then passes itself as a callback and `JobQueue` and `Job` as args to this callback. It feels like falling down a tunnel with mirrors ;)

lemontree210 (commented):
Not much I can contribute here except probably ask for more clarity as to what benefits using this new method offers.

telegram/ext/_jobqueue.py
```diff
@@ -158,6 +158,30 @@ def application(self) -> "Application[Any, CCT, Any, Any, Any, JobQueue[CCT]]":
...
 
+    @staticmethod
+    async def job_callback(job_queue: "JobQueue[CCT]", job: "Job[CCT]") -> None:
+        """This method is used as a callback for the APScheduler jobs.
+
+        More precisely, the ``func`` argument of :class:`apscheduler.job.Job` is set to this method
+        and the ``arg`` argument is set to a tuple containing the :class:`JobQueue`
+        itself and the :class:`~telegram.ext.Job` instance.
+
+        Tip:
```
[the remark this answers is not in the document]
Bibo-Joshi: Felt like one to me :D Don't think that there is a clear distinction …

telegram/ext/_jobqueue.py
```diff
@@ -158,6 +158,30 @@ def application(self) -> "Application[Any, CCT, Any, Any, Any, JobQueue[CCT]]":
...
+            This method is a static method rather than a bound method. This makes the arguments
+            more transparent and allows for easier handling of PTBs integration of APScheduler
+            when utilizing advanced features of APScheduler.
+
+        Hint:
+            This method is effectively a shortcut for :meth:`telegram.ext.Job.run`.
```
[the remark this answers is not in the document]
Bibo-Joshi: rephrased to "wrapper"

telegram/ext/_jobqueue.py
Bibo-Joshi: That's pretty much the point of this PR :D As the PRs linked in the description show, the current status is that

* what we pass to APS is intransparent to the PTB users, making it hard for them to implement a PTB-conformant JobStore
* Because we use `Job.run` - a bound method - as callback, we have to rely on `function.__self__` in `from_aps_job` but that breaks as soon as the link between the instance of `ext.Job` and `Job.run` get's lost, i.e. if you call `Job.run(some_job)` instead of `some_job.run()`. This makes it event harder to implement APS-logic, especially since `_from_aps_job` is protected so far

The downside of this PR is that we expose the "backend" (APS) more to the user, which makes it a bit harder for us to ditch/upgrade APS if ever needed (APS is preparing v4 since before we incorporated it in ptbv13 :D). Then again, we have a fancy new stability policy in place which makes it rather clear what we'll have to do in that case.

Bibo-Joshi:
Another comment on this.

This PR does *not* ensure that the classes `Job` and `JobQueue` itself can be easily persisted/pickled. This is because

* The `ext.Job` instance has a reference to the `aps.Job` instance and this cycling reference can be a problem
* The `JobQueue` has a reference to the APS Scheduler and a weak reference to the `Application` both of which are surely not easy to persist

IMO making `ext.{Job, JobQueue}` pickable out ouf the box is a non-goal of this PR. Serializing a `APS Job` will still be possible with reasonable effort with the following two considerations:

* the first argument of the APS Job callback will be the `JobQueue`. By storing a reference to that in the `JobStore` itself, it will be straight forward to re-insert that on deserializing the APS Job
* the second argument of the APS Job callback will be the `ext.Job`. Instead of serializing that, one can instead serialize the attributes (`callback`, `chat_id`, …)

telegram/ext/_jobqueue.py
```diff
@@ -821,8 +842,25 @@ def next_t(self) -> Optional[datetime.datetime]:
...
 
     @classmethod
-    def _from_aps_job(cls, job: "APSJob") -> "Job[CCT]":
-        return job.func.__self__
+    def from_aps_job(cls, job: "APSJob") -> "Job[CCT]":
+        """Provides the :class:`telegram.ext.Job` that is associated with the given APScheduler
+        job.
+
+        Tip:
+            This method can be useful when using advanced APScheduler features along with
```
harshil21, line 850: maybe add an `Example:` block here? I think rescheduling a job is fairly common no?

telegram/ext/_jobqueue.py
```diff
@@ -821,8 +842,25 @@ def next_t(self) -> Optional[datetime.datetime]:
...
+
+        Args:
+            job (:class:`apscheduler.job.Job`): The APScheduler job
+
+        Returns:
+            :class:`telegram.ext.Job`
+        """
+        job_ = job.args[1]
+        job_._job = job  # pylint: disable=protected-access
+        return job_
```
harshil21, line 863: csi comments / rename variables here? `job`, `_job`, `job_` need to be different from each other haha

telegram/ext/_jobqueue.py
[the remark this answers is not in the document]
harshil21: Tip is okay with me

telegram/ext/_jobqueue.py
Bibo-Joshi: rescheduling doesn't require the method though … just `ext_job(.job).modify(…)`

telegram/ext/_jobqueue.py
harshil21: Perhaps changing the argument name `job` ->  `aps_job` will be better

telegram/ext/_jobqueue.py
harshil21: oh right, this returns a telegram.ext Job. Okay then nothing more is needed

harshil21 (approved):
great, lgtm

telegram/ext/_jobqueue.py
lemontree210: Oh, sorry I didn't reply explicitly that I read and acknowledged this :)

lemontree210 (approved):

PR #3709, opened 2023-05-15
Fix incorrect and inconsistent timeout types

revolter:

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -3586,7 +3586,7 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
         self,
         offset: Optional[int] = None,
         limit: Optional[int] = None,
-        timeout: Optional[float] = None,
+        timeout: Optional[int] = None,
         allowed_updates: Optional[Sequence[str]] = None,
         *,
         read_timeout: float = 2,
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -673,7 +673,7 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
         Args:
             poll_interval (:obj:`float`, optional): Time to wait between polling updates from
                 Telegram in seconds. Default is ``0.0``.
-            timeout (:obj:`float`, optional): Passed to
+            timeout (:obj:`int`, optional): Passed to
                 :paramref:`telegram.Bot.get_updates.timeout`. Default is ``10`` seconds.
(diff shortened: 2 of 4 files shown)


Bibo-Joshi:
Hi. I'm not quite sure what this PR is aiming at. Would you mind explaining where you see inconsistency in the types? IISC we use `float` everywhere.

revolter:
Yes, sure, sorry for not explaining it from the start.

So, knowing that the [official `getUpdates` API](https://core.telegram.org/bots/api#getupdates) declares `timeout` as `Integer`, here are some of the inconsistencies:
1. The `Updater.start_polling` correctly declares it as `int`:

   https://github.com/python-telegram-bot/python-telegram-bot/blob/57c780c62f5c1a3ee8441d6b690dd3f5a3d22288/telegram/ext/_updater.py#L187

   but it's incorrectly documented as `float`:

   https://github.com/python-telegram-bot/python-telegram-bot/blob/57c780c62f5c1a3ee8441d6b690dd3f5a3d22288/telegram/ext/_updater.py#L205
2. The `Bot.get_updates` incorrectly declares it as `float`:

   https://github.com/python-telegram-bot/python-telegram-bot/blob/57c780c62f5c1a3ee8441d6b690dd3f5a3d22288/telegram/_bot.py#L3589

   although it's correctly documented as `int`:

   https://github.com/python-telegram-bot/python-telegram-bot/blob/57c780c62f5c1a3ee8441d6b690dd3f5a3d22288/telegram/_bot.py#L3624

So, the bottom line is that all of them should be `int`, which is what this PR does.

telegram/_bot.py
```diff
@@ -3661,7 +3661,7 @@ async def get_updates(
             await self._post(
                 "getUpdates",
                 data,
-                read_timeout=read_timeout + timeout if timeout else read_timeout,
+                read_timeout=read_timeout + float(timeout) if timeout else read_timeout,
```
Bibo-Joshi, line 3664: float + integer should work without explicit conversion, shoudn't it? 🤔

Bibo-Joshi (commented):
Thanks for the explanation and the catch! Yes, I would like to have the merged, then :) Would you mind resolving the merge conflicts?

telegram/_bot.py
revolter: I don't know, actually 😅 I'm coming from Swift, and I thought it doesn't work like that. So I'll revert this change 👍🏻

Bibo-Joshi:
Than you for the contribution!

revolter:
And thank you for merging!

PR #3716, opened 2023-05-20
Auto-Update Changed Version in Other Files After Dependabot PRs

Bibo-Joshi:
This new workflow automatically copies versions changed by depandabot to the pre-commit config, the readme and any other file where the version might appear. It's a mit naive, b/c it really just searches for exact matches of the version number, but I guess the cases were we have that exact version number appear elsewhere in the repo are seldom enough. Plus, we can review.

Live examples of this workflow can be seen at https://github.com/Bibo-Joshi/actions-tests/pulls (where you can also see that I got this working [*almost* immediately](https://github.com/Bibo-Joshi/actions-tests/commits/main) 🫠)


harshil21 (commented):
wow nice! Can't we tell it which files to look for though, so error rate can become 0%

Bibo-Joshi:
That was also my inital thought, but that might lead to false negatives. As those can't be seen in a PR diff, I decided that false positives might be the better.

harshil21 (approved):
Alright then, LGTM

PR #3719, opened 2023-05-24
Upgrade Furo to support sphinx 7

harshil21:
Upgrades the furo version, so we can now also upgrade the sphinx version.


Bibo-Joshi:
Would you mind just also bumping sphinx in this PR? :)

Issue #3732, opened 2023-06-02
Adopt Py3.8+ features: Drop workaround in `PicklePersistence`


Bibo-Joshi:
With dropping python 3.7 it's possible to remove a py-37-specific workorund in the `PicklePersistence` implementation. More details can be found [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/bf54599618c50b472791e5a172c2e73e8548604e/telegram/ext/_picklepersistence.py#L77-L83). We should revisit that implementation and remove all python-3.7 specific code. The corresponding test cases also needs to be adapted to drop any python 3.7 related clauses.

A PR in this direction is very welcome. As long as #3728 is open, it should be based on the `drop-py-37` branch. If you would like to send a PR, please be sure to read our [contribution guide](https://docs.python-telegram-bot.org/en/stable/contributing.html) and leave a short comment so that we can assign you.

PR #3740, opened 2023-06-03
Remove Python 3.7 workaround for `PicklePersistence`

harshil21:
Closes https://github.com/python-telegram-bot/python-telegram-bot/issues/3732 

I don't think anything in tests need to change?


telegram/ext/_picklepersistence.py
```diff
@@ -49,47 +47,34 @@ def _reconstruct_to(cls: Type[TelegramObj], kwargs: dict) -> TelegramObj:
     """
     This method is used for unpickling. The data, which is in the form a dictionary, is
     converted back into a class. Works mostly the same as :meth:`TelegramObject.__setstate__`.
-    This function should be kept in place for backwards compatibility even if the pickling logic
-    is changed, since `_custom_reduction` places references to this function into the pickled data.
```
Bibo-Joshi, line 53: Why was this comment removed? The contents `_custom_recdution` of just went into `reducer_override` and this still inserts a refercne to `_reconstruct_to` …

Bibo-Joshi (commented):
I guess the tests don't have explicit cases for py 3.7, right. Howover, I am wondering if it is possible to pickle data in 3.7 and unpickle it in 3.8 🤔 Could you check?

telegram/ext/_picklepersistence.py
harshil21: I'm confused by the statement "This function should be kept in place for backwards compatibility". This function needs to be there for 3.8+ as well, since the return of `reducer_override` [needs](https://docs.python.org/3/library/pickle.html#object.__reduce__) a callable object anyway.

Bibo-Joshi: image

You're confused? ;P Regardless of other reasons to keep this method, the point is the following:

1. you pickle stuff. while pickling, a reference to `telegram.ext._picklepersisence.__reconstruct_to` is inserted
2. We upgrade you the library, removing `__reconstruct_to`
3. you update ptb locally and try to load the data
4. reference to `telegram.ext._picklepersisence.__reconstruct_to` can not be resolved
5. 💥

harshil21: > image
> 
> You're confused? ;P

weren't my words 🫣 https://github.com/python-telegram-bot/python-telegram-bot/pull/2893#discussion_r808418514, 
I had thought that comment pertained to the py3.7 special case

> the point is the following:

Ah I get it now, I just thought that python internally did something different but now that I say that out loud I realize that sounds stupid haha

I'll add back the _custom_reduction function again.

Bibo-Joshi: > weren't my words

nononononono, only git blame matters ;P

👍

Bibo-Joshi (approved):
Changes LGTM 👍 Did you have any chance to test if py37-pickled files can be loaded in py38?

PR #3742, opened 2023-06-03
Adopt Py3.8+ features: `missing_ok` Parameter

### LINKED ISSUE #3730: Adopt Py3.8+ features: `missing_ok` Parameter
Python 3.8 adds a `missing_ok` argument to the method [`Path.unlink`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink). This method is used in our test suite at a few places. Setting `missing_ok=True` may help avoiding unexpected errors in the test runs.

Trifase:
Fixes #3730


harshil21 (approved):
LGTM, thanks

Bibo-Joshi:
Thanks for the contribution!

PR #3748, opened 2023-06-06
add walrus operator where possible

thefunkycat:
I have tried to add the walrus operator wherever possible.

I went through most files.
There were a couple more instances I found where I could have added it, but it was unnecessary and would not have necessarily improved readability or code quality.

I was more lenient when it came to adding the walrus operator to tests


thefunkycat:
wait, I branched from the wrong branch ... let me fix that

PR #3728, opened 2023-06-02
Drop Support for Python 3.7

Bibo-Joshi:
As Python 3.7 [reaches EOL at the end of this month](https://devguide.python.org/versions/), it's time to say goodbye.

Before merging this PR, we should try to include as many py3.8-only features as reasonably possible without too much delay. This includes the following issues:

- [x] #3729
- [x] #3730
- [x] #3731
- [x] #3732
- [x] #3733
- [x] #3734
- [x] #3735
- [x] Moreover, we should check if there are any remaining python 3.7 related special cases in the test suite.
- [x] Revert https://github.com/python-telegram-bot/python-telegram-bot/pull/3728/commits/03d3c33570fe7cabc51ca5fc275d49704944fdc1

PRs to mention:

- [ ] #3742 
- [ ] #3749 
- [ ] #3740
- [ ] #3754 
- [ ] #3753 
- [ ] #3764 
- [ ] #3762 
- [ ] #3759

PR #3751, opened 2023-06-09
Add Python 3.12 beta to test matrix

harshil21:
To be merged after #3728

Things changed for us:
- Tests: Fix regex pattern match in test_picklepersistence: https://github.com/python/cpython/pull/31782
- Tests: Slightly better formatted error message for missing abstractmethods. The test also now uses [`__abstractmethods__`](https://peps.python.org/pep-3119/) to dynamically get all defined abstract methods.
- `app.create_task` should not accept generators. See below post


harshil21:
@harshil21 <pre>def gen():
    yield
    
await asyncio.create_task(gen())

</pre>does not work in python 3.12. Changing into an <code>async def</code> does not work either.

The <a href="https://docs.python.org/3.12/whatsnew/3.12.html#asyncio">3.12 changelog notes</a>:

<i>&quot;asyncio.iscoroutine() now returns False for generators as asyncio does not support legacy generator-based coroutines.&quot;

</i>Does this mean we should drop support for generators in <code>app.create_task</code> too?

Bibo-Joshi:
> Does this mean we should drop support for generators in app.create_task too?

Not for py 3.11-, as that would be a breach of our stability policy. We can think about deprecating it, but dropping support unannounced is a no-go.
For 3.12+, to stay in line with our stability poclicy, we should try to still support it + add deprecation notices. If there is no feasible workaround, we ofc don't have a chance.

harshil21:
@Bibo-Joshi okay, now how do I specify different type hints for the <code>coroutine</code> parameter for different python versions? Should I just use a <code>if sys.version_info.. : @overload def create_task(...)</code> thingy?

and what kind of workaround do you suggest for py3.12+? I don&#x27;t see any

Bibo-Joshi:
> @Bibo-Joshi okay, now how do I specify different type hints for the <code>coroutine</code> parameter for different python versions? Should I just use a <code>if sys.version_info.. : @overload def create_task(...)</code> thingy?

Sounds good if that works reasonably well. Don't know if it's possible to make only the overload part the content of the if clause or if you have to define the method twice 

> and what kind of workaround do you suggest for py3.12+? I don&#x27;t see any

I don't have anything in mind TBH, haven't looked into it yet

telegram/ext/_application.py
```diff
@@ -78,6 +79,19 @@
 _STOP_SIGNAL = object()
 _DEFAULT_0 = DefaultValue(0)
 
+
+# Since python 3.12, the coroutine passed to create_task should not be an async generator. Remove
+# this check when we drop support for python 3.11.
+if sys.version_info >= (3, 12):
+    _CoroType = Union["asyncio.Future[object]", Awaitable[RT]]
+    _ErrorCoroType = Optional[Union[Optional["asyncio.Future[object]"], Awaitable[RT]]]
```
harshil21, line 87: Can anyone confirm these are the correct type hints?

telegram/ext/_application.py
```diff
@@ -1019,14 +1035,15 @@ def __create_task_done_callback(self, task: asyncio.Task) -> None:
...
         update: Optional[object] = None,
         is_error_handler: bool = False,
     ) -> RT:
         try:
-            if isinstance(coroutine, Generator):
+            # Generator-based coroutines are not supported in Python 3.12+
+            if isinstance(coroutine, Generator) and sys.version_info < (3, 12):
                 return await asyncio.create_task(coroutine)
-            return await coroutine
+            return await coroutine  # type: ignore[misc]
```
harshil21, line 1046: Because I had to ignore this here. 
`Incompatible types in "await" (actual type "Union[Generator[Optional[Future[object]], None, Any], Awaitable[Any]]", expected type "Awaitable[Any]")  [misc]
`

harshil21 (commented):
I double checked by running mypy in different python versions, and also checked if VSCode shows a type error when passing an (async) generator to `app.create_task` for py > 3.12.

telegram/ext/_application.py
```diff
@@ -78,6 +79,19 @@
 _STOP_SIGNAL = object()
 _DEFAULT_0 = DefaultValue(0)
 
+
+# Since python 3.12, the coroutine passed to create_task should not be an async generator. Remove
+# this check when we drop support for python 3.11.
+if sys.version_info >= (3, 12):
```
Bibo-Joshi, line 85: Can you open a new column in the python versions project and add this?

telegram/ext/_application.py
Bibo-Joshi: Will try to have a oder look on sunday

telegram/ext/_application.py
```diff
@@ -78,6 +79,19 @@
...
+    _CoroType = Union["asyncio.Future[object]", Awaitable[RT]]
+    _ErrorCoroType = Optional[Union[Optional["asyncio.Future[object]"], Awaitable[RT]]]
+else:
+    _CoroType = Union[Generator[Optional["asyncio.Future[object]"], None, RT], Awaitable[RT]]
+    _ErrorCoroType = Optional[
+        Union[Generator[Optional["asyncio.Future[object]"], None, RT], Awaitable[RT]]
+    ]
```
Bibo-Joshi, line 92: = Optional[_CoroType] ?

Bibo-Joshi:
.

harshil21:
pre-commit.ci run

telegram/ext/_application.py
```diff
@@ -78,6 +79,17 @@
 _STOP_SIGNAL = object()
 _DEFAULT_0 = DefaultValue(0)
 
+
+# Since python 3.12, the coroutine passed to create_task should not be an async generator. Remove
+# this check when we drop support for python 3.11.
+if sys.version_info >= (3, 12):
+    _CoroType = Union["asyncio.Future[object]", Awaitable[RT]]
```
Bibo-Joshi, line 86: `asyncio.Future` is not listed in the typeshed repo:

https://github.com/python/typeshed/blob/d492efeaea288177e9eca2734d1d463fbb6fff60/stdlib/asyncio/__init__.pyi#L37-L43

https://github.com/python/typeshed/blob/d492efeaea288177e9eca2734d1d463fbb6fff60/stdlib/asyncio/tasks.pyi#L314-L321

and we also didn't have it before #3543 …

telegram/ext/_application.py
```diff
@@ -1019,14 +1033,15 @@ def __create_task_done_callback(self, task: asyncio.Task) -> None:
...
         self,
-        coroutine: Union[Generator[Optional["asyncio.Future[object]"], None, RT], Awaitable[RT]],
+        coroutine: _CoroType,
         update: Optional[object] = None,
         is_error_handler: bool = False,
     ) -> RT:
         try:
-            if isinstance(coroutine, Generator):
+            # Generator-based coroutines are not supported in Python 3.12+
+            if sys.version_info < (3, 12) and isinstance(coroutine, Generator):
```
Bibo-Joshi, line 1042: Can we issue a deprecation warning here?

Bibo-Joshi (changes requested):

telegram/ext/_application.py
harshil21: Okay, so I replace asyncio.Future with asyncio.Coroutine correct? And we should still leave Awaitable as it is right?

telegram/ext/_application.py
harshil21: good idea, will try

telegram/ext/_application.py
Bibo-Joshi: My undestanding would be that we don't want `Awaitable` either 🤔

harshil21: so #3540 was incorrectly reported? will there be a typing issue from mypy if we run the example in that issue with asyncio.create_task instead of app.create_task in py 3.12? I can check in a bit

Bibo-Joshi: > so https://github.com/python-telegram-bot/python-telegram-bot/issues/3540 was incorrectly reported?

maybe. I went through the discussions of #3540 and #3543 a bit more again and I guess the important parts are:

> asyncio.create_task only accepts Coroutines, and Futures are not acceptable:

>  Awaitable is more suitable, considering both asyncio.Future and typing.Coroutine are sub-classes of it.

> That way, Application.create_task would fully cover the functionality of asyncio.gather plus also supporting Future.

That last one is a comment from my side. I would still say that providing a bit more than `asyncio.create_task` doesn't hurt unless it means additional effort for us. If that's the case, we can stick to `Awaitable`. If supporting `Future` does not work easily in py3.12, we'll have to deprecate that as well.

harshil21: Didn't see any issues with also supporting `Future`. So I kept only `Awaitable` since `Coroutine` and `Future` are subclasses of it.

telegram/ext/_application.py
```diff
@@ -950,7 +963,7 @@ def __run(
 
     def create_task(
         self,
-        coroutine: Union[Generator[Optional["asyncio.Future[object]"], None, RT], Awaitable[RT]],
+        coroutine: _CoroType,
```
Bibo-Joshi, line 966: Here you need to specify the generic `RT` in the definiton of `_CoroType`, i.e. `_CoroType[RT]`. same for other usages of `_(Error)CoroType`. That should then satisfy pyright :)

Bibo-Joshi (approved):

PR #3753, opened 2023-06-12
Make use of `@final`, `Final` and `Literal`

### LINKED ISSUE #3735: Adopt Py3.8+ features: New Typing features
Python 3.8 adds the

- [`@final` decorator](https://docs.python.org/3/library/typing.html#typing.final)
- [`Final` type](https://docs.python.org/3/library/typing.html#typing.Final)
- [`Literal` type](https://docs.python.org/3/library/typing.html#typing.Literal)

to the `typing` module. This allows to make some of our type hints more precise.
We should review our code bases and

* use the `@final` decorator for classes and methods that should not be subclassed/overridden, e.g. in interface classes like `Base{Persistence, Handler, UpdateProcessor, Request}`
* Use `Final` rather than `ClassVar` for class attributes that are not supposed to be overridden, e.g. `Chat.PRIVATE`. Also check if there are other attributes that should be considered final
* Use `Literal` where it makes sense. These may be very few places, considering that it does not play well in combination with enums (https://github.com/python/typing/issues/781)

Bibo-Joshi:
Closes #3735

* Mostly changes `ClassVar` to `Final` for all constants
* Adds some limited usage of the `@final` decorator for `Base{UpdateProcessor, Request}`
* Uses `Literal` for the `http_version` argument of `HTTPXRequest`. Otherwise, `Literal` is not used due to python/typing/issues/781

If there are more places, where `@final` or `Literal` are useful, please do point them out :)

### DIFF
--- a/telegram/_update.py
+++ b/telegram/_update.py
@@ -18,7 +18,7 @@
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains an object that represents a Telegram Update."""
 
-from typing import TYPE_CHECKING, ClassVar, List, Optional
+from typing import TYPE_CHECKING, Final, List, Optional
 
 from telegram import constants
 from telegram._callbackquery import CallbackQuery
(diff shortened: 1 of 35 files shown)


telegram/_update.py
```diff
@@ -171,63 +171,63 @@ class Update(TelegramObject):
...
+    MY_CHAT_MEMBER: Final[str] = constants.UpdateType.MY_CHAT_MEMBER
     """:const:`telegram.constants.UpdateType.MY_CHAT_MEMBER`
 
     .. versionadded:: 13.5"""
-    CHAT_MEMBER: ClassVar[str] = constants.UpdateType.CHAT_MEMBER
+    CHAT_MEMBER: Final[str] = constants.UpdateType.CHAT_MEMBER
     """:const:`telegram.constants.UpdateType.CHAT_MEMBER`
 
     .. versionadded:: 13.5"""
     CHAT_JOIN_REQUEST = constants.UpdateType.CHAT_JOIN_REQUEST
```
lemontree210, line 226:
```suggestion
    CHAT_JOIN_REQUEST: Final[str] = constants.UpdateType.CHAT_JOIN_REQUEST
```

telegram/ext/_applicationbuilder.py
```diff
@@ -564,7 +565,7 @@ def pool_timeout(self: BuilderType, pool_timeout: Optional[float]) -> BuilderTyp
         self._pool_timeout = pool_timeout
         return self
 
-    def http_version(self: BuilderType, http_version: str) -> BuilderType:
+    def http_version(self: BuilderType, http_version: Literal["1.1", "2"]) -> BuilderType:
```
lemontree210, line 568: I would suggest doing something like
```py
HTTPVersion = Literal["1.1", "2"]
...
def http_version(self: BuilderType, http_version: HTTPVersion)
```
since it's repeated at least twice, but I'm not sure whether mypy will complain or not.

I think, BTW, that PyCharm still has poor support for literals, unless that changed in the newest version. Like you iterate over strings that clearly conform with the limitations of a given literal, and it still complains that you pass in a `str` when a `Literal` is expected.

lemontree210 (changes requested):
I don't have the required in-depth knowledge to suggest more cases of `@final` that may have been missed

lemontree210 (approved):

PR #3754, opened 2023-06-13
Drop `CancelledError`

### LINKED ISSUE #3731: Adopt Py3.8+ features: `CancelledError` is a subclass of `BaseException`
Python 3.8 makes [`asyncio.CancelledError`](https://docs.python.org/3/library/asyncio-exceptions.html?highlight=cancellederror#asyncio.CancelledError) as subclass of [`BaseException`](https://docs.python.org/3/library/exceptions.html?highlight=baseexception#BaseException). As a consequence, it is no longer catched by `except Exception` clauses and we no longer need special casing in the style of `except asyncio.CancelledError`.
We review our code base for explicit usages of `asyncio.CancelledError` and check which ones of them we can drop.

clot27:
closes #3731 
Used grep to find all usescases of CancelledError

### DIFF
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -1027,10 +1027,6 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
             if isinstance(coroutine, Generator):
                 return await asyncio.create_task(coroutine)
             return await coroutine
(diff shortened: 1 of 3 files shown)


Bibo-Joshi:
Before merging, I'd like to have it double checked that unit tests exit for all the cases were `CancelledError` was used so far

PR #3759, opened 2023-06-17
Pass `name` to `asyncio.create_task`

### LINKED ISSUE #3733: Adopt Py3.8+ features: Names for asyncio tasks
Python 3.8 adds a `name ` argument to the function [`asyncio.create_task`](https://docs.python.org/3/library/asyncio-task.html?highlight=create_task#asyncio.create_task). This allows to give all tasks an interpretable name, which can help with debugging and also provides more transparency for users.
We should

* revisit our code base to check all uses of `asyncio.create_task` and provide meaningful names
* if possible, also extend the unit tests to assert that the name is set on the tasks. I'm not entirely sure if that's possible

harshil21:
Closes #3733 

I couldn't test the `name` in `Job._run` since that task is instantly awaited (why?)


Bibo-Joshi:
@harshil21 IISC the point is that the call to application.create_task ensures that the job will be awaited on shutdown...

telegram/ext/_application.py
```diff
@@ -1076,6 +1082,7 @@ async def _update_fetcher(self) -> None:
                     self.create_task(
                         self.__process_update_wrapper(update),
                         update=update,
+                        name=f"Application:process_concurrent_update_{self.update_queue.qsize()}",
```
Bibo-Joshi, line 1085: bot.id for consistency? Why the updates queue size? I would find the update ID more helpful...

telegram/ext/_application.py
```diff
@@ -1132,7 +1139,11 @@ async def process_update(self, update: object) -> None:
                             and self.bot.defaults
                             and not self.bot.defaults.block
                         ):
-                            self.create_task(coroutine, update=update)
+                            self.create_task(
+                                coroutine,
+                                update=update,
+                                name=f"Application:process_update_non_blocking:{handler}",
```
Bibo-Joshi, line 1145: Handlers usually don't have readible representations. Is that worth adding? Bot.id (also for the rest) and update ID?

telegram/ext/_conversationhandler.py
```diff
@@ -831,6 +831,7 @@ async def handle_update(  # type: ignore[override]
                         update, application, handler_check_result, context
                     ),
                     update=update,
+                    name="ConversationHandler:handle_update:non_blocking_cb",
```
Bibo-Joshi, line 834: Update ID in the names in CH as well?

telegram/ext/_jobqueue.py
```diff
@@ -797,7 +797,9 @@ async def _run(
             await context.refresh_data()
             await self.callback(context)
         except Exception as exc:
-            await application.create_task(application.process_error(None, exc, job=self))
+            await application.create_task(
+                application.process_error(None, exc, job=self), name="Job:_run:process_error"
```
Bibo-Joshi, line 801: No need for the underscore IMO. Job ID?

Bibo-Joshi (commented):
Thanks for the PR! :) Left a few suggestions below. Sorry for the brief wording, I'm not at a keyboard...

telegram/ext/_application.py
harshil21: > I would find the update ID more helpful

I wanted to do that, but the update queue can contains updates of other types also (like in the customwebhookbot example), so that would result in an error.

The only other thing which makes sense is using the queue size

telegram/ext/_application.py
harshil21: I tried to include a dynamic part to every task name (so that it's unique and easy to distinguish while inspecting all running tasks). 

Can't use update id for the same reason above.

> Handlers usually don't have readible representations

Maybe we should make them readable? Should I open a new issue?

telegram/ext/_conversationhandler.py
harshil21: looks like we only allow type `Update` for CH, so I can add that here yes.

telegram/ext/_application.py
Bibo-Joshi: Hah, very good point! Tbh I don't see what the queue size would be good for though. It doesn't help identifying which update was processed... One could use the ID if available and the type otherwise. Or we leave out both

telegram/ext/_application.py
Bibo-Joshi: Fair point. Yes, let's open an issue

telegram/ext/_application.py
harshil21: i'll leave both then

telegram/ext/_application.py
```diff
@@ -1076,6 +1082,8 @@ async def _update_fetcher(self) -> None:
                     self.create_task(
                         self.__process_update_wrapper(update),
                         update=update,
+                        name=f"Application:{self.bot.id}:process_concurrent_update_"
+                        f"{self.update_queue.qsize()}",
```
Bibo-Joshi, line 1086: https://github.com/python-telegram-bot/python-telegram-bot/pull/3759#discussion_r1238700959 ? ;P

telegram/ext/_application.py
```diff
@@ -1076,6 +1082,8 @@ async def _update_fetcher(self) -> None:
                     self.create_task(
                         self.__process_update_wrapper(update),
                         update=update,
+                        name=f"Application:{self.bot.id}:process_concurrent_update_"
```
Bibo-Joshi, line 1085: One more thought on the `bot.id`: IIRC we do not prevent `application.create_task` from being called before `application.initialize`, right? If so, the `bot.id` would not be available yet …

telegram/ext/_application.py
harshil21: oh my bad thought I removed it

telegram/ext/_application.py
harshil21: yeah but this task will only be created in `_update_fetcher`, which is called in `start()` which has a check to confirm that the app has been initialized.

Bibo-Joshi: ah okay, then we're good I guess :)

PR #3762, opened 2023-06-18
Use @final and Literal in more places

harshil21:
Extension of #3753

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -32,6 +32,7 @@ from typing import (
     Callable,
     Dict,
     List,
+    Literal,
     NoReturn,
     Optional,
     Sequence,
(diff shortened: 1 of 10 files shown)


telegram/_bot.py
```diff
@@ -6602,7 +6603,7 @@ async def send_poll(
         is_anonymous: Optional[bool] = None,
         type: Optional[str] = None,  # pylint: disable=redefined-builtin
         allows_multiple_answers: Optional[bool] = None,
-        correct_option_id: Optional[int] = None,
+        correct_option_id: Optional[Literal[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]] = None,
```
lemontree210, line 6606: same suggestion as with the other PR: doing something like
```py
PossibleOptionID = Literal[0, 1 ....]
```
and same for Markdown version

PR #3764, opened 2023-06-18
fix error with `Literal`, causing tests/docs to fail

lemontree210:
Fix test failure caused by introduction of `Literal` types.

### DIFF
--- a/docs/auxil/admonition_inserter.py
+++ b/docs/auxil/admonition_inserter.py
@@ -528,7 +528,11 @@ class AdmonitionInserter:
         # For custom generics like telegram.ext._application.Application[~BT, ~CCT, ~UD...].
         # This must come before the check for isinstance(type) because GenericAlias can also be
         # recognized as type if it belongs to <class 'types.GenericAlias'>.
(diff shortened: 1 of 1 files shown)


harshil21 (approved):
docs successfully built locally for me

Bibo-Joshi (commented):
Can you maybe extend the CSI comment about why Literal needs to be treated here as well?

lemontree210:
> Can you maybe extend the CSI comment about why Literal needs to be treated here as well?

To me, it looks rather natural in this sequence and it might look a bit strange if I explain only the `Literal` bit:
```py
(
            "<class 'typing._GenericAlias'>",
            "<class 'types.GenericAlias'>",
            "<class 'typing._LiteralGenericAlias'>",
        )
```
They're all Generic aliases.

OTOH, it's hard to imagine that we'll ever put a telegram class inside a literal. Something like:
```py
Persistence = Literal["PicklePersistence", "DictPersistence"]
```
In this case, I could just create a separate condition for `Literal` and skip it altogether without even checking what's inside.

PR #3767, opened 2023-06-20
Remove obsolete note from contributing guidelines

harshil21:
The note about `**kwargs` was obsolete. I didn't modify it by mentioning `api_kwargs` since that is already present in the PR checklist which we mention in the guidelines already.


Bibo-Joshi:
Nice finding! Can we move this into doc fixes though?

PR #3769, opened 2023-06-22
fix CI issue (`_nonwrappers`)

lemontree210:
plugin_github_group.py triggers exception when running CI tasks

`AttributeError: '_HookCaller' object has no attribute '_nonwrappers'`

This attribute is indeed absent, however the attribute `._hookimpls` contains the necessary keys.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,7 +1,7 @@
 pre-commit  # needed for pre-commit hooks in the git commit command
 
 # For the test suite
-pytest==7.3.1
+pytest==7.3.2
 pytest-asyncio==0.21.0  # needed because pytest doesn't come with native support for coroutines as tests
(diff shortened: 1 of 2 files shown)


harshil21:
3.11 macos tests show passing even though there was an internal pytest error (looks like it was flaky, and not related to this PR). 

I believe it showed passing cause https://github.com/python-telegram-bot/python-telegram-bot/blob/master/.github/workflows/test.yml#L69 ran with no tests and then gave an exit code of 0.

Bibo-Joshi:
Ah, the pleasures of using protected members 🫠 Replacing one protected member with another one doesn't sound overly desirable to me TBH. Is there a way to do wtih with proper public API? How terrible do the GitHub Actions logs look like if we remove this custom plugin all together?

Bibo-Joshi:
superseded by #3776

PR #3777, opened 2023-06-26
Remove Test Artifacts Automatically

Bibo-Joshi:
Makes sure that some artifacts that currently always clutter my root directory are properly removed.

Already makes use of the `missing_ok` parameter in anticipation of #3728 and should hence be merged after that.


Bibo-Joshi:
> can't we use the [`tmp_path`](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) pytest fixture?

Nice idea! writing to those files gives me `permisison denied` errors currently. Will have a look if I can fix that.

harshil21 (approved):

PR #3792, opened 2023-07-05
Test and Document `Exception.__cause__` on `NetworkError`

### LINKED ISSUE #3778: [FEATURE] More fine grained network exceptions

We currently only have `NetworkError` and `TimeoutError(NetworkError)` to indicate exceptions that are due to natworking problems (in contrast to exceptions issued by TG). This is not very fine grained and can make advanced exception handling difficult.

I'd like to add more subclasses. Looking it httpx, they have a [large number of them](https://github.com/encode/httpx/blob/2e2949c8ea88668f4cc5e509342b57e60fe4d86a/httpx/_exceptions.py#L1-L32). Not all of them add much value for us IMO. In the following I list all of the httpx excetpions and check the ones that I think would make sense to have for us:

- [ ] HTTPError
  - [ ] RequestError
    - [ ] TransportError
      - [x] TimeoutException ( (corresponds to `tg.TimeoutError`)
        - [x] ConnectTimeout
        - [x] ReadTimeout
        - [x] WriteTimeout
        - [x] PoolTimeout
      - [x] NetworkError (corresponds to `tg.NetworkError`)
        - [x] ConnectError
        - [x] ReadError
        - [x] WriteError
        - [x] CloseError
      - [ ] ProtocolError
        - [ ] LocalProtocolError
        - [ ] RemoteProtocolError
      - [x] ProxyError
      - [ ] UnsupportedProtocol
    - [ ] DecodingError
    - [ ] TooManyRedirects
  - [ ] HTTPStatusError
- [ ] InvalidURL
- [ ] CookieConflict
- [ ] StreamError
  - [ ] StreamConsumed
  - [ ] StreamClosed
  - [ ] ResponseNotRead
  - [ ] RequestNotRead

_No response_

### Additional context

_No response_

Bibo-Joshi:
Closes #3778 as suggested by Noam in the dev chat.

Tests for `{Base, HTTPX}Request` not extended yet.

### DIFF
--- a/telegram/error.py
+++ b/telegram/error.py
@@ -118,11 +118,30 @@ class InvalidToken(TelegramError):
 class NetworkError(TelegramError):
     """Base class for exceptions due to networking errors.
 
+    Args:
+        cause (:obj:`Exception`, optional): The original exception that caused this error as issued
+            by the library used for network communication.
+
+            .. versionadded:: NEXT.VERSION
+
+    Attributes:
+        cause (:obj:`Exception`): Optional. The original exception that caused this error as issued
+            by the library used for network communication.
+
+            .. versionadded:: NEXT.VERSION
+
     Examples:
         :any:`Raw API Bot <examples.rawapibot>`
     """
 
-    __slots__ = ()
+    __slots__ = ("cause",)
+
+    def __init__(self, message: str, cause: Optional[Exception] = None) -> None:
+        super().__init__(message)
+        self.cause: Optional[Exception] = cause
+
+    def __reduce__(self) -> Tuple[type, Tuple[str, Optional[Exception]]]:  # type: ignore[override]
+        return self.__class__, (self.message, self.cause)
 
 
 class BadRequest(NetworkError):
@@ -142,8 +161,8 @@ class TimedOut(NetworkError):
 
     __slots__ = ()
 
-    def __init__(self, message: Optional[str] = None) -> None:
-        super().__init__(message or "Timed out")
+    def __init__(self, message: Optional[str] = None, cause: Optional[Exception] = None) -> None:
+        super().__init__(message=message or "Timed out", cause=cause)
 
 
 class ChatMigrated(TelegramError):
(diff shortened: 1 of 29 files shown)


harshil21 (approved):

lemontree210 (commented):
I have a general question: it was my understanding that if we `raise` some exception `from` another exception, we don't need to specify any `cause` explicitly. I think that's the way it's formulated in the [docs](https://docs.python.org/3/library/exceptions.html#exception-context) and I personally have found `raise ... from ...` quite informative on its own. 

If a user wants to check the cause explicitly, I think they can use `.__cause__` (as per the same docs).

So what am I missing here? :)

Bibo-Joshi:
Huh. I really just didn't know that. Makes sense that `raise … from` has to be stored somehow, but I wasn't aware that this is documented. A bit weird though that `__cause__` is not documented as proper attribute 🤔 Moreorever, that attribute apparently is not pickled: https://github.com/python/cpython/issues/73652.
I'd be okay with the non-pickable thing if we agree that `__cause__` can bes considered to be public python API …

telegram/error.py
```diff
@@ -118,6 +118,13 @@ def __init__(self, message: Optional[str] = None) -> None:
 class NetworkError(TelegramError):
     """Base class for exceptions due to networking errors.
 
+    Tip:
+        This exception (and its subclasses) usually originate from the networking backend
```
lemontree210, line 122:
```suggestion
        This exception (and its subclasses) usually originates from the networking backend
```

lemontree210: or no parentheses :)

telegram/error.py
```diff
@@ -118,6 +118,13 @@ def __init__(self, message: Optional[str] = None) -> None:
...
+        used by :class:`~telegram.request.HTTPXRequest`/a custom implementation of
```
lemontree210, line 123:
```suggestion
        used by :class:`~telegram.request.HTTPXRequest`, a custom implementation of
```

lemontree210 (approved):
Some punctuation nitpicks in docstring, maybe not even justified :)

telegram/error.py
harshil21: no, I think he meant "or".
```suggestion
        used by :class:`~telegram.request.HTTPXRequest`, or a custom implementation of
```

harshil21 (approved):

PR #3797, opened 2023-07-12
add notes on running `post_` methods explicitly

lemontree210:
As discussed in dev group, I added a note stating that in cases like [custom webhook example](https://docs.python-telegram-bot.org/en/stable/examples.customwebhookbot.html) user should explicitly call `post_` methods.

### DIFF
--- a/telegram/ext/_applicationbuilder.py
+++ b/telegram/ext/_applicationbuilder.py
@@ -1095,7 +1095,8 @@ class ApplicationBuilder(Generic[BT, CCT, UD, CD, BD, JQ]):
         Note:
             If you implement custom logic that implies that you will **not** be using
             :meth:`Application.run_polling` or :meth:`Application.run_webhook` to run
-            your application (e.g. in :any:`Custom Webhook Example <examples.customwebhookbot>`),
+            your application (like e.g. in
+            :any:`Custom Webhook Bot Example <examples.customwebhookbot>`),
             make sure that you explicitly call this method.
 
         .. seealso:: :meth:`post_stop`, :meth:`post_shutdown`
@@ -1139,7 +1140,8 @@ class ApplicationBuilder(Generic[BT, CCT, UD, CD, BD, JQ]):
         Note:
             If you implement custom logic that implies that you will **not** be using
             :meth:`Application.run_polling` or :meth:`Application.run_webhook` to run
-            your application (e.g. in :any:`Custom Webhook Example <examples.customwebhookbot>`),
+            your application (like e.g. in
+            :any:`Custom Webhook Bot Example <examples.customwebhookbot>`),
             make sure that you explicitly call this method.
 
         .. seealso:: :meth:`post_init`, :meth:`post_stop`
@@ -1185,7 +1187,8 @@ class ApplicationBuilder(Generic[BT, CCT, UD, CD, BD, JQ]):
         Note:
             If you implement custom logic that implies that you will **not** be using
             :meth:`Application.run_polling` or :meth:`Application.run_webhook` to run
-            your application (e.g. in :any:`Custom Webhook Example <examples.customwebhookbot>`),
+            your application (like e.g. in
+            :any:`Custom Webhook Bot Example <examples.customwebhookbot>`),
             make sure that you explicitly call this method.
 
         .. seealso:: :meth:`post_init`, :meth:`post_shutdown`
--- a/telegram/ext/_applicationbuilder.py
+++ b/telegram/ext/_applicationbuilder.py
@@ -1092,6 +1092,12 @@ class ApplicationBuilder(Generic[BT, CCT, UD, CD, BD, JQ]):
 
                 application = Application.builder().token("TOKEN").post_init(post_init).build()
 
+        Note:
+            If you implement custom logic that implies that you will **not** be using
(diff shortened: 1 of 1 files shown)


telegram/ext/_applicationbuilder.py
```diff
@@ -1092,6 +1092,13 @@ async def post_init(application: Application) -> None:
 
                 application = Application.builder().token("TOKEN").post_init(post_init).build()
 
+        Note:
+            If you implement custom logic that implies that you will **not** be using
+            :meth:`Application.run_polling` or :meth:`Application.run_webhook` to run
+            your application (like e.g. in
+            :any:`Custom Webhook Bot Example <examples.customwebhookbot>`),
+            make sure that you explicitly call this method.
```
Bibo-Joshi, line 1100:
```suggestion
            :any:`Custom Webhook Bot Example <examples.customwebhookbot>`),.
```
There is not really a need to set the function on the `ApplicationBuilder` in this case. you can just manually do `await some_function(application)` if you're doing custom logic anyways :)

Can we use a sphinx insertion thingy to avoid repitition?

lemontree210: Hmm...  makes sense :)

As for the insertion - it doesn't work if I mention methods there. Sphinx complains about it.

Bibo-Joshi: ah, you'll have to use the full name then, see e.g. :meth:~telegram.ext.Application.run_polling

lemontree210: OK, done. It didn't like the `:any:` link to the example either, so I had to replace it with an https hyperlink

harshil21 (commented):
tbh I think the note is redundant. The opening paragraph in `post_*` methods already tell that it is only executed by `run_polling/webhook`, and clicking on that link shows the order of execution ("The order of run_polling is as follows" line).

How do you feel about raising a warning if `app.start()` was called and `app.post_init()` was set but not called? This would need additional logic to work (some class variables we should set when the function has finished running?), so it might be worth discussing if we even want that.

lemontree210:
> tbh I think the note is redundant. The opening paragraph in `post_*` methods already tell that it is only executed by `run_polling/webhook`, and clicking on that link shows the order of execution ("The order of run_polling is as follows" line).

OK, I see what you mean. I actually don't insist on adding my note. I realize now that by following this path, I could have figured out earlier that I have to `await` my function manually. I guess it's once again about our expectations from the user and how much "superfluous" (with or without quotation marks) information we want to give them in our docs. I definitely see that what is said in the note can be inferred from the info already present.

lemontree210:
> How do you feel about raising a warning if `app.start()` was called and `app.post_init()` was set but not called? This would need additional logic to work (some class variables we should set when the function has finished running?), so it might be worth discussing if we even want that.

That's an interesting idea, similar to a warning when user has declared some states but is not using some of them. I too think, however, that the implementation will be rather tedious. Do we have any idea about how many users actually implement this custom logic (e.g. because of several webhooks like in my case)?

docs/substitutions/global.rst
```diff
@@ -57,3 +57,5 @@
 .. |captionentitiesattr| replace:: Tuple of special entities that appear in the caption, which can be specified instead of ``parse_mode``.
 
 .. |datetime_localization| replace:: The default timezone of the bot is used for localization, which is UTC unless :attr:`telegram.ext.Defaults.tzinfo` is used.
+
+.. |post_methods_note| replace:: If you implement custom logic that implies that you will **not** be using :class:`~telegram.ext.Application`'s methods :meth:`~telegram.ext.Application.run_polling` or :meth:`~telegram.ext.Application.run_webhook` to run your application (like it's done in `Custom Webhook Bot Example <https://docs.python-telegram-bot.org/en/stable/examples.customwebhookbot.html>`__), the callback you set in this method **will not be called automatically**. So instead of setting a callback with this method, you have to explicitly ``await`` the function that you want to run at this stage of your application's life (in the `example mentioned above <https://docs.python-telegram-bot.org/en/stable/examples.customwebhookbot.html>`__, that would be in ``async with application`` context manager).
```
Bibo-Joshi, line 61:
```suggestion
.. |post_methods_note| replace:: If you implement custom logic that implies that you will **not** be using :class:`~telegram.ext.Application`'s methods :meth:`~telegram.ext.Application.run_polling` or :meth:`~telegram.ext.Application.run_webhook` to run your application (like it's done in the :any:`Custom Webhook Bot Example <examples.customwebhookbot>`_), the callback you set in this method **will not be called automatically**. So instead of setting a callback with this method, you have to explicitly ``await`` the function that you want to run at this stage of your application's life (in the :any:`example mentioned above <examples.customwebhookbot>`, that would be in ``async with application`` context manager).
```

Bibo-Joshi (approved):
I'm fine with the proposed content changes as they are and I personally don't find them way too verbose (although @harshil21 ofc made a valid point). 
I would guess that programmatically checking and issuing a warning is not worth the effort …

docs/substitutions/global.rst
lemontree210: Nah, I tried this, it didn't work. Sphinx didn't accept `:any:` links in the global substitution.

Bibo-Joshi: *sad noises*

PR #3809, opened 2023-07-20
Update bot.py

david-shiko:
Fix Bot.get_user_profile_photos return type according to the docs:
https://docs.python-telegram-bot.org/en/v13.13/telegram.bot.html?highlight=get_user_profile_photos#telegram.Bot.get_user_profile_photos

Old: `Optional[telegram.UserProfilePhotos]`
New: `telegram.UserProfilePhotos`


harshil21:
hi,

thanks for the PR, but we cannot accept this since we have discontinued the v13.x branch since v20 stable. This means no bug, API and security fixes. 

This is already fixed in v20.

PR #3812, opened 2023-07-25
Feature #3798: Check commands for args

### LINKED ISSUE #3798: [FEATURE] Add args present argument to command filter

There is currently no filter defined by us which simply checks if a command message has args, e.g. `start payload`

Add one argument to the command filter called `has_args` which checks if something comes after the command in the message.

### Additional context

We could also start discussing if we want to provide some way to check the content of args, like a pattern match.

thatguylah:
closes #3798

CommandHandler allows an optional `has_args` argument that defaults to `None`, but can optionally accept either `True`, `False` or any `int`. 
- Used an `Enum` for mapping this logic inside an internal method `_check_correct_args` of `class CommandHandler` 
-  In `check_update` method, right before sending to `filters` for filters parsing, it `_check_correct_args` and if `_valid` then it does nothing and passes to `filters` 
- if however `_valid` is False, it early returns `None` before it hits filters parsing. 
- Implemented test cases in `test_commandhandler.py` for `has_args` 
- All test cases passed. 👍 

Let me know if i need to refactor or add to more documentation that i am not aware about. ☺️

### DIFF
--- a/telegram/ext/_commandhandler.py
+++ b/telegram/ext/_commandhandler.py
@@ -148,13 +148,14 @@ class CommandHandler(BaseHandler[Update, CCT]):
 
         condition = (
             ArgsCondition.HAS_ARGS_NONE
-            if self.has_args is None
+            if (self.has_args is None)
             else ArgsCondition.HAS_ARGS_TRUE_ARGS_PRESENT
-            if self.has_args and args
+            if (self.has_args is True and args)
+            # Must specify is True, otherwise if has_args is int, it will default to Truthy value.
             else ArgsCondition.HAS_ARGS_FALSE_ARGS_ABSENT
-            if not self.has_args and not args
+            if (self.has_args is False and not args)
             else ArgsCondition.HAS_ARGS_NUM
-            if isinstance(self.has_args, int) and len(args) == self.has_args
+            if (isinstance(self.has_args, int) and len(args) == self.has_args)
             else ArgsCondition.HAS_INVALID_ARGS
         )
 
@@ -196,9 +197,6 @@ class CommandHandler(BaseHandler[Update, CCT]):
                 _valid = self._check_correct_args(args)
                 if _valid is False:
                     return None
-                # if _valid: TODO: Implement condition if has_args and args are True.
-                #     raise ValueError("CommandHandler: args are not valid")
-                #     return None  # args does not match what developer specified, returning None.
 
                 filter_result = self.filters.check_update(update)
                 if filter_result:
--- a/telegram/ext/_commandhandler.py
+++ b/telegram/ext/_commandhandler.py
@@ -18,6 +18,7 @@
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains the CommandHandler class."""
 import re
+from enum import Enum, auto
 from typing import TYPE_CHECKING, Any, FrozenSet, List, Optional, Tuple, TypeVar, Union
 
 from telegram import MessageEntity, Update
@@ -102,7 +103,7 @@ class CommandHandler(BaseHandler[Update, CCT]):
             :meth:`telegram.ext.Application.process_update`.
     """
 
-    __slots__ = ("commands", "filters")
+    __slots__ = ("commands", "filters", "has_args")
 
     def __init__(
         self,
@@ -110,6 +111,7 @@ class CommandHandler(BaseHandler[Update, CCT]):
         callback: HandlerCallback[Update, CCT, RT],
         filters: Optional[filters_module.BaseFilter] = None,
         block: DVType[bool] = DEFAULT_TRUE,
+        has_args: Union[bool, None, int] = None,
     ):
         super().__init__(callback, block=block)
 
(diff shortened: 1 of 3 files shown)


thatguylah:
Also, do you perhaps think instead of returning `None` in `check_update` we should raise an `Exception` instead? @Bibo-Joshi

telegram/ext/_commandhandler.py
```diff
@@ -88,6 +89,12 @@ async def callback(update: Update, context: CallbackContext)
             :meth:`telegram.ext.Application.process_update`. Defaults to :obj:`True`.
 
             .. seealso:: :wiki:`Concurrency`
+        has_args (:obj:`bool` | :obj:`int`, optional):
+            Determines whether the command handler should process the update or not.
+            If :obj:`True`, the handler will process any non-zero number of args.
+            If :obj:`False`, the handler will only process if there are no args.
+            if :obj:`int`, the handler will only process if there are exactly that many args.
+            Defaults to :obj:`None`, which means the handler will process any or no args.
```
Bibo-Joshi, line 97:
```suggestion
        has_args (:obj:`bool` | :obj:`int`, optional):
            Specifies criteria on the arguments of the command that this handler should check for.
            If :obj:`True`, the handler will process the update if the command has any non-zero number of arguments.
            If :obj:`False`, the handler will only process the update if there are no arguments following the command.
            If a non-negative :obj:`int` is passed, the handler will only process the update if there are exactly that many arguments specified for the command.
            Defaults to :obj:`None`, which means the handler will not check the number of arguments.
            
            .. versionadded:: NEXT.VERSION
```

telegram/ext/_commandhandler.py
```diff
@@ -100,16 +107,20 @@ async def callback(update: Update, context: CallbackContext)
         block (:obj:`bool`): Determines whether the return value of the callback should be
             awaited before processing the next handler in
             :meth:`telegram.ext.Application.process_update`.
+        has_args (:obj:`bool` | :obj:`int` | None):
+            Optional argument, otherwise all implementations of :class:`CommandHandler` will break.
+            Defaults to :obj:`None`, which means the handler will process any args or no args.
```
Bibo-Joshi, line 112:
```suggestion
        has_args (:obj:`bool` | :obj:`int`):
            Optional. Only allow commands whose arguments satisfy this criterion. See :param:`has_args` for details.
            
            .. versionadded:: NEXT.VERSION
```

telegram/ext/_commandhandler.py
```diff
@@ -126,6 +137,39 @@ def __init__(
...
+            if (self.has_args is None)
+            else ArgsCondition.HAS_ARGS_TRUE_ARGS_PRESENT
+            if (self.has_args is True and args)
+            # Must specify is True, otherwise if has_args is int, it will default to Truthy value.
+            else ArgsCondition.HAS_ARGS_FALSE_ARGS_ABSENT
+            if (self.has_args is False and not args)
+            else ArgsCondition.HAS_ARGS_NUM
+            if (isinstance(self.has_args, int) and len(args) == self.has_args)
+            else ArgsCondition.HAS_INVALID_ARGS
+        )
```
Bibo-Joshi, line 169: The enum is a nice idea, but it seems like rather heavy overkill for this use case :D especially since there are only 2 values that the enums are mapped to.
The important part are the if-clauses you specified and those look good to me. I suggest to simplify this method to something like this:

```python
if (cond 1) or (cond 2) or (cond 3) or (cond 4):
    return True
return False
```

telegram/ext/_commandhandler.py
```diff
@@ -159,6 +203,10 @@ def check_update(
                 ):
                     return None
 
+                _valid = self._check_correct_args(args)
+                if _valid is False:
+                    return None
```
Bibo-Joshi, line 208:
```suggestion
                if not self._check_correct_args(args):
                    return None
```

tests/ext/test_commandhandler.py
```diff
@@ -259,3 +263,22 @@ async def test_context_multiple_regex(self, app, command):
...
+        handler_int_two = CommandHandler(["test"], self.callback_basic, has_args=2)
+
+        assert is_match(handler_true, make_command_update("/test helloworld", bot=bot))
+        assert not is_match(handler_true, make_command_update("/test", bot=bot))
+
+        assert is_match(handler_false, make_command_update("/test", bot=bot))
+        assert not is_match(handler_false, make_command_update("/test helloworld", bot=bot))
+
+        assert is_match(handler_int_one, make_command_update("/test helloworld", bot=bot))
+        assert not is_match(handler_int_one, make_command_update("/test hello world", bot=bot))
```
Bibo-Joshi, line 281:
```suggestion
        assert not is_match(handler_int_one, make_command_update("/test hello world", bot=bot))
        assert not is_match(handler_int_one, make_command_update("/test", bot=bot))
```

same for `hander_int_two` :)

telegram/ext/_commandhandler.py
```diff
@@ -126,6 +137,39 @@ def __init__(
             filters if filters is not None else filters_module.UpdateType.MESSAGES
         )
 
+        self.has_args = has_args
```
Bibo-Joshi, line 140: it would be good to check if `has_args` is negative and raise a `ValueError` if it is. We'd also need a corresponding unit test

Bibo-Joshi (changes requested):
Hi. Thakks for the PR! The overall gist of the changes look good :) I left some comments below.

> Also, do you perhaps think instead of returning `None` in `check_update` we should raise an `Exception` instead?

No, that is not desired. This method it designed to determine whether the handler should handle the update and the return value must indicate that. Raising an exception would mean that it can not be determined whether the update should be handled.

Bibo-Joshi:
Closing due to inactivity

PR #3816, opened 2023-07-27
document more about `proxy_url` and `get_updates_proxy_url`

trim21:

### DIFF
--- a/telegram/ext/_applicationbuilder.py
+++ b/telegram/ext/_applicationbuilder.py
@@ -487,6 +487,11 @@ class ApplicationBuilder(Generic[BT, CCT, UD, CD, BD, JQ]):
         """Sets the proxy for the :paramref:`~telegram.request.HTTPXRequest.proxy_url`
         parameter of :attr:`telegram.Bot.request`. Defaults to :obj:`None`.
 
+        Note:
+            this is not the proxy used for all http requests.
+
+        .. seealso:: :meth:`get_updates_proxy_url`
+
         Args:
             proxy_url (:obj:`str`): The URL to the proxy server. See
                 :paramref:`telegram.request.HTTPXRequest.proxy_url` for more information.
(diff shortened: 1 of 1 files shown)


telegram/ext/_applicationbuilder.py
```diff
@@ -640,6 +645,10 @@ def get_updates_proxy_url(self: BuilderType, get_updates_proxy_url: str) -> Buil
         """Sets the proxy for the :paramref:`telegram.request.HTTPXRequest.proxy_url`
         parameter which is used for :meth:`telegram.Bot.get_updates`. Defaults to :obj:`None`.
 
+        proxy used to get updates.
```
harshil21, line 648:
```suggestion
```

already explained by this method's docstring.

telegram/ext/_applicationbuilder.py
```diff
@@ -487,6 +487,11 @@ def proxy_url(self: BuilderType, proxy_url: str) -> BuilderType:
         """Sets the proxy for the :paramref:`~telegram.request.HTTPXRequest.proxy_url`
         parameter of :attr:`telegram.Bot.request`. Defaults to :obj:`None`.
 
+        Note:
+            this is not the proxy used for all http requests.
```
harshil21, line 491:
```suggestion
```
I don't think this note is necessary because of the 'see also' you added below

harshil21 (changes requested):
Thanks for the PR. I would be okay with adding this 'see also' note in all the `(get_updates_)*` methods in AppBuilder. What say others?

telegram/ext/_applicationbuilder.py
trim21: As a user personally I think it's necessary to note this...

Bibo-Joshi:
I would prefer to just add a `.. seealso:: get_updates_proxy_url` (+ vice versa + for all other request related methods that have two variants)

telegram/ext/_applicationbuilder.py
harshil21: hinrich also voted to remove this note (https://github.com/python-telegram-bot/python-telegram-bot/pull/3816#issuecomment-1654517762)

Bibo-Joshi (commented):
Thanks for the updates! Would you mind also adding similar changes to the other `get_updates_*` methods and their counterparts?

trim21:
> Thanks for the updates! Would you mind also adding similar changes to the other `get_updates_*` methods and their counterparts?

I do mind...

Bibo-Joshi:
Fair enough, thanks for the contribution in any case!

PR #3820, opened 2023-07-29
Provide Versions of `customwebhookbot.py` with Different Frameworks

### LINKED ISSUE #3717: Provide the `customwebhookbot.py` example with different frameworks

Currently, we provide an example of setting up a custom webhook for bots using Starlette. While this helps many people write their own custom webhook implementations, we also frequently receive questions about Flask and Django in our Telegram group. People face difficulties writing their own custom webhook implementations with these frameworks because they do not support asyncio (I guess so).

Adding examples to the documentation for those two frameworks would help people better understand how to integrate them with PTB.

_No response_

### Additional context

If we provide custom webhook bot examples for Starlette, Flask, and Django, it would be a cool UX improvement to allow users to choose the framework in the rendered documentation.

Bibo-Joshi:
Closes #3717 by providing versions with flask, quart and Django. Other examples can ofc be added later as well.

I have used neither of the frameworks before, so the examples are surely not optimal from their perspective and I'm happy for improving comments. Especially for Django I am sure that an actual Django project would set this up very differently as (AFAIK) Django usually makes heavy use of several modules & files.

Moreover, I want to point out that ASGI servers like uvicorn are more frequently started from the command line than from within a python script, but I kept to this scheme because

1. that way we don't have to provide additional info on how to run the script (it's just `python script.py` as per usual)
2. I would have to figure out how to start & stop the ptb application properly in such a setup …

Docs build it available [here](https://docs.python-telegram-bot.org/en/doc-webhook-examples/examples.customwebhookbot.html).


lemontree210 (changes requested):
This looks so great! We're becoming incredibly developer-friendly 😆 
I am only suggesting some changes to the comments and, while we're at it, docstrings.

lemontree210:
Yes, a full-fledged Django project would have its own configuration files. This is an interesting question: on the one hand, we have to provide a short example, and that's why we're using `django.conf.settings`, as per [Django docs](https://docs.djangoproject.com/en/4.2/topics/settings/). On the other hand, though, the reason we are providing examples for other frameworks should be that the user already uses one of them in their code and hence, in case of Django, has all these settings in separate files. This might mean that, for a user that doesn't care which app to use for routing of requests, any framework will do, but a user that already uses Django might not find the Django example useful. Then, instead of one example with some lines changed for the respective framework, this actually should expand into something of a tutorial (if you're using Django, but add these lines to `urlpatterns`, then add this to your settings file etc.). Which, in turn, sort of subverts the idea of having a simple usage example.

examples/customwebhookbot_django.py
```diff
@@ -0,0 +1,201 @@
...
+        update: object,
+        application: "Application",
+    ) -> "CustomContext":
+        if isinstance(update, WebhookUpdate):
+            return cls(application=application, user_id=update.user_id)
+        return super().from_update(update, application)
+
+
+async def start(update: Update, context: CustomContext) -> None:
+    """Display a message with instructions on how to use this bot."""
```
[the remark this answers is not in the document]
Bibo-Joshi: Is there a deepear reason to this other than personal preference? 🤔

Bibo-Joshi:
> This might mean that, for a user that doesn't care which app to use for routing of requests, any framework will do, but a user that already uses Django might not find the Django example useful.

This is a very valid argument, but at least for now I would give more weight to your last point 

> Which, in turn, sort of subverts the idea of having a simple usage example.

I think whenever two frameworks are being used together one will have to find a midway of usability and best practice. This is a first step in this direction. Maybe we can just include a disclaimer, maybe something along the lines

> The following examples show how different Python web frameworks can be used alongside with PTB. This can be useful for two use cases:
>
> 1. For extending the functionality of your existing bot to handling updates of external services
> 2. For extending the functionality of your exisiting web application to also include chat bot functionality
>
> How the PTB and web framework components of the examples below are viewed surely depends on which use case one has in mind. We are fully aware that a combinations of PTB with web frameworks will always mean finding a trade off between usability and best practices for both PTB and the web framework and these examples are certainly far from optimal solutions. Please understand them as starting points and use your expertice of the web framework of your choosing to build up on them. You are of course also very welcome to help improve these examples!

?

lemontree210:
> I think whenever two frameworks are being used together one will have to find a midway of usability and best practice. This is a first step in this direction. Maybe we can just include a disclaimer, maybe something along the lines:

Great disclaimer, agreed :)

examples/customwebhookbot_django.py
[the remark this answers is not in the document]
harshil21: If anything, according to PEP 257 it should be "display" https://peps.python.org/pep-0257/#one-line-docstrings

docs/source/examples.customwebhookbot.rst
```diff
@@ -2,7 +2,20 @@
 =======================
 
 This example is available for different web frameworks.
-Please select the framework you would like to use.
+You can select your preferred framework by opening one of the tabs above the code example.
+
+.. hint::
+
+    The following examples show how different Python web frameworks can be used alongside with PTB.
```
lemontree210, line 9:
```suggestion
    The following examples show how different Python web frameworks can be used alongside PTB.
```

docs/source/examples.customwebhookbot.rst
```diff
@@ -2,7 +2,20 @@
...
+    This can be useful for two use cases:
+
+    1. For extending the functionality of your existing bot to handling updates of external services
+    2. For extending the functionality of your exisiting web application to also include chat bot functionality
+
+    How the PTB and web framework components of the examples below are viewed surely depends on which use case one has in mind.
+    We are fully aware that a combinations of PTB with web frameworks will always mean finding a trade off between usability and best practices for both PTB and the web framework and these examples are certainly far from optimal solutions.
```
lemontree210, line 16:
```suggestion
    We are fully aware that a combination of PTB with web frameworks will always mean finding a tradeoff between usability and best practices for both PTB and the web framework and these examples are certainly far from optimal solutions.
```

lemontree210 (approved):
Great, just a couple of typos in the disclaimer :)

examples/customwebhookbot_django.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env python
+# This program is dedicated to the public domain under the CC0 license.
+# pylint: disable=import-error,wrong-import-position
```
harshil21, line 3: I think this can be removed after dropping the version check?

examples/customwebhookbot_django.py
```diff
@@ -0,0 +1,198 @@
...
+# register handlers
+ptb_application.add_handler(CommandHandler("start", start))
+ptb_application.add_handler(TypeHandler(type=WebhookUpdate, callback=webhook_update))
+
+urlpatterns = [
+    path("telegram", telegram, name="Telegram updates"),
+    path("submitpayload", custom_updates, name="custom updates"),
+    path("healthcheck", health, name="health check"),
+]
+settings.configure(ROOT_URLCONF=__name__, SECRET_KEY=uuid4().hex)
```
harshil21, line 165: why is this outside `main()`?

examples/customwebhookbot_django.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+async def main() -> None:
+    """Finalize configuration and run the applications."""
+    url = "https://domain.tld"
+    admin_chat_id = 123456
+    port = 8000
```
harshil21, line 172: I would prefer these constants be captialized and moved to right after the imports, so we can know immediately what to change.

Also an another constant for bot token.

examples/customwebhookbot_django.py
```diff
@@ -0,0 +1,198 @@
...
+
+    # save the values in `bot_data` such that we may easily access them in the callbacks
+    ptb_application.bot_data["url"] = url
+    ptb_application.bot_data["admin_chat_id"] = admin_chat_id
```
harshil21, line 176: is it better practice to put this in `post_init()`?

examples/customwebhookbot_django.py
```diff
@@ -0,0 +1,198 @@
...
+        config=uvicorn.Config(
+            app=get_asgi_application(),
+            port=port,
+            use_colors=False,
+            host="127.0.0.1",
+        )
+    )
+
+    # Pass webhook settings to telegram
+    await ptb_application.bot.set_webhook(url=f"{url}/telegram", allowed_updates=Update.ALL_TYPES)
```
harshil21, line 188: this should also go in post_init?

examples/customwebhookbot_flask.py
```diff
@@ -0,0 +1,188 @@
...
+
+    # register handlers
+    application.add_handler(CommandHandler("start", start))
+    application.add_handler(TypeHandler(type=WebhookUpdate, callback=webhook_update))
+
+    # Pass webhook settings to telegram
+    await application.bot.set_webhook(url=f"{url}/telegram", allowed_updates=Update.ALL_TYPES)
+
+    # Set up webserver
+    flask_app = Flask(__name__)
```
harshil21, line 136: i'd rather have all flask decorated functions outside main() tbh, looks cleaner

harshil21 (changes requested):
Wow, didn't realize how similar these examples would be! Unfortuantely I can't setup a webhook right now so can't run these examples. I assume you tested all of them? I also like the inline tabs idea, that's neat

Review comments below are mostly stylistic/design based.

examples/customwebhookbot_django.py
```diff
@@ -0,0 +1,198 @@
...
+    from telegram import __version_info__
+except ImportError:
+    __version_info__ = (0, 0, 0, 0, 0)  # type: ignore[assignment]
+
+if __version_info__ < (20, 0, 0, "alpha", 1):
+    raise RuntimeError(
+        f"This example is not compatible with your current PTB version {TG_VER}. To view the "
+        f"{TG_VER} version of this example, "
+        f"visit https://docs.python-telegram-bot.org/en/v{TG_VER}/examples.html"
+    )
```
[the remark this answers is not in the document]
Bibo-Joshi: I vaguely recall a decision to keep this until a v21 release is there … or until the end of the year? let's move this discussion to the dev chat though :)

examples/customwebhookbot_django.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env python
+# This program is dedicated to the public domain under the CC0 license.
+# pylint: disable=import-error,wrong-import-position
+"""
+Simple example of a bot that uses a custom webhook setup and handles custom updates.
+For the custom webhook setup, the libraries `Django` and `uvicorn` are used. Please
+install them as `pip install Django~=4.2.3 uvicorn~=0.23.1`.
```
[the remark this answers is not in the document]
Bibo-Joshi: Luickly `~=` is still compatible 😋 I can update that sure, but my goal wouldn't be to keep these dependencies up to date all the time …

examples/customwebhookbot_django.py
Bibo-Joshi: b/c djange treats the file as o module and needs to be able to import `urlpatterns` from it …

examples/customwebhookbot_django.py
Bibo-Joshi: doesn't need any `await` so doesn't make a difference IMO

examples/customwebhookbot_django.py
Bibo-Joshi: see above

examples/customwebhookbot_flask.py
Bibo-Joshi: that makes a bit more convoluted to define the `application` (as can be seen in the django version) … From my POV you have to choose between two evils in any case …

harshil21: ok, I can't run the examples and don't have enough experience with the framework, so i'll go with your pov.

harshil21:
> I assume you tested all of them? 

?

Bibo-Joshi:
@harshil21 Yes

PR #3826, opened 2023-08-04
Add `__repr__` methods for selected classes

### LINKED ISSUE #3770: [FEATURE] Provide a "official" string representation for most classes

We should define a [`__repr__`](https://docs.python.org/3/reference/datamodel.html#object.__repr__) for most telegram classes, since it is useful for debugging/inspection purposes. 

Define a `__repr__` in all base classes. Typically, the representation should be of the form `<...some useful description...>`. We can just have it so that it returns the expression which can be used to recreate the object.

_No response_

### Additional context

First suggested in https://github.com/python-telegram-bot/python-telegram-bot/pull/3759#discussion_r1235805437

lemontree210:
closes #3770


lemontree210:
1. I put `__repr__` methods right after `__init__`, but for `Job` I put it in the end because that's where `__eq__` and `__hash__` were.
2. Maybe I should add angular brackets to make output look like automatic `__repr__`
3. Not sure how I can access the whole schedule from `Job`. I can only access the time of next job execution, I think.

harshil21:
> 1. I put `__repr__` methods right after `__init__`, but for `Job` I put it in the end because that's where `__eq__` and `__hash__` were.

I actually highly prefer having magic methods in the beginning of a class, since they define the behaviour of a class. Don't know why none of the linters support such a thing. Let's make a new issue to do this if everyone thinks the same.

Bibo-Joshi:
> > 1. I put `__repr__` methods right after `__init__`, but for `Job` I put it in the end because that's where `__eq__` and `__hash__` were.
> 
> I actually highly prefer having magic methods in the beginning of a class, since they define the behaviour of a class. Don't know why none of the linters support such a thing. Let's make a new issue to do this if everyone thinks the same.

I also like having all magic methods in one place. Since `__init__` already is one, the start makes sense, but I'm also fine with the end usually … For this PR putting `__repr__` after `__init__` or where other magic methods already are sounds good 👍 

> 2. Maybe I should add angular brackets to make output look like automatic `__repr__`

I don't see really the benefit of that …

> 3. Not sure how I can access the whole schedule from `Job`. I can only access the time of next job execution, I think.

shouldn't that be available as `job.trigger`?

lemontree210:
I should create some tests for these methods, right?

telegram/ext/_basehandler.py
```diff
@@ -95,6 +95,9 @@ def __init__(
         self.callback: HandlerCallback[UT, CCT, RT] = callback
         self.block: DVType[bool] = block
 
+    def __repr__(self) -> str:
+        return f"Handler {self.__class__.__name__}. Callback: {self.callback.__qualname__}."
```
Bibo-Joshi, line 99: Maybe just `__name__ with callback`? "Handler MessageHandler" reads a bit like duplication

telegram/ext/_jobqueue.py
```diff
@@ -766,6 +769,13 @@ def __init__(
 
         self._job = cast("APSJob", None)  # skipcq: PTC-W0052
 
+    def __repr__(self) -> str:
+        name_infix = f"Job name {self.name}. " if self.name != self.callback.__name__ else ""
+        return (
+            f"Job ID {self.job.id}. {name_infix}Callback name {self.callback.__name__}. "
+            f"APS job trigger: {self.job.trigger}. Next run: {self.next_t}."
```
Bibo-Joshi, line 776: The "next run" information is valid only temporary, i.e. it depends on the time the job is printed (and iirc it.s not even available until the scheduler started). I vote to leave it out.

telegram/ext/_updater.py
```diff
@@ -122,6 +122,12 @@ def __init__(
         self.__lock = asyncio.Lock()
         self.__polling_task: Optional[asyncio.Task] = None
 
+    def __repr__(self) -> str:
+        return (
+            f"Updater for bot {self.bot}. Number of updates currently in the queue: "
```
Bibo-Joshi, line 127: I vote to leave out the queue status - argumentation see `Job.__repr__`

telegram/ext/_conversationhandler.py
```diff
@@ -440,6 +440,13 @@ def __init__(
                     stacklevel=2,
                 )
 
+    def __repr__(self) -> str:
+        name = self.name or "(no name)"
+        return (
+            f"ConversationHandler {name} with {len(self.states)} states, "
+            f"{len(self.entry_points)} entry point(s) and {len(self.fallbacks)} fallbacks"
+        )
```
Bibo-Joshi, line 448: This info seems a bit exessive to me tbh. How about including the states themself, maybe as list truncated to 3 items?

Bibo-Joshi (commented):
> I should create some tests for these methods, right?

Yes please! :) Even if we don't document this behavior I'd like to avoid regressions.

I only just now looked at the changes for the first time and I see that your style of `__repr__`  indeed reads like a short description. It definetily has benefits in terms of being easier to understand 👍 The downside is that it's easier to get a bit lengthy so that I would suggest to double check the wording & printed information in terms of length. I also now better understand your question about `<…>` :D That does in fact make more sense to me now as with this style it is more important to have some markers for the start and end of the representation. You could write a small helper for `f"<{stuff}>"` to make it easy to apply the same style everywhere.

TBH I always had something like `Application(bot=…)` in mind, but I do see that this assumption is in no way the only (or even "best") thing to do. The upside is that it's shorter, the downside is that's a bit less easy to understand. A downside is that it suggests that you can instantiate the object like that which is not always the case. Using e.g. square brackets instead could alleviate the problem a bit, but I admit that I like the general descriptive style a bit better for these "highlevel" classes (note that for the TG classes which are mostly "data classes", the `ClassName(…)` repr still makes sense for me).

lemontree210:
> note that for the TG classes which are mostly "data classes", the `ClassName(…)` repr still makes sense for me)

I don' think we ever planned to create `__repr__` for Telegram classes though (i.e. subclasses of `TelegramObject`) 🤔 But maybe I should try this style for the other classes as well and see how it goes.

Bibo-Joshi:
> > note that for the TG classes which are mostly "data classes", the `ClassName(…)` repr still makes sense for me)
> 
> I don' think we ever planned to create `__repr__` for Telegram classes though (i.e. subclasses of `TelegramObject`) 🤔 But maybe I should try this style for the other classes as well and see how it goes.

Sure we did:

* #3234 
* https://docs.python-telegram-bot.org/en/stable/telegram.telegramobject.html#telegram.TelegramObject.__repr__

lemontree210:
> A downside is that it suggests that you can instantiate the object like that which is not always the case.

How about we write something like `Application(param1=value1, param2=value2, ... )` - with an ellipsis to make it clear that these params are not enough to instantiate the application?

Bibo-Joshi:
> > A downside is that it suggests that you can instantiate the object like that which is not always the case.
> 
> How about we write something like `Application(param1=value1, param2=value2, ... )` - with an ellipsis to make it clear that these params are not enough to instantiate the application?

I'm not sure if that's much better. Thinking about it, e.g. for Jobs the trigger is actually not an argument that we directly pass to neither `Job` nor `run_*`, so showing it as kind of argument would be rather confusing, I guess 🤔

telegram/ext/_basehandler.py
lemontree210: Done

telegram/ext/_updater.py
lemontree210: Done

telegram/ext/_conversationhandler.py
lemontree210: Done (I used `.items()` for more information)

telegram/ext/_callbackcontext.py
```diff
@@ -149,6 +150,9 @@ def __init__(
             Union[Generator[Optional[Future[object]], None, Any], Awaitable[Any]]
         ] = None
 
+    def __repr__(self) -> str:
+        return build_repr_with_selected_attrs(self, update=self.update)
```
Bibo-Joshi, line 154: One comment: `update` will usually have a very long representation (`Update(id, message=Message(from_user=User(id, …), dat=2023-01-01, chat=Chat(id, title='tiltle', …), …), …)`).

Printing the full object makes it more clear what update this context belongs to, but is every long.
Printing instead only the updates id would make that shorter, but is not as descriptive.

I'm fine either way, just thought about this when I saw the test case

tests/ext/test_conversationhandler.py
```diff
@@ -289,6 +291,29 @@ def test_init_persistent_no_name(self):
                 self.entry_points, states=self.states, fallbacks=[], persistent=True
             )
 
+    def test_repr(self):
+        # This function tests behaviour inherited from BaseHandler.
+        # TODO should this test be repeated in all other subclasses of BaseHandler except for
+        #  ConversationHandler that has its own __repr__?
```
Bibo-Joshi, line 297: IMO it would suffice to test this once in `test_basehandler`.

tests/ext/test_jobqueue.py
```diff
@@ -85,6 +85,22 @@ class TestJobQueue:
...
+        jq = JobQueue()
+        jq.set_application(app)
+        assert repr(jq) == f"JobQueue[application={app!r}]"
+
+        callback = self.job_run_once
+        job = jq.run_once(callback, 0, name="name2")
+        assert repr(job) == (
+            f"Job[id={job.job.id}, name={job.name}, callback=job_run_once, "
+            f"trigger=date["
+            f"{datetime.datetime.now(tz=datetime.timezone.utc).strftime('%Y-%m-%d %H:%M:%S UTC')}"
```
Bibo-Joshi, line 100: this looks like it can be flaky. how about defining a fixed date in the future and using that in both places? i.e.

```python
dtm = datetime.utcnow() + timedelta(year=1)
job = jq.run_once(callback, dtm)
assert repr(job) == (… dtm.strftime(…))
```

tests/ext/test_typehandler.py
```diff
@@ -35,6 +35,16 @@ def test_slot_behaviour(self):
             assert getattr(inst, attr, "err") != "err", f"got extra slot '{attr}'"
         assert len(mro_slots(inst)) == len(set(mro_slots(inst))), "duplicate slot"
 
+    def test_repr(self):
+        # This function tests behaviour inherited from BaseHandler.
+        # TODO should this test be repeated in all other subclasses of BaseHandler except for
```
Bibo-Joshi, line 40: see above

Bibo-Joshi (changes requested):

tests/ext/test_conversationhandler.py
lemontree210: Done

tests/ext/test_typehandler.py
lemontree210: Yes, moved the test from test_typehandler.py to test_basehandler.py (with necessary amendments)

tests/ext/test_jobqueue.py
lemontree210: I thought `0` would eliminate the flakiness 😵‍💫 
Done :)

tests/ext/test_conversationhandler.py
Bibo-Joshi: Sorry, CH should then ofc be tested individually. But I see no need to test in all handler classes that simply inherit from BaseHandler :)

lemontree210: OK, I think the test for CH is already in place :)

telegram/ext/_callbackcontext.py
Bibo-Joshi: That seems a bit excessive to me, tbh 😆

lemontree210: Actually, I just realized that `self.update` is wrong because it references a method that has nothing to do with `Update` object. The second thing I realized is that I can't see any easy way of getting the update :) I can think of introducing a new private attribute in CCT (like `update_id`) and set its value in `from_update()`. As far as I understand, I can't touch the `application.update_queue` because if I call `.get()` on it, the update will be removed from the queue.

Bibo-Joshi: Huh, you're right. Even more: update doesn't have to be an instance of TG.Update and in that case meaningful references are even harder ...

Bibo-Joshi: I guess, then it's more sane to skip `__repr__` for `CallbackContext` for now …

lemontree210: OK, removing

telegram/_utils/repr.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def build_repr_with_selected_attrs(obj: object, **kwargs: Any) -> str:
+    """Create ``__repr__`` string in the style `Classname(arg1=1, arg2=2).`
+
+    Attributes that are to be used in the representation, are passed as kwargs.
+    """
+
+    def stringify(key: str, val: Any) -> str:
+        return f"{key}={val.__qualname__ if callable(val) else val}"
```
Bibo-Joshi, line 36: this could be moved outside of `build_repr_*` and made protected with a leading underscore.

tests/ext/test_conversationhandler.py
Bibo-Joshi, line 297: I guess this can be removed now? :)

Bibo-Joshi (commented):
Thanks for the updates! Just two minor comments, otherwise LGTM :) Let's see if we can another review, though

telegram/_utils/repr.py
lemontree210: Done. It's an interesting, almost holywar-like discussion about functions inside of functions. I've heard many arguments pro and contra over these months, what's yours, in a nutshell, for not having one function inside another?

tests/ext/test_conversationhandler.py
lemontree210: Done, thank you!

telegram/_utils/repr.py
Bibo-Joshi: I would expect that there is a (probably tiny) overhead in defining the inner function on every call to the outer function. But I have to "proper" argumentation at hand for my suggestion :D

telegram/ext/_conversationhandler.py
```diff
@@ -440,6 +441,20 @@ def __init__(
                     stacklevel=2,
                 )
 
+    def __repr__(self) -> str:
+        truncation_threshold = 3
+        states_string = (
+            f"{', '.join(tuple(str(item) for item in self.states.items())[:truncation_threshold])}"
+        )
+        if len(self.states) > truncation_threshold:
```
Bibo-Joshi, line 449: ah, codecov complains that the (implicit) `else` clause here is not covered …

lemontree210: Done

telegram/ext/_conversationhandler.py
```diff
@@ -440,6 +441,20 @@ def __init__(
                     stacklevel=2,
                 )
 
+    def __repr__(self) -> str:
+        truncation_threshold = 3
+        states_string = (
+            f"{', '.join(tuple(str(item) for item in self.states.items())[:truncation_threshold])}"
```
harshil21, line 447: why not keep the states_string as a `str(self.states)`? I like seeing the ":" and curly brackets in the representation. For showing only 3 items we can do something like 

`states_string = str(dict(list(self.states.items())[:3]))` 
truncation part: 
`states_string = states_string[:-1] + ", ...}"`

tests/ext/test_jobqueue.py
```diff
@@ -85,6 +85,23 @@ class TestJobQueue:
         " We recommend double checking if the passed value is correct."
     )
 
+    async def test_repr(self, app):
+        import datetime as dtm
```
harshil21, line 89: this import is already defined at the top?

harshil21 (commented):
Nice PR! I see that it has gone through several iterations already, here's my comments now :) That square bracket approach in the repr is easy to understand 👍🏽

tests/ext/test_jobqueue.py
lemontree210: Yep. Removed

tests/ext/test_application.py
```diff
@@ -201,6 +201,28 @@ async def post_stop(application: Application) -> None:
...
         assert isinstance(app.user_data[1], dict)
 
+    @pytest.mark.filterwarnings("ignore: `Application` instances should")
+    async def test_repr(self, one_time_bot):
+        update_queue = asyncio.Queue()
+        job_queue = JobQueue()
+        context_types = ContextTypes()
+        update_processor = SimpleUpdateProcessor(1)
+
+        app = Application(
```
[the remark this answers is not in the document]
lemontree210: Yes we can :)

telegram/ext/_conversationhandler.py
lemontree210: Done

harshil21 (approved):
good feature to have!

Poolitzer (approved):
I like it

Bibo-Joshi:
I was about to hit merge - but then I remembered one more thing 🙈 For `TelegramObject` we documented all the dunder methods and for consistency I would vote to do the same here. However, the code change itself is already approved and I don't want to stretch this PR needlessly. We can do this as a follow-up PR. @lemontree210 please let me now what you prefer

lemontree210:
> For `TelegramObject` we documented all the dunder methods and for consistency I would vote to do the same here. 

consistency is key! :)

tests/ext/test_application.py
```diff
@@ -201,6 +201,10 @@ async def post_stop(application: Application) -> None:
         assert isinstance(app.chat_data[1], dict)
         assert isinstance(app.user_data[1], dict)
 
+    @pytest.mark.filterwarnings("ignore: `Application` instances should")
```
harshil21, line 204: is that ignore still needed?

lemontree210: I guess not! :)

Poolitzer (approved):
LGTM

Bibo-Joshi:
DS failure unrelated - see #3888. Merging.

PR #3847, opened 2023-08-17
Use python 3.12rc1 in tests & add classifier

harshil21:
Uses latest rc in tests.

We should also build wheels for py 3.12 on next release, as [recommended by python devs](https://pythoninsider.blogspot.com/2023/08/python-3120-release-candidate-1-released.html)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -109,6 +109,7 @@ def get_setup_kwargs(raw=False):
             "Programming Language :: Python :: 3.9",
             "Programming Language :: Python :: 3.10",
             "Programming Language :: Python :: 3.11",
+            "Programming Language :: Python :: 3.12",
         ],
         python_requires=">=3.8",
     )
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -109,6 +109,7 @@ def get_setup_kwargs(raw=False):
             "Programming Language :: Python :: 3.9",
             "Programming Language :: Python :: 3.10",
             "Programming Language :: Python :: 3.11",
+            "Programming Language :: Python :: 3.12",
```
Bibo-Joshi, line 112: If we add this then we should also update the version number in the readme :)

Bibo-Joshi (commented):
Luckily we have wheels independent of the python version, so nothing to change for PyPi at all 🐧

setup.py
harshil21: the badge can't be updated to show 3.12, maybe it reads this setup.py change and updates it?

Bibo-Joshi: The badge reads the python version from PyPi. What I mean is the version number in the introduction readme, but I just see that "3.8+" already covers that 🐳

PR #3857, opened 2023-08-24
Add new MENTION filter [feature]

### LINKED ISSUE #3799: [FEATURE] Add bot mentioned filter

There is currently no filter defined by us which simply checks if message contains a bot mention, so `@botusername message`.

Make a new filter which checks if a message contains the mention of a bot. So first check if the message has the [`telegram.constants.MessageEntityType.MENTION`](https://docs.python-telegram-bot.org/en/stable/telegram.constants.html#telegram.constants.MessageEntityType.MENTION), and then check if the username is the one of the bot

### Additional context

Maybe having the only_start parameter here would be nice, same as with the command filter.

## Assignees:

~Current assignee: @ElmurodovJavohir - #3857. If they don't commit/reply by Oct 7th, the next in line to tackle this issue is:~

~- @Prathmesh-rajurkar~
~- @AbhigyanBaruah~

Issue is being tackled by dev team due to the continued inactivity from the contributors.

ElmurodovJavohir:
closes #3799


ElmurodovJavohir:
Some test failed is it okay, or should I fix tests

ElmurodovJavohir:
@Bibo-Joshi I have added unit test, and fixed docs build. But others test failing, that I not changed or effect, is it okay, or am I wrong any thing.

telegram/ext/filters.py
```diff
@@ -2545,3 +2547,51 @@ def filter(self, message: Message) -> bool:
...
+        )
+
+    def filter(self, message: Message) -> bool:
+        if not any(entity.type == self.entity_type for entity in message.entities):
+            return False
+
+        if self.username:
+            if message.text and f"@{self.username}" in message.text:
+                return True
+            return False
```
harshil21, line 2585: codecov is failing because this line isn't tested

ElmurodovJavohir: Fixed

telegram/ext/filters.py
```diff
@@ -2545,3 +2547,51 @@ def filter(self, message: Message) -> bool:
...
+            exact matches are allowed. If not specified, will allow any text message with mention.
+    """
+
+    __slots__ = ("username",)
+    entity_type: str = MessageEntity.MENTION
+
+    def __init__(self, username: Optional[Union[List[str], Tuple[str, ...]]] = None):
+        self.username: Optional[Sequence[str]] = username
+        super().__init__(
+            name=f"filters.Regex(r'@{username}') & filters.Entity({self.entity_type})"
```
Bibo-Joshi, line 2573: Please use somethling like `f"filters.Mention({username})"`, as `Regex(…) & Entity(…)` is not what this filter is doing

telegram/ext/filters.py
```diff
@@ -2545,3 +2547,51 @@ def filter(self, message: Message) -> bool:
...
+            if username
+            else f"filters.Entity({self.entity_type})"
+        )
+
+    def filter(self, message: Message) -> bool:
```
Bibo-Joshi, line 2578: please make sure that both mention and text mention entities are detected as explained it https://github.com/python-telegram-bot/python-telegram-bot/issues/3799#issuecomment-1636387465

telegram/ext/filters.py
```diff
@@ -2545,3 +2547,51 @@ def filter(self, message: Message) -> bool:
...
+
+MENTION = Mention()
+"""
+Shortcut for :class:`telegram.ext.filters.Mention()`.
+
+Examples:
+    To allow any text message that contains mention, simply use
+    ``MessageHandler(filters.MENTION, callback_method)`` or
+    ``MessageHandler(filters.MENTION("username"), callback_method)``.
+"""
```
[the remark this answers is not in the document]
ElmurodovJavohir: You mean username should be required ?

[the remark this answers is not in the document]
harshil21: > You mean username should be required ?

yes, doesn't make sense to have `username` optional since that is already covered by `filters.Entity`.

telegram/ext/filters.py
```diff
@@ -2545,3 +2548,74 @@ def filter(self, message: Message) -> bool:
...
+
+MENTION = Mention
+"""
+Shortcut for :class:`telegram.ext.filters.Mention()`.
+
+Examples:
+    To allow any text message that contains mention, simply use
+    ``MessageHandler(filters.MENTION("@username"), callback_method)`` or
+    ``MessageHandler(filters.MENTION(["@username",`123456`,`user`), callback_method)``.
+"""
```
Bibo-Joshi, line 2621: https://github.com/python-telegram-bot/python-telegram-bot/pull/3857#discussion_r1305794900

please remove the completely

telegram/ext/filters.py
```diff
@@ -2545,3 +2548,74 @@ def filter(self, message: Message) -> bool:
...
+        self._mentions = self._parse_mentions(mentions)
+        super().__init__(name=f"filters.Mention({mentions})")
+
+    @staticmethod
+    def _parse_mentions(
+        mentions: SCT[Union[int, str, TGUser]]
+    ) -> Iterable[Union[int, str, TGUser]]:
+        if isinstance(mentions, Iterable) and not isinstance(mentions, str):
+            return set(mentions)
+        return set(mentions)
```
Bibo-Joshi, line 2583:
```suggestion
        return {mentions}
```

telegram/ext/filters.py
```diff
@@ -2545,3 +2548,74 @@ def filter(self, message: Message) -> bool:
...
+
+    @property
+    def mentions(self) -> FrozenSet[Union[int, str, TGUser]]:
```
Bibo-Joshi, line 2586: Please add a docstring

telegram/ext/filters.py
```diff
@@ -2545,3 +2548,74 @@ def filter(self, message: Message) -> bool:
...
+            and message.text
+            and mention in message.text
+        ):
+            return True
+        return False
+
+    def filter(self, message: Message) -> bool:
+        if any(self.check_mention(message, mention) for mention in self.mentions):
+            return True
+        return False
```
Bibo-Joshi, line 2610: Nice way of writing the check in a consisce way 👍 

```suggestion
        return any(self.check_mention(message, mention) for mention in self.mentions)
```

telegram/ext/filters.py
```diff
@@ -2545,3 +2548,74 @@ def filter(self, message: Message) -> bool:
...
+    """Mention messages. If list of username or chat_id or `User` objects passed, it filters
+    messages to only allow those whose contains one of username  or chat_id or `User`.
+    Note:
+        Usernames only allowed with @ prefix. Example: "@username"
+        Chat id only allowed with positive numbers. Example: 123456
+    Examples:
+        A simple use case for passing username is to allow only messages that were sent with
+        mention username:
+        MessageHandler(filters.Mention("@username"), callback_method)
+        MessageHandler(filters.Mention(["@username",`123456`,`user`), callback_method)
```
Bibo-Joshi, line 2563:
```suggestion
        MessageHandler(filters.Mention(["@username",  123456, user), callback_method)
```

Let's pass chat_ids as integers and not as strings.

telegram/ext/filters.py
```diff
@@ -2545,3 +2548,74 @@ def filter(self, message: Message) -> bool:
...
+    @property
+    def mentions(self) -> FrozenSet[Union[int, str, TGUser]]:
+        return frozenset(self._mentions)
+
+    def check_mention(self, message: Message, mention: Union[int, str, TGUser]) -> bool:
+        if isinstance(mention, TGUser) and bool(
+            any(mention.id == e.user.id for e in message.entities if e.user)
+        ):
+            return True
+        if str(mention).isdigit() and bool(
```
Bibo-Joshi, line 2594:
```suggestion
        if isinstance(mention, int) and bool(
```

telegram/ext/filters.py
```diff
@@ -2545,3 +2548,74 @@ def filter(self, message: Message) -> bool:
...
+        if isinstance(mentions, Iterable) and not isinstance(mentions, str):
+            return set(mentions)
+        return set(mentions)
+
+    @property
+    def mentions(self) -> FrozenSet[Union[int, str, TGUser]]:
+        return frozenset(self._mentions)
+
+    def check_mention(self, message: Message, mention: Union[int, str, TGUser]) -> bool:
+        if isinstance(mention, TGUser) and bool(
```
Bibo-Joshi, line 2590: you should also check if `mention.username` is in the `mention` type mentions

telegram/ext/filters.py
```diff
@@ -2545,3 +2548,74 @@ def filter(self, message: Message) -> bool:
...
+            return True
+        if str(mention).isdigit() and bool(
+            any(mention == e.user.id for e in message.entities if e.user)
+        ):
+            return True
+        if (
+            isinstance(mention, str)
+            and bool(any(e.type == MessageEntity.MENTION for e in message.entities))
+            and message.text
+            and mention in message.text
```
Bibo-Joshi, line 2602: Unfortunately this check doesn't suffice. The following message would pass the filter `Mention(mentions="@notpresent")`:

> This message mentions only @Bibo-Joshi, but `@notpresent` is still part of the text.

Plesae use [`Message.parse_entities`](https://docs.python-telegram-bot.org/en/stable/telegram.message.html#telegram.Message.parse_entities) to check if the username is actually a mention in the message :)

telegram/ext/filters.py
```diff
@@ -2545,3 +2548,74 @@ def filter(self, message: Message) -> bool:
...
+    ) -> Iterable[Union[int, str, TGUser]]:
+        if isinstance(mentions, Iterable) and not isinstance(mentions, str):
+            return set(mentions)
+        return set(mentions)
+
+    @property
+    def mentions(self) -> FrozenSet[Union[int, str, TGUser]]:
+        return frozenset(self._mentions)
+
+    def check_mention(self, message: Message, mention: Union[int, str, TGUser]) -> bool:
```
Bibo-Joshi, line 2589:
```suggestion
    @staticmethod
    def _check_mention(message: Message, mention: Union[int, str, TGUser]) -> bool:
```

No need to make the method part of documented behavior :)

Bibo-Joshi (changes requested):
Thanks for the updates! I left another batch of comments. I didn't have a look at the tests yet …

telegram/ext/filters.py
```diff
@@ -2545,3 +2546,60 @@ def filter(self, message: Message) -> bool:
...
 VOICE = _Voice("filters.VOICE")
 """Messages that contain :attr:`telegram.Message.voice`."""
+
+
+class Mention(MessageFilter):
+    """Mention messages. If list of username or chat_id or :attr:`telegram.User` objects passed,
+    it filters messages to only allow those whose contains one of username
+    or chat_id or :attr:`telegram.User`.
+    Note:
+        Usernames only allowed with @ prefix. Example: "@username"
```
Bibo-Joshi, line 2556: I just see that we discard leading '@' in all other filters that accept usernames, i.e. it doesn't matter if you pass the username with or without '@'. I would like to do the same here.

telegram/ext/filters.py
```diff
@@ -2545,3 +2546,60 @@ def filter(self, message: Message) -> bool:
...
+        A simple use case for passing username is to allow only messages that were sent with
+        mention username:
+        MessageHandler(filters.Mention("@username"), callback_method)
+        MessageHandler(filters.Mention(["@username",123456, `user` ), callback_method)
+    Args:
+        mentions (Union[:obj:`str` ,:obj:`str` ,:attr:`telegram.User` ] ||
+        Collection[Union[:obj:`str` ,:obj`str` ,:attr:`telegram.User` ]]):
+        Which messages to allow with username or chat_id or :attr:`telegram.User` object.
+        If one of them matched it will allow. If not specified,
+        will allow any text message with mention.
```
Bibo-Joshi, line 2568:
```suggestion
        If one of them matched it will allow.
```

The argument is not optional :)

telegram/ext/filters.py
```diff
@@ -2545,3 +2546,60 @@ def filter(self, message: Message) -> bool:
...
+
+    @staticmethod
+    def _check_mention(message: Message, mention: Union[int, str, TGUser]) -> bool:
+        if isinstance(mention, TGUser):
+            return bool(any(mention.id == e.user.id for e in message.entities if e.user))
+        if isinstance(mention, int):
+            return bool(any(mention == e.user.id for e in message.entities if e.user))
+        return bool(
+            isinstance(mention, str)
+            and bool(any(e.type == MessageEntity.MENTION for e in message.parse_entities()))
```
Bibo-Joshi, line 2599: Please have a close look at the docs of `Message.parse_entities()` to see what it returns. Hint: It's not a list of message entities ;)

Bibo-Joshi (changes requested):

ElmurodovJavohir:
@Bibo-Joshi Thanks your feedbacks, I will fix it

telegram/ext/filters.py
ElmurodovJavohir: Oh, After using in real example I understand you, and fixed this issue

telegram/ext/filters.py
ElmurodovJavohir: Fixed

telegram/ext/filters.py
ElmurodovJavohir: I have fixed. I check all argument str, add '@'  if not added.

.templates/$$var_template/$$var_filename.js
```diff
@@ -0,0 +1,2 @@
+export default function $$var_textInFile() {
```
Bibo-Joshi, line 1: Please remove the `.templates` directory and the `template.config.js` file.

telegram/ext/filters.py
```diff
@@ -2545,3 +2546,72 @@ def filter(self, message: Message) -> bool:
...
+        super().__init__(name=f"filters.Mention({mentions})")
+
+    @staticmethod
+    def _parse_mentions(mentions: SCT[Union[int, str, TGUser]]) -> Set[Union[int, str, TGUser]]:
+        if isinstance(mentions, Iterable) and not isinstance(mentions, str):
+            return set(mentions)
+        return {mentions}
+
+    @staticmethod
+    def _fix_mention_username(mention: Union[int, str, TGUser]) -> Union[int, str, TGUser]:
```
Bibo-Joshi, line 2582:
```suggestion
    def _parse_mention(mention: Union[int, str, TGUser]) -> Union[int, str, TGUser]:
```

telegram/ext/filters.py
```diff
@@ -2545,3 +2546,72 @@ def filter(self, message: Message) -> bool:
...
+    """
+
+    __slots__ = ("mention", "_mentions")
+
+    def __init__(self, mentions: SCT[Union[int, str, TGUser]]):
+        self._mentions = self._parse_mentions(mentions)
+        super().__init__(name=f"filters.Mention({mentions})")
+
+    @staticmethod
+    def _parse_mentions(mentions: SCT[Union[int, str, TGUser]]) -> Set[Union[int, str, TGUser]]:
```
Bibo-Joshi, line 2576: you can call `_fix_mention_username` already in this method. THen you don't have to call it on every call to the `mentions` property

telegram/ext/filters.py
```diff
@@ -2545,3 +2546,72 @@ def filter(self, message: Message) -> bool:
...
+            return set(mentions)
+        return {mentions}
+
+    @staticmethod
+    def _fix_mention_username(mention: Union[int, str, TGUser]) -> Union[int, str, TGUser]:
+        if not isinstance(mention, str):
+            return mention
+        if mention.startswith("@"):
+            return mention
+        return f"@{mention}"
```
Bibo-Joshi, line 2587: This *adds* a leading `@` rather than removing it. Removing it is more sane IMO, since `User.username` does not have the leading `@`

telegram/ext/filters.py
```diff
@@ -2545,3 +2546,72 @@ def filter(self, message: Message) -> bool:
...
+            frozenset(Union[:obj:`int` , :obj:`str`, :attr:`telegram.User` ])
+        """
+        # correcting str usernames.
+        _mentions = {self._fix_mention_username(mention) for mention in self._mentions}
+        return frozenset(_mentions)
+
+    @staticmethod
+    def _check_mention(message: Message, mention: Union[int, str, TGUser]) -> bool:
+        if isinstance(mention, TGUser):
+            return bool(any(mention.id == e.user.id for e in message.entities if e.user))
```
Bibo-Joshi, line 2602: one should also check if `mention.username` is in one of the `MessageEntities` of type `MessageEntity.MENTION`

telegram/ext/filters.py
```diff
@@ -2545,3 +2546,72 @@ def filter(self, message: Message) -> bool:
...
+        if isinstance(mention, int):
+            return bool(any(mention == e.user.id for e in message.entities if e.user))
+        return bool(
+            isinstance(mention, str)
```
Bibo-Joshi, line 2606: Is there a specifi reason for this type check? Seems superfluous to me …

telegram/ext/filters.py
```diff
@@ -2545,3 +2546,72 @@ def filter(self, message: Message) -> bool:
...
+            and bool(
+                any(
+                    (entity.type == MessageEntity.MENTION and message.text == entity_value)
+                    for entity, entity_value in message.parse_entities().items()
```
Bibo-Joshi, line 2610:
```suggestion
                    mention == entity_value
                    for entity_value in message.parse_entities(types=[MessageEntity.MENTION]).values()
```
`parse_entities` can already filter that for you :)

please double check how the leading `@` behaves here

Bibo-Joshi (changes requested):
Thanks for the updates. I left another batch of comments. Didn't look at the tests yet …

telegram/ext/filters.py
ElmurodovJavohir: What about

Bibo-Joshi:
Due to inactivity, closing in favor of #3941

PR #3868, opened 2023-08-31
Verify type hints for Bot method & class parameters

harshil21:
Expands `test_official` by verifying that the type hints in our lib is same as TG API's. 
Also changes type hint of `user_id` in bot methods from `int | str` to just `int` as specified by the API.

~The change is breaking because `PassportFile.file_date` now accepts a datetime only, and its `de_json` now converts int -> datetime.~ No longer breaking.

Tests pass locally, seem to be failing here mysteriously

After review, before merge:
- [x] remove print statements


harshil21:
suggestions for optimizations / a different approach welcome btw

tests/test_official.py
```diff
@@ -264,9 +287,222 @@ def check_defaults_type(ptb_param: inspect.Parameter) -> bool:
...
+        else:
+            mapped_type = (
+                getattr(telegram, tg_param_type),  # This will fail if it's not from telegram mod
+                ForwardRef(tg_param_type),
+                tg_param_type,  # for some reason, some annotations are not ForwardRefs, i.e. str.
+            )
+        print("len mapped 0 ", mapped_type)
+    elif len(mapped) == 1:
+        print("len mapped 1 ", mapped)
+        mapped_type = mapped.pop()
```
Bibo-Joshi, line 337: what about `len(mapped) > 1`?

tests/test_official.py
```diff
@@ -264,9 +287,222 @@ def check_defaults_type(ptb_param: inspect.Parameter) -> bool:
...
+        ptb_annotation = ptb_param.annotation  # if it's not nested, just use the annotation
+
+    print(f"{tg_param_type=} ||| {mapped_type=} ||| {ptb_annotation=}")
+
+    if isinstance(ptb_annotation, list):
+        # Some cleaning:
+        # Remove 'Optional[...]' from the annotation if it's present. We do it this way since: 1)
+        # we already check if argument should be optional or not + type checkers will complain.
+        # 2) we don't want to deal with Optional[..] skewing the equality check of our annotation
+        # with the official API, i.e. Optional[int] == int.
```
Bibo-Joshi, line 350: I don't really get this second part. Why do we want to treat `Optional[int]` as equivalent to `int`?

tests/test_official.py
```diff
@@ -264,9 +287,222 @@ def check_defaults_type(ptb_param: inspect.Parameter) -> bool:
...
+    # Now let's do the checking, starting with "Array of ..." types.
+    if "Array of " in tg_param_type:
+        print("array of ", ptb_param.annotation)
+
+        # For exceptions just check if they contain the annotation
+        array_of_exceptions = {  # We accept more types than the official API
+            "results": "InlineQueryResult",
+            "commands": "BotCommand",
+            "keyboard": "KeyboardButton",
+        }
```
Bibo-Joshi, line 380: This should be configurabe somewhere more visible and IMO we should also specify which types we allow in addition

tests/test_official.py
```diff
@@ -264,9 +287,222 @@ def check_defaults_type(ptb_param: inspect.Parameter) -> bool:
...
+        if ptb_param.name in array_of_exceptions:
+            return array_of_exceptions[ptb_param.name] in str(ptb_annotation)
+
+        # let's import obj
```
Bibo-Joshi, line 384: why "import"?

tests/test_official.py
```diff
@@ -264,9 +287,222 @@ def check_defaults_type(ptb_param: inspect.Parameter) -> bool:
...
+        return any(mapped_type[o] == ptb_annotation for o in unionized_objs)
+
+    DEFAULT_VAL_PARAMS = {
+        "parse_mode",
+        "disable_notification",
+        "allow_sending_without_reply",
+        "protect_content",
+        "disable_web_page_preview",
+        "explanation_parse_mode",
+    }
```
Bibo-Joshi, line 420: Can we directly iterate over the attributes of `Defaults` here? Avoids duplication …

tests/test_official.py
```diff
@@ -264,9 +287,222 @@ def check_defaults_type(ptb_param: inspect.Parameter) -> bool:
...
+        # Check if it's DVInput or ODVInput
+        for param_type in [DVInput, ODVInput]:
+            parsed = param_type[mapped_type]
+            if ptb_annotation == parsed:
+                return True
+        # print(f"cause {mapped_type} != {ptb_annotation} for {ptb_param.name}")
+        return False
+
+    # Special case for send_* methods where we accept more types than the official API:
+    additional_types = {
```
Bibo-Joshi, line 433: This should be configurabe somewhere more visible

tests/test_official.py
```diff
@@ -264,9 +287,222 @@ def check_defaults_type(ptb_param: inspect.Parameter) -> bool:
...
+        ("media", True): str,  # actual: Union[str, InputMedia*, FileInput]
+        ("data", True): str,  # actual: Union[IdDocumentData, PersonalDetails, ResidentialAddress]
+    }
+
+    for (param_name, expected_class), exception_type in exceptions.items():
+        if ptb_param.name == param_name and is_class is expected_class:
+            ptb_annotation = exception_type
+
+    # Special case for datetimes
+    if re.search(r"([_]+|\b)date[^\w]*\b", ptb_param.name):
```
Bibo-Joshi, line 469: How about instead searching for "unix time" in the tg dowstring?

Bibo-Joshi (commented):
Pfew, that is quite some code 🤯 I went through it once and left a bunch of comments, but I probably didn't get everything :D 
I feel like as a follow-up of this PR, a refactoring of test_official into multiple modules is in order. I'll create an issue for that.

telegram/_bot.py
```diff
@@ -2521,7 +2521,7 @@ async def send_contact(
     @_log
     async def send_game(
         self,
-        chat_id: Union[int, str],
+        chat_id: int,
```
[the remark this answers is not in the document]
harshil21: yes, just a type hint change.

telegram/_passport/encryptedpassportelement.py
```diff
@@ -151,7 +151,7 @@ def __init__(
         self,
         type: str,  # pylint: disable=redefined-builtin
         hash: str,  # pylint: disable=redefined-builtin
-        data: Optional[PersonalDetails] = None,
+        data: Optional[Union[PersonalDetails, IdDocumentData, ResidentialAddress]] = None,
```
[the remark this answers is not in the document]
harshil21: hm no, de_json handles the case appropriately it seems, so this is only a type hint issue.

telegram/_passport/passportelementerrors.py
```diff
@@ -357,28 +363,33 @@ class PassportElementErrorTranslationFiles(PassportElementError):
...
+        file_hashes (Sequence[:obj:`str`]): List of base64-encoded file hashes.
         message (:obj:`str`): Error message.
 
     Attributes:
         type (:obj:`str`): Type of element of the user's Telegram Passport which has the issue,
             one of ``"passport"``, ``"driver_license"``, ``"identity_card"``,
             ``"internal_passport"``, ``"utility_bill"``, ``"bank_statement"``,
             ``"rental_agreement"``, ``"passport_registration"``, ``"temporary_registration"``.
-        file_hashes (List[:obj:`str`]): List of base64-encoded file hashes.
+        file_hashes (Tuple[:obj:`str`]): List of base64-encoded file hashes.
```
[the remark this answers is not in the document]
harshil21: so when accessed, do you want it to return a list or a tuple? I guess keep it as list then, right?

telegram/_passport/passportfile.py
```diff
@@ -111,6 +111,9 @@ def de_json_decrypted(
         if not data:
             return None
 
+        loc_tzinfo = extract_tzinfo_from_defaults(bot)
+
+        data["file_date"] = from_timestamp(data.get("last_error_date"), tzinfo=loc_tzinfo)
```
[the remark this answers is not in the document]
harshil21: same question. It continue doing the old thing but now raise a warning that it will change in v21 right?

telegram/_payment/orderinfo.py
```diff
@@ -56,15 +56,15 @@ def __init__(
...
+        shipping_address: Optional[ShippingAddress] = None,
         *,
         api_kwargs: Optional[JSONDict] = None,
     ):
         super().__init__(api_kwargs=api_kwargs)
         self.name: Optional[str] = name
         self.phone_number: Optional[str] = phone_number
         self.email: Optional[str] = email
-        self.shipping_address: Optional[str] = shipping_address
+        self.shipping_address: Optional[ShippingAddress] = shipping_address
```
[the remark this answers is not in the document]
harshil21: nope, tests didn't fail and de_json handles it correctly

tests/README.rst
```diff
@@ -72,7 +72,7 @@ complete and correct. To run it, export an environment variable first:
 
     $ export TEST_OFFICIAL=true
 
-and then run ``pytest tests/test_official.py``.
+and then run ``pytest tests/test_official.py``. Note: You need py 3.11+ to run this test.
```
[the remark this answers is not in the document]
harshil21: I put 3.11 since the test suite runs on that. Technically would need at least 3.9+ because of the new type hinting syntax

telegram/_bot.py
```diff
@@ -2826,7 +2826,7 @@ async def answer_inline_query(
     @_log
     async def get_user_profile_photos(
         self,
-        user_id: Union[str, int],
+        user_id: int,
```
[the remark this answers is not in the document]
harshil21: yea so the type hint is corrected to match the docs

tests/test_official.py
```diff
@@ -264,9 +287,222 @@ def check_defaults_type(ptb_param: inspect.Parameter) -> bool:
...
+    """
+    # In order to evaluate the type annotation, we need to first have a mapping of the types
+    # specified in the official API to our types. The keys are types in the column of official API.
+    TYPE_MAPPING: dict[str, set[Any]] = {
+        "Integer or String": {int | str},
+        "Integer": {int},
+        "String": {str},
+        r"Boolean|True": {bool},
+        r"Float(?: number)?": {float},
+        r"Array of (?:Array of )?[\w\,\s]*": {Sequence},
```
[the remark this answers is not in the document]
harshil21: > Can't we be more specific here 

I'd like to do that elegantly but I don't see a easy way, in the end it becomes what I already did below.

> 1) distinguishing 1D from 2D arrays

I did that first here, but then we now need `if` checks for when the values are tuples like: `{(Sequence, Sequence)}` and that again reaches the same level of complexity.

> 2) knownig the type of the entries? 

had an idea of manually instantating `typing._GenericAlias` and then substituting the actual object in but again complicated.

If you have a better way, i'm all ears.

> I would at least leave a comment about that here.

👍🏽

tests/test_official.py
harshil21: should never happen, I can add an assert for that

telegram/_passport/passportelementerrors.py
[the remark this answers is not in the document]
Bibo-Joshi: yes

telegram/_passport/passportfile.py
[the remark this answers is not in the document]
Bibo-Joshi: yes. The warning can say when it was deprecated and point to the stability policy

tests/test_official.py
harshil21: I phrased it poorly but essentially when the API docs tells a parameter is Optional, we have to annotate it as `Optional[obj]`. But we want to check if our `obj` is same as the API `obj`, and wrapping that in `Optional` is annoying since python evaluates `Optional[obj] != obj` (obviously)

tests/README.rst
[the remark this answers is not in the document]
Bibo-Joshi: Is the `|` syntax required? I saw that you used it somewhere in the code … if that's hard to forgoe, keep it as is, but please adjust to py3.9+ here, then :)

tests/test_official.py
harshil21: specify how? as a comment? If we also hard code the other types, then we have more maintainance work

tests/test_official.py
harshil21: remnant of old approach, will remove

tests/test_official.py
Bibo-Joshi: ah, okay. then a bit rephrasing would be good, yes :D

tests/test_official.py
harshil21: a quick ctrl+f tells me it's possible to do that as well. I think I'll add an `or` here just to be more careful

tests/README.rst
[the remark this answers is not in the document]
harshil21: yep, `_unionizer` depends on it. I don't think there's a easy way to use `Union` itself is there? Also sorry looks like that feature is 3.10+ https://docs.python.org/3/library/stdtypes.html#union-type

telegram/_passport/passportfile.py
harshil21, line 116:
```suggestion
        data["file_date"] = from_timestamp(data.get("file_date"), tzinfo=loc_tzinfo)
```
(note to self: d on't forget this glaring mistake + add a test for this + keep all this commented until next major version)

tests/test_official.py
```diff
@@ -264,9 +287,222 @@ def check_defaults_type(ptb_param: inspect.Parameter) -> bool:
...
+        r"InputFile(?: or String)?": {FileInput},
```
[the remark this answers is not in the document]
harshil21: The docs for that say "The path of the file (as string or [pathlib.Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path) object)", so string part of `FileInput` is correct?

tests/_passport/test_passportelementerrorfiles.py
```diff
@@ -58,12 +59,17 @@ def test_to_dict(self, passport_element_error_files):
         assert isinstance(passport_element_error_files_dict, dict)
         assert passport_element_error_files_dict["source"] == passport_element_error_files.source
         assert passport_element_error_files_dict["type"] == passport_element_error_files.type
-        assert (
-            passport_element_error_files_dict["file_hashes"]
-            == passport_element_error_files.file_hashes
-        )
```
harshil21, line 64: making it this a property means making `file_hashes` private and thus it is removed from `to_dict`. Should I override to_dict to correct this then?

tests/README.rst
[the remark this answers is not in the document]
Bibo-Joshi: Okay, then let's just state 3.10+ here :)

tests/test_official.py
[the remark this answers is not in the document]
Bibo-Joshi: Sunds like you have tried a lot already. I'm good with just having a comment, then :)

tests/test_official.py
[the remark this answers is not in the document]
Bibo-Joshi: Ah, yes, you're right :)

telegram/_passport/passportelementerrors.py
```diff
@@ -167,35 +168,54 @@ class PassportElementErrorFiles(PassportElementError):
...
+        data = super().to_dict(recursive)
+        data["file_hashes"] = self._file_hashes
+        return data
+
+    @property
+    def file_hashes(self) -> List[str]:
+        """List of base64-encoded file hashes.
+
+        .. deprecated:: NEXT.VERSION
+            This attribute will return a tuple instead of a list in v22.
```
Bibo-Joshi, line 210:
```suggestion
            This attribute will return a tuple instead of a list in future versions.
```

let's try to hardcode these as seldom as possible :D if we want to be more precise, we can link to the stability policy.

same below.

telegram/_passport/passportelementerrors.py
```diff
@@ -363,36 +383,57 @@ class PassportElementErrorTranslationFiles(PassportElementError):
...
+        """
+        warn(
+            "The attribute `file_hashes` will return a tuple instead of a list in v22. See the "
```
Bibo-Joshi, line 429: .

telegram/_passport/passportfile.py
```diff
@@ -45,7 +45,10 @@ class PassportFile(TelegramObject):
             is supposed to be the same over time and for different bots.
             Can't be used to download or reuse the file.
         file_size (:obj:`int`): File size in bytes.
-        file_date (:class:`datetime.datetime`): Datetime when the file was uploaded.
+        file_date (:obj:`int`): Unix time when the file was uploaded.
+
+            .. deprecated:: NEXT.VERSION
+                This argument will only accept a datetime instead of an integer in v22.
```
Bibo-Joshi, line 51: .

telegram/_passport/passportfile.py
```diff
@@ -90,6 +92,26 @@ def __init__(
...
+
+    @property
+    def file_date(self) -> int:
+        """:obj:`int`: Unix time when the file was uploaded.
+
+        .. deprecated:: NEXT.VERSION
+            This attribute will return a datetime instead of a integer in v22.
+        """
+        warn(
+            "The attribute `file_date` will return a datetime instead of an integer in v22.",
```
Bibo-Joshi, line 109: .

tests/conftest.py
```diff
@@ -50,6 +50,13 @@
     import pytz
 
 
+# Don't collect `test_official.py` on Python 3.10- since it uses newer features like X | Y syntax.
+# Docs: https://docs.pytest.org/en/7.1.x/example/pythoncollection.html#customizing-test-collection
+collect_ignore = []
+if sys.version_info < (3, 10):
+    collect_ignore.append("test_official.py")
```
Bibo-Joshi, line 57: Nice addition 👍 Is it possible (without too much effort) to issue a warning in case someone has set the TEST_OFFICIAL flag, but is running on 3.10-?

Bibo-Joshi (commented):
Thanks for the updates :)

tests/conftest.py
harshil21: should be possible

telegram/_passport/passportelementerrors.py
harshil21: but we will remove it in v22 right? Also why is the stability policy not in linked in the README.md?

Bibo-Joshi: Yes, that is the plan, I would just avoid hard-coding as much as possible. Just in case something unforseen happens :)
linking in the readme is a good idea, can go into the doc-fixes branch 👍

harshil21: I will go with "future major versions"

tests/conftest.py
harshil21: ok nvm, pytest doesn't actually show any output unless you use the `-s` flag. Tried using `warnings.warn`, print() and `logging.info`

Bibo-Joshi: shame. But okay, doesn't matter much :) Thanks for checking this out!

Bibo-Joshi (approved):
Thanks for the new updates :)  I don't think I have any more comments left - great job with Harshil, I continue to be amazed by your patience with test optimization 👏 

Does anyone else want to review?

Poolitzer (approved):
I didnt catch an error, great job with the official extension!

telegram/_bot.py
```diff
@@ -3412,7 +3412,7 @@ async def edit_message_reply_markup(
         chat_id: Optional[Union[str, int]] = None,
         message_id: Optional[int] = None,
         inline_message_id: Optional[str] = None,
-        reply_markup: Optional["InlineKeyboardMarkup"] = None,
+        reply_markup: Optional[InlineKeyboardMarkup] = None,
```
lemontree210, line 3415: It's interesting that in some places you are converting normal references into ForwardRefs, and here the direction is opposite.

tests/test_official.py
```diff
@@ -264,11 +321,196 @@ def check_defaults_type(ptb_param: inspect.Parameter) -> bool:
...
+    ):
+        print("additional_types ")
+        mapped_type = mapped_type | ADDITIONAL_TYPES[ptb_param.name]
+
+    for (param_name, expected_class), exception_type in EXCEPTIONS.items():
+        if ptb_param.name == param_name and is_class is expected_class:
+            ptb_annotation = exception_type
+
+    # Special case for datetimes
+    if re.search(r"([_]+|\b)date[^\w]*\b", ptb_param.name) or "Unix time" in tg_parameter[-1]:
```
lemontree210, line 471: I remember that @Bibo-Joshi usually asks for a `re.VERBOSE` in similar cases 😉

lemontree210 (commented):
I am NOT picking on function docstrings containing a third-person verb instead of imperative 😆 

Very impressive indeed. I think that there could be some intersections in functionality with our admonition inserting tools, but that's just a thought.

telegram/_bot.py
harshil21: was a consequence of an old approach, now they can just continue to be ForwardRefs

lemontree210 (approved):
Thank you!

Bibo-Joshi:
Deepsource failure is unrelated. Merging once other tests pass

Bibo-Joshi:
Awesome addition! thanks a lot :)

PR #3882, opened 2023-09-09
Use `black`'s `--preview` flag once as a test

harshil21:
I noticed multiple times that `black` doesn't always automatically fix issues where the code has exceeded the line length. This seemed to happen in strings like:

``` python
warn(
"blah blah this string will not be formatted no matter how long it gets",
...
)
```

Which made me find the fix for it: https://github.com/psf/black/issues/1802#issuecomment-1580967937

~So this PR enables that and it also fixes a few more things, neat!~


Bibo-Joshi:
Hey, interesting findings! I agree that the future features under the `--preview` flag continue to improve the black style. Black not splitting long string automatically surely is a bit tedious right now, though not really a deal breaker IMO :D

The description of [the preview feater](https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html#preview-style) states

> *Experimental*, potentially disruptive style changes are gathered under the `--preview` CLI flag. At the end of each year, these changes *may* be adopted into the default style, as described in [The Black Code Style](https://black.readthedocs.io/en/stable/the_black_code_style/index.html).

This sounds to me like functionality behind the `--preview` flag is more likely to be buggy, may change more frequently and may in the end not make it into the stable style. TBH the benefit of the gradually nicer style tweaks is worth this instability risk :/ 
A one-time change is ofc fine with me as long as the current black version accepts these changes :)

Bibo-Joshi:
Sorry harshil but I think there is a misunderstanding here. My point is that I don't like the changes in pre-commit-config and pyproject.toml very much …

PR #3883, opened 2023-09-10
Rearrange magic methods to the top

harshil21:
Simply rearranges the magic methods to be at the top of a class since they define class behaviour.

Ref: https://github.com/python-telegram-bot/python-telegram-bot/pull/3826#issuecomment-1666593803

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -308,6 +308,54 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
 
         self._freeze()
 
+    async def __aenter__(self: BT) -> BT:
+        try:
+            await self.initialize()
+            return self
+        except Exception as exc:
+            await self.shutdown()
+            raise exc
+
+    async def __aexit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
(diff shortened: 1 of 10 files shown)


Poolitzer:
I mean... If you rearrange them anyway... maybe we want them in alphabetical order? :D

harshil21:
Yeah I thought of it but it's too much maintainence work for little benefit imo

telegram/_bot.py
```diff
@@ -308,6 +308,54 @@ def __init__(
 
         self._freeze()
 
+    async def __aenter__(self: BT) -> BT:
+        try:
```
Bibo-Joshi, line 312: Having just written https://github.com/python-telegram-bot/python-telegram-bot/pull/3826#issuecomment-1714435152 I see that these methods do not have a docstring - although they are implicitly documented in the class docstring. If we decide that they should get a docstring, we can do that in a separate PR, just wanted your opinion on it :)

Bibo-Joshi (approved):
LGTM - just one related question and conflict resolving is needed :)

telegram/_bot.py
harshil21: yeah I think they should get a separate docstring, makes it more consistent.

Bibo-Joshi: https://github.com/python-telegram-bot/python-telegram-bot/issues/3886

PR #3906, opened 2023-09-29
try fixing failing tests

Bibo-Joshi:
tests are currently failing on master …

### DIFF
--- a/tests/_files/test_photo.py
+++ b/tests/_files/test_photo.py
@@ -99,7 +99,7 @@ class TestPhotoWithoutRequest(TestPhotoBase):
         assert photo.file_size in self.file_size
         assert thumb.width == 90
         assert thumb.height == 90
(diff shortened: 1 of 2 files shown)


tests/_files/test_photo.py
```diff
@@ -99,7 +99,7 @@ def test_expected_values(self, photo, thumb):
         assert photo.file_size in self.file_size
         assert thumb.width == 90
         assert thumb.height == 90
-        assert thumb.file_size == 1477
+        assert thumb.file_size == 1475
```
harshil21, line 102:
```suggestion
        assert thumb.file_size == 1477
```
i didn't see any test failing because of this?

Bibo-Joshi: The scheduled run on master failed. But apparently the value is not fixed. Will try with a list of options like we do in other places

PR #3907, opened 2023-09-29
Add docs for ctx managers

### LINKED ISSUE #3886: [Documentation]: Context manager dunder Methods
The dunder methods `__aenter__` and `__aexit__` that are implemented in several classes in `tg.(ext)` (Updater, Application, Bot, BaseRequest, BaseUpdateProcessor) should get docstrings that explain how these methods behave.

* Since the behavior is very similar in most cases, usage of sphinx insertions may be beneficial
* The notes that we currently have in the class docstring regarding context manager behavior can possibly be (party) replaced by references to these methods
* Sphinx needs to be configured to show these docstrings

clot27:
closes #3886 
I wonder if we should add docstring for args in `__aexit__`?


telegram/_bot.py
```diff
@@ -323,6 +335,12 @@ async def __aexit__(
         exc_val: Optional[BaseException],
         exc_tb: Optional[TracebackType],
     ) -> None:
+        """
+        |async_context_manager| shuts down the Bot.
```
harshil21, line 339: same here with the `:ref:` and in other classes

harshil21 (commented):
docstring nitpicks mostly:

docs/source/conf.py
```diff
@@ -77,6 +77,13 @@
 # and we document the types anyway
 autodoc_typehints = "none"
 
+# Show docstring for special members
+autodoc_special_members = ["__aenter__", "__aexit__"]
+
+autodoc_default_options = {
+    "special-members": True,
+}
+
```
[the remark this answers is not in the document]
clot27: This fails the docs building, I have tested by removing the list but still it produces some warnings, I came with other solution,,

docs/source/conf.py
```diff
@@ -77,6 +77,14 @@
...
 autodoc_typehints = "none"
 
+# Show docstring for special members
+autodoc_special_members = ["__aenter__", "__aexit__"]
+
+autodoc_default_options = {
+    "special-members": True,
+    "exclude-members": "__init__",
+}
+
```
Bibo-Joshi, line 87: Nice - does that mean that we can remove lines like 

https://github.com/python-telegram-bot/python-telegram-bot/blob/1ef242a17e9bcb7a9229664795da9677c2eeb72f/docs/source/telegram.telegramobject.rst?plain=1#L7

?

I don't guide unnderstand the first two lines though. That option doesn't seem to exist (I only found mentions in https://github.com/sphinx-doc/sphinx/issues/4075 and https://github.com/sphinx-doc/sphinx/pull/4076, but the latter was closed in favor of https://github.com/sphinx-doc/sphinx/pull/5215) and lines 84 already seems to care of it?

telegram/_bot.py
```diff
@@ -310,6 +310,15 @@ def __init__(
...
 
     async def __aenter__(self: BT) -> BT:
+        """
+        |async_context_manager| :meth:`initializes <initialize>` the Bot.
+
+        Returns:
+            The initialized Bot instance.
+
+        Raises:
+            :exc:`Exception`: If an exception is raised during initialization.
```
Bibo-Joshi, line 320: IMO it would be good to mention that `shutdown` is called in this case. similar for the other places :)

Bibo-Joshi (commented):
Thanks for your PR! In addition to my below comments:

* I see no added value in documenting the arguments of `__aexit__`, as they are defined by Python itself. If you like to, you can add something like `Please refer to the [Python documentation](https://docs.python.org/3.11/reference/datamodel.html?highlight=__aexit__#object.__aexit__) for information about the accepted arguments`
* We have these parts "Instances of this class can be used as asyncio context managers, where …" in the class docstrings of the affected classes. Could you add a "seealso aenter & exit" below these?

docs/source/conf.py
Bibo-Joshi: I tested locally and I obsevered:

* `autodoc_special_members = ["__aenter__", "__aexit__"]` has no effect - after removing that line, the methods are still listed in the documentation
* `"special-members": True` does indeed include all dunder methods. After reomving the line indicated above, the methods still show in the docs. Since most of our dunder methods already have docstrings, this does sound like a good thing to me.

I double checked and the only public classes with dundermethods without docstrings are Bot (`__eq__` & `__hash__`) and the classes in the `error` module. I would therefore propose:

1. remove lines 80 & 81 and then merge this PR
2. in a follow-up PR (or just in doc-fixes branch):
    1. add docstrings where missing on dunder methods
    2. Remove the `:special-members:` setting from the `*.rst` files

Would that be okay with you? cc @harshil21

harshil21: yep, sounds good. I also observed that `autodoc_special_members` variable has no effect.

clot27: great! I would be going with the follow up pr also

docs/source/conf.py
```diff
@@ -77,8 +77,15 @@
...
 
+# Show docstring for special members
+autodoc_default_options = {
+    "special-members": True,
+    "exclude-members": "__init__",
+}
+
 # Fail on warnings & unresolved references etc
 nitpicky = True
+nitpick_ignore = [("py:class", "type")]
```
harshil21, line 88: ! why? I removed it locally and didn't see anything fail related to this

telegram/_bot.py
```diff
@@ -310,6 +312,16 @@ def __init__(
...
     async def __aenter__(self: BT) -> BT:
+        """
+        |async_context_manager| :meth:`initializes <initialize>` the Bot.
+
+        Returns:
+            The initialized Bot instance.
+
+        Raises:
+            :exc:`Exception`: If an exception is raised during initialization, :meth:`shutdown`
+            is called in this case.
```
harshil21, line 323:
```suggestion
                is called in this case.
```
need to indent the next line to fix doc build. Apply this for all the `__aenter__`'s

docs/source/conf.py
clot27: yea, I was just trying to fix that build warning..

Bibo-Joshi (approved):
Thanks for the updates! :)

docs/source/conf.py
Bibo-Joshi: Openend #3926 for this :) @clot27 would you like to address that yourself?

clot27: yes sure

PR #3929, opened 2023-10-13
Add dunder method docstring

### LINKED ISSUE #3926: [Documentation] Dunder methods
Follow-up of #3907.

It should be double checked where we have dunder/magic methods of public classes that do not have a proper docstring yet. On first glance, I saw `Bot.__{eq, hash}__` as well as some methods in the `__error__` module.
These methods should get descriptive docstrings (even if it's just pointing to the base class) with argument descriptions where applicable.

Moreover, the `:special-members:` directives can be removed everywhere in the `.rst` files.

clot27:
Closes #3926


telegram/ext/filters.py
```diff
@@ -183,18 +183,39 @@ def __init__(self, name: Optional[str] = None, data_filter: bool = False):
...
     def __invert__(self) -> "BaseFilter":
+        """Defines `NOT` bitwise operator for :class:`BaseFilter` object.
+
+        Returns:
+           :obj:`BaseFilter`
+        """
         return _InvertedFilter(self)
 
     def __repr__(self) -> str:
+        """:obj:`str`: Name for this filter."""
```
Bibo-Joshi, line 218: * Please add a `Returns` directive
* Please point to the `name` property

telegram/ext/_jobqueue.py
```diff
@@ -779,6 +779,15 @@ def __init__(
         self._job = cast("APSJob", None)  # skipcq: PTC-W0052
 
     def __getattr__(self, item: str) -> object:
+        """Get the specific attribute of :class:`telegram.ext.Job` object.
```
Bibo-Joshi, line 782: This is not quite try - first we check if this object itself has such an attribute, as you have written in the `Raises` clause. Maybe you can also formulate this in the style of https://docs.python-telegram-bot.org/en/stable/telegram.telegramobject.html#telegram.TelegramObject.__getstate__ ?

telegram/error.py
```diff
@@ -78,12 +78,28 @@ def __init__(self, message: str):
         self.message: str = msg
 
     def __str__(self) -> str:
+        """Get the string representation of exception message.
```
Bibo-Joshi, line 81:
```suggestion
        """Gives the string representation of exceptions message.
```

telegram/error.py
```diff
@@ -78,12 +78,28 @@ def __init__(self, message: str):
...
+
+        Returns:
+           :obj:`str`
+        """
         return self.message
 
     def __repr__(self) -> str:
+        """Get the unambiguous string representation of exception.
```
Bibo-Joshi, line 89:
```suggestion
        """Gives an unambiguous string representation of the exception.
```

telegram/error.py
```diff
@@ -78,12 +78,28 @@ def __init__(self, message: str):
...
+
+        Returns:
+           :obj:`str`
+        """
         return f"{self.__class__.__name__}('{self.message}')"
 
     def __reduce__(self) -> Tuple[type, Tuple[str]]:
+        """Defines how to serialize the exception for pickle. See
```
Bibo-Joshi, line 97: Can you link to the pickle module as well? :)

telegram/_bot.py
```diff
@@ -363,11 +363,24 @@ def __deepcopy__(self, memodict: Dict[int, object]) -> NoReturn:
         raise TypeError("Bot objects cannot be deepcopied!")
 
     def __eq__(self, other: object) -> bool:
+        """Defines equality condition for the :class:`telegram.Bot` object.
+        Two objects of this class are considered to be equal if their all parameters
+        are identical.
+
+        Returns:
+            :obj:`True` if both objects have all parameters identical. :obj:`False` otherwise.
```
Bibo-Joshi, line 371:
```suggestion
        Two objects of this class are considered to be equal if their  attributes :attr:`bot`
        are equal.

        Returns:
            :obj:`True` if both attributes :attr:`bot` are equal. :obj:`False` otherwise.
```

telegram/_bot.py
```diff
@@ -363,11 +363,24 @@ def __deepcopy__(self, memodict: Dict[int, object]) -> NoReturn:
...
         if isinstance(other, self.__class__):
             return self.bot == other.bot
         return False
 
     def __hash__(self) -> int:
+        """Builds a hash value for this object such that the hash of two objects is
+        equal if and only if the objects are equal in terms of :meth:`__eq__`.
+
+        Returns:
+            :obj:`int` The hash value of the object.
```
Bibo-Joshi, line 382: IMO pointing to `TelegramObject.__hash__` is enough here, similar as done e.g. for https://docs.python-telegram-bot.org/en/stable/telegram.message.html#telegram.Message.de_json

Bibo-Joshi (changes requested):
Thanks for the nice PR! I left some comments below :)

One (unrelated) thing that I noticed is that `Bot.__eq__` doesn't comply with the documentation of [`TelegramObject.__eq__`](https://docs.python-telegram-bot.org/en/stable/telegram.telegramobject.html#telegram.TelegramObject.__eq__), i.e. it doesn't fall back to `super().__eq__`. Since so for `Bot.__eq__` is not explicitly documented, it might be wise to adjust this before adding documentation would make this a (very minor) breaking change. What do you think?
cc @Poolitzer @harshil21

telegram/ext/filters.py
```diff
@@ -183,18 +183,57 @@ def __init__(self, name: Optional[str] = None, data_filter: bool = False):
         self._data_filter = data_filter
 
     def __and__(self, other: "BaseFilter") -> "BaseFilter":
+        """Defines `AND` bitwise operator for :class:`BaseFilter` object.
+        The combined filter accepts an update only if it is accepted by both filters.
+        For exampe, filters.PHOTO & filters.CAPTION will only accept messages that contain
```
Bibo-Joshi, line 188:
```suggestion
        For exampe, ``filters.PHOTO & filters.CAPTION`` will only accept messages that contain
```

telegram/ext/_jobqueue.py
```diff
@@ -779,6 +779,19 @@ def __init__(
         self._job = cast("APSJob", None)  # skipcq: PTC-W0052
 
     def __getattr__(self, item: str) -> object:
+        """Overrides :py:meth:`object.__getattr__` to get specific attribute of
+        object :class:`telegram.ext.Job` object if exists.
```
Bibo-Joshi, line 783:
```suggestion
        the :class:`telegram.ext.Job` object or of its attribute :class:`apscheduler.job.Job`, if exists.
```

Sorry, my previous comment wasn't well formulated …

Bibo-Joshi (commented):
Thanks for the updates! Just two smaller comments below :)  What dou you think about https://github.com/python-telegram-bot/python-telegram-bot/pull/3929#pullrequestreview-1704967873 ?

telegram/ext/_defaults.py
```diff
@@ -119,6 +125,13 @@ def __hash__(self) -> int:
         )
 
     def __eq__(self, other: object) -> bool:
+        """Defines equality condition for the :class:`Defaults` object.
+        Two objects of this class are considered to be equal if their all parameters
```
harshil21, line 129:
```suggestion
        Two objects of this class are considered to be equal if all their parameters
```

telegram/ext/_jobqueue.py
```diff
@@ -787,11 +800,25 @@ def __getattr__(self, item: str) -> object:
             ) from exc
 
     def __eq__(self, other: object) -> bool:
+        """Defines equality condition for the :class:`telegram.ext.Job` object.
+        Two objects of this class are considered to be equal if their :paramref:`id`
```
harshil21, line 804:
```suggestion
        Two objects of this class are considered to be equal if their :class:`id <apscheduler.job.Job>`
```

telegram/ext/_jobqueue.py
```diff
@@ -787,11 +800,25 @@ def __getattr__(self, item: str) -> object:
...
+        are equal.
+
+        Returns:
+            :obj:`True` if both objects have :paramref:`id` parameters identical.
+                :obj:`False` otherwise.
```
harshil21, line 809:
```suggestion
            :obj:`True` if both objects have :paramref:`id` parameters identical.
            :obj:`False` otherwise.
```
rendering issue

telegram/ext/filters.py
```diff
@@ -183,18 +183,57 @@ def __init__(self, name: Optional[str] = None, data_filter: bool = False):
...
+        both a photo and a caption
```
harshil21, line 189:
```suggestion
        both a photo and a caption.
```

telegram/ext/filters.py
```diff
@@ -183,18 +183,57 @@ def __init__(self, name: Optional[str] = None, data_filter: bool = False):
...
     def __repr__(self) -> str:
+        """Gives name for this filter.
+
+        .. seealso::
+               :meth:`name`
+
+        Returns:
+            :obj:`str`:
+        """
         return self.name
```
harshil21, line 237: `BaseFilter.__repr__` feels pretty useless. It should be updated in line with #3826 isn't it?

harshil21 (changes requested):
> Since so for Bot.__eq__ is not explicitly documented, it might be wise to adjust this before adding documentation would make this a (very minor) breaking change. What do you think?

I agree, should be updated, though we will need to add an `_id_attrs` for `Bot` as well,  what should that include? 


Also `telegram.ext.updater.rst` still has the `:special-members:` in it, should be deleted.

Also I believe that a side effect of adding the special-members derivative in the `conf.py` file was that now we have all these extra documented methods (see `BaseFilter` docs):
image

telegram/ext/filters.py
Bibo-Joshi: Why is it useless? We have defined well readible names for all filters and some of these are in fact rather elaborate: https://github.com/python-telegram-bot/python-telegram-bot/blob/a0c81ec3d46fe1889ab3a9aca40b56f33784eabb/tests/ext/test_filters.py#L2037

Bibo-Joshi:
> > Since so for Bot.**eq** is not explicitly documented, it might be wise to adjust this before adding documentation would make this a (very minor) breaking change. What do you think?
> 
> I agree, should be updated, though we will need to add an `_id_attrs` for `Bot` as well, what should that include?

`_id_attrs` unfortunately don't work for the `Bot` class, as we're comparing with the `bot` propery. However, that's only cached once `Bot.initialize` is called, so not on `__init__`.  Replacing `return False` with `return super().__eq__(other)` (and adjusting tests) would be enough IMO.

> Also `telegram.ext.updater.rst` still has the `:special-members:` in it, should be deleted.

good catch 👍 

> Also I believe that a side effect of adding the special-members derivative in the `conf.py` file was that now we have all these extra documented methods (see `BaseFilter` docs): image

adding `object` to `:inherited-members` in `filters.rst` should exlude these :)

telegram/ext/filters.py
harshil21: hm yeah I just forgot about that, I probably shouldn't review when I'm sleepy :P

Bibo-Joshi:
> > Also I believe that a side effect of adding the special-members derivative in the `conf.py` file was that now we have all these extra documented methods (see `BaseFilter` docs):
> 
> adding `object` to `:inherited-members` in `filters.rst` should exlude these :)

I see that there is also `__weakref__` in `Bot` :/ This is getting messier than I thought and it's also somewhat hard to catch … at least with 10min trial & error I didn't find a very good solution yet …

telegram/_bot.py
```diff
@@ -356,11 +356,19 @@ def __deepcopy__(self, memodict: Dict[int, object]) -> NoReturn:
         raise TypeError("Bot objects cannot be deepcopied!")
 
     def __eq__(self, other: object) -> bool:
+        """Defines equality condition for the :class:`telegram.Bot` object.
+        Two objects of this class are considered to be equal if their attributes
+        :attr:`bot` are equal.
+
+        Returns:
+            :obj:`True` if both attributes :attr:`bot` are equal. :obj:`False` otherwise.
+        """
```
Bibo-Joshi, line 365: https://github.com/python-telegram-bot/python-telegram-bot/pull/3929#pullrequestreview-1704967873

Would you like to address this in this pr? if not, please open an issue and put it on the next milestone so that we don't forget about it

clot27: @Bibo-Joshi https://github.com/python-telegram-bot/python-telegram-bot/pull/3929#issuecomment-1793701253 seems breaking to me, it equates Bot to ExtBot (via `==`) as it checks the .bot argument and both objects have same .bot arguments, but hashes are different...?

Bibo-Joshi: Good obsveration! I see that `__class__` is also part of `TelegramObject.__hash__`, but I guess there are no relevant cases for other classes where one would compage parent class objects with child class objects.
In the case of `Bot`, one could replace `self.__class__` with `Bot` in both `__eq__` and `__hash__` - that way `Bot` and `ExtBot` would be considered "the same", which I would see as expected behavior. Looking at `__hash__`, I also note that it would be cleaner to also relay to `super()` in case `self._bot_user is None`
I still see changes to this as non-breaking as we haven't specified the behavior on this level so far.

tests/test_bot.py
```diff
@@ -316,22 +316,25 @@ async def shutdown():
...
 
             assert a != d
             assert hash(a) != hash(d)
 
             assert a != e
             assert hash(a) != hash(e)
 
+            a._bot_user = None
+            assert hash(a) != hash(b)
+
```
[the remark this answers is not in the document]
clot27: We cant to do the equality check if the object isnt initialized..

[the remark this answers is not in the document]
Bibo-Joshi: hah, good point :D

PR #3935, opened 2023-10-19
Add Parameter `socket_options` to `HTTPXRequest`

### LINKED ISSUE #2965: [FEATURE] Set SocketOpts
To improve the networking backend, we would like to set socketopts if possible. Before the `asyncio` switch, we did something like

https://github.com/python-telegram-bot/python-telegram-bot/blob/ad5a1fb1f6a01f5f662e9e610ddb77fb57558e44/telegram/request/_httpxrequest.py#L164-L168

It would be nice if we could do that for httpx as well at some point.

Bibo-Joshi:
Closes #2965

TBH I'm not quite sure if I got the parameter description right, as I've never really worked with socket options. @tsnoam if you could kindly have a look at the descrption, that would be a great help :)

- [ ] Once merged, the dev wiki section on https://github.com/python-telegram-bot/python-telegram-bot/wiki/Handling-network-errors#socket-options should be updated

### DIFF
--- a/telegram/request/_httpxrequest.py
+++ b/telegram/request/_httpxrequest.py
@@ -147,15 +147,13 @@ class HTTPXRequest(BaseRequest):
             if socket_options
             else None
         )
-        # See https://github.com/python-telegram-bot/python-telegram-bot/pull/3542
-        # for why we need to use `dict()` here.
-        self._client_kwargs = dict(  # pylint: disable=use-dict-literal
-            timeout=timeout,
-            proxies=proxy_url,
-            limits=limits,
-            transport=transport,
+        self._client_kwargs = {
+            "timeout": timeout,
+            "proxies": proxy_url,
+            "limits": limits,
+            "transport": transport,
             **http_kwargs,
-        )
+        }
 
         try:
             self._client = self._build_client()
--- a/telegram/request/_httpxrequest.py
+++ b/telegram/request/_httpxrequest.py
@@ -17,7 +17,7 @@
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains methods to make POST and GET requests using the httpx library."""
-from typing import Optional, Tuple
+from typing import Collection, Optional, Tuple, Union
 
 import httpx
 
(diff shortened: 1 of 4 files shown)


telegram/request/_httpxrequest.py
```diff
@@ -122,15 +138,22 @@ def __init__(
...
+            httpx.AsyncHTTPTransport(
+                socket_options=socket_options,
+            )
+            if socket_options
+            else None
+        )
         # See https://github.com/python-telegram-bot/python-telegram-bot/pull/3542
         # for why we need to use `dict()` here.
-        self._client_kwargs = dict(  # pylint: disable=use-dict-literal  # noqa: C408
+        self._client_kwargs = dict(  # pylint: disable=use-dict-literal
```
tsnoam, line 151: tbh, this `disable=use-dict-literal` makes no sense. probably was a different problem.

not that it really matters. just an academic discussion.

telegram/request/_httpxrequest.py
```diff
@@ -91,6 +97,15 @@ class HTTPXRequest(BaseRequest):
 
             .. versionchanged:: 20.5
                 Accept ``"2"`` as a valid value.
+        socket_options (Collection[Tuple], optional): A collection of socket options to set on
+            the underlying socket. For each entry, the first two elements must be integers
+            representing the socket level and the socket option, respectively. The third element
+            must be either an integer or a bytes-like object representing the value to set.
+            Alternatively, the entry can be a 4-tuple, where the third element is :obj:`None` and
+            the fourth element is an integer representing the timeout to set.
+            Defaults to :obj:`None`.
```
tsnoam, line 106: Where was this text taken from? It sounds too confusing.

I would just go "This object is passed as is to XXX and the implementation details are OS specific. This is an advanced section, make sure you know what you're doing"

or you can say something like "see [httpx.SOCKET_OPTIONS] for more details"

tsnoam (approved):
lgtm.

still, check the comments. you might want to address some of them.

telegram/request/_httpxrequest.py
Bibo-Joshi: TBH I had quite a bit of help from GitHub Copilot on this :D But redirecting to httpx docs might be a saner option 👍

harshil21 (approved):

PR #3939, opened 2023-10-21
Rename Parameter `proxy_url` to `proxy` and Support `httpx.Proxy` as Input

### LINKED ISSUE #3844: [FEATURE] SSL verification + custom proxies

There is no way to disable ssl verification for proxies and pass a custom httpx.Proxy, instead of just a string.

Some proxies work as a man-in-the-middle and may lead to errors, like: ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1006). 

Due to this reason, I think it's necessary to allow control over ssl verification in HTTPXRequest. Just add an argument like `ssl_verify: bool = True` and add `verify=ssl_verify` to `self._client_kwargs`.

Another thing I'd love to see, is to change the type of:
`proxy_url: Optional[str] = None,`
to
`proxy_url: Optional[Union[str, httpx.Proxy] = None,`

This is due to some proxies requiring authorization through headers, which could be passed as:
```
httpx.Proxy(
    url=proxy_url,
    headers={
        "x-secret-header": "secret value"
    },
)
```
(moreover it might be used for passing logging headers to the proxy and so on)

_No response_

### Additional context

_No response_

Bibo-Joshi:
Closes #3844

### DIFF
--- a/telegram/ext/_applicationbuilder.py
+++ b/telegram/ext/_applicationbuilder.py
@@ -32,9 +32,12 @@ from typing import (
     Union,
 )
 
+import httpx
+
 from telegram._bot import Bot
 from telegram._utils.defaultvalue import DEFAULT_FALSE, DEFAULT_NONE, DefaultValue
 from telegram._utils.types import DVInput, DVType, FilePathInput, HTTPVersion, ODVInput
+from telegram._utils.warnings import warn
 from telegram.ext._application import Application
 from telegram.ext._baseupdateprocessor import BaseUpdateProcessor, SimpleUpdateProcessor
 from telegram.ext._contexttypes import ContextTypes
@@ -44,6 +47,7 @@ from telegram.ext._updater import Updater
 from telegram.ext._utils.types import BD, BT, CCT, CD, JQ, UD
 from telegram.request import BaseRequest
 from telegram.request._httpxrequest import HTTPXRequest
+from telegram.warnings import PTBDeprecationWarning
 
 if TYPE_CHECKING:
     from telegram.ext import BasePersistence, BaseRateLimiter, CallbackContext, Defaults
@@ -66,14 +70,14 @@ _BOT_CHECKS = [
     ("request", "request instance"),
     ("get_updates_request", "get_updates_request instance"),
     ("connection_pool_size", "connection_pool_size"),
-    ("proxy_url", "proxy_url"),
+    ("proxy", "proxy"),
     ("pool_timeout", "pool_timeout"),
     ("connect_timeout", "connect_timeout"),
     ("read_timeout", "read_timeout"),
     ("write_timeout", "write_timeout"),
     ("http_version", "http_version"),
     ("get_updates_connection_pool_size", "get_updates_connection_pool_size"),
-    ("get_updates_proxy_url", "get_updates_proxy_url"),
+    ("get_updates_proxy", "get_updates_proxy"),
     ("get_updates_pool_timeout", "get_updates_pool_timeout"),
     ("get_updates_connect_timeout", "get_updates_connect_timeout"),
     ("get_updates_read_timeout", "get_updates_read_timeout"),
@@ -136,7 +140,7 @@ class ApplicationBuilder(Generic[BT, CCT, UD, CD, BD, JQ]):
         "_get_updates_connect_timeout",
         "_get_updates_connection_pool_size",
         "_get_updates_pool_timeout",
-        "_get_updates_proxy_url",
+        "_get_updates_proxy",
         "_get_updates_read_timeout",
         "_get_updates_request",
         "_get_updates_write_timeout",
@@ -149,7 +153,7 @@ class ApplicationBuilder(Generic[BT, CCT, UD, CD, BD, JQ]):
         "_post_stop",
         "_private_key",
         "_private_key_password",
-        "_proxy_url",
+        "_proxy",
         "_rate_limiter",
         "_read_timeout",
         "_request",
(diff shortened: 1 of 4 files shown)


telegram/request/_httpxrequest.py
```diff
@@ -91,6 +86,20 @@ class HTTPXRequest(BaseRequest):
 
             .. versionchanged:: 20.5
                 Accept ``"2"`` as a valid value.
+        proxy (:obj:`str` | ``httpx.Proxy``, optional): The URL to a proxy server or an
```
tsnoam, line 89: Looking again at httpx code, the proxy parameter has a much wider definition:
```python
URLTypes = Union["URL", str]
ProxiesTypes = Union[URLTypes, "Proxy", Dict[URLTypes, Union[None, URLTypes, "Proxy"]]]
```

Since ptb doesn't need a dict, consider at least chaning to 
```suggestion
        proxy (:obj:`str` | ``httpx.Proxy`` | ``httpx.URL``, optional): The URL to a proxy server or an
```

telegram/request/_httpxrequest.py
```diff
@@ -91,6 +86,20 @@ class HTTPXRequest(BaseRequest):
...
+            ``httpx.Proxy`` object. For example ``'http://127.0.0.1:3128'`` or
+            ``'socks5://127.0.0.1:3128'``. Defaults to :obj:`None`.
+
+            Note:
+                * The proxy URL can also be set via the environment variables ``HTTPS_PROXY`` or
+                  ``ALL_PROXY``. See `the docs of httpx`_ for more info.
+                * For Socks5 support, additional dependencies are required. Make sure to install
+                  PTB via :command:`pip install "python-telegram-bot[socks]"` in this case.
```
tsnoam, line 97:
```suggestion
                * For Socks5 support, additional dependencies are required. They can be installed using the
                  command :command:`pip install "python-telegram-bot[socks]"`.
```

tsnoam (approved):
lgtm. see comments if you care to change anything.

telegram/request/_httpxrequest.py
Bibo-Joshi: I can add the URL type, yes. I'll skip the dict types though as discussed in #3844 :)

telegram/request/_httpxrequest.py
```diff
@@ -107,6 +102,20 @@ class HTTPXRequest(BaseRequest):
...
+        proxy (:obj:`str` | ``httpx.Proxy`` | ``httpx.URL``, optional): The URL to a proxy server,
+            a ``httpx.Proxy`` object or a ``httpx.URL`` object. For example
+            ``'http://127.0.0.1:3128'`` or ``'socks5://127.0.0.1:3128'``. Defaults to :obj:`None`.
+
+            Note:
+                * The proxy URL can also be set via the environment variables ``HTTPS_PROXY`` or
+                  ``ALL_PROXY``. See `the docs of httpx`_ for more info.
+                * For Socks5 support, additional dependencies are required. Make sure to install
+                  PTB via :command:`pip install "python-telegram-bot[socks]"` in this case.
+                * Socks5 proxies can not be set via environment variables.
```
harshil21, line 114: I would also add a note here linking to #3844 as an example of advanced use cases, since the httpx documentation is non existant for `httpx.Proxy` and `httpx.URL`.

telegram/request/_httpxrequest.py
```diff
@@ -115,14 +124,27 @@ class HTTPXRequest(BaseRequest):
     def __init__(
         self,
         connection_pool_size: int = 1,
-        proxy_url: Optional[str] = None,
+        proxy_url: Optional[Union[str, httpx.Proxy]] = None,
```
harshil21, line 127:
```suggestion
        proxy_url: Optional[Union[str, httpx.Proxy, httpx.URL]] = None,
```
and URL?

harshil21 (changes requested):
The `ApplicationBuilder.proxy_url` part should be changed as well? Shouldn't the test fail because of the uncaught warning raised?

telegram/request/_httpxrequest.py
Bibo-Joshi: I added a new bullet point with a description. IMO the discussion in #3844 is a tad to long to be a proper reference in the documentation :D

Bibo-Joshi:
> The `ApplicationBuilder.proxy_url` part should be changed as well? Shouldn't the tests fail because of the uncaught warning raised?

Sorry, hadn't seen this. Will change appbilder. PTB warnings are explicitly ignored in `conftest.py` so that we don't *have* to catch them on every test

telegram/ext/_applicationbuilder.py
```diff
@@ -655,21 +683,47 @@ def get_updates_connection_pool_size(
...
+        Args:
+            get_updates_proxy_url (:obj:`str` | ``httpx.Proxy`` | ``httpx.URL``): See
+                :paramref:`telegram.ext.ApplicationBuilder.get_updates_proxy.get_updates_proxy`.
+
+        Returns:
+            :class:`ApplicationBuilder`: The same builder with the updated argument.
+        """
+        warn(
+            "`ApplicationBuilder.get_updates_proxy_url` is deprecated since version "
+            "Next.VERSION. Use `ApplicationBuilder.proxy` instead.",
```
clot27, line 701:
```suggestion
            "Next.VERSION. Use `ApplicationBuilder.proxy` instead.",
```
NEXT.VERSION

harshil21:
```suggestion
            "NEXT.VERSION. Use `ApplicationBuilder.get_updates_proxy` instead.",
```
and this also :)

telegram/ext/_applicationbuilder.py
```diff
@@ -486,21 +490,45 @@ def connection_pool_size(self: BuilderType, connection_pool_size: int) -> Builde
...
+        Args:
+            proxy_url (:obj:`str` | ``httpx.Proxy`` | ``httpx.URL``): See
+                :paramref:`telegram.ext.ApplicationBuilder.proxy.proxy`.
+
+        Returns:
+            :class:`ApplicationBuilder`: The same builder with the updated argument.
+        """
+        warn(
+            "`ApplicationBuilder.proxy_url` is deprecated since version "
+            "Next.VERSION. Use `ApplicationBuilder.proxy` instead.",
```
harshil21, line 508:
```suggestion
            "NEXT.VERSION. Use `ApplicationBuilder.proxy` instead.",
```

harshil21 (changes requested):
Tests look good, just some typos:

harshil21 (approved):
lgtm

PR #3941, opened 2023-10-22
Add `filters.Mention`

### LINKED ISSUE #3799: [FEATURE] Add bot mentioned filter

Bibo-Joshi:
Closes #3799 

Supersedes #3857 

-  [x] Added new classes & modules to the docs and all suitable ``__all__`` s


clot27 (approved):

telegram/ext/filters.py
```diff
@@ -2560,3 +2561,70 @@ def filter(self, message: Message) -> bool:
 
 VOICE = _Voice("filters.VOICE")
 """Messages that contain :attr:`telegram.Message.voice`."""
+
+
+class Mention(MessageFilter):
```
harshil21, line 2566: Need to move this class since the objects in this file are ordered alphabetically

harshil21 (changes requested):
Code looks good!

harshil21 (approved):

PR #3943, opened 2023-10-23
Add `ApplicationBuilder.(get_updates_)socket_options`

Bibo-Joshi:
* Follow up for #3935
* Also adds tests that make sure that `ApplicationBuilder` has all necessary methods

-  ~Added new classes & modules to the docs and all suitable ``__all__`` s~


lemontree210 (commented):
I did look through the code and it looks fine but I don't feel like I'm qualified enough to judge :)

tests/ext/test_applicationbuilder.py
```diff
@@ -379,6 +416,36 @@ class Client:
...
+            "request"
+        ).get_updates_socket_options(((4, 5, 6),)).get_updates_connection_pool_size(
+            "get_updates"
+        ).build()
+
+        for kwargs in httpx_request_kwargs:
+            if kwargs.get("connection_pool_size") == "request":
+                assert kwargs.get("socket_options") == ((1, 2, 3),)
+            else:
+                assert kwargs.get("socket_options") == ((4, 5, 6),)
```
harshil21, line 447:
```suggestion
                assert kwargs.get("get_updates_socket_options") == ((4, 5, 6),)
```
shouldn't it be this?

harshil21 (commented):
code in general looks good to me.

tests/ext/test_applicationbuilder.py
harshil21: oh nvm it is kwargs from HTTPXRequest, that parameter doesn't exist

PR #3952, opened 2023-10-27
Improve `write_timeout` Handling for Media Methods

Bibo-Joshi:
Second part of #3864

Unfortunately I realised that the approach outlined in #3864 is not as backward compatible as I initially thought, but I think I found a rather clean deprecation layer

-  ~Added new classes & modules to the docs and all suitable ``__all__`` s~


Poolitzer (approved):
LGTM

docs/auxil/sphinx_hooks.py
```diff
@@ -107,19 +107,31 @@ def autodoc_process_docstring(
...
-                keyword_args[i - insert_index].format(
+            to_insert = keyword_args[i - insert_index]
+
+            if (
+                "post.write_timeout`. Defaults to" in to_insert
+                and method_name in media_write_timeout_deprecation_methods
+            ):
+                effective_insert = media_write_timeout_deprecation
+            else:
+                effective_insert = (to_insert,)
```
harshil21, line 122:
```suggestion
                effective_insert: list[str] = media_write_timeout_deprecation
            else:
                effective_insert: list[str] = list(to_insert)
```
unified types would be nice and could avoid potential bugs in the future

docs/auxil/kwargs_insertion.py
```diff
@@ -45,7 +45,33 @@
...
+    "create_new_sticker_set",
+]
+media_write_timeout_deprecation = [
+    "    write_timeout (:obj:`float` | :obj:`None`, optional): Value to pass to "
+    "        :paramref:`telegram.request.BaseRequest.post.write_timeout`. Defaults to ``20``."
+    "",
+    "",
+    "       .. deprecated:: NEXT.VERSION",
+    "           In future versions, the default value will be changed to "
+    "           :attr:`~telegram.request.BaseRequest.DEFAULT_NONE`.",
```
harshil21, line 71: technically the parameter was already changed to be `DEFAULT_NONE`  so we can't say "in future versions"

harshil21 (commented):
logic looks good, just few minor comments

docs/auxil/kwargs_insertion.py
Bibo-Joshi: mh, I don't really mean the value behing the "=" in the signature but rather the value that will be passed to `BaseRequest.post`. And that's still 20s after this PR. I tried to specify a bit

harshil21 (approved):

Issue #3397, opened 2022-11-28
[BUG] Application gets stuck while manually stopping in async environment


0x4c47:
### Steps to Reproduce

1. Create an async environment with signal handling and task cancellation to run the Telegram bot in. Working code example:
```python
import asyncio
import logging
import signal
import telegram.ext
import telegram.ext.filters

logging.basicConfig(level=logging.DEBUG)
log = logging.getLogger("custom")
app = telegram.ext.ApplicationBuilder().token("<token here>").build()

async def main():
    loop = asyncio.get_event_loop()
    loop.add_signal_handler(signal.SIGTERM, lambda: asyncio.create_task(shutdown(loop)))

    await app.initialize()
    await app.start()
    await app.updater.start_polling()
    try:
        while True:
            log.info("Now running in main loop...")
            await asyncio.sleep(10000)
    except asyncio.CancelledError:
        log.warning("Telegram wait loop got cancelled, stopping...")
        await app.updater.stop()
        await app.stop()
        await app.shutdown()

async def shutdown(loop):
    log.warning("Received SIGTERM, shutting down...")
    tasks = [task for task in asyncio.all_tasks() if task is not asyncio.current_task()]
    [task.cancel() for task in tasks]
    log.warning("Waiting for async tasks to be cancelled...")
    await asyncio.gather(*tasks)

asyncio.run(main())
```
2. Send a SIGTERM to the process: `kill -s SIGTERM <pid>`

### Expected behaviour

Application stops correctly.

### Actual behaviour

Application gets stuck while stopping and never stops, see logs. After the last log line, nothing happens, no matter how long I'm waiting. I have to manually kill the process with another signal like SIGKILL to stop it (ungracefully).

### Operating System

Linux (tested on Fedora 37, Ubuntu 22.04 and within a Debian docker container)

### Version of Python, python-telegram-bot & dependencies

```shell
python-telegram-bot 20.0a6
Bot API 6.3
Python 3.11.0 (main, Oct 24 2022, 00:00:00) [GCC 12.2.1 20220819 (Red Hat 12.2.1-2)]
```

### Relevant log output

```python
[...]
INFO:custom:Now running in main loop...
DEBUG:telegram.ext._updater:Start network loop retry getting Updates
DEBUG:telegram._bot:Entering: get_updates
WARNING:custom:Received SIGTERM, shutting down...
WARNING:custom:Waiting for async tasks to be cancelled...
[...]

Issue #3893, opened 2023-09-16
[BUG] Exception while shutdown


et-ness:
### Steps to Reproduce

1. run an example
2. stop with CTRL+C
3. randomly the shutdown got an exception

### Expected behaviour

Close the bot without exceptions

### Actual behaviour

These exceptions is very similar to issue #1563 if I run the same code on python3.9 I got the exception described in #1563 while if I run on python3.11 sometime I got these exceptions:

<details>

```shell
^CTraceback (most recent call last):
  File "/usr/lib/python3/dist-packages/anyio/streams/tls.py", line 130, in _call_sslobject_method
    result = func(*args)
             ^^^^^^^^^^^
  File "/usr/lib/python3.11/ssl.py", line 921, in read
    v = self._sslobj.read(len)
        ^^^^^^^^^^^^^^^^^^^^^^
ssl.SSLWantReadError: The operation did not complete (read) (_ssl.c:2546)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/httpcore/backends/asyncio.py", line 34, in read
    return await self._stream.receive(max_bytes=max_bytes)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/anyio/streams/tls.py", line 195, in receive
    data = await self._call_sslobject_method(self._ssl_object.read, max_bytes)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/anyio/streams/tls.py", line 137, in _call_sslobject_method
    data = await self.transport_stream.receive()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/anyio/_backends/_asyncio.py", line 1265, in receive
    await self._protocol.read_event.wait()
  File "/usr/lib/python3.11/asyncio/locks.py", line 213, in wait
    await fut
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/httpcore/_exceptions.py", line 10, in map_exceptions
    yield
[...]

PR #3959, opened 2023-11-03
handle _get_updates_cleanup() timeout error

guillemap:
This prevents the application from hanging at updater.stop() process in some cases where it is stopped using ctrl+c.

The error is found when manually initializing the application to run it alongside other asyncio frameworks, as described [here](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Frequently-requested-design-patterns#running-ptb-alongside-other-asyncio-frameworks). In this case, sometimes when stopping with crtl+c the app hangs at `await application.updater.stop()` and never reaches `await application.stop()`, while showing the warning:

> Fetching updates got a asyncio.CancelledError. Ignoring as this task may onlybe closed via \`Application.stop`.

Potentially fixes https://github.com/python-telegram-bot/python-telegram-bot/issues/3397

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -55,6 +55,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `franciscod <https://github.com/franciscod>`_
 - `gamgi <https://github.com/gamgi>`_
 - `Gauthamram Ravichandran <https://github.com/GauthamramRavichandran>`_
+- `guillemap <https://github.com/guillemap>`_
 - `Harshil <https://github.com/harshil21>`_
 - `Hugo Damer <https://github.com/HakimusGIT>`_
 - `ihoru <https://github.com/ihoru>`_
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -1146,7 +1146,7 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
                 # application.stop() will wait for it's clean shutdown.
                 _LOGGER.warning(
                     "Fetching updates got a asyncio.CancelledError. Ignoring as this task may only"
-                    "be closed via `Application.stop`."
+                    " be closed via `Application.stop`."
                 )
 
     async def __process_update_wrapper(self, update: object) -> None:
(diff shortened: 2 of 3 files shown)


Bibo-Joshi:
Hi. Thanks for your PR. I guess you mean "potentially fixing #3893" rather than #3397?
However, what your PR does is "treat the symptoms", but unfortunately it does not really solve the issue that surfaced in #3893. I'm very hesitant with this approach, tbh. A try-except at that place in *addition* to a proper fixe for #3893 would be indeed desirable, but adding it without an actual fix (or propared infrastructure for a fix) does not seem a good solution to me.
In any case, I strongly suggest to make the logging level WARNING instead of DEBUG, as there is indeed something happening that the user should be aware of.

guillemap:
I agree.
My need to submit this PR comes from the same issue as https://github.com/python-telegram-bot/python-telegram-bot/issues/3397 and seems totally realted to https://github.com/python-telegram-bot/python-telegram-bot/issues/3893.

What's happening along these lines of code is that the callback is called with a timeout=0. I don't know exactly why, but in combination with these semi-random conditions (as explained in https://github.com/python-telegram-bot/python-telegram-bot/issues/3893) the get_updates() tends to throw the TimedOut exception. Then as it's not handled the application hangs and cannot stop properly.

My suggestion is to:

1. Put a bigger timeout value. I have tested and it also fixes the issue in the sense that it doesn't throw the TimedOut exception. 
2. Also maintain the exception handling to avoid the application to hang there.

Bibo-Joshi:
> 1. Put a bigger timeout value. I have tested and it also fixes the issue in the sense that it doesn't throw the TimedOut exception. 

That is (very roughly) the idea outlined in #3893, yes (although it's the `read_timeout` argument that needs to be adapted and not the `timeout` argument)

> 2. Also maintain the exception handling to avoid the application to hang there.

As I said above, the additional exception handling does sound like a good idea to be, but I'm not comfortable with putting into place without adressing 1 first.

----

Please also note that a workaround is mentioned in https://github.com/python-telegram-bot/python-telegram-bot/issues/3893#issuecomment-1735346440

Bibo-Joshi:
Closing in favor of #3963

PR #3962, opened 2023-11-04
Fix Persistency Issue with Ended Conversations

Bibo-Joshi:
Fixes a bug in persistency of conversations:

CH.END is persited as other states, however `CH._update_state` expects to completely delete keys once they reach that state. So I now call `CH._state_update` manually in `_initialize_persistence`.

Tests still missng.

Reference: https://t.me/pythontelegrambotgroup/707714

- ~[ ] Added ``.. versionadded:: NEXT.VERSION``, ``.. versionchanged:: NEXT.VERSION`` or ``.. deprecated:: NEXT.VERSION`` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)~
- [x] Added myself alphabetically to ``AUTHORS.rst`` (optional)
- ~[ ] Added new classes & modules to the docs and all suitable ``__all__`` s~

### DIFF
--- a/telegram/ext/_conversationhandler.py
+++ b/telegram/ext/_conversationhandler.py
@@ -626,6 +626,8 @@ class ConversationHandler(BaseHandler[Update, CCT]):
 
         # Since CH.END is stored as normal state, we need to properly parse it here in order to
         # actually end the conversation, i.e. delete the key from the _conversations dict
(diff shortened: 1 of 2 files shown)


Poolitzer (approved):
LGTM

PR #3965, opened 2023-11-04
Improve Insertion of Kwargs into `Bot` Methods

Bibo-Joshi:
Fixes the issue that the kwargs that we auto-insert intsead of hard-coding are not properly indexed and do not show up in

* the rules-bot search
* in the pop-up windows search in the docs

nice side-effect: we can now write the insertion as google style docstring

Should be aligned with #3952, i.e. whichever is merged second must adapt.

### DIFF
--- a/docs/auxil/kwargs_insertion.py
+++ b/docs/auxil/kwargs_insertion.py
@@ -18,42 +18,31 @@
 import inspect
 
 keyword_args = [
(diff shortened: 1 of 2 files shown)


harshil21 (approved):
Perfect! I had tried this earlier but it didn't work because I missed the `priority` setting, TIL.

PR #3972, opened 2023-11-11
Adding parameter test_env to bot

JoelGotsch:
allowing to use the test environment via the new test_env parameter for the application builder.
This enables testing Web Apps as documented here: https://core.telegram.org/bots/webapps#testing-mini-apps

I tried to document as well as possible, but could need some help for coming up with tests for that.

### DIFF
--- a/docs/source/inclusions/bot_methods.rst
+++ b/docs/source/inclusions/bot_methods.rst
@@ -390,6 +390,8 @@
       - The last name of the bot
     * - :attr:`~telegram.Bot.local_mode`
       - Whether the bot is running in local mode
+    * - :attr:`~telegram.Bot.test_env`
+      - Whether the bot should use the test endpoints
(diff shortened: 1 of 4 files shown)


Bibo-Joshi:
Hi. Thanks for taking the time to create this PR! I see that you've even dug into `ApplicationBuilder` to add a corresponding method 👍 
Still, tbh I'm very hesitant to consider this PR for merge, because for #3355 we have not yet had a decision to actually put it on the roadmap, nor which technical solution we would like to go with in that case. In particular, the mentioned open question is still open.

Bibo-Joshi:
I'll go ahead and close this PR due to inactivity. I'll linkto it in #3355 so that eventual future work may be based on it if we at some point decide to go ahead with that. Again, thanks for making the effort!

Issue #3971, opened 2023-11-11
[BUG] Exception when starting bot with Persistence enabled where non-default `store_data` value is provided


msenior85:
### Steps to Reproduce

1.  Create a telegram application with persistence enabled and `store_data` argument of the `Persistence` class having a non-default value
```python
persistence = PicklePersistence(
    filepath=".bot_data",
    store_data=PersistenceInput.bot_data,
)
app = (
    ApplicationBuilder()
    .token("BOT_TOKEN")
    .persistence(persistence)
    .build()
)
```
2. Run the bot
3. `AttributeError` Exception will be raised when starting the bot

### Expected behaviour

Bot will start up without exception and start polling the telegram server for updates.

### Actual behaviour

Exception is raised when starting the bot
```bash
Traceback (most recent call last):
  File "/home/george/Projects/echo/bot.py", line 51, in <module>
    main()
  File "/home/george/Projects/echo/bot.py", line 43, in main
    .build()
  File "/home/george/Projects/telegram-quiz-bot/venv/lib/python3.10/site-packages/telegram/ext/_applicationbuilder.py", line 328, in build
    persistence.set_bot(bot)
  File "/home/george/Projects/telegram-quiz-bot/venv/lib/python3.10/site-packages/telegram/ext/_basepersistence.py", line 181, in set_bot
    if self.store_data.callback_data and (not isinstance(bot, ExtBot)):
AttributeError: '_collections._tuplegetter' object has no attribute 'callback_data'
```

### Operating System

Ubuntu 22.04

### Version of Python, python-telegram-bot & dependencies

```shell
python-telegram-bot 20.6
Bot API 6.9
Python 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
```

### Relevant log output

```python
Traceback (most recent call last):
  File "/home/george/Projects/echo/bot.py", line 51, in <module>
    main()
  File "/home/george/Projects/echo/bot.py", line 43, in main
    .build()
  File "/home/george/Projects/telegram-quiz-bot/venv/lib/python3.10/site-packages/telegram/ext/_applicationbuilder.py", line 328, in build
    persistence.set_bot(bot)
  File "/home/george/Projects/telegram-quiz-bot/venv/lib/python3.10/site-packages/telegram/ext/_basepersistence.py", line 181, in set_bot
    if self.store_data.callback_data and (not isinstance(bot, ExtBot)):
[...]

PR #3973, opened 2023-11-12
Add persistent bot example showing how to use persistence input #3971

msenior85:
Add an example bot showing how to use `PersistenceInput` class when creating an application that saves one of more but not all of `{bot, chat, user, callback}_data`.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -55,6 +55,7 @@ The following wonderful people contributed directly or indirectly to this projec
 - `franciscod <https://github.com/franciscod>`_
 - `gamgi <https://github.com/gamgi>`_
 - `Gauthamram Ravichandran <https://github.com/GauthamramRavichandran>`_
+- `George Muholi <https://github.com/msenior85>`_
 - `Harshil <https://github.com/harshil21>`_
 - `Hugo Damer <https://github.com/HakimusGIT>`_
 - `ihoru <https://github.com/ihoru>`_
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
Hi. Thanks for taking the time to send a PR. However, I see no added value in your changes for PTBs documentation because

1. We already have examples on persistent conversations in place. If we would want to showcase the usage of `PersistenceInput`, it would be enough to do that in those examples.
2. IMO there is no need to have an explicit example about how to use the `PersistenceInput` class in your code. The docs of [`BasePersistence`](https://docs.python-telegram-bot.org/en/stable/telegram.ext.basepersistence.html#telegram.ext.BasePersistence.params.store_data) state that the `store_data` argument expects an object of type `PersistenceInput` and [the docs of that class](https://docs.python-telegram-bot.org/en/stable/telegram.ext.persistenceinput.html#telegram.ext.PersistenceInput) show you which arguments it accepts. This kind of documentation is standard bot within PTB and also within other libraries. See e.g. the [docs of `Appliction.add_handler`](https://docs.python-telegram-bot.org/en/stable/telegram.ext.application.html#telegram.ext.Application.add_handler) or [the docs of the widely used `requests` library](https://requests.readthedocs.io/en/latest/api/#requests.adapters.HTTPAdapter.add_headers). I consider understanding the concepts of "accepted types for a parameter" and "instantiating an instance of a specified class" to be part of the fundamentals of object oriented programming in python. Teaching those is beyond the scope of PTBs documentation.

PR #3977, opened 2023-11-21
Add `do_process` in `BaseHandler`

Antares0982:
In the `process_update` method of the class `Application`, the handler is called through a "check then run" approach, where the "check" and "run" steps are two different calls. If this piece of code is moved to the `BaseHandler` as a method `do_process`, users can choose to override this method to implement a customized processing of the update. For instance, by inheriting the `ConversationHandler` and overriding it like this:

```python
async def do_process(self, context, update, app):
    async with self._some_async_lock:
        super().do_process(context, update, app)
```

One can achieve a simple (but inefficient) concurrent version of `ConversationHandler`. With a slightly more complex implementation an efficient `ConversationHandler` is also possible.

This PR only refactors the code and does not introduce new things.

### DIFF
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -1179,31 +1179,12 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
         for handlers in self.handlers.values():
             try:
                 for handler in handlers:
-                    check = handler.check_update(update)  # Should the handler handle this update?
-                    if not (check is None or check is False):  # if yes,
-                        if not context:  # build a context if not already built
-                            context = self.context_types.context.from_update(update, self)
-                            await context.refresh_data()
-                        coroutine: Coroutine = handler.handle_update(update, self, check, context)
-
-                        if not handler.block or (  # if handler is running with block=False,
-                            handler.block is DEFAULT_TRUE
-                            and isinstance(self.bot, ExtBot)
-                            and self.bot.defaults
-                            and not self.bot.defaults.block
-                        ):
-                            self.create_task(
-                                coroutine,
-                                update=update,
-                                name=(
-                                    f"Application:{self.bot.id}:process_update_non_blocking"
-                                    f":{handler}"
-                                ),
-                            )
-                        else:
-                            any_blocking = True
-                            await coroutine
-                        break  # Only a max of 1 handler per group is handled
+                    processed, context, _cur_blocking = await handler.do_process(
+                        context, update, self
+                    )
+                    if processed:
+                        any_blocking |= _cur_blocking
+                        break
 
             # Stop processing with any other handler.
             except ApplicationHandlerStop:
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
Hi. Thanks for the PR! Even though you mostly moved around existing logic, at first glance the change still looks like a significant impact on the PTB logic to me as it effectively removes the separation between `check_update` and `process_update`. E.g. users could override do_process without calling `super()` or either of `{check, process_update}`.
I also don't quite understand what concurrency use cases you have in mind for `ConversationHandler`. What exactly do you want to achieve that's not already doable via [ptbs built-in concurrency feautures](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Concurrency)?

Antares0982:
> I also don't quite understand what concurrency use cases you have in mind for `ConversationHandler`. What exactly do you want to achieve that's not already doable via [ptbs built-in concurrency feautures](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Concurrency)?

The document of `ConversationHandler` says that
> Warning:
    ConversationHandler heavily relies on incoming updates being processed one by one. When using this handler, telegram.ext.ApplicationBuilder.concurrent_updates should
    be set to False.

For example the entry of a `ConversationHandler` is `/start` and `concurrent_updates` is set to `True`. if there are multiple `/start`s in a very short time, the `ConversationHandler` will be triggered more than once (you may already know that).

Without refactoring this, it is difficult to implement a concurrent `ConversationHandler`. The user may need to subclass the `Application`, override `process_update` and call `application_class()` in `ApplicationBuilder` to build the instance. If the user wants the "check and run" still works for most handlers, he has to write more codes. Or, the user can acquire lock in `check_update` and release it after `handle_update`, but need to be very careful when exception happens because the lock is acquired and released in two different functions (cannot use the `async with` syntax).

> Even though you mostly moved around existing logic, at first glance the change still looks like a significant impact on the PTB logic to me as it effectively removes the separation between `check_update` and `process_update`. E.g. users could override do_process without calling `super()` or either of `{check, process_update}`. 

We should trust the user is experienced enough to override the (a bit complicated) method `do_process`. The user is free to override without calling  either of `{check, process_update}`, if he certainly knows what he's doing.  We can rename the method to something like `handler_do_process_internal` to avoid overriding it accidently.

> Note also that your changes at least change _some_ logic as several tests are failing ;)

That's weird, it seems [this commit](https://github.com/Antares0982/python-telegram-bot/commit/a6618e901306a069fb6df6c783b49a2c9bf02ed5) is failing the unit test, but it actually did nothing more than breaking one method into two. Did I miss any thing obvious here? :(

Bibo-Joshi:
> The document of ConversationHandler says that

I see. Let me mention that by now we have the setup of `BaseUpdateProcessor` which allows to customize the setup of concurrent update handling, which allows e.g. to process updates of different chats concurrently while updates within each chat are processed sequentially. This can be a way of enabling concurrency of update processing while still ensuring sequential processing of `ConverstionHandler`

> We should trust the user is experienced enough to override the (a bit complicated) method `do_process`.

Frankly, I tend to disagree :D The logic of how PTB processes updates tries to be adaptable to a variety of use cases but if you don't fully understand it, it's easy to trigger side effects that you haven't considered and that are hard to debug. From my experience of the last few years I must say that I do feel like there is a significant number of people who I think would fall into these traps. Things that come to my mind spontaniously are:

* using a naive singleton lock in CH + concurrent updates can easily lead to effectively sequential processing b/c the lock will be aquired by all different conversations being handled by the same CH: If CH is processing an update of user1, the updates of users2, 3, … will all have to wait even to be *checked* until user1 is handled
* making extensive use of I/O operations in `do_process` will slow down the loop of checking which handler should handle an update. This can e.g. lead to inline queries having time out before being handed to the correct handler
* making extensive use of bot methods like `get_chat_admins` can lead to exceeding flood limits quickly and/or also slowing downs the loop as described above

I'm certain that there are other implications that I'm not seeing right now. As punshline: I think this needs elaborate discussion and so far I'm not entirely convinced of the added benefit.

Antares0982:
> I see. Let me mention that by now we have the setup of `BaseUpdateProcessor` which allows to customize the setup of concurrent update handling, which allows e.g. to process updates of different chats concurrently while updates within each chat are processed sequentially. This can be a way of enabling concurrency of update processing while still ensuring sequential processing of `ConverstionHandler`

I think the ideal case is that we only acquire lock for specific types of handlers (which are likely to have lower priority to other handlers), while other updates are still processed concurrently (per chat per user).

> > We should trust the user is experienced enough to override the (a bit complicated) method `do_process`.
> 
> Frankly, I tend to disagree :D The logic of how PTB processes updates tries to be adaptable to a variety of use cases but if you don't fully understand it, it's easy to trigger side effects that you haven't considered and that are hard to debug. From my experience of the last few years I must say that I do feel like there is a significant number of people who I think would fall into these traps. Things that come to my mind spontaniously are:
> 
> * using a naive singleton lock in CH + concurrent updates can easily lead to effectively sequential processing b/c the lock will be aquired by all different conversations being handled by the same CH: If CH is processing an update of user1, the updates of users2, 3, … will all have to wait even to be _checked_ until user1 is handled
> * making extensive use of I/O operations in `do_process` will slow down the loop of checking which handler should handle an update. This can e.g. lead to inline queries having time out before being handed to the correct handler
> * making extensive use of bot methods like `get_chat_admins` can lead to exceeding flood limits quickly and/or also slowing downs the loop as described above
> 
> I'm certain that there are other implications that I'm not seeing right now. As punshline: I think this needs elaborate discussion and so far I'm not entirely convinced of the added benefit.

I see your point. I have to admit that this refactor is violating the aesthetics, the package itself should only provide simple and general code. I would like to know that if there is a simpler way to achive this

> I think the ideal case is that we only acquire lock for specific types of handlers (which are likely to have lower priority to other handlers), while other updates are still processed concurrently (per chat per user).

PR #3986, opened 2023-11-26
Feat: Introduce unix mode for webhook

### LINKED ISSUE #3978: [FEATURE] bind webhook server to unix socket?

The usual thing to do with python web servers is to put them behind a reverse proxy, such as NGINX. And the usual thing to do to a server behind NGINX is to run it locally bound to a unix socket. However, there seems to be no way to do just that currently in PTB, as far as I can see in the `webhookhandler.py` file?

I think I'd be nice to have an additional parameter in the `WebhookServer` constructor, such as `unix_socket`, which would bind the server to the specified unix socket instead of the host-port pair.

_No response_

### Additional context

[Here is the docs confirming, that currently you'd use NGINX only with a regular localhost connection.](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Webhooks#using-nginx-with-one-domainport-for-all-bots)

Poolitzer:
This closes #3978. It might break on windows with the Path solution I came up with for the input to tornado.


telegram/ext/_updater.py
```diff
@@ -53,6 +53,7 @@
 
 
 _UpdaterType = TypeVar("_UpdaterType", bound="Updater")  # pylint: disable=invalid-name
+_DefaultIP = DefaultValue("127.0.0.1")
```
Bibo-Joshi, line 56: If this is supposed to be used outside of this module, it should not start with an underscore. TBH I personally would move this to `defaultvalue.py` b/c that's where we define the other singletons, but that's not a deabreaker for me :D

telegram/ext/_application.py
```diff
@@ -911,6 +912,11 @@ def run_webhook(
                 header isn't set or it is set to a wrong token.
 
                 .. versionadded:: 20.0
+            unix (:class:`pathlib.Path` | :obj:`str`, optional): Path to the unix socket file. Path
+                can be empty in which case the file will be created. When using this param, you
```
Bibo-Joshi, line 916:
```suggestion
                does not need to exist, in which case the file will be created. When using this param, you
```

"can be empty" sounds like "you can pass an empty string '' here" to me 🤔

telegram/ext/_updater.py
```diff
@@ -499,6 +501,11 @@ async def start_webhook(
                 header isn't set or it is set to a wrong token.
 
                 .. versionadded:: 20.0
+            unix (:class:`pathlib.Path` | :obj:`str`, optional): Path to the unix socket file. Path
+                can be empty in which case the file will be created. When using this param, you
+                need to set the :paramref:`webhook_url`!
```
Bibo-Joshi, line 506: please also explain that this is mutually exclusive with listen, port & url_path. The notes on webhook_url & the mutual excluisveness can go into an "Attention" (danger/caution/…) annotation.

telegram/ext/_utils/webhookhandler.py
```diff
@@ -50,21 +54,37 @@ class WebhookServer:
...
         self._shutdown_lock = asyncio.Lock()
 
     async def serve_forever(self, ready: Optional[asyncio.Event] = None) -> None:
         async with self._server_lock:
-            self._http_server.listen(self.port, address=self.listen)
+            if self.unix:
+                socket = bind_unix_socket(str(self.unix))
+                self._http_server.add_socket(socket)
+            else:
+                self._http_server.listen(self.port, address=DefaultValue.get_value(self.listen))
```
Bibo-Joshi, line 87: Can you move the `get_value` to `updater.py`? that would be cleaner IMO.

telegram/ext/_utils/webhookhandler.py
```diff
@@ -50,21 +54,37 @@ class WebhookServer:
...
+        port: int,
+        webhook_app: "WebhookAppClass",
+        ssl_ctx: Optional[SSLContext],
+        unix: Optional[Union[str, Path]] = None,
     ):
+        if not isinstance(listen, DefaultValue) and unix:
+            raise RuntimeError(
+                "You can not pass unix and listen, only use one. Unix if you want to initialize a"
+                " unix socket, or listen for a standard TCP server"
+            )
```
Bibo-Joshi, line 72: I would prefer for this to happen in `Updater` already. IMO it's not webhookhandlers responsibility do to argument parsing.

Another question/observation: IISC `port` and `url_path` are also mutually exclusive to `unix`, right? if so, these should be checked here as well and for this should also be `DefaultValue` objects I guess 🤔

telegram/ext/_application.py
Poolitzer: Depending on the os I bet you can

telegram/ext/_updater.py
Poolitzer: no I want to move this

telegram/ext/_utils/webhookhandler.py
Poolitzer: so much default

telegram/ext/_updater.py
```diff
@@ -456,8 +456,8 @@ async def _get_updates_cleanup() -> None:
 
     async def start_webhook(
         self,
-        listen: str = "127.0.0.1",
-        port: int = 80,
+        listen: DVType[str] = DEFAULT_IP,
+        port: DVType[int] = DEFAULT_80,
         url_path: str = "",
```
Bibo-Joshi, line 461: Is there a specific reason why you didn't make this a `DefaultValue` (you even created `DEFAULT_EMPTY_STRING`) and also didn't include it in the docstrings and runtime checks? 🤔

tests/auxil/networking.py
```diff
@@ -111,7 +112,11 @@ async def send_webhook_message(
 
     url = f"http://{ip}:{port}/{url_path}"
 
-    async with AsyncClient() as client:
+    transport = None
+    if unix:
+        transport = AsyncHTTPTransport(uds=unix)
```
Bibo-Joshi, line 117:
```suggestion
    transport = AsyncHTTPTransport(uds=unix) if unix else None
```

tests/ext/test_updater.py
```diff
@@ -73,6 +74,12 @@ def _reset(self):
         self.cb_handler_called = None
         self.test_flag = False
 
+    @pytest.fixture()
+    def file_path(self) -> str:
+        path = TEST_DATA_PATH / "test.sock"
+        yield str(path)
+        path.unlink(missing_ok=True)
+
```
Bibo-Joshi, line 82: don't [pytests built-in features](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) suffice for this?

tests/ext/test_updater.py
```diff
@@ -679,9 +686,13 @@ async def delete_webhook(*args, **kwargs):
...
     @pytest.mark.parametrize("drop_pending_updates", [True, False])
     @pytest.mark.parametrize("secret_token", ["SecretToken", None])
+    @pytest.mark.parametrize("unix", [None, True])
     async def test_webhook_basic(
-        self, monkeypatch, updater, drop_pending_updates, ext_bot, secret_token
+        self, monkeypatch, updater, drop_pending_updates, ext_bot, secret_token, unix, file_path
     ):
+        # Skipping unix test on windows since they fail
+        if unix and platform.system() == "Windows":
+            pytest.skip("Windows doesn't support unix bind")
```
Bibo-Joshi, line 695: why not make this a fixture like [here](https://github.com/python-telegram-bot/python-telegram-bot/blob/592c6cc6d349f3d1b84c270fb9efce283e0e7f28/tests/ext/test_ratelimiter.py#L144-L147)?

tests/ext/test_updater.py
```diff
@@ -755,19 +792,75 @@ async def set_webhook(*args, **kwargs):
...
+                unix=file_path if unix else None,
+            )
             assert (await updater.update_queue.get()).to_dict() == update.to_dict()
             await updater.stop()
             assert not updater.running
 
+    @pytest.mark.skipif(
+        platform.system() == "Windows",
+        reason="Windows doesn't support unix bind",
+    )
```
Bibo-Joshi, line 828: these tests should pass on all OSes, shouldn't they?

tests/ext/test_updater.py
```diff
@@ -755,19 +792,75 @@ async def set_webhook(*args, **kwargs):
...
+    async def test_unix_webhook_raises(self, updater):
```
Bibo-Joshi, line 829:
```suggestion
    async def test_unix_webhook_mutually_exclusive_params(self, updater):
```

tests/ext/test_updater.py
```diff
@@ -755,19 +792,75 @@ async def set_webhook(*args, **kwargs):
...
+            return True
+
+        monkeypatch.setattr(updater.bot, "set_webhook", set_webhook)
+
+        async with updater:
+            await updater.start_webhook(unix=file_path, webhook_url="string")
+
+            update = make_message_update("Webhook")
+            await send_webhook_message("ip", 123, update.to_json(), unix=file_path)
+            await updater.stop()
```
Bibo-Joshi, line 853: What does this test? :D It's clear to me neither from the test name nor from the content …

Bibo-Joshi (changes requested):

telegram/ext/_updater.py
Poolitzer: yes, url_path is supported for unix servers. I forgot to remove the default

tests/ext/test_updater.py
Poolitzer: where was this when I created this code grrrr

tests/ext/test_updater.py
Poolitzer: because all other params should be tested on windows, and I would hate to copy paste the code just to have the skip as a decorator

tests/ext/test_updater.py
Poolitzer: right.

tests/ext/test_updater.py
Poolitzer: sorry leftover from me wondering if path works on windows.

tests/ext/test_updater.py
Bibo-Joshi: Here https://t.me/c/1494805131/38624 🤡

tests/ext/test_updater.py
Bibo-Joshi: ah, fair enough :D

tests/auxil/networking.py
```diff
@@ -90,7 +91,7 @@ async def send_webhook_message(
     content_type: str = "application/json",
     get_method: Optional[str] = None,
     secret_token: Optional[str] = None,
-    unix: Optional[str] = None,
+    unix: Optional[Path] = None,
```
Bibo-Joshi, line 94: maybe allow both str & Path and convert just to `Path(unix)`? We even have a type alias prepared for that somewhere.

telegram/ext/_updater.py
Bibo-Joshi: Huh? Can you give an example of how a http request to a socket file with url_path would look like? Is `url_path` a good name in this case? Consider me unknowledgable in this regard, so just for my understanding …

tests/auxil/networking.py
Poolitzer: Does that matter for an internal testing method?

telegram/ext/_updater.py
Poolitzer: We still start a http server, just instead of listening to an IP, it listens to a file. But one could still define /url_path/to_here, for the http server it doesn't matter if the full call is 127.0.0.1/url_path/to_here or unix.socket/url_path/to_here. And since we support it for the IP binding, I thought it makes sense to also pass it for unix. I don't personally see why anyone would use this but pfff.

We actually already use this in the test if you read carefully, url_path is set to "TOKEN".

Bibo-Joshi: https://media.giphy.com/media/lXu72d4iKwqek/giphy.gif

fancy

Bibo-Joshi (approved):

PR #3987, opened 2023-11-26
Feat: Better handling of non Updates in webhooks

### LINKED ISSUE #3979: UnboundLocalError when submitting smth else then an Update to the webhook
https://github.com/python-telegram-bot/python-telegram-bot/blob/dd9af64a5c99c2518ef86027e008cdb8659dd1f5/telegram/ext/_utils/webhookhandler.py#L141-L148

I realized that when putting a not decodable "Update", the appropriate log is shown, but since the code continues to run, the other party receives a non descriptive 500 Internal Server Error and in the traceback an UnboundLocalError appears.

This could be easily fixed with adding `update = None` to line 149, but this means the sending party receives 200 OK while it is certainly not OK.

However, if I add a descriptive error here, bad actors could realize that they hit a telegram bot webhook and try to abuse it/spam it.  Do we want that?

If people want to protect their webhooks, they should add the secret anyway. There we however raise if the wrong/none token has been supplied, already exposing that a PTB webhook is located behind the URL...

<details>
 <summary>Traceback if you want to look at it</summary>

```
DEBUG:telegram.ext.Updater:0.0.0.0 - Exception in TelegramHandler
  File "/home/poolitzer/python-telegram-bot/venv/lib/python3.11/site-packages/tornado/web.py", line 1786, in _execute
    result = await result
             ^^^^^^^^^^^^
  File "/home/poolitzer/python-telegram-bot/telegram/ext/_utils/webhookhandler.py", line 159, in post
    if update:
       ^^^^^^
UnboundLocalError: cannot access local variable 'update' where it is not associated with a value
```

</details>

Poolitzer:
This closes #3979

### DIFF
--- a/telegram/ext/_utils/webhookhandler.py
+++ b/telegram/ext/_utils/webhookhandler.py
@@ -148,7 +148,7 @@ class TelegramHandler(tornado.web.RequestHandler):
             )
             raise tornado.web.HTTPError(
                 HTTPStatus.BAD_REQUEST, reason="Update could not be processed"
-            ) from None
+            ) from exc
 
         if update:
             _LOGGER.debug(
--- a/telegram/ext/_utils/webhookhandler.py
+++ b/telegram/ext/_utils/webhookhandler.py
@@ -146,6 +146,9 @@ class TelegramHandler(tornado.web.RequestHandler):
                 "Received data was *not* processed!",
                 exc_info=exc,
             )
+            raise tornado.web.HTTPError(
+                HTTPStatus.BAD_REQUEST, reason="Update could not be processed"
+            ) from None
 
         if update:
             _LOGGER.debug(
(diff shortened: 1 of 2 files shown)


telegram/ext/_utils/webhookhandler.py
```diff
@@ -146,6 +146,9 @@ async def post(self) -> None:
                 "Received data was *not* processed!",
                 exc_info=exc,
             )
+            raise tornado.web.HTTPError(
+                HTTPStatus.BAD_REQUEST, reason="Update could not be processed"
+            ) from None
```
Bibo-Joshi, line 151: why `from None` and not `from exc`?

Poolitzer: hm, thought it wouldn´t make a difference to the end user but you are right its better with.

Bibo-Joshi (approved):

PR #3773, opened 2023-06-23
Invoice Payload Filtering

thefunkycat:
Closes #3752 

I've made two commits:
- one updates the successful payment message filter to also accept a list of strings to be able to filter 
- the other adds a pattern argument to the precheckoutqueryhandler and checks updates based on that.

Looking forward to a review, will gladly make any changes required.

PR #4005, opened 2023-12-12
Add `invoice_payload` filtering

### LINKED ISSUE #3752: Filter support for invoice_payload in PreCheckoutQueryHandler and successful_payment message

There is currently no solution to filter for `invoice_payload` in the previously mentioned updates.

I would like to change the SuccessfulPayment message filter to the same type of filter we have with the [text one](https://docs.python-telegram-bot.org/telegram.ext.filters.html#telegram.ext.filters.Text). And I would like to have a pattern argument in the PreCheckoutQueryHandler, similar to the [callbackqueryhandler](https://docs.python-telegram-bot.org/en/stable/telegram.ext.callbackqueryhandler.html).

Filtering has to happen in the callback. That works, but is not our design philosophy.

### Additional context

_No response_

aelkheir:
Closes #3752

- [x]  Added ``.. versionadded:: NEXT.VERSION``, ``.. versionchanged:: NEXT.VERSION`` or ``.. deprecated:: NEXT.VERSION`` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)
- [x]  Documented code changes according to the `CSI standard <https://standards.mousepawmedia.com/en/stable/csi.html>`__
- [ ]  ~Added myself alphabetically to ``AUTHORS.rst`` (optional)~
- [x]  Added new classes & modules to the docs and all suitable ``__all__`` s
- [ ]  Checked the `Stability Policy <https://docs.python-telegram-bot.org/stability_policy.html>`_ in case of deprecations or changes to documented behavior

why not add `collect_additional_context` in `PreCheckoutQueryHandler` to store the match since we're adding a `pattern` argument. i can do it if you approve.

### DIFF
--- a/telegram/ext/_precheckoutqueryhandler.py
+++ b/telegram/ext/_precheckoutqueryhandler.py
@@ -20,7 +20,7 @@
 
 
 import re
-from typing import Optional, TypeVar, Union
+from typing import Optional, Pattern, TypeVar, Union
 
 from telegram import Update
 from telegram._utils.defaultvalue import DEFAULT_TRUE
@@ -77,7 +77,7 @@ class PreCheckoutQueryHandler(BaseHandler[Update, CCT]):
     def __init__(
         self,
         callback: HandlerCallback[Update, CCT, RT],
-        pattern: Optional[Union[str, re.Pattern[str]]] = None,
+        pattern: Optional[Union[str, Pattern[str]]] = None,
         block: DVType[bool] = DEFAULT_TRUE,
     ):
         super().__init__(callback, block=block)
@@ -85,7 +85,7 @@ class PreCheckoutQueryHandler(BaseHandler[Update, CCT]):
         if isinstance(pattern, str):
             pattern = re.compile(pattern)
 
-        self.pattern: Optional[Union[str, re.Pattern[str]]] = pattern
+        self.pattern: Optional[Union[str, Pattern[str]]] = pattern
 
     def check_update(self, update: object) -> bool:
         """Determines whether an update should be passed to this handler's :attr:`callback`.
--- a/telegram/ext/_precheckoutqueryhandler.py
+++ b/telegram/ext/_precheckoutqueryhandler.py
@@ -100,8 +100,6 @@ class PreCheckoutQueryHandler(BaseHandler[Update, CCT]):
         if isinstance(update, Update) and update.pre_checkout_query:
             invoice_payload = update.pre_checkout_query.invoice_payload
             if self.pattern:
-                if invoice_payload is None:
-                    return False
                 if re.match(self.pattern, invoice_payload):
                     return True
             else:
--- a/telegram/ext/_precheckoutqueryhandler.py
+++ b/telegram/ext/_precheckoutqueryhandler.py
@@ -50,9 +50,8 @@ class PreCheckoutQueryHandler(BaseHandler[Update, CCT]):
 
             The return value of the callback is usually ignored except for the special case of
             :class:`telegram.ext.ConversationHandler`.
-        pattern (:func:`re.Pattern <re.compile>` | :obj:`callable` | :obj:`type`): Optional.
-            Regex pattern, callback or type to test
-            :attr:`telegram.PreCheckoutQuery.invoice_payload` against.
+        pattern (:obj:`str` | :func:`re.Pattern <re.compile>`, optional): Optional. Regex pattern
+            to test :attr:`telegram.PreCheckoutQuery.invoice_payload` against.
 
             .. versionadded:: NEXT.VERSION
         block (:obj:`bool`, optional): Determines whether the return value of the callback should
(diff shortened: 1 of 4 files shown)


Bibo-Joshi:
Thanks for the PR! I have two general comments that I'd like you to address before I review in details:

* the `pattern` argument for `PreCheckoutQueryHanlder` accepting a `type` makes little sense, as the invoice payload will alsways be a string. I would even argue that we don't need to allow callables. This was added to `CallbackQueryHandler` only for usage with arbitrary callback_data and not for string data
* Please add unit tests for the code changes and ensure that the existing ones don't fail. See also https://github.com/python-telegram-bot/python-telegram-bot/issues/3752#issuecomment-1598921068. If you're new to writing unit tests and have problems getting started, that's ofc ok and I recommend to reach out via https://t.me/pythontelegrambotchannel/105 in this case.

aelkheir:
Thanks for the feedback.
> This was added to CallbackQueryHandler only for usage with arbitrary callback_data

Aha got it, I did not understand why it was there

>  I would even argue that we don't need to allow callables.

Ok. I thought it was needed because of this https://github.com/python-telegram-bot/python-telegram-bot/pull/3773#discussion_r1249788373


> Please add unit tests for the code changes and ensure that the existing ones don't fail.

On it.

aelkheir:
I create some test. I see some checks are still failing. Doc build is failing, although it built fine locally.

aelkheir:
Couldn't figure out why pytest (3.8, macos-latest) is still failing. But I think its ready for review. @Bibo-Joshi

tests/ext/test_filters.py
```diff
@@ -1877,6 +1878,13 @@ def test_filters_successful_payment(self, update):
         update.message.successful_payment = "test"
         assert filters.SUCCESSFUL_PAYMENT.check_update(update)
 
+    def test_filters_successful_payment_payloads(self, update):
+        update.message.successful_payment = SuccessfulPayment(
+            "USD", 100, "custom-payload", "123", "123"
+        )
+        assert filters.SuccessfulPayment(("custom-payload",)).check_update(update)
+        assert not filters.SuccessfulPayment(["test1"]).check_update(update)
```
Bibo-Joshi, line 1886: please also add a test checking the string representation of the new class. There are a few tests with "repr" in their name in this file that you can look at for inspiration :)
It would also be nice

* if you can add a line with `SuccesfulPayment().check_update` here for completenes so that this is not only implicitly tested via `SUCCESFUL_PAYMENT`
* you can test with a message taht doesn't contain a successful payment at all

tests/ext/test_precheckoutqueryhandler.py
```diff
@@ -103,6 +106,14 @@ async def callback(self, update, context):
             and isinstance(update.pre_checkout_query, PreCheckoutQuery)
         )
 
+    def test_with_pattern(self, pre_checkout_query):
+        handler = PreCheckoutQueryHandler(self.callback, pattern=".*voice.*")
```
Bibo-Joshi, line 110: please also test passing a compiled pattern :)

Bibo-Joshi (changes requested):
Thanks for the nice PR! I left a few smaller comments below :) the tests on macOS recently tend to fail due to flakyness in time-based jobs …

tests/ext/test_filters.py
aelkheir: > add a test checking the string representation of the new class.

Is it okay to keep the str repr of the filter, as it for e.g will return, `filters.SuccessfulPayment(['payload'])`, or should it be modified to match some other filters that implement a `name` property to remove the square brackets and single quotes, so like, `filters.SuccessfulPayment(payload1, payload2)`

tests/ext/test_precheckoutqueryhandler.py
```diff
@@ -103,6 +107,23 @@ async def callback(self, update, context):
...
+
+        assert handler.check_update(pre_checkout_query)
+
+        pre_checkout_query.pre_checkout_query.invoice_payload = "nothing here"
+        assert not handler.check_update(pre_checkout_query)
+
+    def test_with_compiled_pattern(self, pre_checkout_query):
```
aelkheir, line 118: I created a new test for compiled pattern. If that's not desired, I can rewrite it in the above test.

aelkheir:
Applied changes, ready for more

telegram/ext/_precheckoutqueryhandler.py
```diff
@@ -75,15 +75,12 @@ async def callback(update: Update, context: CallbackContext)
     def __init__(
         self,
         callback: HandlerCallback[Update, CCT, RT],
-        pattern: Optional[Union[str, Pattern[str]]] = None,
         block: DVType[bool] = DEFAULT_TRUE,
+        pattern: Optional[Pattern[str]] = None,
```
Bibo-Joshi, line 79:
```suggestion
        pattern: Optional[Union[str, Pattern[str]]] = None,
```

accepting a string as input is okay, my point in https://github.com/python-telegram-bot/python-telegram-bot/pull/4005#discussion_r1427212996 was just that we can convert strings to patterns on init already :) sorry if that was unclear.

tests/ext/test_filters.py
Bibo-Joshi: Sounds fine to me. I haven't found other filters that do what you describe, tbh :D

tests/ext/test_precheckoutqueryhandler.py
Bibo-Joshi: It does the job and I'm happy to keep it as is. adding a note here just in case you're interested in pytests capabilities: One can unify the two tests with `pytest.mark.parametrize`.

Bibo-Joshi (approved):
Thanks for the updates and sorry for the late review. LGTM now :) Let me check if one of the other team members can also have a second look.

tests/ext/test_filters.py
aelkheir: I agree it's not necessary. However the filter I was referring to was `filters._ChatUserBaseFilter`.
https://github.com/python-telegram-bot/python-telegram-bot/blob/2345bfbb5337f6e5d63b9d56bbf3fa75646c1b35/telegram/ext/filters.py#L802-L807

Poolitzer (approved):
LGTM

Bibo-Joshi:
Thanks very much for the nice contribution :)

aelkheir:
Ha! No really, thank _you_. I learnt something new.

Issue #4008, opened 2023-12-13
[QUESTION] JobQueue warning message still present


intersecato:
### Issue I am facing

In version `20.7` I still see the warning message that warns of the alignment of the days parameter in JobQueue with cron. In this regard, I had read about a pull request (https://github.com/python-telegram-bot/python-telegram-bot/pull/3046) in which it was specified that it was a message that would be removed in the following version, but it is still present. I don't know if it's intentional or if there is a way to hide it, it's starting to bother me.

Also this commit (https://github.com/python-telegram-bot/python-telegram-bot/pull/3046/commits/97697eee9516fa9ccfa81fee0b8e9fea3863837e) I think it was done to remember it

### Traceback to the issue

```python
PTBUserWarning: Prior to v20.0 the `days` parameter was not aligned to that of cron's weekday scheme. We recommend double checking if the passed value is correct.
```

### Operating System

Linux

### Version of Python, python-telegram-bot & dependencies

```shell
python-telegram-bot 20.7
Bot API 6.9
Python 3.12.0
```

Bibo-Joshi:
Hi. What [this comment](https://github.com/python-telegram-bot/python-telegram-bot/pull/3046#discussion_r880888219) says is that the warning should be removed on the next *major* version, i.e. v21. Since PTB is using Pythons standart [`warnings`](https://docs.python.org/3.11/library/warnings.html?highlight=warnings#module-warnings) module, warnings can also be ignored  via Pythons standard features. We give a short primer on this on [this wiki page](../wiki/Exceptions%2C-Warnings-and-Logging) which should help ease your sore eye ;)

PR #4011, opened 2023-12-15
Removed deprecated warning message

Konano:
In the _jobqueue.py file, we removed a deprecated warning message that is no longer needed after v20.0. This warning was about the `days` parameter not being aligned with cron's weekday scheme.

### DIFF
--- a/telegram/ext/_jobqueue.py
+++ b/telegram/ext/_jobqueue.py
@@ -589,13 +589,6 @@ class JobQueue(Generic[CCT]):
             queue.
 
         """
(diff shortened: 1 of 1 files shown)


Poolitzer:
Hello, thanks for your intention to contribute to our small project. As written out in [this comment](https://github.com/python-telegram-bot/python-telegram-bot/issues/4008#issuecomment-1854578113) of #4008, we will remove the warning in the next _major_ version. Till then you can follow the guide in that comment to disable the warning in your project.

PR #4002, opened 2023-12-10
Documentation Improvements

Bibo-Joshi:
Gathering smaller documentation improvements to be merged before the next release

- [ ] mention #4079
- [ ] #4104

PR #4034, opened 2023-12-29
API 7.0

### LINKED ISSUE #4033: [FEATURE] API 7.0

This is the biggest bot API update ever!, hence opening this issue to keep the track of all changes.

> [!IMPORTANT]  
> * To double check if you have covered all relevant changes (also in documentation), please see https://t.me/bot_api_changes/102 and https://t.me/bot_api_changes/103 (and maybe following).
> * Please copy the checklist from #4034 to your PR and make sure to cover everything :)
> * Please base your branch on the [`api-7.0`](https://github.com/python-telegram-bot/python-telegram-bot/tree/api-7.0) branch
> * If this is your first contribution, please be sure to read the [contribution guide](https://docs.python-telegram-bot.org/en/v20.7/contributing.html) and feel free to reach out with any questions via https://t.me/pythontelegrambotchannel/105

Reactions
- [x]  Added the classes [ReactionTypeEmoji](https://core.telegram.org/bots/api#reactiontypeemoji) and [ReactionTypeCustomEmoji](https://core.telegram.org/bots/api#reactiontypecustomemoji) representing different types of reaction.
- [x] Added updates about a reaction change on a message with non-anonymous reactions, represented by the class [MessageReactionUpdated](https://core.telegram.org/bots/api#messagereactionupdated) and the field message_reaction in the class [Update](https://core.telegram.org/bots/api#update). The bot must explicitly allow the update to receive it. @Poolitzer 
- [x] Added updates about reaction changes on a message with anonymous reactions, represented by the class [MessageReactionCountUpdated](https://core.telegram.org/bots/api#messagereactioncountupdated) and the field message_reaction_count in the class [Update](https://core.telegram.org/bots/api#update). The bot must explicitly allow the update to receive it. @Poolitzer 
- [x] Added the method [setMessageReaction](https://core.telegram.org/bots/api#setmessagereaction) that allows bots to react to messages. @Poolitzer 
- [x] Added the field available_reactions to the class [Chat](https://core.telegram.org/bots/api#chat) @Poolitzer 

Replies 2.0
- [x] Added the class [ExternalReplyInfo](https://core.telegram.org/bots/api#externalreplyinfo) and the field external_reply of type [ExternalReplyInfo](https://core.telegram.org/bots/api#externalreplyinfo) to the class [Message](https://core.telegram.org/bots/api#message), containing information about a message that is replied to by the current message, but can be from another chat or forum topic. @clot27 
- [x] Added the class [TextQuote](https://core.telegram.org/bots/api#textquote) and the field quote of type [TextQuote](https://core.telegram.org/bots/api#textquote) to the class Message, which contains the part of the replied message text or caption that is quoted in the current message. @clot27 
- [x] Added the class [ReplyParameters](https://core.telegram.org/bots/api#replyparameters) and replaced parameters reply_to_message_id and allow_sending_without_reply in the methods [copyMessage](https://core.telegram.org/bots/api#copymessage), [sendMessage](https://core.telegram.org/bots/api#sendmessage), [sendPhoto](https://core.telegram.org/bots/api#sendphoto), [sendVideo](https://core.telegram.org/bots/api#sendvideo), [sendAnimation](https://core.telegram.org/bots/api#sendanimation), [sendAudio](https://core.telegram.org/bots/api#sendaudio), [sendDocument](https://core.telegram.org/bots/api#senddocument), [sendSticker](https://core.telegram.org/bots/api#sendsticker), [sendVideoNote](https://core.telegram.org/bots/api#sendvideonote), [sendVoice](https://core.telegram.org/bots/api#sendvoice), [sendLocation](https://core.telegram.org/bots/api#sendlocation), [sendVenue](https://core.telegram.org/bots/api#sendvenue), [sendContact](https://core.telegram.org/bots/api#sendcontact), [sendPoll](https://core.telegram.org/bots/api#sendpoll), [sendDice](https://core.telegram.org/bots/api#senddice), [sendInvoice](https://core.telegram.org/bots/api#sendinvoice), [sendGame](https://core.telegram.org/bots/api#sendgame), and [sendMediaGroup](https://core.telegram.org/bots/api#sendmediagroup) with the field reply_parameters of type [ReplyParameters](https://core.telegram.org/bots/api#replyparameters). @clot27 

Link Preview Customization - @harshil21 
- [x] Added the class [LinkPreviewOptions](https://core.telegram.org/bots/api#linkpreviewoptions) and replaced the parameter disable_web_page_preview with link_preview_options in the methods [sendMessage](https://core.telegram.org/bots/api#sendmessage) and [editMessageText](https://core.telegram.org/bots/api#editmessagetext).
- [x] Replaced the field disable_web_page_preview with link_preview_options in the class [InputTextMessageContent](https://core.telegram.org/bots/api#inputtextmessagecontent).
- [x] Added the field link_preview_options to the class [Message](https://core.telegram.org/bots/api#message) with information about the link preview options used to send the message.

Block Quotation
- [x] Added support for “blockquote” entities in received messages. @aelkheir 
- [x] Added support for “blockquote” entity parsing in “MarkdownV2” and “HTML” parse modes @aelkheir 
- [x] Allowed to explicitly specify “blockquote” entities in formatted texts. @aelkheir 
- [ ] Update `Message.text_(html|markdown(_v2)?)` - see https://github.com/python-telegram-bot/python-telegram-bot/pull/4038#issuecomment-1881689873 and #4061

Multiple Message Actions @clot27 
- [x] Added the method [deleteMessages](https://core.telegram.org/bots/api#deletemessages) to allow the deletion of multiple messages in a single request. @clot27 
- [x] Added the method [forwardMessages](https://core.telegram.org/bots/api#forwardmessages) for forwarding of multiple messages in a single request. @clot27 
- [x] Added the method [copyMessages](https://core.telegram.org/bots/api#copymessages) for copying of multiple messages in a single request. @clot27 

Request for multiple users  @Bibo-Joshi
- [x] Renamed the class KeyboardButtonRequestUser to [KeyboardButtonRequestUsers](https://core.telegram.org/bots/api#keyboardbuttonrequestusers) and added the field max_quantity to it.
- [x] Renamed the field request_user in the class [KeyboardButton](https://core.telegram.org/bots/api#keyboardbutton) to request_users. The old name will still work for backward compatibility.
- [x] Added the class [UsersShared](https://core.telegram.org/bots/api#usersshared).
- [x] Replaced the field user_shared in the class [Message](https://core.telegram.org/bots/api#message) with the field users_shared.

Chat Boost - @harshil21 
- [x] Added updates about chat boost changes, represented by the classes [ChatBoostUpdated](https://core.telegram.org/bots/api#chatboostupdated) and [ChatBoostRemoved](https://core.telegram.org/bots/api#chatboostremoved) and the fields chat_boost and removed_chat_boost in the class [Update](https://core.telegram.org/bots/api#update). The bot must be an administrator in the chat to receive these updates.
- [x] Added the classes [ChatBoostSourcePremium](https://core.telegram.org/bots/api#chatboostsourcepremium), [ChatBoostSourceGiftCode](https://core.telegram.org/bots/api#chatboostsourcegiftcode) and [ChatBoostSourceGiveaway](https://core.telegram.org/bots/api#chatboostsourcegiveaway), representing different sources of a chat boost.
- [x] Added the method [getUserChatBoosts](https://core.telegram.org/bots/api#getuserchatboosts) for obtaining the list of all active boosts a user has contributed to a chat.

Giveaway @Bibo-Joshi 
- [x] Added the class [Giveaway](https://core.telegram.org/bots/api#giveaway) and the field giveaway to the class [Message](https://core.telegram.org/bots/api#message) for messages about scheduled giveaways.
- [x] Added the class [GiveawayCreated](https://core.telegram.org/bots/api#giveawaycreated) and the field giveaway_created to the class [Message](https://core.telegram.org/bots/api#message) for service messages about the creation of a scheduled giveaway.
- [x] Added the class [GiveawayWinners](https://core.telegram.org/bots/api#giveawaywinners) and the field giveaway_winners to the class [Message](https://core.telegram.org/bots/api#message) for messages about the completion of a giveaway with public winners.
- [x] Added the class [GiveawayCompleted](https://core.telegram.org/bots/api#giveawaycompleted) and the field giveaway_completed to the class [Message](https://core.telegram.org/bots/api#message) for service messages about the completion of a giveaway without public winners.

~Web App Changes~
~- [ ] Added the field SettingsButton to the class [WebApp](https://core.telegram.org/bots/webapps#initializing-mini-apps).~
~- [ ] Added the fields header_bg_color, accent_text_color, section_bg_color, section_header_text_color, subtitle_text_color, destructive_text_color to the class [ThemeParams](https://core.telegram.org/bots/webapps#themeparams).~

Other Changes
- [x] Added support for the fields emoji_status_custom_emoji_id and emoji_status_expiration_date in the class [Chat](https://core.telegram.org/bots/api#chat) for non-private chats. @Bibo-Joshi 
- [x] Added the fields accent_color_id, background_custom_emoji_id, profile_accent_color_id, and profile_background_custom_emoji_id to the class [Chat](https://core.telegram.org/bots/api#chat). @Bibo-Joshi 
- [x] Added the field has_visible_history to the class [Chat](https://core.telegram.org/bots/api#chat). @Bibo-Joshi 
- [x] Added the class [MessageOrigin](https://core.telegram.org/bots/api#messageorigin) and replaced the fields forward_from, forward_from_chat, forward_from_message_id, forward_signature, forward_sender_name, and forward_date with the field forward_origin of type [MessageOrigin](https://core.telegram.org/bots/api#messageorigin) in the class [Message](https://core.telegram.org/bots/api#message). @clot27 
- [x] Improved documentation for the field message of the class [callbackQuery](https://core.telegram.org/bots/api#callbackquery) and the field pinned_message of the class [Message](https://core.telegram.org/bots/api#message) by adding the classes [MaybeInaccessibleMessage](https://core.telegram.org/bots/api#maybeinaccessiblemessage) and [InaccessibleMessage](https://core.telegram.org/bots/api#inaccessiblemessage). @Poolitzer 

If you want to contribute, you can start from the [Contribution Guide](https://docs.python-telegram-bot.org/en/v20.7/contributing.html).

### Additional context

Please comment below before start working on any part of the update so as to keep things in track and avoid double work.

clot27:
Closes #4033 

- [x] Re-enable checking of defaults handling for default_quote in `Message.reply_*`

### Checklist

- [x] Added myself alphabetically to ``AUTHORS.rst`` (optional)
- [x] Create PR to remove functionality deprecated in this PR: see #4099

- New classes:

- Added new shortcuts:

   - [x]  In :class:`~telegram.Chat` & :class:`~telegram.User` for all methods that accept ``chat/user_id``
   - [x]  In :class:`~telegram.Message` for all methods that accept ``chat_id`` and ``message_id``
   - [x]  For new :class:`~telegram.Message` shortcuts: Added ``quote`` argument if methods accepts ``reply_to_message_id``
   - [x] In :class:`~telegram.CallbackQuery` for all methods that accept either ``chat_id`` and ``message_id`` or ``inline_message_id``

-  If relevant:

   - [x]  Add new message types to :attr:`telegram.Message.effective_attachment`
   - [x] Added new handlers for new update types
   - [x] Add the handlers to the warning loop in the :class:`~telegram.ext.ConversationHandler`
   - [x] Added new filters for new message (sub)types
   - [x] Updated the Bot API version number in all places: ``README.rst`` and ``README_RAW.rst`` (including the badge), as well as ``telegram.constants.BOT_API_VERSION_INFO``
   - [x] Added logic for arbitrary callback data in :class:`telegram.ext.ExtBot` for new methods that either accept a ``reply_markup`` in some form or have a return type that is/contains :class:`~telegram.Message`


tests/auxil/bot_method_checks.py
```diff
@@ -315,6 +315,7 @@ async def check_defaults_handling(
     defaults_no_custom_defaults = Defaults()
     kwargs = {kwarg: "custom_default" for kwarg in inspect.signature(Defaults).parameters}
     kwargs["tzinfo"] = pytz.timezone("America/New_York")
+    kwargs.pop("disable_web_page_preview")  # mutually exclusive with link_preview_options
```
harshil21, line 318: @Bibo-Joshi is this an acceptable way to pass the `check_defaults_handling` test? Asking cause I'm not really familiar with that test.

telegram/ext/_defaults.py
```diff
@@ -80,25 +94,42 @@ def __init__(
...
+            raise ValueError(
+                "`disable_web_page_preview` and `link_preview_options` are mutually exclusive."
+            )
+
+        if disable_web_page_preview is not None:
+            warn(
+                "`Defaults.disable_web_page_preview` is deprecated. Use "
+                "`Defaults.link_preview_options` instead.",
+                category=PTBDeprecationWarning,
+                stacklevel=2,
```
harshil21, line 117: reminder to self to test stacklevel

tests/auxil/bot_method_checks.py
Bibo-Joshi: as long as `disable_web_page_preview` needs defaults handling, it should be covered in some way. But I grant you that `check_defaults_handling` is a mess 🙈

harshil21:
quick observation (not related to api 7.0 but should probably be fixed before release): Docs for e.g. `constants` are showing methods like `__new__` and `__format__`

Bibo-Joshi:
> quick observation (not related to api 7.0 but should probably be fixed before release): Docs for e.g. `constants` are showing methods like `__new__` and `__format__`

https://github.com/python-telegram-bot/python-telegram-bot/pull/4002/commits/2866a0c7c77b93b6e3b90415fca495daf73cfa70, https://github.com/python-telegram-bot/python-telegram-bot/pull/4002#issuecomment-1877548925 :)

docs/source/telegram.at-tree.rst
```diff
@@ -71,21 +80,32 @@ Available Types
     telegram.messageautodeletetimerchanged
     telegram.messageentity
     telegram.messageid
+    telegram.messageorigin
+    telegram.messageoriginuser
+    telegram.messageoriginhiddenuser
+    telegram.messageoriginchat
+    telegram.messageoriginchannel
```
Bibo-Joshi, line 87: @clot please sort alphabetically

telegram/__init__.py
```diff
@@ -138,6 +146,11 @@
     "MessageAutoDeleteTimerChanged",
     "MessageEntity",
     "MessageId",
+    "MessageOrigin",
+    "MessageOriginUser",
+    "MessageOriginHiddenUser",
+    "MessageOriginChat",
+    "MessageOriginChannel",
```
Bibo-Joshi, line 153: @clot please sort alphabetically

telegram/_bot.py
```diff
@@ -828,8 +833,23 @@ async def send_message(
...
-                this message.
+                this message. Mutually exclusive with :paramref:`link_preview_options`.
+
+                .. versionchanged:: NEXT.VERSION
+                    Bot API 7.0 introduced :paramref:`link_preview_options` replacing this
+                    argument. PTB will automatically convert this argument to that one, but
+                    you should update your code to use the new argument.
+
+                .. deprecated:: NEXT.VERSION
+                    In future versions, this argument will become keyword only.
```
Bibo-Joshi, line 851:
```suggestion
                    In future versions, this argument will become a keyword-only argument.
```

telegram/_bot.py
```diff
@@ -828,8 +833,23 @@ async def send_message(
...
+                .. versionadded:: NEXT.VERSION
+
             disable_web_page_preview (:obj:`bool`, optional): Disables link previews for links in
-                this message.
+                this message. Mutually exclusive with :paramref:`link_preview_options`.
+
+                .. versionchanged:: NEXT.VERSION
+                    Bot API 7.0 introduced :paramref:`link_preview_options` replacing this
+                    argument. PTB will automatically convert this argument to that one, but
+                    you should update your code to use the new argument.
```
Bibo-Joshi, line 848: I should convert only if I want to use the advanced features, right? Since we're planning to keep dwpp for convenience?

telegram/_bot.py
```diff
@@ -848,11 +868,23 @@ async def send_message(
...
+        if not isinstance(disable_web_page_preview, DefaultValue) and not isinstance(
+            link_preview_options, DefaultValue
+        ):
+            raise ValueError(
+                "`disable_web_page_preview` and `link_preview_options` are mutually  exclusive."
+            )
+        # Convert to LinkPreviewOptions:
+        if not isinstance(disable_web_page_preview, DefaultValue):
+            link_preview_options = LinkPreviewOptions(is_disabled=disable_web_page_preview)
+
```
Bibo-Joshi, line 887: @harshil21 would a helper function be worth the while?

telegram/_bot.py
```diff
@@ -930,6 +962,46 @@ async def delete_message(
...
+    ) -> bool:
+        """
+        Use this method to delete multiple messages simultaneously. If some of the specified
+        messages can't be found, they are skipped.
+
+        .. versionadded:: NEXT.VERSION
+
+        Args:
+            chat_id (:obj:`int` | :obj:`str`): |chat_id_channel|
+            message_ids (Sequence[:obj:`int`]): Identifiers of 1-100 messages to delete.
```
Bibo-Joshi, line 985: We'll need new constants for these limits (1 & 100). Maybe we have a `constants.BulkRequestLimit` since the limits for `(delete|copy|forward)_messages` are all the same …

telegram/_bot.py
```diff
@@ -996,6 +1068,67 @@ async def forward_message(
...
+        read_timeout: ODVInput[float] = DEFAULT_NONE,
+        write_timeout: ODVInput[float] = DEFAULT_NONE,
+        connect_timeout: ODVInput[float] = DEFAULT_NONE,
+        pool_timeout: ODVInput[float] = DEFAULT_NONE,
+        api_kwargs: Optional[JSONDict] = None,
+    ) -> Tuple[MessageId, ...]:
+        """
+        Use this method to forward messages of any kind. If some of the specified messages can't be
+        found or forwarded, they are skipped. Service messages and messages with protected content
+        can't be forwarded. Album grouping is kept for forwarded messages.
```
Bibo-Joshi, line 1090: We should update our wiki entry on handling media groups :D (if done before release, this should be done in the dev-wiki)

telegram/_chat.py
```diff
@@ -2380,6 +2519,92 @@ async def copy_message(
...
+        """Shortcut for::
+
+             await bot.copy_messages(chat_id=update.effective_chat.id, *args, **kwargs)
+
+        For the documentation of the arguments, please see :meth:`telegram.Bot.copy_messages`.
+
+        .. seealso:: :meth:`copy_message`, :meth:`send_copy`, :meth:`copy_messages`.
+
+        Returns:
+            Tuple[:class:`telegram.MessageId`]: On success, returns a tuple of `MessageId` of the
```
Bibo-Joshi, line 2546:
```suggestion
            Tuple[:class:`telegram.MessageId`]: On success, returns a tuple of :class:`MessageId` of the
```

please also check in other places

telegram/_chat.py
```diff
@@ -2464,6 +2689,92 @@ async def forward_to(
             message_thread_id=message_thread_id,
         )
 
+    async def forwards_from(
```
Bibo-Joshi, line 2692:
```suggestion
    async def forward_messages_from(
```

The pular form of "forward" doesn't really stick with me TBH, some for `forwards_to` 🤔

telegram/_chat.py
```diff
@@ -2464,6 +2689,92 @@ async def forward_to(
...
+             await bot.forward_messages(chat_id=update.effective_chat.id, *args, **kwargs)
+
+        For the documentation of the arguments, please see :meth:`telegram.Bot.forward_messages`.
+
+        .. seealso:: :meth:`forward_to`, :meth:`forward_from`, :meth:`forwards_to`.
+
+        .. versionadded:: NEXT.VERSION
+
+        Returns:
+            Tuple[:class:`telegram.Message`]: On success, a tuple of ``MessageId`` of sent messages
```
Bibo-Joshi, line 2717:
```suggestion
            Tuple[:class:`telegram.MessageID`]: On success, a tuple of :class:`~telegram.MessageId` of sent messages
```
Please also check in other places

telegram/_chatboost.py
```diff
@@ -0,0 +1,408 @@
...
+    from telegram import Bot
+
+
+class ChatBoostSource(TelegramObject):
+    """
+    Base class for Telegram ChatBoostSource objects. It can be one of:
+
+    :class:`telegram.ChatBoostSourcePremium`
+    :class:`telegram.ChatBoostSourceGiftCode`
+    :class:`telegram.ChatBoostSourceGiveaway`
```
Bibo-Joshi, line 40: Can we make this an itemized list?

telegram/_chatboost.py
```diff
@@ -0,0 +1,408 @@
...
+
+    Attributes:
+        source (:obj:`str`): The source of the chat boost. Can be one of:
+            :attr:`~telegram.ChatBoostSource.PREMIUM`, :attr:`~telegram.ChatBoostSource.GIFT_CODE`,
+            or :attr:`~telegram.ChatBoostSource.GIVEAWAY`.
+    """
+
+    __slots__ = ("source",)
+
+    PREMIUM: Final[str] = "premium"  # TODO: Make it a reference to constants
```
Bibo-Joshi, line 60: leaving a dot :)

telegram/_chatboost.py
```diff
@@ -0,0 +1,408 @@
...
+    GIFT_CODE: Final[str] = "gift_code"
+    """:const:`telegram.constants.ChatBoostSources.GIFT_CODE`"""
+    GIVEAWAY: Final[str] = "giveaway"
+    """:const:`telegram.constants.ChatBoostSources.GIVEAWAY`"""
+
+    def __init__(self, source: str, *, api_kwargs: Optional[JSONDict] = None):
+        super().__init__(api_kwargs=api_kwargs)
+
+        # Required by all subclasses:
+        self.source: str = source
```
Bibo-Joshi, line 71: once the constants for the sources are implemented, please convert to enum here like done e.g. for `Chat.type` (and add corresponding test)

telegram/_chatboost.py
```diff
@@ -0,0 +1,408 @@
...
+
+class ChatBoostSourcePremium(ChatBoostSource):
+    """
+    The boost was obtained by subscribing to Telegram Premium or by gifting a Telegram Premium
+    subscription to another user.
+
+    .. versionadded:: NEXT.VERSION
+
+    Args:
+        source (:obj:`str`): The source of the chat boost. Always
```
Bibo-Joshi, line 107: no source argument needed. attribute can be removed as well, b/c documented in base class :) some for the others

telegram/_chatboost.py
```diff
@@ -0,0 +1,408 @@
...
+        source (:class:`telegram.ChatBoostSource`): Source of the added boost.
+    """
+
+    __slots__ = ("boost_id", "add_date", "expiration_date", "source")
+
+    def __init__(
+        self,
+        boost_id: str,
+        add_date: int,
+        expiration_date: int,
```
Bibo-Joshi, line 229:
```suggestion
        add_date: datetime,
        expiration_date: datetime,
```
please also update the docstring of `add_date` accordingly. the "Unix timestamp" can be removed as well

telegram/_chatboost.py
```diff
@@ -0,0 +1,408 @@
...
+    def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["ChatBoost"]:
+        """See :meth:`telegram.TelegramObject.de_json`."""
+        data = cls._parse_data(data)
+
+        if not data:
+            return None
+
+        data["source"] = ChatBoostSource.de_json(data.get("source"), bot)
+        loc_tzinfo = extract_tzinfo_from_defaults(bot)
+        data["expiration_date"] = from_timestamp(data["expiration_date"], tzinfo=loc_tzinfo)
```
Bibo-Joshi, line 254: `add_date` needs to be converted as well

telegram/_chatboost.py
```diff
@@ -0,0 +1,408 @@
...
+        source (:class:`telegram.ChatBoostSource`): Source of the removed boost.
+    """
+
+    __slots__ = ("chat", "boost_id", "remove_date", "source")
+
+    def __init__(
+        self,
+        chat: Chat,
+        boost_id: str,
+        remove_date: int,
```
Bibo-Joshi, line 336: type hint and docstring as above

telegram/_giveaway.py
```diff
@@ -0,0 +1,336 @@
...
+    """This object represents a service message about the creation of a scheduled giveaway.
+    Currently holds no information.
+    """
+
+    __slots__ = ()
+
+    def __init__(self, *, api_kwargs: Optional[JSONDict] = None) -> None:
+        super().__init__(api_kwargs=api_kwargs)
+
+        self._freeze()
```
Bibo-Joshi, line 155: can probably be removed

telegram/_giveaway.py
```diff
@@ -0,0 +1,336 @@
...
+
+        # Get the local timezone from the bot if it has defaults
+        loc_tzinfo = extract_tzinfo_from_defaults(bot)
+
+        data["chats"] = tuple(Chat.de_list(data.get("chats"), bot))
+        data["winners_selection_date"] = from_timestamp(
+            data.get("winners_selection_date"), tzinfo=loc_tzinfo
+        )
+
+        return cls(**data)
```
Bibo-Joshi, line 142: use `super().de_json` instead everywhere

telegram/_inline/inputtextmessagecontent.py
```diff
@@ -62,29 +79,63 @@ class InputTextMessageContent(InputMessageContent):
...
+            and link_preview_options is not DEFAULT_NONE
+        ):
+            raise ValueError(
+                "`disable_web_page_preview` and `link_preview_options` are mutually exclusive."
+            )
+
+        # Convert to LinkPreviewOptions:
+        if not isinstance(disable_web_page_preview, DefaultValue):
+            link_preview_options = LinkPreviewOptions(is_disabled=disable_web_page_preview)
+
```
Bibo-Joshi, line 114: can this be combined with a helper function also used in `Bot`?

telegram/_inline/inputtextmessagecontent.py
```diff
@@ -62,29 +79,63 @@ class InputTextMessageContent(InputMessageContent):
...
+        """Optional[:obj:`bool`]: Disables link previews for links in the sent message.
+
+        .. deprecated:: NEXT.VERSION
+        """
+        warn(
+            "The `disable_web_page_preview` attribute is deprecated. "
+            "Use `link_preview_options` instead.",
+            PTBDeprecationWarning,
+            stacklevel=2,
+        )
```
Bibo-Joshi, line 136: Please use the helper function from `tg._utils.warnings_transition`

Bibo-Joshi (changes requested):
First review for 31/60 files :) I gave up on tagging everyone, please check which comments apply to your changes :)

telegram/_bot.py
Bibo-Joshi, line 1: `set_message_reaction` hasn't been added yet, but while answering https://stackoverflow.com/questions/77761967 I had the idea that we could offer convenience functionality for that method:

* accept a single object instead of a list for `reaction` and convert that to a list
* accept simple strings instead of `ReactionTypeEmoji`. Then you'd have to build an object if you want to use customemoji
* Since TG explictly specifies the allowed emoji, we could even accept strings for custom emoji as well and do something like

```python
reaction = [RecationTypeEmoji(string) if string in normal_emoji else ReactionTypeCustomEmoji(string)]
```

telegram/_reply.py
```diff
@@ -0,0 +1,431 @@
...
+
+if TYPE_CHECKING:
+    from telegram import Bot
+
+
+class ExternalReplyInfo(TelegramObject):
+    """
+    This object contains information about a message that is being replied to, which may
+    come from another chat or forum topic.
+
```
Bibo-Joshi, line 55: please add note on object comparability, also for other classes

telegram/_reply.py
```diff
@@ -0,0 +1,431 @@
...
+    """
+    Describes reply parameters for the message that is being sent.
+
+    .. versionadded:: NEXT.VERSION
+
+    Args:
+        message_id (:obj:`int`): Identifier of the message that will be replied to in the current
+            chat, or in the chat :paramref:`chat_id` if it is specified.
+        chat_id (:obj:`int` | :obj:`str`, optional): If the message to be replied to is from a
+            different chat, |channel_id_channel|
```
Bibo-Joshi, line 344:
```suggestion
            different chat, |chat_id_channel|
```
?

telegram/constants.py
```diff
@@ -1754,3 +2587,50 @@ class ForumTopicLimit(IntEnum):
...
+
+class Date(Enum):
+    """This enum contains the date literal for :class:`telegram.InaccessibleMessage`
+
+    .. versionadded:: NEXT.VERSION
+    """
+
+    __slots__ = ()
+
+    ZERO_DATE = datetime.datetime(1970, 1, 1, tzinfo=UTC)
```
Bibo-Joshi, line 2600: just double checking: is this enough to make the object properly tz-aware in the pytz-sense? Otherwise we could use something like 

https://github.com/python-telegram-bot/python-telegram-bot/blob/a8f1164b0cb9bb1e26920a03ae5c88b167bd3041/tests/_utils/test_datetime.py#L58-L62

telegram/constants.py
```diff
@@ -1754,3 +2587,50 @@ class ForumTopicLimit(IntEnum):
     * :paramref:`~telegram.Bot.edit_general_forum_topic.name` parameter of
       :meth:`telegram.Bot.edit_general_forum_topic`
     """
+
+
+class Date(Enum):
+    """This enum contains the date literal for :class:`telegram.InaccessibleMessage`
+
+    .. versionadded:: NEXT.VERSION
+    """
```
Bibo-Joshi, line 2596: Having an enum with only a single member seems a bit like overkill tbh. `SUPPORTED_WEBHOOK_PORTS` & the bot api versions also still are non-enum cnostants

telegram/constants.py
```diff
@@ -1754,3 +2587,50 @@ class ForumTopicLimit(IntEnum):
...
+class ReactionEmojis(StringEnum):
+    """This enum contains the available emojis of :class:`telegram.ReactionTypeEmoji`. The enum
+    members of this enumeration are instances of :class:`str` and can be treated as such.
+
+    .. versionadded:: NEXT.VERSION
+
+    The descriptions were taken from emojipedia.org
+    """
+
+    # TODO if we want to have these descriptions, we should ask for permission from emojipedia
```
Bibo-Joshi, line 2628: Interesting! Maybe the unicode consortium also provides a description? I would assume that those are intended for public use … In any case I don't think that we'll run into trouble …

telegram/constants.py
```diff
@@ -1754,3 +2587,50 @@ class ForumTopicLimit(IntEnum):
...
+
+    __slots__ = ()
+
+    EMOJI = "emoji"
+    """:obj:`str`: A :class:`telegram.ReactionType` with a normal emoji."""
+    CUSTOM_EMOJI = "custom_emoji"
+    """:obj:`str`: A :class:`telegram.ReactionType` with a custom emoji."""
+
+
+class ReactionEmojis(StringEnum):
```
Bibo-Joshi, line 2619:
```suggestion
class ReactionEmoji(StringEnum):
```

List of members is not yet complete

Bibo-Joshi (changes requested):
10 more files reviewed, +26 comments 😄 
I must say I'm really impressed on how much code we managed to generate in such a short time and that at high quality!

telegram/_bot.py
harshil21: correct, but don't we want to recommend to use that new argument anyway?

telegram/_bot.py
harshil21: hmm so far we only have 2 instances. If another one comes up, it's worth

telegram/_chatboost.py
harshil21: yep, realized that and many changes waiting to be pushed once I figure out the tests

telegram/_chatboost.py
harshil21: yep yep did that gonna push soon

telegram/_inline/inputtextmessagecontent.py
harshil21: hmm good point, not a high priority right now let me check soon

telegram/_chat.py
clot27: Yeah seemed odd to me too, tagging pool because he suggested the names the last time we added forward/copy shortcuts. cc @Poolitzer

telegram/_messageorigin.py
```diff
@@ -0,0 +1,256 @@
...
+        date (:obj:`datetime.datetime`): Date the message was sent originally in Unix time.
+        sender_user (:class:`telegram.User`): User that sent the message originally.
+    """
+
+    __slots__ = ("sender_user",)
+
+    def __init__(
+        self,
+        date: datetime.datetime,
+        sender_user: User,
```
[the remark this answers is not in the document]
clot27: I'd prefer those in `MO.de_json()` with if statements...

telegram/_bot.py
Poolitzer: will do

telegram/_chat.py
Poolitzer: I don't mind :D. But I think a bigger difference then a s is probably good

telegram/_message.py
```diff
@@ -75,24 +81,132 @@
...
+
+    Objects of this class are comparable in terms of equality. Two objects of this class are
+    considered equal, if their :attr:`message_id` and :attr:`chat` are equal
+
+    Args:
+        message_id (:obj:`int`): Unique message identifier.
+        date (:class:`datetime.datetime`): Date the message was sent in Unix time or 0 in Unix
+            time. Converted to :class:`datetime.datetime`
+
+            |datetime_localization| (Does not apply to the 0, which will always be in UTC).
```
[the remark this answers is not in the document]
Poolitzer: I thought it be more appropriate, since not an actual time is shared, just unix set to 0.

telegram/constants.py
Poolitzer: they do not only the official names

telegram/_message.py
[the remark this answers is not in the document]
Poolitzer: Immediately depreciation sounds good to me

telegram/_bot.py
Poolitzer: and I need to update my bot again grrr

telegram/_message.py
```diff
@@ -717,7 +921,9 @@ def __init__(
         channel_chat_created: Optional[bool] = None,
         migrate_to_chat_id: Optional[int] = None,
         migrate_from_chat_id: Optional[int] = None,
-        pinned_message: Optional["Message"] = None,
+        # TODO changing this results in a couple mypy errors. Not touching them till we decide
+        # on how to continue with this
+        pinned_message: Optional[MaybeInaccessibleMessage] = None,
```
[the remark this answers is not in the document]
Poolitzer:
```
telegram/ext/_callbackdatacache.py:405: error: Argument 1 to "__process_message" of "CallbackDataCache" has incompatible type "MaybeInaccessibleMessage"; expected "Message"  [arg-type]
telegram/ext/_extbot.py:510: error: Argument 1 to "process_message" of "CallbackDataCache" has incompatible type "MaybeInaccessibleMessage"; expected "Message"  [arg-type]
telegram/ext/_extbot.py:513: error: Argument 1 to "process_message" of "CallbackDataCache" has incompatible type "MaybeInaccessibleMessage"; expected "Message"  [arg-type]

```

I get this

telegram/constants.py
Poolitzer: According to https://stackoverflow.com/a/50710825 at least.

telegram/constants.py
Poolitzer: It needs to be an enum, otherwise I can't literal it

telegram/constants.py
Poolitzer: Will wait with that till we have an official permission to use the description

telegram/_bot.py
Bibo-Joshi: I think there was one in `InputMedia*`?

telegram/_inline/inputtextmessagecontent.py
Bibo-Joshi: ah, here it is :D

telegram/_message.py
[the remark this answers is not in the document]
Bibo-Joshi: that is a valid point in time though … https://www.epochconverter.com/ automatically converts it to your timezone as well.

telegram/_message.py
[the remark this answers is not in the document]
Bibo-Joshi: ah, good points 👍 IMO the MRO is the right one as it reflects the changes in the Bot API. adding some `isinstance` checks to the exsting CDC-logic should fix the issues , I guess

telegram/_messageorigin.py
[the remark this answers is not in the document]
Bibo-Joshi: In `MO.de_json` I can see no call to `User.de_json` either 🤔  doing that there is ofc fine with me as well 👍

telegram/ext/_messagereactionhandler.py
```diff
@@ -0,0 +1,203 @@
...
+                and not self._user_usernames
+            ):
+                return True
+            # this filter must chain False, because it filters both chat and users, which wouldn't
+            # work if we return True
+            # TODO this breaks mypy and I need help with the best solution to not make it break
+            if self._chat_ids and update.effective_chat.id not in self._chat_ids:
+                return False
+            if self._chat_usernames and update.effective_chat.username not in self._chat_usernames:
+                return False
```
Bibo-Joshi, line 187: I don't quite understand . from the above docstring i would have expected the following logic:

1. determine user_id, user_username, chat_id, chat_username
2. for all 4: if entity in self._corresponding_frozenset: return True
3. return False

Bibo-Joshi (changes requested):
Yeay, I reached the tests 🥳  (and that's where I stop for today)

telegram/_messageorigin.py
```diff
@@ -0,0 +1,281 @@
...
+            cls.CHANNEL: MessageOriginChannel,
+        }
+        if cls is MessageOrigin and data.get("type") in _class_mapping:
+            return _class_mapping[data.pop("type")].de_json(data=data, bot=bot)
+
+        loc_tzinfo = extract_tzinfo_from_defaults(bot)
+        data["date"] = from_timestamp(data.get("date"), tzinfo=loc_tzinfo)
+
+        if "sender_user" in data:
+            data["sender_user"] = User.de_json(data.get("sender_user"), bot)
```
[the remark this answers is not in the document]
clot27: Dont see the point if it is neither efficient (at least nothing I saw proves it) nor shorter...

telegram/ext/_messagereactionhandler.py
Bibo-Joshi: @Poolitzer I adapted this quite a bit now. Please double check if the new logic does what you wanted it to do.

telegram/_message.py
```diff
@@ -75,24 +81,132 @@
...
+        chat (:class:`telegram.Chat`): Conversation the message belongs to.
+    """
+
+    __slots__ = ("chat", "message_id", "date")
+
+    def __init__(
+        self,
+        chat: Chat,
+        message_id: int,
+        # TODO Should this also accept the literal
```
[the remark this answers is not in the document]
Bibo-Joshi: https://peps.python.org/pep-0586/#legal-parameters-for-literal-at-type-check-time

`Literal` won't work with dtms

Bibo-Joshi:
I added a warning to `update.effective_message` saying that this property will return a `MaybeAccessibleMessage` in the future instead of `Message`. I did this because:

* `update.callback_query.message` now has this type
* `update.effective_message` can already be `None`, so users that want their static type checker & linter to be happy already have to apply some casting/type-ignoring/isinstance-checks. Hence, this won't change very much for them.

For now, `update.callback_query.message` is returned only if that really is a `Message` object. Please make yourself noticed if you disagree :)

telegram/constants.py
Bibo-Joshi: The [CLDR names](https://cldr.unicode.org/index) seem to be the names that are listed on emojipedia as "unicode name". The data can be downloaded [here](https://cldr.unicode.org/index/downloads), which points to https://github.com/unicode-org/cldr, which has a [do-whatever-you-want license](https://github.com/unicode-org/cldr/blob/main/LICENSE).

The cldr names are also listed on https://www.unicode.org/emoji/charts/full-emoji-list.html

Bibo-Joshi:
I had some more insights in effective_message.

I ran a simple example bot and noticed that we use effective_message at several internal places which all issue warnings now. Muting them would be quite some effort and I would say that it's okay to bother the users with them for a few versions if we don't find an elegant solution.

This made me realize that if you use effective_message, you most likely want to do something with it. That made me see that all the Message.reply_* functions can in theory work on InaccessibleMessage as well as chat_id and message_id are given. The question is, if bots can reply to or forward inaccessible messages - I would doubt it, but will double check. If this is not possible, then there is actually rather few things that one can do with these objects and in this case I'll revert my position and say that effective_message should only return accessible messages.

telegram/ext/_extbot.py
```diff
@@ -427,6 +431,29 @@ def _insert_defaults(self, data: Dict[str, object]) -> None:
...
+            # 4)
+            # We need handle the LinkPreviewOptions case:
+            # If Defaults.LPO is set, and LPO is passed in the bot method we should fuse
+            # them, giving precedence to passed values.
+            # Defaults.LPO(True, "google.com", True) & # LPO=LPO(True, ..., False) ->
+            # LPO(True, "google.com", False)
+            elif (
+                isinstance(val, LinkPreviewOptions)
+                and self.defaults
+                and (passed_lpo := data.get(key, False)) is not DEFAULT_NONE
```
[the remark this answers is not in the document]
harshil21: I just wanted to check if something was passed, and since we don't do runtime type checking I thought this was okay. Can change it to `val`.

telegram/ext/_defaults.py
```diff
@@ -178,11 +209,11 @@ def disable_notification(self, value: object) -> NoReturn:
         )
 
     @property
-    def disable_web_page_preview(self) -> Optional[bool]:
-        """:obj:`bool`: Optional. Disables link previews for links in this
-        message.
+    def disable_web_page_preview(self) -> ODVInput[bool]:
+        """:obj:`bool`: Optional. Disables link previews for links in all outgoing
+        messages.
```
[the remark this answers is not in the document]
harshil21: your point of needing to build `Defaults` only once makes me want to deprecate DWPP. I'll add the depr note to the attr docstring but I don't think the attribute access warning is needed since we already warn when they build the Defaults object (in the `__init__`)

harshil21:
> Please make yourself noticed if you disagree :)

I do feel that `effective_message` should always be a "fire and forget" kind of property. It defeats the purpose of the property if the user again has to write more code to get the actual underlying `Message`. So changing that for the small possibility of having an `InaccessibleMessage` isn't worth refactoring codebases to use additional checks imo.

What do you think about raising a warning in `effective_message` when `update.callback_query.message` is an `InaccessibleMessage`? So only those who want to handle InaccessibleMessages separately, they now can, meanwhile for the rest everything stays the same.

Bibo-Joshi:
> What do you think about raising a warning in `effective_message` when `update.callback_query.message` is an `InaccessibleMessage`? So only those who want to handle InaccessibleMessages separately, they now can, meanwhile for the rest everything stays the same.

Sounds sane 👍 will do

telegram/_callbackquery.py
```diff
@@ -238,10 +253,13 @@ async def edit_message_text(
                 chat_id=None,
                 message_id=None,
             )
+        if not isinstance(self.message, Message):
```
harshil21, line 256: why not have a method like `get_message()` (akin to get_bot()) which raises an error if the message is Inaccessible?

harshil21 (commented):
will slowly review more this weekend

telegram/_callbackquery.py
Bibo-Joshi: for internal use that's a nice idea, I agree 👍 I wouldn't make it public though, b/c there are three different cases (Message, InaccessibleMessage, None) and depending on the use case people might be expecting different behavior …

telegram/_message.py
```diff
@@ -1539,10 +1547,113 @@ def _quote(self, quote: Optional[bool], reply_to_message_id: Optional[int]) -> O
...
+
+        effective_index = 0 or index
+
+        matches = list(re.finditer(re.escape(quote), text))
+        if (length := len(matches)) < effective_index:
+            raise ValueError(
+                f"You requested the {index}-nth occurrence of '{quote}', but this text appears "
+                f"only {length} times."
+            )
+        # apply utf-16 conversion
```
Bibo-Joshi, line 1568: todo

telegram/_message.py
```diff
@@ -1539,10 +1547,113 @@ def _quote(self, quote: Optional[bool], reply_to_message_id: Optional[int]) -> O
...
                 default_quote = None
             if (default_quote is None and self.chat.type != Chat.PRIVATE) or default_quote:
-                return self.message_id
+                return ReplyParameters(self.message_id)
 
         return None
 
+    def compute_quote_position_and_entities(
+        self, quote: str, index: Optional[int] = None
+    ) -> Tuple[int, Tuple[MessageEntity, ...]]:
```
Bibo-Joshi, line 1556: needs a docstring. This should also point out that `quote` is expected to plain text without formatting entities as explained in https://github.com/python-telegram-bot/python-telegram-bot/issues/4033#issuecomment-1874358806 as well as list the `{Runtime, Value}Error` with reasons

telegram/_message.py
```diff
@@ -1539,10 +1547,113 @@ def _quote(self, quote: Optional[bool], reply_to_message_id: Optional[int]) -> O
...
+            .. versionadded:: NEXT.VERSION
+
+        Args:
+            quote (:obj:`str`, optional): Quoted part of the message to be replied to; 0-1024
+                characters after entities parsing. The quote must be an exact substring of the
+                message to be replied to, including bold, italic, underline, strikethrough,
+                spoiler, and custom_emoji entities. The message will fail to send if the quote
+                isn't found in the original message.
+            quote_index (:obj:`int`, optional): Position of the quote in the original message in
+                UTF-16 code units.
```
Bibo-Joshi, line 1598: No, these arguments will be passed to `compute_quote_position_and_entities` and the point is that `compute_quote_position_and_entities` returns the entities and quote position such that the user doesn't have to include the formatting entities in `quote` and doesn't have to comute the utf-16 position ;)

telegram/_message.py
```diff
@@ -1539,10 +1547,113 @@ def _quote(self, quote: Optional[bool], reply_to_message_id: Optional[int]) -> O
...
+            target_chat_id (:obj:`int` | :obj:`str`, optional): If the message to be replied to is
+                from a different chat, |chat_id_channel|
```
Bibo-Joshi, line 1600: No, this is the `target_chat_id` is the id of the chat that you want to send the message to. This is different from `ReplyParameters.chat_id` (where you got this docstring from)

telegram/_message.py
```diff
@@ -1539,10 +1547,113 @@ def _quote(self, quote: Optional[bool], reply_to_message_id: Optional[int]) -> O
...
+            allow_sending_without_reply (:obj:`bool`, optional): |allow_sending_without_reply|
+            message_thread_id (:obj:`int`, optional): |message_thread_id|
+
+        Returns:
+            :obj:`dict`: On success, a dict containing information about ``reply_parameters`` is
+            returned.
+        """
+        kwargs: _ReplyKwargs = {
+            "chat_id": target_chat_id or self.chat_id
+        }  # type: ignore [typeddict-item]
```
Bibo-Joshi, line 1610: if it helps with mypy, you can also construct `kwargs` at one place at the end of the method

telegram/_message.py
```diff
@@ -1572,23 +1683,26 @@ async def reply_text(
         For the documentation of the arguments, please see :meth:`telegram.Bot.send_message`.
 
         Keyword Args:
-            quote (:obj:`bool`, optional): If set to :obj:`True`, the message is sent as an actual
-                reply to this message. If ``reply_to_message_id`` is passed, this parameter will be
-                ignored. Default: :obj:`True` in group chats and :obj:`False` in private chats.
+            quote (:obj:`bool`, optional): |reply_quote|
```
Bibo-Joshi, line 1686: please add deprecation notice, both as warning (probably in `_parse_quote_argumets`) and as docstring

telegram/_message.py
```diff
@@ -1539,10 +1547,113 @@ def _quote(self, quote: Optional[bool], reply_to_message_id: Optional[int]) -> O
...
+        return kwargs
+
+    async def _parse_quote_arguments(
+        self,
+        do_quote: Optional[Union[bool, _ReplyKwargs]],
+        quote: Optional[bool],
+        reply_to_message_id: Optional[int],
+    ) -> Tuple[Union[str, int], ReplyParameters]:
+        if quote and do_quote:
+            raise ValueError("Mutually exclusive")
```
Bibo-Joshi, line 1642:
```suggestion
            raise ValueError("The arguments `quote` and `do_quote` are mutually exclusive")
```

tests/test_message.py
```diff
@@ -70,11 +70,7 @@
 from telegram.ext import Defaults
 from telegram.warnings import PTBDeprecationWarning
 from tests._passport.test_passport import RAW_PASSPORT_DATA
-from tests.auxil.bot_method_checks import (
-    check_defaults_handling,
-    check_shortcut_call,
-    check_shortcut_signature,
-)
```
Bibo-Joshi, line 68: I left `check_defaults_handling` in here on purpose - I just commented out the usages so that we can see the code coverage reports for once ;) Please re-add and un-comment the usages

docs/substitutions/global.rst
```diff
@@ -66,8 +66,10 @@
...
 
-.. |rtm_aswr_deprecated| replace:: Bot API 7.0 introduced :paramref:`~telegram.Bot.send_message.reply_parameters` replacing this argument. PTB will automatically convert this argument to that one, but you should update your code to use the new argument.
+.. |rtm_aswr_deprecated| replace:: replacing this argument. PTB will automatically convert this argument to that one, but you should update your code to use the new argument.
 
 .. |keyword_only_arg| replace:: In future versions, this argument will become a keyword-only argument.
 
-.. |mutually_exc_rp| replace:: Mutually exclusive with :paramref:`~telegram.Bot.send_message.reply_parameters`.
+.. |reply_quote| replace:: If set to :obj:`True`, the message is sent as an actual reply to this message. If ``reply_to_message_id`` is passed, this parameter will be ignored. Default: :obj:`True` in group chats and :obj:`False` in private chats.
+
+.. |do_quote| replace:: If set to :obj:`True`, the replied message is quoted. For a dict, it must be the output of :meth:`~telegram.Message.build_reply_arguments` to specify exact ``reply_parameters``.
```
Bibo-Joshi, line 75: please add part

"If ``reply_to_message_id`` is passed, this parameter will be ignored. Default: :obj:`True` in group chats and :obj:`False` in private chats."

this is still relevant for the new argument

Bibo-Joshi (changes requested):
Thanks for the commit @clot27 . Unfortunately you didn't address the things that were still missing in my proof of concept ;)

Apart from my comments, `Defaults.quote` needs to be depercated in favor of a new `Defaults.do_quote`. And ofc we'll need tests. For the `compute_quote_position_and_entities` method, I'd like to have tests with real requests (i.e. not just monkey patched).

Please reach out if you'd like some help with these todos :)

telegram/_message.py
clot27: Aha, sorry for the confusion. Wasnt following the shortcut discussion closesly 🥶

telegram/ext/_defaults.py
Bibo-Joshi: is this still relevant?

telegram/_message.py
```diff
@@ -1556,6 +1556,23 @@ def _quote(
...
         self, quote: str, index: Optional[int] = None
     ) -> Tuple[int, Tuple[MessageEntity, ...]]:
+        """
+        Use this function to compute position and entities of quote.
+
+        Args:
+            quote (:obj:`str`): Part of the message which is to be quoted. This is
+                expected to have plain text without formatting entities.
+            index (:obj:`int`, optional): Position of the quote in the original message in
+                UTF-16 code units.
```
Bibo-Joshi, line 1566: This one is still not right. What we want here is the number of the appearance of the text. E.g. if the message has the text

> ABC DEF ABC DEF

calling `compute_quote_position("ABC")` or `compute_quote_position("ABC", 0)` should quote the first "ABC", but calling `compute_quote_position("ABC", 1)` should quote the second "ABC" (standard zero-based indexing)

the UTF-16 offset will have to be computed and is part of the output. Computing this for the user is an important part of this convenience functionality b/c emojis & stuff are not quite as easy to handle in utf-16. that's also why the "apply utf-16 conversion" is still open below ;) have a look at `_parse_html/md_entities`, where utf-16 handling is used as well

clot27: Hmm should we also add examples regarding the that to docstrings?

Bibo-Joshi:
I'll ignore deesource - the code of `Message._parse_markdown` wasn't changed much, mostly indentation

telegram/ext/filters.py
```diff
@@ -1848,31 +1886,35 @@ class _All(UpdateFilter):
...
                 or StatusUpdate.DELETE_CHAT_PHOTO.check_update(update)
-                or StatusUpdate.CHAT_CREATED.check_update(update)
+                or StatusUpdate.FORUM_TOPIC_CLOSED.check_update(update)
+                or StatusUpdate.FORUM_TOPIC_CREATED.check_update(update)
+                or StatusUpdate.FORUM_TOPIC_EDITED.check_update(update)
+                or StatusUpdate.FORUM_TOPIC_REOPENED.check_update(update)
+                or StatusUpdate.GENERAL_FORUM_TOPIC_HIDDEN.check_update(update)
+                or StatusUpdate.GENERAL_FORUM_TOPIC_UNHIDDEN.check_update(update)
+                or StatusUpdate.GIVEAWAY_COMPLETED.check_update(update)
+                or StatusUpdate.GIVEAWAY_CREATED.check_update(update)
```
harshil21, line 1901: shouldn't GIVEAWAY_WINNERS also be included in this?

PR #4038, opened 2023-12-30
`blockquote` message entity api-7.0.

### LINKED ISSUE #4033: [FEATURE] API 7.0

aelkheir:
Closes #4033

### Scope: Block Quotation

- [x] Added support for “blockquote” entities in received messages. @aelkheir
- [x]  Added support for “blockquote” entity parsing in “MarkdownV2” and “HTML” parse modes @aelkheir
- [x]  Allowed to explicitly specify “blockquote” entities in formatted texts. @aelkheir

### Check-list for PRs

- [x]  Documented code changes according to the `CSI standard <https://standards.mousepawmedia.com/en/stable/csi.html>`__
- [ ] Added new classes & modules to the docs and all suitable ``__all__`` s

- If relevant:

### DIFF
--- a/telegram/_message.py
+++ b/telegram/_message.py
@@ -3698,7 +3698,10 @@ class Message(TelegramObject):
                         raise ValueError(
                             "Blockquote entities are not supported for Markdown version 1"
                         )
-                    insert = f">{escaped_text}"
+                    prefix = ">" if entity.offset == 0 else "\n>"
+                    message_length = len(str(message_text, encoding="utf-16-le"))  # type: ignore
+                    suffix = "" if entity.offset + entity.length - 1 == message_length else "\n"
+                    insert = prefix + escaped_text.replace("\n", "\n>") + suffix
                 elif entity.type == MessageEntity.CUSTOM_EMOJI:
                     if version == 1:
                         raise ValueError(
--- a/telegram/_message.py
+++ b/telegram/_message.py
@@ -3563,8 +3563,7 @@ class Message(MaybeInaccessibleMessage):
         if message_text is None:
             return None
 
-        message_text = message_text.encode("utf-16-le")  # type: ignore
-
+        utf_16_text = message_text.encode("utf-16-le")
         html_text = ""
         last_offset = 0
 
@@ -3572,83 +3571,69 @@ class Message(MaybeInaccessibleMessage):
         parsed_entities = []
 
         for entity, text in sorted_entities:
-            if entity not in parsed_entities:
-                nested_entities = {
-                    e: t
-                    for (e, t) in sorted_entities
-                    if e.offset >= entity.offset
-                    and e.offset + e.length <= entity.offset + entity.length
-                    and e != entity
-                }
-                parsed_entities.extend(list(nested_entities.keys()))
+            if entity in parsed_entities:
+                continue
 
-                orig_text = text
+            nested_entities = {
+                e: t
+                for (e, t) in sorted_entities
+                if e.offset >= entity.offset
+                and e.offset + e.length <= entity.offset + entity.length
+                and e != entity
+            }
+            parsed_entities.extend(list(nested_entities.keys()))
+
+            if nested_entities:
+                escaped_text = Message._parse_html(
+                    text, nested_entities, urled=urled, offset=entity.offset
+                )
+            else:
                 escaped_text = escape(text)
 
-                if nested_entities:
-                    escaped_text = Message._parse_html(
-                        orig_text, nested_entities, urled=urled, offset=entity.offset
-                    )
-
-                if entity.type == MessageEntity.TEXT_LINK:
-                    insert = f'<a href="{entity.url}">{escaped_text}</a>'
-                elif entity.type == MessageEntity.TEXT_MENTION and entity.user:
-                    insert = f'<a href="tg://user?id={entity.user.id}">{escaped_text}</a>'
-                elif entity.type == MessageEntity.URL and urled:
-                    insert = f'<a href="{escaped_text}">{escaped_text}</a>'
-                elif entity.type == MessageEntity.BLOCKQUOTE:
-                    insert = f"<blockquote>{escaped_text}</blockquote>"
-                elif entity.type == MessageEntity.BOLD:
-                    insert = f"<b>{escaped_text}</b>"
-                elif entity.type == MessageEntity.ITALIC:
-                    insert = f"<i>{escaped_text}</i>"
-                elif entity.type == MessageEntity.CODE:
-                    insert = f"<code>{escaped_text}</code>"
-                elif entity.type == MessageEntity.PRE:
-                    if entity.language:
-                        insert = (
-                            f'<pre><code class="{entity.language}">{escaped_text}</code></pre>'
-                        )
-                    else:
-                        insert = f"<pre>{escaped_text}</pre>"
-                elif entity.type == MessageEntity.UNDERLINE:
-                    insert = f"<u>{escaped_text}</u>"
-                elif entity.type == MessageEntity.STRIKETHROUGH:
-                    insert = f"<s>{escaped_text}</s>"
-                elif entity.type == MessageEntity.SPOILER:
-                    insert = f'<span class="tg-spoiler">{escaped_text}</span>'
-                elif entity.type == MessageEntity.CUSTOM_EMOJI:
-                    insert = (
-                        f'<tg-emoji emoji-id="{entity.custom_emoji_id}">{escaped_text}</tg-emoji>'
-                    )
+            if entity.type == MessageEntity.TEXT_LINK:
+                insert = f'<a href="{entity.url}">{escaped_text}</a>'
+            elif entity.type == MessageEntity.TEXT_MENTION and entity.user:
+                insert = f'<a href="tg://user?id={entity.user.id}">{escaped_text}</a>'
+            elif entity.type == MessageEntity.URL and urled:
+                insert = f'<a href="{escaped_text}">{escaped_text}</a>'
+            elif entity.type == MessageEntity.BLOCKQUOTE:
+                insert = f"<blockquote>{escaped_text}</blockquote>"
+            elif entity.type == MessageEntity.BOLD:
+                insert = f"<b>{escaped_text}</b>"
+            elif entity.type == MessageEntity.ITALIC:
+                insert = f"<i>{escaped_text}</i>"
+            elif entity.type == MessageEntity.CODE:
+                insert = f"<code>{escaped_text}</code>"
+            elif entity.type == MessageEntity.PRE:
+                if entity.language:
+                    insert = f'<pre><code class="{entity.language}">{escaped_text}</code></pre>'
                 else:
-                    insert = escaped_text
+                    insert = f"<pre>{escaped_text}</pre>"
+            elif entity.type == MessageEntity.UNDERLINE:
+                insert = f"<u>{escaped_text}</u>"
+            elif entity.type == MessageEntity.STRIKETHROUGH:
+                insert = f"<s>{escaped_text}</s>"
+            elif entity.type == MessageEntity.SPOILER:
+                insert = f'<span class="tg-spoiler">{escaped_text}</span>'
+            elif entity.type == MessageEntity.CUSTOM_EMOJI:
+                insert = f'<tg-emoji emoji-id="{entity.custom_emoji_id}">{escaped_text}</tg-emoji>'
+            else:
+                insert = escaped_text
 
-                if offset == 0:
-                    html_text += (
-                        escape(
-                            message_text[  # type: ignore
-                                last_offset * 2 : (entity.offset - offset) * 2
-                            ].decode("utf-16-le")
-                        )
-                        + insert
-                    )
-                else:
-                    html_text += (
-                        message_text[  # type: ignore
-                            last_offset * 2 : (entity.offset - offset) * 2
-                        ].decode("utf-16-le")
-                        + insert
-                    )
-
-                last_offset = entity.offset - offset + entity.length
-
-        if offset == 0:
-            html_text += escape(
-                message_text[last_offset * 2 :].decode("utf-16-le")  # type: ignore
+            # Make sure to escape the text that is not part of the entity
+            # if we're in a nested entity, this is still required, since in that case this
+            # text is part of the parent entity
+            html_text += (
+                escape(
+                    utf_16_text[last_offset * 2 : (entity.offset - offset) * 2].decode("utf-16-le")
+                )
+                + insert
             )
-        else:
-            html_text += message_text[last_offset * 2 :].decode("utf-16-le")  # type: ignore
+
+            last_offset = entity.offset - offset + entity.length
+
+        # see comment above
+        html_text += escape(utf_16_text[last_offset * 2 :].decode("utf-16-le"))  # type: ignor
 
         return html_text
 
@@ -3748,13 +3733,22 @@ class Message(MaybeInaccessibleMessage):
         version: MarkdownVersion = 1,
         offset: int = 0,
     ) -> Optional[str]:
-        version = int(version)  # type: ignore
+        if version == 1:
+            for entity_type in (
+                MessageEntity.UNDERLINE,
+                MessageEntity.STRIKETHROUGH,
+                MessageEntity.SPOILER,
+                MessageEntity.BLOCKQUOTE,
+                MessageEntity.CUSTOM_EMOJI,
+            ):
+                if any(entity.type == entity_type for entity in entities):
+                    name = entity_type.name.title().replace("_", " ")  # type:ignore[attr-defined]
+                    raise ValueError(f"{name} entities are not supported for Markdown version 1")
 
         if message_text is None:
             return None
 
-        message_text = message_text.encode("utf-16-le")  # type: ignore
-
+        utf_16_text = message_text.encode("utf-16-le")
         markdown_text = ""
         last_offset = 0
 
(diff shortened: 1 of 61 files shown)


Bibo-Joshi (changes requested):
Thanks for the PR! The changes so far look good. Can you please also update `Message.(text|caption)_(html|markdown).*` and the corresponding tests?

aelkheir:
Roger!

telegram/_message.py
```diff
@@ -663,6 +663,10 @@ class Message(TelegramObject):
     .. |custom_emoji_no_md1_support| replace:: Since custom emoji entities are not supported by
        :attr:`~telegram.constants.ParseMode.MARKDOWN`, this method now raises a
        :exc:`ValueError` when encountering a custom emoji.
+
+    .. |blockquote_no_md1_support| replace:: Since block quotation entities are not supported
+       by :attr:`~telegram.constants.ParseMode.MARKDOWN`, this method now raises a
+       :exc:`ValueError` when encountering a block quotation.
```
Bibo-Joshi, line 669: This sounds sane to me, since there is no kind of fallback option. Just for completeness I want to mention that for the custom emojis (where we did have a fallback option), we did not raise an exception immediately, see https://github.com/python-telegram-bot/python-telegram-bot/pull/3673/files#diff-a9dec124bab0bbe6efcdb93c525ae03d28c86809642c8550d2057b19db8d7bf3. cc-ing @harshil21 @clot so that other dev team members can give their opinion as well

telegram/_message.py
```diff
@@ -3675,6 +3693,12 @@ def _parse_markdown(
                             "Spoiler entities are not supported for Markdown version 1"
                         )
                     insert = f"||{escaped_text}||"
+                elif entity.type == MessageEntity.BLOCKQUOTE:
+                    if version == 1:
+                        raise ValueError(
+                            "Blockquote entities are not supported for Markdown version 1"
+                        )
+                    insert = f">{escaped_text}"
```
Bibo-Joshi, line 3701: needs multiline handling as already being discussed in https://t.me/pythontelegrambotdev/965. Please update the tests to cover these cases :)

Bibo-Joshi (changes requested):
Thanks for the updates :) just two comments

telegram/_message.py
aelkheir: Sorry for the late reply.
So, I came up with this, which I think would fit most cases

```python
insert = "\n>" + escaped_text.replace("\n", "\n>") + "\n"
```
I'm assuming that the output of `_parse_(html, markdown*)` should produce the same visual structure as the original message if/when used like `message.reply_text(message.text_markdown/html)`

these are the reasons for the extra `\n`s
- each quote line/paragraph, including the first, must be in its own new line. So trying to send `">line 1>line2"` or `"abc >def"` as text with `pars_mode=mdv2`, will raise `telegram.error.BadRequest: Can't parse entities: character '>' ...`
- the parsed quote must be separated from the following entities/text with a line break . This is to prevent the last line of the quote from including the rest of the message text as part of the quote.

Issues am facing is when the message starts (or ends) with a blockquote entity. There will be an extra unneeded blank lines.
I guess I can trim them at `message.(text/caption)_(html/markdown)*`, but don't like the repetition. would appreciate any ideas, or even a completely different approach

telegram/_message.py
harshil21: Agree with that, it is called as version "1" for a reason, raising the error is the way to go specially after we did that warning also

telegram/_message.py
Bibo-Joshi: Not to worry. Thanks for your elaborations! I haven't looked into it deeply yet, but a spontanuous idea would be that one should be able to use `MessageEntity.{offset, length}` to determine if the quoted text is at the start/end of the message. Also I'm wondering if `escaped_text.replace("\n", "\n>")` works as intended if the quote-text itself contains "\n", i.e. something like

> This is \n a quote

🤔

telegram/_message.py
clot27: Agree!

telegram/_message.py
aelkheir: > one should be able to use `MessageEntity.{offset, length} to ...`

Oh nice, problem solved then.

> Also I'm wondering if `escaped_text.replace("\n", "\n>")` works as intended

Hmm, apparently it does. Telegram escapes `\n`s when its meant literally. and surprisingly `string.replace("\n", "...")` will not match escaped `\\n`s

```bash
$ python
Python 3.11.4 ...
>>> """line 1
... line \\n 2
... line 3""".replace("\n", '  **  ')
'line 1  **  line \\n 2  **  line 3'
>>>
```

Bibo-Joshi: okay, that's good, then :) adding this case to the tests would still be sane IMO :D

Bibo-Joshi (commented):
Thanks for the updates! I did some manual testing and unfortunately it seems that the situation is even more involved than I though :/ My findings so far:

* inserting `\n` everywhere introuduces unwanted linebreaks
* in markdown `>` must always be the first character in the line to work for the quote block. Hence, whether or not we need to insert an `\n` depends on the text *before* the entity. E.g. `"ABC<blockquote>DEF</blockquote>"` works fine, but `"ABC>DEF"` does not - we'd have to use `"ABC\n>DEF"`. This renders in the same way, but the text of the message is in fact different (by a `\n`)
* if the block quote also contains a nested entity, then something in the handling for that un-escapes the line breaks leading to `\n` in the `text_md_v2` instead of the original `\\n`
* One thing that MD apparently just can not represent is 2 block quotes following each other directly without blank line. `"<blockquote>ABC</blockquote><blockquote>DEF</blockquote>"` works for HTML, but `">ABC\n>DEF" renders as one two-line quote, while `">ABC\n\n>DEF"` rendors as two one-line quotes separated by a blank line. This also needs checkinf if two block quote entities are following each other directly and TBH af first glance I think the most sane thing to do is to raise an exception and not produce any output.

I'm not sure how we can proceed here. But then again, it's late and I might see more clearly tomorrow …

For reference: I tried to test by checking if sending a message with the produced `text_*` has the same message entities & text as the original message. MWE:

<details>

```python
from telegram import Message, Update
from telegram.ext import CallbackContext, Application, MessageHandler, filters


async def handle(update: Update, context: CallbackContext):
    message = update.effective_message

    print(repr(message.text))
    print(repr(message.entities))
    print(repr(message.text_html))
    print(repr(message.text_markdown_v2))

    reply_md = await message.reply_markdown_v2(message.text_markdown_v2)
    reply_html = await message.reply_html(message.text_html)
    for reply, name in ((reply_md, "markdown"), (reply_html, "HTML")):
        print("-" * 80)
        print(f"Replying with {name}:")
        print(reply.text == message.text, repr(reply.text), repr(message.text), sep="\n")
        print(
            reply.entities == message.entities,
            repr(reply.entities),
            repr(message.entities),
            sep="\n",
        )
        print(
            reply.text_html == message.text_html,
            repr(reply.text_html),
            repr(message.text_html),
            sep="\n",
        )
        print(
            reply.text_markdown_v2 == message.text_markdown_v2,
            repr(reply.text_markdown_v2),
            repr(message.text_markdown_v2),
            sep="\n",
        )


if __name__ == "__main__":
    application = (
        Application.builder()
        .token("TOKEN")
        .build()
    )
    application.add_handler(MessageHandler(filters.ALL, handle))
    application.run_polling()

```

</details>

Test messages that I used:

image
image

Bibo-Joshi:
I dug around more, ended up refactoring a bit of code took the liberty to push my changes to your branch. Managed to fix the nested entities stuff (and I'm wondering why the `if offset == 0` never caused problems so far).

What's still missing is throwing an exception for two consequcitve block quotes in MarkdownV2.

Moreover, even though block quote parsing seems to be working correctly now, I think there is a bug on TG side. Sending a text of the form ">A\nB" leads to TG detecting a Message entity of length 2 instead of 1, which includes the "\n". You can reproduce this via

https://api.telegram.org/botTOKEN/sendMessage?chat_id=CHAT_ID&text=%3EA%0AB&parse_mode=MarkdownV2

If you try to compute the `text_md_v2` for the resulting message, the B will be moved into the block quote.

For completeness, let me also share my renewed testing script:


<details>

```python
import asyncio
import json
from pathlib import Path
from pprint import pprint

from telegram import Message, Update, Bot
from telegram.constants import ParseMode
from telegram.ext import CallbackContext, Application, MessageHandler, filters


def get_message(bot: Bot) -> Message:
    return Update.de_json(json.loads(Path("response_3.json").read_bytes()), bot).effective_message


async def main():
    async with Bot("TOKEN") as bot:
        message = get_message(bot)
        await run_tests(message)


def get_formatted_text(message: Message, parse_mode: ParseMode) -> str:
    return getattr(message, f"text_{parse_mode.name.lower()}")


async def run_tests(message: Message) -> None:
    print(repr(message.text))
    pprint(repr(message.entities))

    for parse_mode in (ParseMode.HTML, ParseMode.MARKDOWN_V2):
        print("-" * 80)
        print(f"Replying with {parse_mode}:")
        await message.get_bot().send_message(
            chat_id=1145108092,
            text=f"============ Replying with {parse_mode}:",
        )
        formatted_text = get_formatted_text(message, parse_mode)
        print(repr(formatted_text))
        reply = await message.get_bot().send_message(
            chat_id=1145108092,
            text=formatted_text,
            parse_mode=parse_mode,
        )
        print(reply.text == message.text, repr(reply.text), repr(message.text), sep="\n")
        print(
            reply.entities == message.entities,
            repr(reply.entities),
            repr(message.entities),
            sep="\n",
        )
        pprint(reply.parse_entities())
        print(
            reply.text_html == message.text_html,
            repr(reply.text_html),
            repr(message.text_html),
            sep="\n",
        )
        print(
            reply.text_markdown_v2 == message.text_markdown_v2,
            (reply.text_markdown_v2),
            (message.text_markdown_v2),
            sep="\n====\n",
        )
        next_reply = await reply.reply_text(
            get_formatted_text(reply, parse_mode), parse_mode=parse_mode, quote=True
        )
        await next_reply.reply_text(
            get_formatted_text(next_reply, parse_mode), parse_mode=parse_mode, quote=True
        )


if __name__ == "__main__":
    asyncio.run(main())
```

response.json:

```json
{
    "update_id": 219020534,
    "message": {
        "message_id": 31843,
        "from": {
            "id": 1145108092,
            "is_bot": false,
            "first_name": "Hinrich",
            "last_name": "Mahler (@Bibo-Joshi)",
            "username": "BiboJoshi",
            "language_code": "de"
        },
        "chat": {
            "id": 1145108092,
            "first_name": "Hinrich",
            "last_name": "Mahler (@Bibo-Joshi)",
            "username": "BiboJoshi",
            "type": "private"
        },
        "date": 1704553973,
        "text": "block \\n quote\nABC\nblock \\n quote\nwith \\n line break\nDEF\nblock \\n quote\nwith \\n line breakp",
        "entities": [
            {
                "offset": 0,
                "length": 14,
                "type": "blockquote"
            },
            {
                "offset": 9,
                "length": 5,
                "type": "underline"
            },
            {
                "offset": 19,
                "length": 33,
                "type": "blockquote"
            },
            {
                "offset": 28,
                "length": 5,
                "type": "italic"
            },
            {
                "offset": 28,
                "length": 5,
                "type": "strikethrough"
            },
            {
                "offset": 57,
                "length": 34,
                "type": "blockquote"
            },
            {
                "offset": 66,
                "length": 5,
                "type": "spoiler"
            }
        ]
    }
}
```

</details>

Bibo-Joshi:
Ah, what I forgot is handling of cases that result from things like `ABC<blockquote>DEF</blockquote>GHI`, where we need to insert newline characters.

For the potential bug in TG sidr mentioned above, I guess we can try to work around it but checking if the text if the block quote ends on a newline character. Funny things may still happen if the quote is located at the end of the message 🤔 in any case I think its worth pointing oit this behavior to TG.

In fact one could also point them to the first past of this comment and ask them to unify the handling of these cases, always breaking lines before and after block quotes ...

Bibo-Joshi:
I opened https://github.com/tdlib/telegram-bot-api/issues/515

Bibo-Joshi:
A first reply on https://github.com/tdlib/telegram-bot-api/issues/515 is there. Punshlines:

* We're discuraged from using mechanism like `text_hmtl/markdown` that convert `MessageEntities` back into markup syntax and the newline behavior 
* code blocks and block quotes start start & end of new lines, but in the end the rendering is client side decision

While somewhat expected, this reply is still a bummer for me and I see the need for an internal discussion on how to proceed with `text_html/markdown`.

For this PR, I will revert to the stage that "just" added the `MessageEntity` extensions and open a new WIP-PR with the current state for reference & further discussion. Sorry that my request to implement `text_html/markdown` took this turn @aelkheir

aelkheir:
No problem! I'd be happy to help with the outcome solution of the discussions. As always, thank you.

PR #4039, opened 2023-12-31
Fixed a small problem for a wrong type hint of the filters parameter.

Palaptin:
Made the type Optional, since None is also allowed for the parameter filters.


Bibo-Joshi:
Thanks for the contirbution!

PR #4051, opened 2024-01-03
Use Recommended `pre-commit` Mirror for `black`

Bibo-Joshi:
See

* https://github.com/psf/black/blob/main/.pre-commit-hooks.yaml#L1-L2
* https://black.readthedocs.io/en/stable/integrations/source_version_control.html
* https://github.com/psf/black-pre-commit-mirror

Most importantly, note that

> mypyc wheels go brr

😄

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
     autoupdate_schedule: monthly
 
 repos:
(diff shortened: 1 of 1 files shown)


Poolitzer:
If they didn't go brr we wouldn't be here, so that is good :D

PR #4052, opened 2024-01-03
Introduce `sort-all` Hook for `pre-commit`

Bibo-Joshi:
I found https://pypi.org/project/sort-all/ and found it really useful for keeping the `__all__` variables sorted.
Admittedly, it's not the best maintained package, but the core functionality works fine. One downside is that comments in the list will be removed, but so far this doesn't affect us.
I would indeed hope that something similar can be set up for `__slots__`.


Poolitzer:
It's not a lot of code maybe it makes sense to build smth for us with slots and all?

harshil21:
@Poolitzer that&#x27;s what I was thinking as well, maybe open an issue in Ruff, cause it will get implemented pretty quickly if accepted

Bibo-Joshi:
> It's not a lot of code maybe it makes sense to build smth for us with slots and all?

Have already forked :D https://github.com/Bibo-Joshi/sort-all

> that's what I was thinking as well, maybe open an issue in Ruff, cause it will get implemented pretty quickly if accepted

also a nice idea! would you mind doing that? Things to mention IMO are

* the `sort-all` package for reference
* comment-handling
* extensibility for general iterables of strings with user defined names

---

If you like the idea in general, I'd like to go ahead with the package that we have so far :)

Bibo-Joshi:
@Bibo-Joshi @clot27 points to https://github.com/astral-sh/ruff/issues/1198 in which it&#x27;s mentioned that isort can do this already. I&#x27;ll have to check

Bibo-Joshi:
`isort` does support sorting `__all__` via the `sort_reexports` option even though this option is not documted in the [options list](https://pycqa.github.io/isort/docs/configuration/options.html) (only [here](https://pycqa.github.io/isort/reference/isort/index.html#class-variables)). One major bummer however is that this doesn't seem to honor the `black` settings, in particular it uses single quotes (`'` instead of `"`) and it collapses multiline definitions (see also https://github.com/PyCQA/isort/issues/2193). One would have to apply `black` again and the `pre-commit` step would always fails.
I would therefore not consider the isort implementation as useful for us ATM.

Bibo-Joshi:
https://github.com/PyCQA/isort/issues/2193 also mentions https://pypi.org/project/asort/. Even though that package is even less maintained than sort-all, it worked well on first try and even kept comments. I'll have another look at it in time

Bibo-Joshi:
Welp, asort completely deletes the content of `__all__` when run against our code base and also messes up several unrelated things. so not an option …

harshil21 (approved):
Looks good to me as a temporary solution. I suggested this package in the ruff issue: https://github.com/astral-sh/ruff/issues/1198#issuecomment-1880114865

Issue #4060, opened 2024-01-08
Move handler-related files to a new subdirectory


Bibo-Joshi:
The [`telegram/ext`](https://github.com/python-telegram-bot/python-telegram-bot/tree/master/telegram/ext) directory is getting a bit crowded. It would be good if all `_*handler.py` files as well as `filters.py` (which is part of the handlers setup) would be moved to a directory `telegram/ext/_handlers`. The imports and `telegram/ext/__init__.py` would have to be adapted in such a way that the public API doesn't change, i.e. `from telegram.ext import BaseHandler, CommandHandler, filters` etc is still the expected way of importing.

ManikMKR11:
Thanks Boss, I am interested here, but I don't understand about this all Thanks ♥️♥️♥️

you means you are thinking to delete the telegram.ext and make new module for work with filters and so on

Bibo-Joshi:
No, deleting `telegram.ext` is not what is meant at all. I explicitly clarified:

> the public API doesn't change, i.e. from telegram.ext import BaseHandler, CommandHandler, filters etc is still the expected way of importing.

lucasmolinari:
Hello :) That's a great oportunity to me to learn about python packages and help you. Can I work on it?

PR #4064, opened 2024-01-11
Move handler related files to _handlers subdirectory

lucasmolinari:
This is related to #4060. 
Created a new directory ```telegram/ext/_handlers``` where all ```_*handler``` related files were moved, including ```filters.py```.
Modified ```telegram/ext/__init__.py``` to point to the new directory.
Some files where pointing exactly to the handlers, for e.g ```telegram\ext\_application.py``` with the import statement ```from telegram.ext._basehandler import BaseHandler``` so I modified those too.


lucasmolinari:
> I tried around a bit and couldn't get it to work either. Also tried to find an explanation for the issue on the web, but didn't really > get that far either. Sometimes, Python import behavior remains a mystery :D

I searched too, there is literally nothing about that in the web, at least I couldn't find it. Tried with AI too but I didn't got lucky :[

I did what you suggested and moved the filters module to ```tg/ext/filters.py``` and removed the underscore from the handler names in ```tg/ext/_handlers```. Also, I noticed that some imports were still not working, so I fixed that with the help of pytest.

Since we are here, are these merge checks supposed to be succesful? Did I messed something up for that to be not working? At least the required ones seems to be fine.

Bibo-Joshi:
> Since we are here, are these merge checks supposed to be succesful? Did I messed something up for that to not be working? At least the required ones seems to be fine.

The `Bot API Tests` is expected to fail until #4034 is merged. The other ones should not fail.
For the unit tests, I had a look at the failures and there seem to be 2 issuesd

* in `test_basehandler.py` we use `from telegram.ext._basehandler import BaseHandler`. That should be changed to `from telegram.ext import BaseHandler`
* in `test_conversationhandler` in the test `test_unknown_state_warning`, we check that some warnings are issued as expected and explicitly use the path to `_conversationhandler.py`. This should be updated as well.

DeepSource reports "Function with cyclomatic complexity higher than threshold" for `ConversationHandler.check_update`, however since you didn't actually touch that, I'm okay with ignoring that.

lucasmolinari:
There it is, sorry about the commits that are only about formatting. First time working with pre-commit :)

Poolitzer:
Only three commits from fighting pre-commit? What are you, some kind of wizard? I usually need at least five.

lucasmolinari:
Thank you both for been supportive. Learned a lot with this simple contribution, next time will be easier!

Bibo-Joshi (approved):
LGTM :) Thanks for the work! @Poolitzer or maybe @clot27 would you mind giving it a once-over as well?

clot27 (approved):
Nice work @lucasmolinari, LGTM!

Bibo-Joshi:
Thanks for the contribution!

PR #4067, opened 2024-01-14
Add Missing Conversions of `type` to Corresponding Enum from `telegram.constants`

Bibo-Joshi:

### DIFF
--- a/telegram/_botcommandscope.py
+++ b/telegram/_botcommandscope.py
@@ -22,6 +22,7 @@ from typing import TYPE_CHECKING, Dict, Final, Optional, Type, Union
 
 from telegram import constants
 from telegram._telegramobject import TelegramObject
+from telegram._utils import enum
 from telegram._utils.types import JSONDict
 
 if TYPE_CHECKING:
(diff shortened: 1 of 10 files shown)


harshil21 (commented):
yeah, I had seen some places where this was missed. Is there a way to easily test this for all classes (test should probably assume that `self.type` needs an enum reference)

Bibo-Joshi:
Added a simple regex testcase and actually found another place to apply the change. I excluded passport stuff - we don't have enums for that and I don't like touching that stuff if not *really* necessary 😬

tests/test_enum_types.py
```diff
@@ -0,0 +1,52 @@
...
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+import re
+from pathlib import Path
+
+telegram_root = Path(__file__).parent.parent / "telegram"
+telegram_ext_root = telegram_root / "ext"
+exclude_dirs = {
+    # We touch passport stuff only if strictly necessary.
+    telegram_root
+    / "_passport",
```
harshil21, line 27:
```suggestion
    telegram_root / "_passport",
```
having this on a new line kinda confused me (specially since I'm reviewing on mobile)

harshil21 (approved):
The general test case looks good and I'm fine with excluding the passport directory!

tests/test_enum_types.py
Bibo-Joshi: black wants it that way …

PR #4075, opened 2024-01-21
Bump ruff and remove sort-all

harshil21:
Removes sort-all since ruff now supports sorting `__all__`. Also adds 2 more ruff rules:

- [flake8 quotes](https://docs.astral.sh/ruff/rules/#flake8-quotes-q)
- [flake8-no-pep-420](https://docs.astral.sh/ruff/rules/#flake8-no-pep420-inp)

Also forces pre-commit to use python 3.12, to ensure consistency between different hooks and between locally run pre-commit and CI pre-commit


harshil21:
huh, my pre-commit did not report those errors (differences in python version?).. also it seems like [`SIM401`](https://docs.astral.sh/ruff/rules/#flake8-simplify-sim) was silently being ignored all this time? It's not a preview rule so not sure how ruff didn't catch that before..

telegram/constants.py
```diff
@@ -29,11 +29,12 @@
     * Most of the constants in this module are grouped into enums.
 """
 # TODO: Remove this when https://github.com/PyCQA/pylint/issues/6887 is resolved.
-# pylint: disable=invalid-enum-extension
+# pylint: disable=invalid-enum-extension,invalid-slots
```
Bibo-Joshi, line 32: what's invalid here?

telegram/ext/_application.py
```diff
@@ -1133,7 +1133,7 @@ async def __create_task_callback(
                 return await asyncio.create_task(coroutine)
             # If user uses generator in python 3.12+, Exception will happen and we cannot do
             # anything about it. (hence the type ignore if mypy is run on python 3.12-)
-            return await coroutine  # type: ignore[misc]
+            return await coroutine
```
Bibo-Joshi, line 1136: This fails the pre-commit check - IIRC mypy wants to remove that when running on py3.1  🤔

Bibo-Joshi (commented):
different python version could certainly be. ruff always wants me to use `typing.Self` when running on py3.11 …

telegram/ext/_application.py
harshil21: oh yeah. I even wrote a comment about it on the above line..

telegram/constants.py
harshil21: `E0238: Invalid __slots__ object (invalid-slots)` for all classes whose base class is `IntEnum`

harshil21:
Interesting thing I found about `pre-commit` - new contributors will not need to run `pre-commit install` whenever they clone our repo, if we do [this](https://pre-commit.com/#automatically-enabling-pre-commit-on-repositories). It will automatically install the hooks during cloning and run them on `git commit`. I can try doing this on another PR.

harshil21:
> Interesting thing I found about `pre-commit` - new contributors will not need to run `pre-commit install` whenever they clone our repo, if we do [this](https://pre-commit.com/#automatically-enabling-pre-commit-on-repositories). It will automatically install the hooks during cloning and run them on `git commit`. I can try doing this on another PR.

well I misunderstood what that option is for. It's only meant for local setups where you want _your_ git to automatically install the `.pre-commit-config.yaml` from repositories to `.git/hooks`. We can't instruct _others'_ git to do that (which makes sense since the `INSTALL_PYTHON` variable in `.git/hooks/pre-commit` differs according to where python is installed in your system).

.pre-commit-config.yaml
```diff
@@ -88,9 +88,6 @@ repos:
...
           - aiolimiter~=1.1.0
--   repo: https://github.com/aio-libs/sort-all/
-    rev: v1.2.0
-    hooks:
-    -   id: sort-all
-        name: sort-all
-        files: ^(telegram|examples|tests)/.*\.py$
+
+default_language_version:
+  python: python3.12
```
Bibo-Joshi, line 93: why 3.12 instead of our minimum supported version 3.8?

harshil21: cause I'm running that version locally 😁
It shouldn't make a big difference since black and ruff have target-versions set

Bibo-Joshi: my thought is rather that ruff might make us do stuff that break on older versions 🤔

harshil21: ok I'll make it 3.8 and check what changes

harshil21: oh well I need to have 3.8 installed for pre-commit to work..

Bibo-Joshi (approved):
LGTM :) Is there anything left from your side or should I merge?

harshil21:
@Bibo-Joshi all good from my side

PR #4079, opened 2024-01-23
chore: :wrench: add option to download docs as zipped html

### LINKED ISSUE #4050: Offer Docs to be downloaded as zip of html files [FEATURE]
*self-explanatory.*

I saw that the documentation website offers the docs as pdf file [link-for-v20.7](https://docs.python-telegram-bot.org/en/v20.7/), which is something I really like because I work most of the time offline and I relay on reading documentations rather than tutorials.
but as I was scrolling the website, I realized that I'm enjoying the experience and the overall design of it, and I wished it was possible to download the entire documentations as html.zip file.

so why don't you start offering the documentation to be downloaded as zip of html files.

kenjitagawa:
closes #4050

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -12,6 +12,7 @@ sphinx:
 # Optionally build your docs in additional formats such as PDF
 formats:
    - pdf
+   - htmlzip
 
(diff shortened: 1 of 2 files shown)


harshil21 (approved):

PR #4084, opened 2024-01-28
Add `Bot.do_api_request`

### LINKED ISSUE #4053: [FEATURE] `Bot.do_api_request`

We currently provide no way for users to use new API endpoints through PTB before we add the in the next release. With `api_kwargs` we have already provided ways for users to use new parameters & attributes, but this part is still missing in terms of API-"upward"-compatibility.
I haven't seen _too_ many requests for this over the years, but at least for the new reactions methods that came up a few times in the last days. Adding a way to do this would IMO be a notable step to make PTB even more future proof.

Add a method `Bot.do_api_request` (or similar name) that accepts

* the name of an endpoint (e.g. `sendMessage`, maybe also allowing python style `send_message` would be good
* a json-dict with value to pass to TG
* optionally a `TelegramObject` subclass that will be used to `de_json` the return value

This method will then execute that request. The documentation should make clear, which utility functions of PTB are working and which are not working. In particular

* defaults handling can not work (if a new method accepts a `parse_mode` parameter, that will either be part of the json-dict - such that no default value is inserted - or not - in wich case we can't know that a default value *should* be inserted)
* convenience input parsing for `InputFile` / local file paths can't work b/c we don't know which parameters should be interpreted as media upload
* setting default write-timeout to the media-write-timeout will work b/c the networking backend should detect that depending on `RequestData.files`
* Converting datetimes to ints will work (including default Timezones)

(These explanations could then in fact also be linked in the docstrings of the `api_kwargs` argument of the other methods)

The method should issue a warning if the Bot class has a method with the name of the requested endpoint, telling the user to use that one instead.

* instead of a json-dict one could just accept arbitrary keywoard arguments, which may be viewed as more convenient by some, but goes against the usual arguments philosophy that we have in PTB so far. This would leave the (very small) risk that a parameter name defined by TG would clash with a parameter name defined by PTB (e.g. the timeouts).
* Users can in theory also directoy use the networking backend (`BaseRequest.request()`), however
    * we do not document (on purpose!) how `RequestData` is constructed
    * they'd still have to build the url on their own
* Instead of issuing warning for already-implemented endpoints, the method could in fact throw an exception. This would make every PTB-API release breaking by design.

### Additional context

One outcome of https://github.com/python-telegram-bot/python-telegram-bot/issues/2162 is

https://github.com/python-telegram-bot/python-telegram-bot/blob/979988add1a56933380db970b4bae8e2582ebeb4/telegram/request/_baserequest.py#L371-L377

If we allow users to manually set the endpoints via a new `Bot.do_api_request`, we might need to rething this part. One possible option would be to rename the error class to `InvalidTokenOrMethod` that we use here. An alternative would be to add special exception handling directly within `Bot.do_api_request`, converting `InvalidToken` into `MethodNotFound` with a message along the line "Assumg that `Bot.initialize()` worked correctly, the provided endpoint must be invalid".

Bibo-Joshi:
Closes #4053

- [x] add rate limit kwargs in ExtBot and make sure that tests also cover that


telegram/_bot.py
```diff
@@ -762,6 +763,101 @@ async def shutdown(self) -> None:
...
+        *,
+        read_timeout: ODVInput[float] = DEFAULT_NONE,
+        write_timeout: ODVInput[float] = DEFAULT_NONE,
+        connect_timeout: ODVInput[float] = DEFAULT_NONE,
+        pool_timeout: ODVInput[float] = DEFAULT_NONE,
+    ) -> Any:
+        """Do a request to the Telegram API.
+
+        This method is here to make it easier to use new API methods that are not yet supported
+        by the python-telegram-bot library.
```
Poolitzer, line 781:
```suggestion
        by this library.
```

telegram/_utils/strings.py
```diff
@@ -0,0 +1,38 @@
...
+    """Converts a snake_case string to camelCase.
+
+    Args:
+        snake_str (:obj:`str`): The string to convert.
+
+    Returns:
+        :obj:`str`: The converted string.
+    """
+    components = snake_str.split("_")
+    return components[0] + "".join(x.title() for x in components[1:])
```
Poolitzer, line 38: I am waiting for some weird method to break this xD

Poolitzer (changes requested):
LGTM just one phrasing request out of pride

telegram/error.py
```diff
@@ -133,6 +134,19 @@ def __init__(self, message: Optional[str] = None) -> None:
...
 
 
+class EndPointNotFound(TelegramError):
+    """Raised when the requested endpoint is not found. Only relevant for
+    :meth:`telegram.Bot.do_api_request`.
+
+    Args:
+        endpoint (:obj:`str`): The endpoint that was not found
+
+            .. versionadded:: 20.0
```
clot27, line 144: Wrong versionadded directive

clot27, line 144: Wrong versionadded directive

tests/test_bot.py
```diff
@@ -170,6 +169,7 @@ def bot_methods(ext_bot=True, include_camel_case=False):
             arg_values.append((cls, name, attribute))
             ids.append(f"{cls.__name__}.{name}")
 
+    print(arg_values)
```
clot27, line 172: can be removed now ig

clot27 (commented):
Great addition! just few nitpicks

PR #4088, opened 2024-02-01
Migrate from `setup.cfg` to `pyproject.toml`

harshil21:
Wanted to do this for a long time. Having two config files is just unnecessary since `pyproject.toml` is now the recommended config file for tools, linters, packaging, etc.

So this PR migrates all the tools to use `pyproject.toml` instead. There are only 3 remaining:

- `license-files`: Seems like this breaks `test_meta.py`? Otherwise it's possible to move it to `pyproject.toml` according to [their docs](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration).
- `sphinx-pypi-upload`: This was added by leandro 8 years ago, back when docs were uploaded not to RTD, but to pythonhosted.org. I recommend removing that dependency and from the `setup.cfg` (`upload_sphinx`, `build_sphinx`). More information on that method can be found [here](https://pythonhosted.org/an_example_pypi_project/buildanduploadsphinx.html#using-sphinx-pypi-upload).
- `flake8`: Unfortunately, flake8 does not support pyproject.toml due to some weird reason (shame!) - https://github.com/PyCQA/flake8/issues/234 but we should be able to remove `flake8` once `ruff` becomes stable.


Poolitzer:
Well we aren't on pythonhosted anymore. I don't see any reason keeping that in our config.

pyproject.toml
```diff
@@ -24,3 +27,99 @@ explicit-preview-rules = true
...
+omit = [
+    "tests/",
+    "telegram/__main__.py"
+]
+
+[tool.coverage.report]
+exclude_also = [
+    "@overload",
+    "if TYPE_CHECKING:"
+]
```
Bibo-Joshi, line 125: missing new line

Bibo-Joshi (commented):
Fully agree on the general idea 👍 

* license-files: If you're able to move to pyproject, I'm ofc happy. test-meta just checks the return code of `python setup.py bdist_dumb`, so running that manually should probably give you an idea of the exception
* sphinx-pypi-upload: I agree with pool, ditch that

harshil21:
> * license-files: If you're able to move to pyproject, I'm ofc happy. test-meta just checks the return code of `python setup.py bdist_dumb`, so running that manually should probably give you an idea of the exception
 
there was some error regarding the parsing of pyproject.toml, I'll have a closer look later

harshil21:
> > * license-files: If you're able to move to pyproject, I'm ofc happy. test-meta just checks the return code of `python setup.py bdist_dumb`, so running that manually should probably give you an idea of the exception
> 
> there was some error regarding the parsing of pyproject.toml, I'll have a closer look later

Leaving this for now, since we will be discussing if it's worth to ditch `setup.py` and `ptb-raw` as discussed internally (I'll open a new issue for that soon)

PR #4100, opened 2024-02-05
Drop Usage of DeepSource

Bibo-Joshi:
After initial code quality improvements by integrating DeepSource, the tool has become somewhat of a burden.

* it regularly highlights rather edge-casy issues as risk
* New code quality rules are introduced by the DS team without a chance for us to pin a specified ruleset (see https://discuss.deepsource.com/t/improve-granular-setting-possibility-to-pin-a-deepsource-version/472)
* Configuration of the tool is somewhat cumbersome

I propose to drop it. Much of it has since also been replaced by ruff.
Once this PR is merged, the GH integration is to be removed.


harshil21:
Agree. The PR should also remove the unwanted deepsource ignore comments (skipcq?)

PR #4104, opened 2024-02-07
Fix documentation version to 20.7

xTudoS:
The documentation previously stated the version as 20.6, which was incorrect. This commit updates the version number in the documentation to reflect the correct version, 20.7.

Changes Made:
- Updated the version number in documentation from 20.6 to 20.7


clot27:
Thanks for the contribution!

PR #4114, opened 2024-02-11
Remove Functionality Deprecated in API 7.0

### LINKED ISSUE #4099: Remove Functionality Deprecated in API 7.0
We should create a PR for this shortly after #4034 is merged - will be easier to do that now rather than when API 7.1 is released (both in terms of time and recalling what we did).
The PR will be merged only once API 7.1 is released.

Note: Only pure Bot API deprecations should be removed by this PR, e.g. the deprecated `Defaults.quote` will have to stay longer.

Bibo-Joshi:
~When ready,~ closes #4099


tests/auxil/bot_method_checks.py
```diff
@@ -549,11 +549,9 @@ async def check_defaults_handling(
     }
     # We tested this for a long time, but Bot API 7.0 deprecated it in favor of
     # reply_parameters. In the transition phase, both exist in a mutually exclusive
-    # way. Testing both cases would require a lot of additional code, so we just
-    # ignore this parameter here until it is removed.
-    # Same for disable_web_page_preview
+    # way. Testing both cases would require a lot of additional code, so we for now are content
```
harshil21, line 552: which both cases?

tests/test_bot.py
```diff
@@ -2155,7 +2155,7 @@ async def test_forward_messages(self, bot, chat_id):
         msg1, msg2 = await tasks
 
         forward_messages = await bot.forward_messages(
-            chat_id, from_chat_id=chat_id, message_ids=(msg1.message_id, msg2.message_id)
+            chat_id, from_chat_id=chat_id, message_ids=sorted((msg1.message_id, msg2.message_id))
```
harshil21, line 2158: any idea why the bot api wants them to be sorted?

Bibo-Joshi: > The identifiers must be specified in a strictly increasing order.

https://core.telegram.org/bots/api#forwardmessages

don't know *why* they want that, though …

telegram/ext/filters.py
```diff
@@ -2146,15 +2146,19 @@ class _UserShared(MessageFilter):
         __slots__ = ()
 
         def filter(self, message: Message) -> bool:
-            return bool(message.user_shared)
+            return bool(message.api_kwargs.get("user_shared"))
 
     USER_SHARED = _UserShared(name="filters.StatusUpdate.USER_SHARED")
-    """Messages that contain :attr:`telegram.Message.user_shared`.
+    """Messages that contain ``"user_shared"`` in :attr:`telegram.TelegramObject.api_kwargs`.
```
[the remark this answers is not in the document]
Bibo-Joshi: We did - see line 2163

tests/auxil/bot_method_checks.py
Bibo-Joshi: passing `allow_sending_without_reply` *and* testing `reply_parameters`

tests/test_bot.py
harshil21: should PTB auto sort the passed message ids?

Bibo-Joshi: mh … my gut says no. Sroting can be expensive and if the user already takes care of it, it's wasted effort. then again it's only 100 entries max and the effort spent on sorting an already sorted list is probably small 🤔

harshil21: well the performance doesn't matter, since the user can do it in their code, or we do it library side. It will take "time" either way. The advantage is the user doesn't have to worry about it

Bibo-Joshi: Performance does matter if thr user always passes sorted lists - then it's unnecessary overhead. I'd like to see some small benchmark before going ahead with this. In any case, it would be a separate pr :)

harshil21 (approved):
I hope nothing was missed

Poolitzer (approved):
LGTM

Bibo-Joshi:
> I hope nothing was missed

Had searched for all "20.8" versionchange annotation and double checked the TG changelog …. Should be fine :)

Issue #4125, opened 2024-02-22
[QUESTION] What is the correct shutdown procedure, when running inside event loop?


kramarov-evg:
### Issue I am facing

I need to keep a persistet aiohttp session, which I try to store inside bot_data and, given it should only be created in async context, I have an async function wrapper around session and application creation. So I need to launch the bot inside an already existing async context, hence, running event loop. I've referred to #3310, #3711 and [wiki on running with other asyncio FWs](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Frequently-requested-design-patterns#running-ptb-alongside-other-asyncio-frameworks). I've understood the following points:

- To launch application in an already running event loop I need to stop using `Application.run_polling` and stick with the sequence of calls it replaces (`Application.initalize -> Application.start -> Application.updater.start_polling -> Application.updater.stop -> Application.stop -> Application.shutdown`)
- First and last calls can be replaced with a context manager (`with application`) making shortening call sequence to `Application.start -> Application.updater.start_polling -> Application.updater.stop -> Application.stop`
- Between `start_polling` and `updater.stop` should be something that would keep the event loop alive to prevent application fom stopping immediately after starting

So I came up with the code below, which for some reason doesn't quit upon Ctrl-C event. Moreover, it takes 2 Ctrl-Cs to stop the program, and the second produces the stacktrace below. And my question is, how to correctly handle termination when running inside already existing event loop?

### Traceback to the issue

```python
Traceback (most recent call last):
  File "/usr/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.9/asyncio/base_events.py", line 629, in run_until_complete
    self.run_forever()
  File "/usr/lib/python3.9/asyncio/base_events.py", line 596, in run_forever
    self._run_once()
  File "/usr/lib/python3.9/asyncio/base_events.py", line 1854, in _run_once
    event_list = self._selector.select(timeout)
  File "/usr/lib/python3.9/selectors.py", line 452, in select
    def select(self, timeout=None):
  File "/home/egor/.vscode-server/extensions/ms-python.debugpy-2024.0.0-linux-arm64/bundled/libs/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_trace_dispatch_regular.py", line 374, in __call__
    return None if event == 'call' else NO_FTRACE
KeyboardInterrupt

During handling of the above exception, another exception occurred:

[...]

PR #4126, opened 2024-02-22
Make `Updater.stop` Independent of `CancelledError`

Bibo-Joshi:
See #4125 for discussion

Since the tests for these things usually don't run well on Windows, I haven't run the unit tests yet - let's see.
Adding new unit tests will also likely be chellanging. If the test coverage doesn't drop, I might be content with leaving as is 😬


Bibo-Joshi:
All right, tests seem to be stuck. will have to debug on linux

Bibo-Joshi:
Test coverage seems to be fine and as expected adjusting the existing tests was already rather annoying, so tbh I'm not eager to try and add another unit test here.
I manually ran the customwebhook as well, which still works fine.

telegram/ext/_updater.py
```diff
@@ -126,6 +126,7 @@ def __init__(
         self._httpd: Optional[WebhookServer] = None
         self.__lock = asyncio.Lock()
         self.__polling_task: Optional[asyncio.Task] = None
+        self.__polling_task_stop_event: asyncio.Event = asyncio.Event()
```
harshil21, line 129: need the slot entry for this...

telegram/ext/_updater.py
```diff
@@ -706,39 +709,58 @@ async def _network_loop_retry(
             description (:obj:`str`): Description text to use for logs and exception raised.
             interval (:obj:`float` | :obj:`int`): Interval to sleep between each call to
                 `action_cb`.
+            stop_event (:class:`asyncio.Event` | :obj:`None`): Event to wait on for stopping the
+                loop. Setting the event will make the loop exit even if action_cb is currently
```
harshil21, line 713:
```suggestion
                loop. Setting the event will make the loop exit even if `action_cb` is currently
```

tests/ext/test_application.py
```diff
@@ -1432,15 +1432,22 @@ async def callback(update, context):
...
 
         async def get_updates(*args, **kwargs):
             if exception_event.is_set():
                 raise exception
+
             # This makes sure that other coroutines have a chance of running as well
-            await asyncio.sleep(0)
+            if exception_testing_done.is_set() and app.updater.running:
+                # the longer sleep makes sure that we can exit also while get_updates is running
+                await asyncio.sleep(20)
```
harshil21, line 1447: does it really have to be 20 secs? Doesn't that mean the test suite is hung up by 20 secs?

tests/ext/test_updater.py
```diff
@@ -611,8 +614,9 @@ async def get_updates(*args, **kwargs):
                 # Unfortunately we need to use the private attribute here to produce the problem
                 updater._running = False
                 second_update_event.set()
+                await asyncio.sleep(1)
+                # await third_update_event.wait()
```
harshil21, line 618: don't need this?

telegram/ext/_updater.py
```diff
@@ -706,39 +709,58 @@ async def _network_loop_retry(
...
+
+        async def do_action() -> bool:
+            if not stop_event:
+                return await action_cb()
+
+            action_cb_task = asyncio.create_task(action_cb())
+            stop_task = asyncio.create_task(stop_event.wait())
+            done, pending = await asyncio.wait(
+                (action_cb_task, stop_task), return_when=asyncio.FIRST_COMPLETED
+            )
```
harshil21, line 726: clever

harshil21 (commented):
code looks good. Didn't have a very close look at the tests.. they are kind of a black box because of so many `Event`'s. I assume you ran an example bot on both Windows and linux and checked if Ctrl+C works well?

tests/ext/test_application.py
Bibo-Joshi: no, what I'm testing here is that calling `Updater.stop` does indeed *not* wait for the 2 seconds to finish, i.e. that  `asyncio.wait` returns the `event.wait()` as done :)

Bibo-Joshi:
> Didn't have a very close look at the tests.. they are kind of a black box because of so many `Event`'s.

They are certainly not very pretty :/

> I assume you ran an example bot on both Windows and linux and checked if Ctrl+C works well?

Yes, I ran echobot & the example from #4125.

harshil21 (approved):

PR #4134, opened 2024-02-25
Handle properties in `__setstate__`

### LINKED ISSUE #4111: [FEATURE] Improve Backward Compatibility for pickling

When TG deprecates attributes, we usually convert them to properties in our classes and issue a warning on access. For this, we rename `Class.attribute` to `Class.._attribute`. This leads to [problems](https://t.me/pythontelegrambotgroup/723964) on un-pickling data that was pickled with previous versions. While this is okay acccording to our [stability policy](https://docs.python-telegram-bot.org/en/v20.8/stability_policy.html#what-doesn-t-this-policy-cover), it would still be nice to handle this better for the users

In `TelegramObjejct.__setstate__`, the current logic is

https://github.com/python-telegram-bot/python-telegram-bot/blob/1cf63c26c59c3db2c938bffe26898216283b60af/telegram/_telegramobject.py#L311-L322

I would like to change this to

1. first try set_attr
2. check if the attribute name is a property
3. if yes, try setting _attributename (unless the name already starts with underscore)
4. if that fails or the attribute is not a property, put the attribute into `api_kwargs` - unless the attribute has a leading underscore. This last part is for when we finally completely remove the property - pickling with the property in place and unpickling without the property in the new version will then simply remove the attribute.

Moreover, in `_apply_api_kwargs`

https://github.com/python-telegram-bot/python-telegram-bot/blob/1cf63c26c59c3db2c938bffe26898216283b60af/telegram/_telegramobject.py#L492-L494

we can first check if tha attribute is a property

Do nothing as we didn't promise that this works.

### Additional context

https://t.me/pythontelegrambotgroup/723964

harshil21:
Closes #4111.

Implementing tests for this was hard! Since our library's code will keep changing, we could not rely on the previous method of pickling a file from an older version and unpickling in our current version. We need to specfically test when we are in a deprecation phase like now. So I came up with using `exec()` statements to fool Python into thinking the class is changing.


telegram/_telegramobject.py
```diff
@@ -312,7 +313,18 @@ def __setstate__(self, state: Dict[str, object]) -> None:
...
             except AttributeError:
-                # catch cases when old attributes are removed from new versions
+                # So an attribute was deprecated and removed from the class. Let's handle this:
+                # 1) Is the attribute now a property with no setter? Let's check that:
+                if isinstance(getattr(self.__class__, key, None), property):
+                    # It is, so let's try to set the "private attribute" instead
+                    try:
+                        setattr(self, f"_{key}", val)
+                    except AttributeError:  # guess we've completely removed it
+                        api_kwargs[key] = val  # add it to api_kwargs as fallback
```
Bibo-Joshi, line 323: very nitpicky: you're mixing inline comments with block comments (block comments are prefered by pep8 and me 🤡 )

tests/test_telegramobject.py
```diff
@@ -358,6 +358,67 @@ async def test_pickle_backwards_compatibility(self):
...
+        pickled_v1 = pickle.dumps(old)
+
+        # After some API changes:
+        v2 = """
+class PicklePropertyTest(TelegramObject):
+    __slots__ = ("_forward_from", "_date", "_new_attr")
+    def __init__(self, forward_from=None, f_date=None, new_attr=None, api_kwargs=None):
+        super().__init__(api_kwargs=api_kwargs)
+        self._forward_from = forward_from
+        self.f_date = f_date
```
Bibo-Joshi, line 387: This is not covered in the tests below - is that on purpose?

telegram/_telegramobject.py
```diff
@@ -490,7 +502,11 @@ def _apply_api_kwargs(self, api_kwargs: JSONDict) -> None:
         """
         # we convert to list to ensure that the list doesn't change length while we loop
         for key in list(api_kwargs.keys()):
-            if getattr(self, key, True) is None:
+            # property attributes are not settable, so we need to set the private attribute
+            if isinstance(getattr(self.__class__, key, None), property):
+                with contextlib.suppress(AttributeError):
+                    setattr(self, f"_{key}", api_kwargs.pop(key))
```
Bibo-Joshi, line 508: Can you leave a comment here on why it's okay to supress the AttributeError?

Bibo-Joshi (changes requested):
Thanks & kudos for the PR! The number of changed lines is not that big but I can imagine that it took quite some headache to get it done 😬 
Just smaller comments

tests/test_telegramobject.py
harshil21: i'm calling v2_unpickle.forward_date, which is using this attribute

Bibo-Joshi (approved):

Issue #4116, opened 2024-02-13
[BUG] Docs confusingly suggest that `message_reaction` and `message_reaction_count` are included in webhook updates by default


mbanerjeepalmer:
### Steps to Reproduce

1. Read docs which say that for `allowed_updates` you should  'Specify an empty sequence to receive all updates except [telegram.Update.chat_member](https://docs.python-telegram-bot.org/en/v20.8/telegram.update.html#telegram.Update.chat_member) (default).'
2. Leave `allowed_updates` empty (i.e. do not specify parameter at all).

### Expected behaviour

Given the python-telegram-bot docs, the user expects the webhook to deliver `message_reaction` and `message_reaction_count`.

### Actual behaviour

The webhook is not set up to deliver message reactions. This is because Telegram's default behaviour is actually the following:
> Specify an empty list to receive all update types except chat_member, message_reaction, and message_reaction_count (default). If not specified, the previous setting will be used.

Source: https://core.telegram.org/bots/api#setwebhook

### Operating System

Ubuntu 22.04.3

### Version of Python, python-telegram-bot & dependencies

```shell
python-telegram-bot 20.7
Bot API 6.9
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
```

### Additional Context

The wording change needs to be made here: https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/_bot.py#L4260

mbanerjeepalmer:
Can do

Trifase:
Please, can you update the docs for `run_polling` too for the same parameter?

harshil21:
hey @mbanerjeepalmer, are you still working on this? We're nearing a new release so it would be nice to include this as well.

mbanerjeepalmer:
@harshil21 PR opened. Black was giving me weird line lengths so there are some unnecessary line breaks, let me know if that needs fixing.

harshil21:
Closed by #4140.

PR #4140, opened 2024-03-01
Webhook excludes message_reaction, and message_reaction_count by default #4116

mbanerjeepalmer:
Fixes https://github.com/python-telegram-bot/python-telegram-bot/issues/4116

I had some weird issues with Black so there are some additional line breaks. Hopefully this won't be reflected in the finished docs.


harshil21 (approved):
Everything looks good! Thanks for the contribution @mbanerjeepalmer!

PR #4157, opened 2024-03-12
Handle `SystemExit` raised in Handlers

### LINKED ISSUE #4155: [Bug] sys.exit hangs in callback
Currently we have support for `sys.exit` implicit mentioned in the docs (e.g. [run_polling](https://docs.python-telegram-bot.org/en/v21.0.1/telegram.ext.application.html#telegram.ext.Application.run_polling), follow the [SystemExit link](https://docs.python.org/3/library/exceptions.html#SystemExit), which then links to `sys.exit`) but actually calling it in a callback results in the code hanging at update_queue.join. This should either be fixed or we should remove "support" for it.

### LINKED ISSUE #4156: [Bug] no clean way to exit application in post_init
We have four ways to exit application, none of them work without an error. We should fix at least one and document which works:

- `app.stop()` returns the error that the application is not running,
- `app.shutdown()` gives "This Updater was not initialized",
- `app.stop_running()` returns because the application is not running,
-  sys.exit gives "coroutine 'Updater.start_polling' was never awaited."

Bibo-Joshi:
closes #4156. closes #4155.
~Have not checked with jobs or error handlers yet.~

MWE for checking the behavior:

<details>

```python
import asyncio
import logging
import sys

from telegram.ext import ApplicationBuilder, ContextTypes, TypeHandler

# Enable logging
logging.basicConfig(
)
logging.getLogger("telegram.ext.Application").setLevel(logging.DEBUG)

async def raise_system_exit(_, context: ContextTypes.DEFAULT_TYPE):
    for ___ in range(5):
        await context.application.update_queue.put("Hello, world!")
    print("raising system exit")
    raise sys.exit(1)


async def post_init(application):
    async def put(app):
        await asyncio.sleep(2)
        print("enqueueing update")
        await app.update_queue.put("Hello, world!")

    application.create_task(put(application))


application = ApplicationBuilder().token("TOKEN").post_init(post_init).build()
application.add_handler(TypeHandler(object, raise_system_exit))

if __name__ == "__main__":
    application.run_polling()
```

</details>


Bibo-Joshi:
Did some more rework and also some testing. Seems to work with handler/error handler callbacks and jobs now. I also realized that when using `block=False` we don't have an issue in the first place: If the SystemExit is raised in a task that apparently is reported directory to the event loop.

Also removed the supression of `CancelledError` as already hinted in #4172

Exetnded example for fiddling around, including both a run_polling and a custom version.

```python
import asyncio
import logging
import sys

from telegram.ext import ApplicationBuilder, ContextTypes, TypeHandler, Application

# Enable logging
logging.basicConfig(
)
logging.getLogger("telegram.ext.Application").setLevel(logging.DEBUG)
logging.getLogger("telegram.ext.Updater").setLevel(logging.DEBUG)
logging.getLogger("httpx").setLevel(logging.WARNING)
logging.getLogger("httpcore").setLevel(logging.WARNING)


async def queue(application: Application):
    for ___ in range(5):
        await application.update_queue.put("Hello, world!")


async def queue_and_raise(context: ContextTypes.DEFAULT_TYPE):
    await queue(context.application)
    print("raising system exit")
    raise sys.exit(1)


async def raise_system_exit(_, context: ContextTypes.DEFAULT_TYPE):
    # await queue_and_raise(context)
    raise RuntimeError("This is a test error")


async def error_handler(_, context: ContextTypes.DEFAULT_TYPE):
    print("error handler got called")
    await queue_and_raise(context)


async def general_handler(update: object, context: ContextTypes.DEFAULT_TYPE):
    # await asyncio.sleep(1)
    # context.application.stop_running()
    print("general handler got called for update", update)


async def job_callback(context: ContextTypes.DEFAULT_TYPE):
    await queue_and_raise(context)


async def post_init(application):
    async def put(app):
        await asyncio.sleep(2)
        print("enqueueing update")
        await app.update_queue.put(-1)

    # application.create_task(put(application))
    application.job_queue.run_once(job_callback, 5)


application = (
    ApplicationBuilder().token("TOKEN")
    .post_init(post_init)
    .build()
)
application.add_handler(TypeHandler(int, raise_system_exit, block=True))
application.add_handler(TypeHandler(object, general_handler))
application.add_error_handler(error_handler, block=False)


def main():
    application.run_polling()


async def main_async():
    await application.initialize()
    await application.start()
    await application.updater.start_polling()
    await application.post_init(application)
    await queue(application)
    try:
        await asyncio.Event().wait()
    except Exception as ex:
        print(f"exception while running server: {ex!r}")
    except BaseException as ex:
        print(f"base exception while running server: {ex!r}")
        raise ex
    finally:
        print("stopping application")
        await application.stop()
        await application.updater.stop()
        await application.shutdown()


if __name__ == "__main__":
    # asyncio.run(main_async())
    main()

```

Bibo-Joshi:
Latest commit also addresses #4156, at least party. By proply closing the `updater_coroutine`, one can now `raise SystemExit` in `post_init` and still get a clean shutdown. I'll still think about if we can make `stop_running` available in a clean fashing in `post_init` as well.

Bibo-Joshi:
Last commits allows calling `stop_running` in `post_init`.
I still have to add tests for raising `SystemExit` in the different callbacks (will have to do so on a linux machine), but the code changes could already be reviewed

Bibo-Joshi:
https://github.com/python-telegram-bot/python-telegram-bot/actions/runs/8723054036/attempts/3#summary-23932096602

this test failure happens only on windows with python 3.11/3.12 and only that one single test. I don't get why sniffio isn't able to detect the asyncio lib in that particluar case and I'm unable to reproduce the problem locally.

The failer doesn't look related to the PR to me, but it's still strange …

Bibo-Joshi:
After removing all other tests, `test_signal_handler` failed only on py11 + windows and that succeeded after rerun … let's try adding rest of the tests again

telegram/ext/_application.py
```diff
@@ -1049,6 +1081,9 @@ def __run(
             updater_coroutine.close()
 
             try:
+                # In we never got to await the coroutine, it's cleaner to close it here
+                # This can be the case e.g. if post_init raises SystemExit
+                updater_coroutine.close()
```
harshil21, line 1086: why do we have to close the updater_coroutine twice in succession?

harshil21 (approved):
Pretty happy with this PR. Though the shutdown logic has gotten complicated over the past few updates (at least to me), I did manually test various cases and it worked as I expected. Tests also seem to have been simplified (no threading?!)

telegram/ext/_application.py
Bibo-Joshi: looks like a merge conflict thingy … removing

harshil21 (approved):
looks good from my side. Another review would help - @Poolitzer do you want to review too since you created the issue?

telegram/ext/_application.py
```diff
@@ -670,14 +672,26 @@ async def stop(self) -> None:
             raise RuntimeError("This Application is not running!")
 
         self._running = False
+        self.__stop_running_marker.clear()
```
Poolitzer, line 675: why do we need this again here

telegram/ext/_application.py
```diff
@@ -733,9 +766,7 @@ def run_polling(
         polling updates from Telegram using :meth:`telegram.ext.Updater.start_polling` and
         a graceful shutdown of the app on exit.
 
-        The app will shut down when :exc:`KeyboardInterrupt` or :exc:`SystemExit` is raised.
-        On unix, the app will also shut down on receiving the signals specified by
-        :paramref:`stop_signals`.
+        |app_run_shutdown|  :paramref:`stop_signals`.
```
Poolitzer, line 769:
```suggestion
        |app_run_shutdown| :paramref:`stop_signals`.
```

Poolitzer (changes requested):
:)

telegram/ext/_application.py
Bibo-Joshi: in case the app is stopped & restartet during the lifetime of the python script, we need to reset the marker. otherwise the second startup will immediately shut down as well

PR #4168, opened 2024-03-23
Add property `Update.effective_sender`

### LINKED ISSUE #4085: [FEATURE] Add Update.effective_sender
We have `Update.effective_user` for the sender of the message. This attribute however misses out on the anonymous sender info provided in `Message.sender_chat`, `MessageReactionUpdated.action_chat` and `Poll.voter_chat`.

I would propose adding a new shortcut to the `Update` class, `effective_sender`. It either returns the sender chats above or effective_user, if they are not set.

aelkheir:
Closes #4085.

- [ ] ~Added myself alphabetically to ``AUTHORS.rst`` (optional)~
- [ ] ~Added new classes & modules to the docs and all suitable ``__all__`` s~
- [ ] ~Checked the `Stability Policy <https://docs.python-telegram-bot.org/stability_policy.html>`_ in case of deprecations or changes to documented behavior~

-  If relevant:

### DIFF
--- a/telegram/_update.py
+++ b/telegram/_update.py
@@ -18,7 +18,7 @@
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains an object that represents a Telegram Update."""
 
-from typing import TYPE_CHECKING, Final, List, Optional
+from typing import TYPE_CHECKING, Final, List, Optional, Union
 
 from telegram import constants
 from telegram._callbackquery import CallbackQuery
@@ -224,6 +224,7 @@ class Update(TelegramObject):
     __slots__ = (
         "_effective_chat",
         "_effective_message",
+        "_effective_sender",
         "_effective_user",
         "callback_query",
         "channel_post",
@@ -371,6 +372,7 @@ class Update(TelegramObject):
         self.message_reaction_count: Optional[MessageReactionCountUpdated] = message_reaction_count
 
         self._effective_user: Optional[User] = None
+        self._effective_sender: Optional[Union["User", "Chat"]] = None
         self._effective_chat: Optional[Chat] = None
         self._effective_message: Optional[Message] = None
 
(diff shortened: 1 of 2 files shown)


telegram/_update.py
```diff
@@ -444,6 +446,56 @@ def effective_user(self) -> Optional["User"]:
...
+
+        Note:
+            * When the user is anonymous this is either :class:`telegram.Chat` or :obj:`None`.
+            * If the user is not anonymous it's either :class:`telegram.User` or :obj:`None`.
+
+        If no user whatsoever is associated with this update, this gives :obj:`None`. This
+        is the case if any of
+
+        * :attr:`channel_post`
+        * :attr:`edited_channel_post`
```
Bibo-Joshi, line 463: I strongly suggest using `message.sender_chat` also in these cases. Not including a field called `sender_chat` in this property doesn't make much sense to me :D

tests/test_update.py
```diff
@@ -261,6 +269,47 @@ def test_effective_user(self, update):
...
+        # remove none anonymous user
+        if message := (update.message or update.edited_message):
+            message._unfreeze()
+            message.from_user = None
+        elif reaction := (update.message_reaction):
+            reaction._unfreeze()
+            reaction.user = None
+        elif answer := (update.poll_answer):
+            answer._unfreeze()
+            answer.user = None
```
Bibo-Joshi, line 297: TBH I don't quite get what this test is covering that the above tests does not cover 🤔

tests/test_update.py
```diff
@@ -261,6 +269,47 @@ def test_effective_user(self, update):
...
+        sender = update.effective_sender
+        if not (
+            update.channel_post is not None
+            or update.edited_channel_post is not None
+            or update.poll is not None
+            or update.chat_boost is not None
+            or update.removed_chat_boost is not None
+            or update.message_reaction_count is not None
+        ):
+            assert sender.id == 1
```
Bibo-Joshi, line 283: I would have expected something like `assert isinstance(sender, Chat/User)` here making sure that we're getting the right attribute for the right use cases 🤔

Bibo-Joshi (changes requested):
Hi, thanks for the PR! Overall change quality looks good, just left a few comments

tests/test_update.py
aelkheir: ahhh, here i was trying to mimic updates from anonymous group members. for example a reaction coming from  anonymous user would have `actor_chat` and not `user`. but since they're both set, i remove `user`.
this may allow us to do `assert isinstance(sender, Chat)` you mentioned

tests/test_update.py
aelkheir: Right. this is why i intended to have two tests, although currently not working as you suggest..

`test_effective_sender_non_anonymous`: makes sure sender is either `User` or `None`
`test_effective_sender_anonymous`: makes sure sender is either `Chat`, `User` or `None`

i'll make them them more clear 😅

tests/test_update.py
```diff
@@ -108,8 +123,8 @@
     {"message": message},
     {"edited_message": message},
     {"callback_query": CallbackQuery(1, User(1, "", False), "chat", message=message)},
-    {"channel_post": message},
-    {"edited_channel_post": message},
+    {"channel_post": channel_post},
+    {"edited_channel_post": channel_post},
```
aelkheir, line 127: Had to create separate `message` objects here for `channel_post`s, since they always have `sender_chat` and should not be deleted as will happen to regular messages below

telegram/_update.py
```diff
@@ -444,6 +446,58 @@ def effective_user(self) -> Optional["User"]:
...
+              :attr:`telegram.Message.from_user` or :attr:`telegram.Message.sender_chat`.
+            * If :attr:`poll_answer` is present, this will give either
+              :attr:`telegram.PollAnswer.user` or :attr:`telegram.PollAnswer.voter_chat`.
+            * If :attr:`channel_post` is present, this will give
+              :attr:`telegram.Message.sender_chat`.
+
+        .. versionadded:: NEXT.VERSION
+        """
+        if self._effective_sender:
+            return self._effective_sender
```
Bibo-Joshi, line 480: codecov says that this line was not covered by the unit tests :)

Bibo-Joshi (changes requested):
Thanks for the updates! One more comment

telegram/_update.py
aelkheir: Done. Sorry for the delay.

Bibo-Joshi:
Thank you very much for contributing this enhancement 🙂

PR #4170, opened 2024-03-24
`Message.reply_*` to reply to same topic when in groups.

### LINKED ISSUE #4139: [FEATURE] Reply functions should reply to the same thread by default even without quoting

When replying to messages from a group that has topics enabled using reply methods such as `Message.reply_html`, the reply is sent to the General topic by default. This is because the `Message.reply_html` method does not preserve the original `message_thread_id`.

Upon closer inspection, I see that this is tied to the `do_quote` parameter - if this parameter is set to `True` or `None` then the message is sent to the correct topic and the original message is quoted. If this parameter is set to `False` then the message is sent to the General topic and the original message is not quoted. There are legitimate use cases where you don't want to quote the original message (e.g. for brevity) but still want the reply to end up in the same thread.

I am using version 20.8 of the library.

In my opinion, the intuitive solution is for the reply to go to the same `message_thread_id` that the message came from by default. This is consistent with the behavior of replying to the same `chat_id` by default. Of course, if the caller wants to pass a different `message_thread_id`, that possibility is still available.

One can update their code base to pass the `message_thread_id` parameter manually to the reply methods, but this seems overkill.

### Additional context

I can implement this myself and send a PR, but I was wondering if there was a specific design intent behind this decision.

aelkheir:
Closes #4139.

- [ ] ~Documented code changes according to the `CSI standard <https://standards.mousepawmedia.com/en/stable/csi.html>`__~
- [ ] ~Added myself alphabetically to ``AUTHORS.rst`` (optional)~
- [ ] ~Added new classes & modules to the docs and all suitable ``__all__`` s~
- [ ] ~Checked the `Stability Policy <https://docs.python-telegram-bot.org/stability_policy.html>`_ in case of deprecations or changes to documented behavior~

### DIFF
--- a/telegram/_message.py
+++ b/telegram/_message.py
@@ -1592,8 +1592,6 @@ class Message(MaybeInaccessibleMessage):
         chat_id, effective_reply_parameters = await self._parse_quote_arguments(
             do_quote, quote, reply_to_message_id, reply_parameters
         )
-        message_thread_id = message_thread_id or self.message_thread_id
-
         return await self.get_bot().send_message(
             chat_id=chat_id,
             text=text,
@@ -1606,7 +1604,7 @@ class Message(MaybeInaccessibleMessage):
             allow_sending_without_reply=allow_sending_without_reply,
             entities=entities,
             protect_content=protect_content,
-            message_thread_id=message_thread_id,
+            message_thread_id=message_thread_id or self.message_thread_id,
             read_timeout=read_timeout,
             write_timeout=write_timeout,
             connect_timeout=connect_timeout,
@@ -1673,8 +1671,6 @@ class Message(MaybeInaccessibleMessage):
         chat_id, effective_reply_parameters = await self._parse_quote_arguments(
             do_quote, quote, reply_to_message_id, reply_parameters
         )
-        message_thread_id = message_thread_id or self.message_thread_id
-
         return await self.get_bot().send_message(
             chat_id=chat_id,
             text=text,
@@ -1687,7 +1683,7 @@ class Message(MaybeInaccessibleMessage):
             allow_sending_without_reply=allow_sending_without_reply,
             entities=entities,
             protect_content=protect_content,
-            message_thread_id=message_thread_id,
+            message_thread_id=message_thread_id or self.message_thread_id,
             read_timeout=read_timeout,
             write_timeout=write_timeout,
             connect_timeout=connect_timeout,
@@ -1750,8 +1746,6 @@ class Message(MaybeInaccessibleMessage):
         chat_id, effective_reply_parameters = await self._parse_quote_arguments(
             do_quote, quote, reply_to_message_id, reply_parameters
         )
-        message_thread_id = message_thread_id or self.message_thread_id
-
         return await self.get_bot().send_message(
             chat_id=chat_id,
             text=text,
@@ -1764,7 +1758,7 @@ class Message(MaybeInaccessibleMessage):
             allow_sending_without_reply=allow_sending_without_reply,
             entities=entities,
             protect_content=protect_content,
-            message_thread_id=message_thread_id,
+            message_thread_id=message_thread_id or self.message_thread_id,
             read_timeout=read_timeout,
             write_timeout=write_timeout,
             connect_timeout=connect_timeout,
@@ -1827,8 +1821,6 @@ class Message(MaybeInaccessibleMessage):
         chat_id, effective_reply_parameters = await self._parse_quote_arguments(
             do_quote, quote, reply_to_message_id, reply_parameters
         )
-        message_thread_id = message_thread_id or self.message_thread_id
-
         return await self.get_bot().send_message(
             chat_id=chat_id,
             text=text,
@@ -1841,7 +1833,7 @@ class Message(MaybeInaccessibleMessage):
             allow_sending_without_reply=allow_sending_without_reply,
             entities=entities,
             protect_content=protect_content,
-            message_thread_id=message_thread_id,
+            message_thread_id=message_thread_id or self.message_thread_id,
             read_timeout=read_timeout,
             write_timeout=write_timeout,
             connect_timeout=connect_timeout,
@@ -1905,8 +1897,6 @@ class Message(MaybeInaccessibleMessage):
         chat_id, effective_reply_parameters = await self._parse_quote_arguments(
             do_quote, quote, reply_to_message_id, reply_parameters
         )
-        message_thread_id = message_thread_id or self.message_thread_id
-
         return await self.get_bot().send_media_group(
             chat_id=chat_id,
             media=media,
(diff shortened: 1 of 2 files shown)


aelkheir:
Just a few thoughts if i'm allowed to:

I don't see much value for having a message_thread_id param for reply_* functions. just as 
there isn't one for chat_id. It's a convenience method and IMO it's safe to assume that the reply should be where the 
original message existed. same chat, same thread/topic. if that's not the case, user can always 
drop down to bot.send_*. but maybe that's too breaking of a change..soo

telegram/_message.py
```diff
@@ -1581,6 +1592,8 @@ async def reply_text(
         chat_id, effective_reply_parameters = await self._parse_quote_arguments(
             do_quote, quote, reply_to_message_id, reply_parameters
         )
+        message_thread_id = message_thread_id or self.message_thread_id
```
Bibo-Joshi, line 1595: this could also be done directly in line 1609, but it doesn't matter much to me

Bibo-Joshi (commented):
Thank you for the PR. The code changes look good so far, looking forward to the unit tests :)

> I don't see much value for having a message_thread_id param for reply_* functions. just as there isn't one for chat_id. It's a convenience method and IMO it's safe to assume that the reply should be where the original message existed. same chat, same thread/topic.
If that's not the case, user can always drop down to bot.send_*. but maybe that's too breaking of a change..soo

Apart from the "breaking change" argument, I would somewhat disagree with the statement "it's safe to assume that the reply should be where the original message existed". With the recent updates, TG has significantly broadened the term "reply" by allowing you to reply to messages in other chats and this is also supported by `reply_*`. Removing the ability to reply in a differen thread would be counterintuitive for me…

aelkheir:
Thanks for the quick review 😄 
> With the recent updates, TG has significantly broadened the term "reply" by allowing you to reply to messages in other chats and this is also supported by reply_*. Removing the ability to reply in a differen thread would be counterintuitive for me…

You have a point. but for reply_* to send messages in other chats, one has to quote the message.
This also led me to discovering a bug in the current code, triggering this callback
```python
async def start(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
    await update.message.reply_text(
        "This is a reply",
        do_quote=update.message.build_reply_arguments(
            target_chat_id=1645307364,
        ),
    )
```
 from within a group topic (other than general) and with ` target_chat_id` being any private chat w/ bot, will raise `telegram.error.BadRequest: Message thread not found`..
looks like message_thread_id present in a non matching target_id confuses telegram.

telegram/_message.py
aelkheir: will do

Bibo-Joshi:
Ah, interesting finding! I guess, we should use `self.message_thread_id` only if `chat_id in {self.chat_id, self.username}`?

aelkheir:
Yeah if I understand:  `target_chat_id in {message.chat_id, message.chat.username}`?. but where should that check happen. somewhere in `bot.send_*` methods? I could file an issue if i'm following correctly.

telegram/_message.py
```diff
@@ -1593,7 +1604,7 @@ async def reply_text(
             allow_sending_without_reply=allow_sending_without_reply,
             entities=entities,
             protect_content=protect_content,
-            message_thread_id=message_thread_id,
+            message_thread_id=message_thread_id or self.message_thread_id,
```
Bibo-Joshi, line 1607:
```suggestion
            message_thread_id=message_thread_id or (self.message_thread_id if chat_id in {self.chat_id, self.username} else None),
```

Bibo-Joshi (commented):
To clarify: This problem happens on your `reply-same-thread` branch or on master?
I don't see why it would happen on master, since so far we don't pass `message_thread_id` to `send_message` within `reply_text` by default.
On your branch it would make sense. In that case, I would edit the logic as indicated below - at this point it would make sense to introduce something like `Message._parse_message_thread_id` though :D

telegram/_message.py
aelkheir: Makes much sense.
Only in my branch, in master if user passed message_thread_id by themselves but the PS. cleared it :)

Bibo-Joshi:
Hi. Just wanted to check if you need any help with the unit tests

aelkheir:
Hi, i was a bit distracted (and excited!) by the ongoing api update :). Will prioritize them.

aelkheir:
Done. Checks are failing because of the `StickerSet` incompatibility issue.

tests/test_message.py
```diff
@@ -482,6 +482,46 @@ async def make_assertion(*args, **kwargs):
...
             pytest.fail(f"reply_parameters is {reply_parameters} but should be 42")
 
+    @staticmethod
+    async def check_thread_id_parsing(
+        message: Message, method, bot_method_name: str, args, monkeypatch
+    ):
+        """Used in testing reply_* below. Makes sure that meassage_thread_id is parsed
+        correctly."""
+
+        async def make_assertion(*args, **kwargs):
```
Bibo-Joshi, line 492:
```suggestion
        async def extract_message_thread_id(*args, **kwargs):
```
Didn't really assert anything:)

tests/test_message.py
```diff
@@ -482,6 +482,46 @@ async def make_assertion(*args, **kwargs):
...
+        assert message_thread_id is None
+
+        message.message_thread_id = 99
+        message_thread_id = await method(*args)
+        assert message_thread_id == 99
+
+        message_thread_id = await method(*args, message_thread_id=50)
+        assert message_thread_id == 50
+
+        if bot_method_name != "send_chat_action":
```
Bibo-Joshi, line 508:
```suggestion
        if bot_method_name == "send_chat_action":
            return
```

Avoids unnecessary nesting :)

tests/test_message.py
```diff
@@ -482,6 +482,46 @@ async def make_assertion(*args, **kwargs):
...
+            )
+            assert message_thread_id is None
+
+            message_thread_id = await method(
+                *args,
+                do_quote=message.build_reply_arguments(
+                    target_chat_id=3,
+                ),
+            )
+            assert message_thread_id == 99
```
Bibo-Joshi, line 523:
```suggestion
                    target_chat_id=message.chat_id,
                ),
            )
            assert message_thread_id == message.message_thread_id
```
Makes it a tad more intuitive what were testing here :)

Bibo-Joshi (changes requested):
Awesome, thanks for the swift updates! Just a bit of nitpicking
For the tests, I can try to provide a temporary fix so that we can continue testing ...

Bibo-Joshi (approved):

Bibo-Joshi:
apparently we don't have code coverage reports when all tests fail. I'll try and make the sticker tests xfail for now ...

Bibo-Joshi:

PR #4189, opened 2024-04-02
feat: add the field `is_from_offline` to the class `Message`

mahdyar:


tests/test_message.py
```diff
@@ -263,6 +263,7 @@ def message(bot):
         {"reply_to_story": Story(Chat(1, Chat.PRIVATE), 0)},
         {"boost_added": ChatBoostAdded(100)},
         {"sender_boost_count": 1},
+        {"is_from_offline": True}
```
harshil21, line 266:
```suggestion
        {"is_from_offline": True},
```

telegram/_message.py
```diff
@@ -888,6 +894,7 @@ class Message(MaybeInaccessibleMessage):
         "group_chat_created",
         "has_media_spoiler",
         "has_protected_content",
+        "is_from_offline",
```
harshil21, line 897: please run `ruff check telegram/ --fix` (see contrib guide on running pre-commit)

harshil21 (changes requested):
thanks for the quick pr!

mahdyar:
> thanks for the quick pr!

Thx for the quick review! :)

harshil21:
One more thing - you also need to add an entry to `constants.MessageType` for the same field. And then make sure test_constants.py passes

harshil21:
> One more thing - you also need to add an entry to `constants.MessageType` for the same field. And then make sure test_constants.py passes

Update: Apparently you shouldn't add it to constants. Instead add an ignore entry to that field in the [test](https://github.com/python-telegram-bot/python-telegram-bot/blob/3ff7a6896c46c4efcd43c03bfd4525f2bc7a037b/tests/test_constants.py#L152)

mahdyar:
> > One more thing - you also need to add an entry to `constants.MessageType` for the same field. And then make sure test_constants.py passes
> 
> Update: Apparently you shouldn't add it to constants. Instead add an ignore entry to that field in the [test](https://github.com/python-telegram-bot/python-telegram-bot/blob/3ff7a6896c46c4efcd43c03bfd4525f2bc7a037b/tests/test_constants.py#L152)

Done.

harshil21:
Test failures are unrelated. Thanks for contributing!

PR #4204, opened 2024-04-13
Propose `check_update` to be sync and async

sergio-bershadsky:
I would love to see `check_update` to be async as it will help to use io call to another services/fs without blocking main thread, thanks!

### DIFF
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -1250,8 +1250,11 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
 
         for handlers in self.handlers.values():
             try:
-                for handler in handlers:
-                    check = handler.check_update(update)  # Should the handler handle this update?
+                for handler in handlers:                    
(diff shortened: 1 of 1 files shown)


Bibo-Joshi:
Hi. Thanks for the PR.

However, `check_update` not being a coroutine function is a concious design decision. The `check_update` should rely only on the information available in the `update`. Keep in mind that handlers are stateless in most cases. If you want to have stateful handling, changing the state should happen within the callback (as it's done e.g. in `ConversationHandler`). Using handlers [in multiple groups](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Frequently-requested-design-patterns#how-to-handle-updates-in-several-handlers) may also help.

If `check_update` were a coroutine function and could do intensive I/O operations that would significantly influence the whole setup of `Application.process_update` and how concurrency handling works in PTB: Imagine a handler with `block=False` where  `check_update` makes more I/O intensive operations than the callback. Then the advantage of `block=False` is completely gone.

Hence, I will have to reject your PR.

PR #4206, opened 2024-04-14
Removed deprecated warning message in new major version

Konano:
In the `_jobqueue.py` file, we removed a deprecated warning message that is no longer needed after v20.0. This warning was about the days parameter not being aligned with cron's weekday scheme.

As written out in [this comment](https://github.com/python-telegram-bot/python-telegram-bot/issues/4008#issuecomment-1854578113) of #4008, this warning needs to be removed in the next major version (v21).


Bibo-Joshi:

PR #4222, opened 2024-04-25
Handle Exceptions in building `CallbackContext`

Bibo-Joshi:
Explicitly catches and logs exceptions in creating Callback context objects. Relevant particularly if the user uses a custom class that overrides from_update. Since from_error calls from_update, exceptions in from_update can currently completely shut down update processing *and* error handling. See https://t.me/pythontelegrambottalk/253238 for reference.
Work in progress, e.g. tests are still missing.

### DIFF
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -1248,13 +1248,24 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
         context = None
         any_blocking = False  # Flag which is set to True if any handler specifies block=True
 
-        for handlers in self.handlers.values():
+        for handlers in self.handlers.values():  # pylint: disable=too-many-nested-blocks
             try:
                 for handler in handlers:
                     check = handler.check_update(update)  # Should the handler handle this update?
                     if not (check is None or check is False):  # if yes,
                         if not context:  # build a context if not already built
-                            context = self.context_types.context.from_update(update, self)
+                            try:
+                                context = self.context_types.context.from_update(update, self)
+                            except Exception as exc:
+                                _LOGGER.critical(
+                                    (
+                                        "Error while building CallbackContext for update %s. "
+                                        "Update will not be processed."
+                                    ),
+                                    update,
+                                    exc_info=exc,
+                                )
+                                return
                             await context.refresh_data()
                         coroutine: Coroutine = handler.handle_update(update, self, check, context)
 
@@ -1809,13 +1820,25 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
                 callback,
                 block,
             ) in self.error_handlers.items():
-                context = self.context_types.context.from_error(
-                    update=update,
-                    error=error,
-                    application=self,
-                    job=job,
-                    coroutine=coroutine,
-                )
+                try:
+                    context = self.context_types.context.from_error(
+                        update=update,
+                        error=error,
+                        application=self,
+                        job=job,
+                        coroutine=coroutine,
+                    )
+                except Exception as exc:
+                    _LOGGER.critical(
+                        (
+                            "Error while building CallbackContext for exception %s. "
+                            "Exception will not be processed by error handlers."
+                        ),
+                        error,
+                        exc_info=exc,
+                    )
+                    return False
+
                 if not block or (  # If error handler has `block=False`, create a Task to run cb
                     block is DEFAULT_TRUE
                     and isinstance(self.bot, ExtBot)
--- a/telegram/ext/_jobqueue.py
+++ b/telegram/ext/_jobqueue.py
@@ -31,6 +31,7 @@ try:
 except ImportError:
     APS_AVAILABLE = False
 
+from telegram._utils.logging import get_logger
 from telegram._utils.repr import build_repr_with_selected_attrs
 from telegram._utils.types import JSONDict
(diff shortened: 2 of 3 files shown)


telegram/ext/_application.py
```diff
@@ -1251,30 +1251,43 @@ async def process_update(self, update: object) -> None:
             try:
                 for handler in handlers:
                     check = handler.check_update(update)  # Should the handler handle this update?
-                    if not (check is None or check is False):  # if yes,
-                        if not context:  # build a context if not already built
+                    if check is None or check is False:
```
harshil21, line 1254:
```suggestion
                    if not check:
```

telegram/ext/_jobqueue.py
```diff
@@ -953,7 +955,16 @@ async def _run(
         self, application: "Application[Any, CCT, Any, Any, Any, JobQueue[CCT]]"
     ) -> None:
         try:
-            context = application.context_types.context.from_job(self, application)
+            try:
+                context = application.context_types.context.from_job(self, application)
+            except Exception as exc:
+                _LOGGER.critical(
```
harshil21, line 961: codecov shows this line wasn't covered

tests/ext/test_application.py
```diff
@@ -2421,3 +2421,81 @@ async def callback(update, context):
...
+            record = caplog.records[0]
+            assert record.name == "telegram.ext.Application"
+            assert record.getMessage().startswith(
+                "Error while building CallbackContext for update 1"
+            )
+            assert record.levelno == logging.CRITICAL
+
+            caplog.clear()
+            with caplog.at_level(logging.CRITICAL):
+                await app.process_update(2)
```
harshil21, line 2459:
```suggestion
            await app.process_update(2)
```
shouldn't the critical warning only be raised when `1` was passed?

tests/ext/test_application.py
```diff
@@ -2421,3 +2421,81 @@ async def callback(update, context):
...
+            record = caplog.records[0]
+            assert record.name == "telegram.ext.Application"
+            assert record.getMessage().startswith(
+                "Error while building CallbackContext for exception 1"
+            )
+            assert record.levelno == logging.CRITICAL
+
+            caplog.clear()
+            with caplog.at_level(logging.CRITICAL):
+                await app.process_error(update=None, error=2)
```
harshil21, line 2498:
```suggestion
            await app.process_error(update=None, error=2)
```
same

tests/ext/test_jobqueue.py
```diff
@@ -646,3 +646,44 @@ async def test_from_aps_job_missing_reference(self, job_queue):
...
+
+    async def test_run_job_exception_in_building_context(
+        self, monkeypatch, job_queue, caplog, app
+    ):
+        # Makes sure that exceptions in building the context don't stop the application
+        exception = ValueError("TestException")
+        original_from_job = CallbackContext.from_job
+
+        def raise_exception(job, application):
+            print(job, job.data, application)
```
harshil21, line 658:
```suggestion
```

tests/ext/test_jobqueue.py
```diff
@@ -646,3 +646,44 @@ async def test_from_aps_job_missing_reference(self, job_queue):
...
+        assert record.name == "telegram.ext.JobQueue"
+        assert record.getMessage().startswith(
+            "Error while building CallbackContext for job job_callback"
+        )
+        assert record.levelno == logging.CRITICAL
+
+        caplog.clear()
+        with caplog.at_level(logging.CRITICAL):
+            job_queue.run_once(job_callback, 0.1, data=2)
+            await asyncio.sleep(0.2)
```
harshil21, line 686:
```suggestion
        job_queue.run_once(job_callback, 0.1, data=2)
        await asyncio.sleep(0.2)
```

harshil21 (commented):
interesting edge case scenario

telegram/ext/_application.py
Bibo-Joshi: No, we can't change this. We explicitly document that `None` or `False` indicate that the handler should not trigger but anything else will be passed to `handle_update` https://docs.python-telegram-bot.org/en/v21.1.1/telegram.ext.basehandler.html#telegram.ext.BaseHandler.check_update. Users could be returning `[]` or some other custom object that evaulates to falsy.

telegram/ext/_jobqueue.py
Bibo-Joshi: That's weird. I blame caching. you even left a comment on the test that covers this :D 

image

tests/ext/test_application.py
Bibo-Joshi: yes, that's the point here. I'm checking that no log is issued when there is no exception :) I'll leave a comment

tests/ext/test_application.py
Bibo-Joshi: same as above

tests/ext/test_jobqueue.py
Bibo-Joshi: Same as above

telegram/ext/_application.py
harshil21: hmm I suspected that but didn't bother to check the docs, my bad.

tests/ext/test_application.py
harshil21: ah that's a bit counterintuitive

harshil21 (approved):

PR #4235, opened 2024-05-05
Bump pytest-asyncio to 0.23.6

Bibo-Joshi:
I noticed that 0.23.* is out since 12/23 but dependabot for some reasons doesn't update to that. wanted to at least check if that's working as well


Bibo-Joshi:
Nevermind, I'm just forgetful https://github.com/python-telegram-bot/python-telegram-bot/pull/4066#issuecomment-1894722749

PR #4246, opened 2024-05-09
Api 7.3 docs

Bibo-Joshi:
Several documentation updates from API 7.3. See the API changelog as well as https://t.me/bot_api_changes/133

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -2287,9 +2287,10 @@ class Bot(TelegramObject, AsyncContextManager["Bot"]):
         """
         Use this method to send audio files, if you want Telegram clients to display the file
         as a playable voice message. For this to work, your audio must be in an ``.ogg`` file
-        encoded with OPUS (other formats may be sent as Audio or Document). Bots can currently
-        send voice messages of up to :tg-const:`telegram.constants.FileSizeLimit.FILESIZE_UPLOAD`
-        in size, this limit may be changed in the future.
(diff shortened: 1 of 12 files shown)


telegram/_bot.py
harshil21, line 1: `send_location.live_period`, `IQRL.live_period`, and  `InputLocationMessageContent.live_period` docstring also changed - "Period in seconds during which the location will be updated (see <a href="https://telegram.org/blog/live-locations">Live Locations</a>, should be between 60 and 86400, or 0x7FFFFFFF for live locations that can be edited indefinitely."

harshil21 (changes requested):
https://t.me/bot_api_changes/134 can also be added

telegram/_bot.py
Bibo-Joshi: That has already been partially added in #4247 that's why I didn.t want to do it here. We'll also need a new constant for that

telegram/_business.py
```diff
@@ -246,7 +246,7 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["BusinessIntr
 
 class BusinessLocation(TelegramObject):
     """
-    This object represents the location of a business account.
+    This object Contains information about the location of a Telegram Business account.
```
aelkheir, line 249:
```suggestion
    This object contains information about the location of a Telegram Business account.
```
Notification falsely says "requested your review", so found this :)

PR #4248, opened 2024-05-09
Api 7.3 InputPollOption

Bibo-Joshi:
### Checklist

- ~[] Added myself alphabetically to ``AUTHORS.rst`` (optional)~

- New classes:

### DIFF
--- a/telegram/_poll.py
+++ b/telegram/_poll.py
@@ -51,8 +51,8 @@ class InputPollOption(TelegramObject):
         text_parse_mode (:obj:`str`, optional): |parse_mode|
             Currently, only custom emoji entities are allowed.
         text_entities (Sequence[:class:`telegram.MessageEntity`], optional): Special entities
-            that appear in the option :param:`text`. It can be specified instead of
-            :param:`text_parse_mode`.
+            that appear in the option :paramref:`text`. It can be specified instead of
+            :paramref:`text_parse_mode`.
             Currently, only custom emoji entities are allowed.
             This list is empty if the text does not contain entities.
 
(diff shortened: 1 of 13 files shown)


telegram/_poll.py
```diff
@@ -28,12 +28,79 @@
...
+        text (:obj:`str`): Option text,
+            :tg-const:`telegram.PollOption.MIN_LENGTH`-:tg-const:`telegram.PollOption.MAX_LENGTH`
+            characters.
+        text_parse_mode (:obj:`str`, optional): |parse_mode|
+            Currently, only custom emoji entities are allowed.
+        text_entities (Sequence[:class:`telegram.MessageEntity`], optional): Special entities
+            that appear in the option :paramref:`text`. It can be specified instead of
+            :paramref:`text_parse_mode`.
+            Currently, only custom emoji entities are allowed.
+            This list is empty if the text does not contain entities.
```
harshil21, line 57: do we need to mention that it's empty? Isn't it obvious?

telegram/_poll.py
```diff
@@ -28,12 +28,79 @@
...
+
+    Attributes:
+        text (:obj:`str`): Option text,
+            :tg-const:`telegram.PollOption.MIN_LENGTH`-:tg-const:`telegram.PollOption.MAX_LENGTH`
+            characters.
+        text_parse_mode (:obj:`str`): Optional. |parse_mode|
+            Currently, only custom emoji entities are allowed.
+        text_entities (Sequence[:class:`telegram.MessageEntity`]): Special entities
```
harshil21, line 65:
```suggestion
        text_entities (Sequence[:class:`telegram.MessageEntity`]): Optional. Special entities
```
or purporsely omitted?

telegram/_poll.py
Bibo-Joshi: the alternative would be that it's `None`, but we explicitly make it an empty sequence. We do that for a couple of array-type attributes, see #3057. However, reading that prs description again, I see that I explicitly opted out of doing that for input-type classes. I'll try to double check if we have some problems with this behavior

telegram/_poll.py
Bibo-Joshi: same reasoning as above

telegram/_poll.py
Bibo-Joshi: Had another closer look. Since #3249, we have *all* array-type attributes as always-tuple. This does not seem to cause any problem. The same PR made `TO.to_dict()` skip empty sequence arguments, so empty tuples are not reported to TG. I don't know why in #3057 I thought this to be a problem. In any case, never change a running system, so I'll leave it as is :D

harshil21 (approved):

PR #4250, opened 2024-05-10
Add new rules to ruff config

harshil21:
Most notably adds the "ARG" rule which I deliberately didn't add previously. https://github.com/python-telegram-bot/python-telegram-bot/pull/4180#discussion_r1554283924 prompted this.

### DIFF
--- a/telegram/_birthdate.py
+++ b/telegram/_birthdate.py
@@ -21,7 +21,6 @@ from datetime import date
 from typing import Optional
 
 from telegram._telegramobject import TelegramObject
-from telegram._utils.datetime import DTM_UTC
 from telegram._utils.types import JSONDict
 
 
--- a/telegram/_birthdate.py
+++ b/telegram/_birthdate.py
@@ -21,6 +21,7 @@ from datetime import date
 from typing import Optional
 
 from telegram._telegramobject import TelegramObject
+from telegram._utils.datetime import DTM_UTC
 from telegram._utils.types import JSONDict
 
 
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -8850,7 +8850,7 @@ CUSTOM_EMOJI_IDENTIFIER_LIMIT` custom emoji identifiers can be specified.
             api_kwargs=api_kwargs,
         )
 
-    def to_dict(self, _: bool = True) -> JSONDict:
+    def to_dict(self, recursive: bool = True) -> JSONDict:  # noqa: ARG002
         """See :meth:`telegram.TelegramObject.to_dict`."""
         data: JSONDict = {"id": self.id, "username": self.username, "first_name": self.first_name}
 
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -8850,7 +8850,7 @@ CUSTOM_EMOJI_IDENTIFIER_LIMIT` custom emoji identifiers can be specified.
             api_kwargs=api_kwargs,
         )
 
-    def to_dict(self, recursive: bool = True) -> JSONDict:
+    def to_dict(self, _: bool = True) -> JSONDict:
         """See :meth:`telegram.TelegramObject.to_dict`."""
         data: JSONDict = {"id": self.id, "username": self.username, "first_name": self.first_name}
 
--- a/telegram/_message.py
+++ b/telegram/_message.py
@@ -198,7 +198,7 @@ class MaybeInaccessibleMessage(TelegramObject):
             data["date"] = from_timestamp(data["date"], tzinfo=loc_tzinfo)
 
         data["chat"] = Chat.de_json(data.get("chat"), bot)
-        return super()._de_json(data=data, bot=bot)
+        return super()._de_json(data=data, bot=bot, api_kwargs=api_kwargs)
 
 
 class InaccessibleMessage(MaybeInaccessibleMessage):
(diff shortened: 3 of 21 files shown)


telegram/_birthdate.py
```diff
@@ -85,4 +86,4 @@ def to_date(self, year: Optional[int] = None) -> datetime:
                 "The `year` argument is required if the `year` attribute was not present."
             )
 
-        return datetime(year or self.year, self.month, self.day)  # type: ignore[arg-type]
```
Bibo-Joshi, line 88: curious: what's this about?

telegram/_bot.py
```diff
@@ -8867,7 +8867,7 @@ async def replace_sticker_in_set(
             api_kwargs=api_kwargs,
         )
 
-    def to_dict(self, recursive: bool = True) -> JSONDict:
```
Bibo-Joshi, line 8870: I don't think this makes sense. If you have an `obj` and only know that it's an instance of `TelegramObject`, you want to be able to call `obj.to_dict(recrusive=False)`. if `isinstance(obj, Bot)`, that won't work this way.

telegram/_message.py
```diff
@@ -198,7 +198,7 @@ def _de_json(
             data["date"] = from_timestamp(data["date"], tzinfo=loc_tzinfo)
 
         data["chat"] = Chat.de_json(data.get("chat"), bot)
-        return super()._de_json(data=data, bot=bot)
+        return super()._de_json(data=data, bot=bot, api_kwargs=api_kwargs)
```
Bibo-Joshi, line 201: wtf, how did we miss this so far? why don't the tests complain about that? o.0

telegram/ext/_baseupdateprocessor.py
```diff
@@ -165,7 +165,7 @@ class SimpleUpdateProcessor(BaseUpdateProcessor):
 
     async def do_process_update(
         self,
-        update: object,
+        update: object,  # noqa: ARG002
```
Bibo-Joshi, line 168: tbh this seems like a bug in ARG002 - is the parent class defines this, ruff shouldn't complain …

telegram/ext/filters.py
```diff
@@ -522,14 +522,14 @@ def name(self) -> str:
...
-    def name(self, name: str) -> NoReturn:
+    def name(self, _: str) -> NoReturn:
         raise RuntimeError("Cannot set name for combined filters.")
 
 
 class _All(MessageFilter):
     __slots__ = ()
 
-    def filter(self, message: Message) -> bool:
+    def filter(self, _: Message) -> bool:
```
Bibo-Joshi, line 532: false positive again

Bibo-Joshi (commented):
interesting updates!

telegram/_birthdate.py
harshil21: a new rule I added complains about datetime objects that don't have a tzinfo. I can ignore it if you like cause I think that's technically a very small breaking change.

telegram/_message.py
harshil21: no idea, I was pretty surprised too

telegram/ext/_baseupdateprocessor.py
harshil21: I'll open an issue on their repo and see what they think

telegram/_birthdate.py
Bibo-Joshi: Ah, wait a minute: I think the real issue here is that we return a date*time* while we should return only a `date`. for datetimes it can arguably make sense that one should not use naive objects, for dates that is not the case. I'd classify this as bug on our side.

telegram/ext/_baseupdateprocessor.py
Bibo-Joshi: I'd still be okay with merging btw

telegram/_birthdate.py
harshil21: oh wow I forgot datetime.date existed lol

Bibo-Joshi (approved):
LEGTM, then :) Let me know if I should merge or if you have additional changes planned

harshil21:
all done from my side. Can be merged

telegram/ext/_baseupdateprocessor.py
Bibo-Joshi: btw, did you open an issue?

harshil21: No, I didn't get the time to. Do you want to open one?

Bibo-Joshi: I doug around a bit and found this discussion here: https://github.com/astral-sh/ruff/issues/10409#issuecomment-1998327059
Unfortunately the [`typing.override`](https://docs.python.org/3.12/library/typing.html#typing.override) decorator is new in Py3.12 and I wasn't able to trick `ruff` into using similar behavior by things like 

```python
try:
    from typing import override
except ImportError:
    override = lambda f: f
```
I guess I'll put it on our python version todo-list :)

Bibo-Joshi: https://github.com/python-telegram-bot/python-telegram-bot/projects/8#card-93004105

harshil21: makes sense actually.

PR #4253, opened 2024-05-11
Support Python 3.13 beta

harshil21:
We should probably wait for `cffi` to release their new version to target 3.13, otherwise passport related functionality won't work. Tracking: https://github.com/python-cffi/cffi/issues/23, https://github.com/python-cffi/cffi/issues/71


harshil21:
If possible we should do #4129 before this is merged so it's easier to change `setup.py`. `test_meta` fails on 3.13 because of the change to `locals()` - https://docs.python.org/3.13/library/functions.html#locals. I don't want to change that part yet because I don't want to potentially break the release

harshil21:
> I don't want to change that part yet because I don't want to potentially break the release

ironically if we don't merge this, then all ptb installations on py 3.13 will fail by default. And we don't have python version specific wheels so we can build it on another version and it should be good to go. 

uses pre-release of cffi 1.17.0 since they don't plan to release the full version until 3.13 rc1

harshil21:
Merged master so we should be good to go. Also verified that `pip install .[passport]` correctly installs cffi 1.17.0 on python 3.13. 

Not sure why type completeness is unhappy. All ruff did was remove some quotes around the annotation and that failed this test apparently.

pyproject.toml
```diff
@@ -36,6 +36,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
```
Bibo-Joshi, line 39: Do you want to list this in beta-support already or should we wait until 3.13 has an official release?

telegram/ext/_application.py
```diff
@@ -251,40 +251,45 @@ class Application(Generic[BT, CCT, UD, CD, BD, JQ], AsyncContextManager["Applica
...
+            "post_stop",
+            "update_queue",
+            "updater",
+            "user_data",
+        )
+        + ("__weakref__",)
+        if sys.version_info >= (3, 13)
+        else ()
     )
+    # Allowing '__weakref__' creation here since we need it for the JobQueue
```
Bibo-Joshi, line 290: can we move this code block to *above* the `+ ("__wearef__")` line or does black move that around?

tests/conftest.py
```diff
@@ -291,6 +292,5 @@ def timezone(tzinfo):
 
 
 @pytest.fixture()
-def tmp_file(tmp_path):
-    with tmp_path / uuid4().hex as file:
-        yield file
+def tmp_file(tmp_path) -> Path:
+    return tmp_path / uuid4().hex
```
Bibo-Joshi, line 296: just curious, what is this change about?

tests/auxil/pytest_classes.py
```diff
@@ -89,6 +89,10 @@ class PytestMessage(Message):
     pass
 
 
+class PytestUpdater(Updater):
```
Bibo-Joshi, line 92: I guess this is only to a add `__dict__` (i.e. circumvent slots)? What changed in py3.13 that we need this in the tests for Updater now?

tests/ext/test_application.py
```diff
@@ -1581,7 +1581,13 @@ def thread_target():
         async def post_init(app: Application) -> None:
             events.append("post_init")
 
-        app = Application.builder().bot(one_time_bot).post_init(post_init).build()
+        app = (
+            Application.builder()
+            .application_class(PytestApplication)
+            .updater(PytestUpdater(one_time_bot, asyncio.Queue()))
```
Bibo-Joshi, line 1587: ah, you use `PytestApplication` in more places, too. What's the reason behind this?

tests/request/test_request.py
```diff
@@ -72,7 +73,7 @@ async def make_assertion(*args, **kwargs):
 
 @pytest.fixture()
 async def httpx_request():
-    async with HTTPXRequest() as rq:
+    async with NonchalantHttpxRequest() as rq:
```
Bibo-Joshi, line 76: is this a change due to py3.13 or is this rather a stability improvement? If so, I would like to keept it for #4271, tbh …

tests/request/test_request.py
```diff
@@ -130,15 +131,13 @@ def test_slot_behaviour(self):
             assert getattr(inst, at, "err") != "err", f"got extra slot '{at}'"
         assert len(mro_slots(inst)) == len(set(mro_slots(inst))), "duplicate slot"
 
-    async def test_context_manager(self, monkeypatch):
+    async def test_context_manager(self, monkeypatch, httpx_request):
```
Bibo-Joshi, line 134: these changes look wrong to me. we monkeypath init & shutdown before using `httpx_request` as context manager - I don't think that we should use an already initialized object for that.

tests/test_bot.py
```diff
@@ -253,7 +253,7 @@ async def initialize(*args, **kwargs):
         async def stop(*args, **kwargs):
             self.test_flag.append("stop")
 
-        temp_bot = PytestBot(token=bot.token)
+        temp_bot = PytestBot(token=bot.token, request=NonchalantHttpxRequest())
```
Bibo-Joshi, line 256: stability improvement?

pyproject.toml
```diff
@@ -82,6 +83,7 @@ job-queue = [
 ]
 passport = [
     "cryptography!=3.4,!=3.4.1,!=3.4.2,!=3.4.3,>=39.0.1",
+    "cffi >= 1.17.0rc1; python_version > '3.12'"
```
Bibo-Joshi, line 86: can you please add a comment aboud cffi being a dependency of crypto that added 3.13 support only in 1.17.0rc1?

telegram/_chatfullinfo.py
```diff
@@ -490,10 +490,10 @@ def __init__(
             self.unrestrict_boost_count: Optional[int] = unrestrict_boost_count
             self.custom_emoji_sticker_set_name: Optional[str] = custom_emoji_sticker_set_name
             self.birthdate: Optional[Birthdate] = birthdate
-            self.personal_chat: Optional["Chat"] = personal_chat
-            self.business_intro: Optional["BusinessIntro"] = business_intro
-            self.business_location: Optional["BusinessLocation"] = business_location
-            self.business_opening_hours: Optional["BusinessOpeningHours"] = business_opening_hours
+            self.personal_chat: Optional[Chat] = personal_chat
```
Bibo-Joshi, line 493: I'm wondering about these change. I gguess you didn't remove *all* double quotes on types - otherwise we'd probably get a bunch of import errors? So these instances where marked as superfluous by ruff? Why was this marked only after adding py3.13 and not earlier? note that the type completeness workflow runs on py3.9. updating pyright in the workflow might also be an option.

Bibo-Joshi (commented):
Hi, thanks for the updates :) I left a bunch of comments🙆‍♂️

pyproject.toml
harshil21: why not, if we are fully python 3.13 compatible, we can list it immedieately no?

tests/conftest.py
harshil21: Tests error out:

```python
    def tmp_file(tmp_path) -> Path:
>       with tmp_path / uuid4().hex as file:
E       TypeError: 'PosixPath' object does not support the context manager protocol
```

tests/auxil/pytest_classes.py
harshil21: > I guess this is only to a add `__dict__` (i.e. circumvent slots)? 

Yes

> What changed in py3.13 that we need this in the tests for Updater now?

`Updater`'s superclass, `AsyncContextManager`, now has slots in python 3.13.

tests/ext/test_application.py
harshil21: since the slots for application now actually don't have a `__dict__`, we can get an error like:

``` python
>       monkeypatch.setattr(
            app,
            "stop",
            call_after(app.stop, lambda _: events.append("app.stop")),
        )
E       AttributeError: 'Application' object attribute 'stop' is read-only
```
so we need PytestApplication to actually be able to monkeypatch again

tests/request/test_request.py
harshil21: yeah I just realized it yesterday. Guess I'll just have to use `NonchalantHTTPXRequest` for that line.

tests/test_bot.py
harshil21: same reason above.

telegram/_chatfullinfo.py
harshil21: > otherwise we'd probably get a bunch of import errors? So these instances where marked as superfluous by ruff?

yeah, these were autocorrected by ruff.

> Why was this marked only after adding py3.13 and not earlier? 

probably because we missed updating the `target_version` for python 3.12 - https://github.com/python-telegram-bot/python-telegram-bot/pull/4253/commits/585b362ad4773c3121408ac2efb9b090fc3ca0f6#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711. Now that it autodetects it, this shouldn't happen again.

> note that the type completeness workflow runs on py3.9. updating pyright in the workflow might also be an option

ok let me upgrade it and check. I can't seem to run it locally btw, don't know what I'm doing wrong: 

```
Module name: "telegram"
Package directory: ""
Module directory: ""
   error: No py.typed file found

Symbols used in public interface:

Symbols exported by "telegram": 0
  With known type: 0
  With ambiguous type: 0
  With unknown type: 0
    (Ignoring unknown types imported from other packages)

```

tests/request/test_request.py
Bibo-Joshi: I see. NonchalantHR xfails on timeout and retry-after, though … I guess for the cases were we monkeypatch the request anyways, it doesn't matter, but please double check if there any new usages of `NHR` where we should actually better care to always make a real request

pyproject.toml
```diff
@@ -83,6 +83,7 @@ job-queue = [
 ]
 passport = [
     "cryptography!=3.4,!=3.4.1,!=3.4.2,!=3.4.3,>=39.0.1",
+    # cffi is a dependency of cryptography and needs a pre-release version to support python 3.13
```
Bibo-Joshi, line 86:
```suggestion
    # cffi is a dependency of cryptography and added support python 3.13 in 1.17.0rc1
```

telegram/_chatfullinfo.py
Bibo-Joshi: > probably because we missed updating the target_version for python 3.12

dedum …

> I can't seem to run it locally btw, don't know what I'm doing wrong:

you need to install ptb locally as done in the workflow, i.e. `pip install .`

The quotes are a bit confusing and I'm looking forward to and https://peps.python.org/pep-0563/ being implemented by default (not only via `__future__`). e.g. in chatfullinfo the business* classes are imported only in the `if TYPE_CHECKING` clause 

https://github.com/python-telegram-bot/python-telegram-bot/blob/eca420ae4db56d104fa19d14a8cade2c709ca24d/telegram/_chatfullinfo.py#L34-L35

so that they are not available at definition time of the class and  `__init__` has to use the quotes

https://github.com/python-telegram-bot/python-telegram-bot/blob/eca420ae4db56d104fa19d14a8cade2c709ca24d/telegram/_chatfullinfo.py#L404-L406

But for some reason, it's okay to drop the quotes in the boday of `__init__`:

https://github.com/python-telegram-bot/python-telegram-bot/blob/eca420ae4db56d104fa19d14a8cade2c709ca24d/telegram/_chatfullinfo.py#L493-L496

Trying to insert `print(BusinessIntro)` or `print(Optional[BI])` in the `__init__` raises a `NameError` as expected. WTF. And why are the quotes removed by ruff only if we add py3.12 support? If that's a new feature in 3.12 that's not supported in 3.11-, ruff should stay away from that or the tests should fail.

So many questions 🤯 Anyways, as long as the tests succedd I'm happy :D

Bibo-Joshi: Regarding type completeness it looks to me like pyright isn't happey with `JobQueue.scheduler for some reason

```
telegram.ext._jobqueue.JobQueue.scheduler
  c:\Users\hinri\PycharmProjects\python-telegram-bot\venv311\Lib\site-packages\telegram\ext\_jobqueue.py:111:14 - error: Type unknown for variable "telegram.ext._jobqueue.JobQueue.scheduler"
```

That makes the type of `JobQueue` partially unknown and also for every class that `JobQueue` is referenced in … I haven't really figured out what pyrights problem is with aps though …

tests/request/test_request.py
harshil21: just checked, none of them make an actual request.

Bibo-Joshi (approved):
LGTM, then, thanks for the work :) Anyhting to add from your side or should I merge?

harshil21:
All good from my side!

telegram/ext/_jobqueue.py
```diff
@@ -106,7 +106,7 @@ def __init__(self) -> None:
 
         self._application: Optional[weakref.ReferenceType[Application]] = None
         self._executor = AsyncIOExecutor()
-        self.scheduler: AsyncIOScheduler = AsyncIOScheduler(**self.scheduler_configuration)
+        self.scheduler: "AsyncIOScheduler" = AsyncIOScheduler(**self.scheduler_configuration)
```
harshil21, line 109: ruff complains about this for me locally (i'm on py 3.13 so that's possibly why now)

harshil21:
also forget to mention that for some reason mypy segfaults on my build of python 3.13b1. Does anyone else have the same issue?

Bibo-Joshi:
Did not try yet ...

PR #4265, opened 2024-05-21
Update Settings for pre-commit.ci

Bibo-Joshi:
This does two things:

* Makes the updates run quarterly instead of weekly. IMO the monthly updates have been a bit too much in the past, especially since these are "just" code quality dependencies
* Instead of using the default commit message `[pre-commit.ci] pre-commit autoupdate` it sets a commit message that is roughly in the same style as the dependabot commit messages. Probably mostly personal preference but at least the `[pre-commit.ci]` part is unnecessary IMO and so far I've always removed it.

See https://pre-commit.ci/#configuration for reference.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,8 @@
 
 ci:
     autofix_prs: false
-    autoupdate_schedule: monthly
+    autoupdate_schedule: quarterly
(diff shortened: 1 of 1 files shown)


harshil21 (approved):
Ok. But if I see a new ruff version with new features i'm definitely sending a pr asap

PR #4266, opened 2024-05-21
Remove Functionality Deprecated in v21.2

### LINKED ISSUE #4244: Remove Functionality Deprecated by API 7.3
Create a PR for that as immediate follow-Up for #4243. That way it's easier than doing that when the next API update drops.

harshil21:
Closes #4244


docs/source/conf.py
```diff
@@ -25,7 +25,7 @@
 release = "21.2"  # telegram.__version__
 
 # If your documentation needs a minimal Sphinx version, state it here.
-needs_sphinx = "6.1.3"
```
Bibo-Joshi, line 28: why is this required?

telegram/_chatfullinfo.py
```diff
@@ -19,58 +19,375 @@
...
 
-class ChatFullInfo(Chat):
+class ChatFullInfo(_ChatBase):
     """
     This object contains full information about a chat.
 
     Objects of this class are comparable in terms of equality. Two objects of this class are
     considered equal, if their :attr:`~telegram.Chat.id` is equal.
 
-    Caution:
```
Bibo-Joshi, line 43: maybe add a versionchanged for this?

telegram/_chatfullinfo.py
```diff
@@ -19,58 +19,375 @@
...
             .. versionadded:: 21.2
+        title (:obj:`str`, optional): Title, for supergroups, channels and group chats.
+        username (:obj:`str`, optional): Username, for private chats, supergroups and channels if
+            available.
+        first_name (:obj:`str`, optional): First name of the other party in a private chat.
+        last_name (:obj:`str`, optional): Last name of the other party in a private chat.
+        is_forum (:obj:`bool`, optional): :obj:`True`, if the supergroup chat is a forum
+            (has topics_ enabled).
+
+            .. versionadded:: 20.0
```
Bibo-Joshi, line 69: I guess we *could* remove all of these versioning directives since `CFI` is actually a new class 🤔 Would also be fine with keeping for completeness

telegram/_chatfullinfo.py
```diff
@@ -120,47 +437,99 @@ def __init__(
...
+        data["business_intro"] = BusinessIntro.de_json(data.get("business_intro"), bot)
+        data["business_location"] = BusinessLocation.de_json(data.get("business_location"), bot)
+        data["business_opening_hours"] = BusinessOpeningHours.de_json(
+            data.get("business_opening_hours"), bot
+        )
+
+        api_kwargs = {}
+        # This is a deprecated field that TG still returns for backwards compatibility
+        # Let's filter it out to speed up the de-json process
+        if (amaa := "all_members_are_administrators") in data:
```
Bibo-Joshi, line 532: could this also still be relevant for the `Chat` class? if so, it could be moved to `_ChatBase`

tests/test_chatfullinfo.py
Bibo-Joshi, line 1: Can we mention somewhere here that all the shortcut methods are tested in `test_chat`?

Bibo-Joshi (commented):
Thanks you for the PR! I left a few comments :)

docs/source/conf.py
harshil21: just updated it because it seemed quite outdated

telegram/_chatfullinfo.py
harshil21: what should I write? that you can access Chat's shortcuts as well?

telegram/_chatfullinfo.py
harshil21: would be nice to have a history

docs/source/conf.py
Bibo-Joshi: If we don't have a fixed need for that, I don't see a strict reason to do that 🤔  in any case, I would rather have that on doc-fixes …

telegram/_chatfullinfo.py
Bibo-Joshi: all right, let's keep then

telegram/_chatfullinfo.py
```diff
@@ -44,6 +44,9 @@ class ChatFullInfo(_ChatBase):
 
     .. versionadded:: 21.2
 
+    .. versionchanged:: NEXT.VERSION
+        You can now also use the shortcut methods of :class:`telegram.Chat` on this object.
+
```
Bibo-Joshi, line 49: these are already available right now b/c CFI is a subclass of Chat … At least the wording is a bit confusing to me. Maybe

```suggestion
    .. versionchanged:: NEXT.VERSION
        Explicit support for all shortcut methods known from :class:`telegram.Chat` on this object. Previously those were only available bcause this class inherited from :class:`telegram.Chat`.
```
?

telegram/_chatfullinfo.py
```diff
@@ -526,10 +529,4 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["ChatFullInfo
...
         )
 
-        api_kwargs = {}
-        # This is a deprecated field that TG still returns for backwards compatibility
-        # Let's filter it out to speed up the de-json process
-        if (amaa := "all_members_are_administrators") in data:
-            api_kwargs[amaa] = data.pop(amaa)
-
-        return super()._de_json(data=data, bot=bot, api_kwargs=api_kwargs)
+        return super().de_json(data=data, bot=bot)
```
Bibo-Joshi, line 532: 🥳

PR #4271, opened 2024-05-22
Make network based tests as concurrent as possible

harshil21:
Ideas implemented:

- ... WIP

Benchmarks:

- WIP


Bibo-Joshi:
> * [ ]  Introduce `pytest-socket` and completely disable network access for `*WithoutRequest` tests

Just question for my understanding: This is intended to make any tests noticed that should instead go to `WithRequest`?

> * [ ]  Extend that naming scheme to `telegram.ext` tests

Instead of splitting test classes for tg.ext, I would rather just disable network access for tg.ext-tests completely and mock every interaction by TG.

> * [ ]  Introduce `pyfakefs` to test suite

Which problem does this solve?

> * [ ]  Check if using `uvloop` in tests provides any measurable speed up.

Sounds like an interseting thing to check out, however I see great value in testing on the default setups on the different os-ses.

> * [ ]  Change CI's pytest  `--dist` to `worksteal`.
> * [ ]  Isolate tests from other tests
> * [ ]  Continue above point and introduce `pytest-randomly` (need to check if pytest-xdist effectively does the job of randomly)

Not sure if I see a notable benefit in this, but I have no objection if you can manage to work around tests cases like `sticker_set` that currently rely on the order :D 

> * [ ]  Profile all tests and see what clogs up and fix that.
> * [ ]  solve the TODO in `test_official.exceptions`
> * [ ]  [`dirty-equals`](https://dirty-equals.helpmanual.io/latest/) sounds like it could be quite useful for us!

sounds interesting :) feel free to try it out!

harshil21:
> question for my understanding: This is intended to make any tests noticed that should instead go to `WithRequest`?

Yes, and also there are currently some tests in there which don't access the network, but whose fixtures do access the network.

> Instead of splitting test classes for tg.ext, I would rather just disable network access for tg.ext-tests completely and mock every interaction by TG.

good idea. Ideally one should be able to test locally without network connection (<1 min with pytest-xdist) and be sure that the ones which do use the network pass.

> Which problem does this solve?

hm, it is supposed to be faster (since there's no disk I/O) and safer, but we don't do much I/O so probably not worth it.

> Not sure if I see a notable benefit in this, but I have no objection if you can manage to work around tests cases like `sticker_set` that currently rely on the order :D

it is one of the possible reasons for the flakiness, and it is good practice to keep tests isolated from each other

Bibo-Joshi:
> > question for my understanding: This is intended to make any tests noticed that should instead go to `WithRequest`?
> 
> Yes, and also there are currently some tests in there which don't access the network, but whose fixtures do access the network.

👍 

> > Instead of splitting test classes for tg.ext, I would rather just disable network access for tg.ext-tests completely and mock every interaction by TG.
> 
> good idea. Ideally one should be able to test locally without network connection (<1 min with pytest-xdist) and be sure that the ones which do use the network pass.

👍 

> > Which problem does this solve?
> 
> hm, it is supposed to be faster (since there's no disk I/O) and safer, but we don't do much I/O so probably not worth it.

I was at least wondering, since we e.g. already use pytests tmp_path fixtures.

> > Not sure if I see a notable benefit in this, but I have no objection if you can manage to work around tests cases like `sticker_set` that currently rely on the order :D
> 
> it is one of the possible reasons for the flakiness, and it is good practice to keep tests isolated from each other

👍

tests/conftest.py
```diff
@@ -180,6 +180,23 @@ async def tz_bot(timezone, bot_info):
         return default_bot
 
 
+@pytest.fixture(scope="session")
+async def tz_bots(tzinfos, bot_info):
```
Bibo-Joshi, line 184: 1) I recommend to add a comment in this fixture about why we use a list here instead of properly parametrizing (speed)
2) do we still need the `tz_bot` fixture for anything else, or can it be removed, then?
3) Can we make sure that the tests at least point out the timezone on failure?

tests/conftest.py
```diff
@@ -309,6 +326,18 @@ def tzinfo(request):
     return BasicTimezone(offset=datetime.timedelta(hours=hours_offset), name=request.param)
 
 
+@pytest.fixture(scope="session")
+def tzinfos():
```
Bibo-Joshi, line 330: what's the benefit of having this as a fixture? IISC it's used only in `tz_bots` and could also just be hardcoded there?

Bibo-Joshi (commented):
Hey Harshil, I had a look at the changes that are here so far.

What would you think of converting the wishlist in the comments into an issue and adressing them in multiple separate PR? that would give as the benefit of the individual points earlier, make reviewing probably easier and would also make it easier to distribute the open points onto different people :)

tests/conftest.py
harshil21: > do we still need the tz_bot fixture for anything else, or can it be removed, then?

we do still use it in the `NonRequest` tests, where the time penalty is very negligible.

> Can we make sure that the tests at least point out the timezone on failure?

that would be hard to do without making it even more complex, but let me have an another look again

tests/conftest.py
harshil21: mm yeah, a bit of a premature optimization at this stage in the PR, but yes I'll keep it mind

harshil21:
created a new issue for the other points. I will have to rebase this branch and remove commits which are not relevant to the purpose of this PR

Bibo-Joshi:
@harshil21 I'd like to close this PR. #4488, #4489 and #4491 covered most.
What's not covered is converting the `tz_bot` fixture to a list. Since #4491 and #4493 take away much of the performance penalty of that fixture, I'd prefer to keep the tests well separated

harshil21:
fair enough.

Sorry I couldn't complete it over several months.

PR #4304, opened 2024-06-18
Log Received Data on Deserialization Errors

Bibo-Joshi:
as discussed offline. makes debugging easier.
unit tests need to be updated

### DIFF
--- a/tests/test_bot.py
+++ b/tests/test_bot.py
@@ -533,6 +533,29 @@ class TestBotWithoutRequest:
             123, "text", api_kwargs={"unknown_kwarg_1": 7, "unknown_kwarg_2": 5}
         )
 
+    async def test_get_updates_deserialization_error(self, bot, monkeypatch, caplog):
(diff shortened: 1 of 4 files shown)


tests/test_bot.py
```diff
@@ -533,6 +533,29 @@ async def post(url, request_data: RequestData, *args, **kwargs):
...
+                HTTPStatus.OK,
+                b'{"ok": true, "result": [{"update_id": "1", "message": "unknown_format"}]}',
+            )
+
+        monkeypatch.setattr(HTTPXRequest, "do_request", faulty_do_request)
+
+        bot = ExtBot(get_updates_request=HTTPXRequest(), token=bot.token)
+
+        with caplog.at_level(logging.CRITICAL), pytest.raises(AttributeError):
+            await bot.get_updates()
```
harshil21, line 548: does the bot not need initializing? If so, it's better to use `PytestExtBot` instead

harshil21 (approved):

PR #4311, opened 2024-06-20
API 7.5 new classes

harshil21:
### Checklist

- New classes:

- Added new shortcuts:

   - [ ]  In :class:`~telegram.Chat` & :class:`~telegram.User` for all methods that accept ``chat/user_id``

-  If relevant:

   - [ ] Added the new method(s) to ``_extbot.py``
   - [ ] Added or updated ``bot_methods.rst``


telegram/__init__.py
```diff
@@ -242,6 +252,19 @@
     "warnings",
 )
 
+from telegram._stars import (
```
Bibo-Joshi, line 255: the rest of the import statements uses `from ._*` and is sorted alphabetically. If possible, let's try to stay in that schema :)

telegram/_stars.py
```diff
@@ -0,0 +1,426 @@
...
+    @classmethod
+    def de_json(
+        cls, data: Optional[JSONDict], bot: "Bot"
+    ) -> Optional["RevenueWithdrawalStateSucceeded"]:
+        data = cls._parse_data(data)
+
+        if not data:
+            return None
+
+        if "date" in data:
```
Bibo-Joshi, line 146: Do we need the if-clause? `date` is a required argument

telegram/_stars.py
```diff
@@ -0,0 +1,426 @@
...
+    .. versionadded:: NEXT.VERSION
+
+    Args:
+        date (:obj:`datetime.datetime`): Date the withdrawal was completed as a datetime object.
+        url (:obj:`str`): An HTTPS URL that can be used to see transaction details.
+
+    Attributes:
+        type (:obj:`str`): The type of the state, always
+            :tg-const:`telegram.RevenueWithdrawalState.SUCCEEDED`.
+        date (:obj:`datetime.datetime`): Date the withdrawal was completed as a datetime object.
```
Bibo-Joshi, line 118: should the date be added to the equality comparison?

telegram/_stars.py
```diff
@@ -0,0 +1,426 @@
...
+    def __init__(self, type: str, *, api_kwargs: Optional[JSONDict] = None) -> None:
+        super().__init__(api_kwargs=api_kwargs)
+        self.type: str = type
+
+        self._id_attrs = (self.type,)
+        self._freeze()
+
+    @classmethod
+    def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["TransactionPartner"]:
+        print(data)
```
Bibo-Joshi, line 208: 🙃

telegram/_stars.py
```diff
@@ -0,0 +1,426 @@
...
+
+    .. versionadded:: NEXT.VERSION
+
+    Args:
+        user (:class:`telegram.User`): Information about the user.
+
+    Attributes:
+        type (:obj:`str`): The type of the transaction partner,
+            always :tg-const:`telegram.TransactionPartner.USER`.
+        user (:class:`telegram.User`): Information about the user.
```
Bibo-Joshi, line 282: should that be added to equality comparison?

telegram/_stars.py
```diff
@@ -0,0 +1,426 @@
...
+
+    def __init__(self, *, api_kwargs: Optional[JSONDict] = None) -> None:
+        super().__init__(type=TransactionPartner.OTHER, api_kwargs=api_kwargs)
+
+
+class StarTransaction(TelegramObject):
+    """Describes a Telegram Star transaction.
+
+    Objects of this class are comparable in terms of equality. Two objects of this class are
+    considered equal, if their :attr:`id`, :attr:`amount`, and :attr:`date` are equal.
```
Bibo-Joshi, line 325: looking at the attribute descriptions I would have selected id, source, receiver. My reasoning:

* id is unique up to describing if its a original transaction or a refund
* the latter is identified by source and receiver

telegram/_stars.py
```diff
@@ -0,0 +1,426 @@
...
+
+        return super().de_json(data=data, bot=bot)  # type: ignore[return-value]
+
+
+class StarTransactions(TelegramObject):
+    """
+    Contains a list of Telegram Star transactions.
+
+    Objects of this class are comparable in terms of equality. Two objects of this class are
+    considered equal, if their :attr:`transactions` is equal.
```
Bibo-Joshi, line 406:
```suggestion
    considered equal, if their :attr:`transactions` are equal.
```
?

Bibo-Joshi (commented):
Hey : ) I noticed the "WIP" label too late, so here you already have somen comment 😅

telegram/_stars.py
harshil21: I just use required arguments by default for comparison, source and receiver are optional so do you still want that?

Bibo-Joshi: using required arguments is the sane thing to do in most cases, true, but IMO more meaningful comparison overrules that. Plus, for mutually exclusive optionals, using all of them is a pattern that we already use e.g. in IKB. If my reasoning makes sense to you, then I'd like to go that way :)

telegram/_stars.py
```diff
@@ -237,8 +237,8 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["TransactionP
             cls.OTHER: TransactionPartnerOther,
         }
 
-        if cls is TransactionPartner and data.get("type") in _class_mapping:
-            return _class_mapping[data.pop("type")].de_json(data=data, bot=bot)
+        if cls is TransactionPartner and data.get("type") in _class_mapping:  # type: ignore
```
harshil21, line 240: had to use the type ignore here, since I'm not doing a:
```python
if not data:
    return None
```
since this breaks the `de_json` for `TransactionPartnerOther` (empty class with only `type` key which is popped so `data` becomes empty and thus `None` is returned). But we used the same approach for `MenuButtonCommands` which works just fine so i'm not sure what I did wrong.

telegram/_stars.py
```diff
@@ -387,31 +416,31 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["StarTransact
         if not data:
             return None
```
Bibo-Joshi, line 417: codecov says this was not covered …

telegram/_stars.py
```diff
@@ -387,31 +416,31 @@ def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["StarTransact
...
-        if "date" in data:
-            # Get the local timezone from the bot if it has defaults
-            loc_tzinfo = extract_tzinfo_from_defaults(bot)
-            data["date"] = from_timestamp(data["date"], tzinfo=loc_tzinfo)
+        # Get the local timezone from the bot if it has defaults
+        loc_tzinfo = extract_tzinfo_from_defaults(bot)
+        data["date"] = from_timestamp(data.get("date", None), tzinfo=loc_tzinfo)
 
         data["source"] = TransactionPartner.de_json(data.get("source"), bot)
+        print(data.get("receiver"))
```
Bibo-Joshi, line 424: 🙃

telegram/_stars.py
```diff
@@ -420,7 +449,17 @@ def __init__(
...
 
         self._id_attrs = (self.transactions,)
         self._freeze()
+
+    @classmethod
+    def de_json(cls, data: Optional[JSONDict], bot: "Bot") -> Optional["StarTransactions"]:
+        data = cls._parse_data(data)
+
+        if not data:
+            return None
```
Bibo-Joshi, line 462: ☂️ 🙂

tests/test_stars.py
```diff
@@ -133,15 +263,16 @@ def test_de_json_star_transaction_localization(self, star_transaction, tz_bot, b
         assert st_bot.date.tzinfo == UTC
         assert st_offset == tz_bot_offset
 
-    def test_to_dict(self, star_transaction):
+    def test_to_dict(self):
+        st = star_transaction()
```
Bibo-Joshi, line 267: did you mean to use this as fixuter instead, i.e. pass it as argument? same in the other tests

Bibo-Joshi (commented):
Hey, thanks for the updates! I went throug the new changes and admittedly only skimmed the tests 😅 
I left a few more comments

telegram/_stars.py
harshil21: aghh, I'm going to tell ruff to fix that

telegram/_stars.py
harshil21: ahh such a small thing which I just didn't notice

telegram/_stars.py
harshil21: we did the same for basically every other class.. hmm..

tests/test_stars.py
harshil21: I had done that before but I actually wanted to use the return value in `TestStarTransactionsBase`, so I kept it as a function

Bibo-Joshi: Mh, I see. Might something like this be of use? https://stackoverflow.com/a/63047695/10606962
I'm okay with keeping as is, but it's at least something that future-us will likely stumble over if we have re-touch tests.

Bibo-Joshi (approved):
LGTM just left me know if you want to change something about the tests or keep it as is :)

harshil21:
I'll keep the tests as is for now.

PR #4317, opened 2024-06-21
PoC: `pytest-socket` like testing

Bibo-Joshi:
spin-off from #4271.

Initial fidings & thoughts:

* `pytest-socket` doesn't seem to play well with our asyncio-loop-settings
* evaluating fixtures that make requests from within `WithoutRequest` tests is currently failing
* at first glance, there doesn't seem to be a large number of tests that make requests even though we don't want them to. It might be worth hiding the logic behind a env-var so that the additional logic doesn't slow down local testing (but reports wrong networking usage in the CI)


Bibo-Joshi:
Further insights:

* If the flag for "internet connection allowed" is a singleton, i.e. globally unique, it's not really possible to have the tests run in parallel via pytest-xdist while still allowing *some* tests to make http requests. If you want to avoid race conditions here by using locks, you basically go back to sequential execution (at least for everything that uses bot `HTTPXRequest`).
* If we want to set the flag on a per-test/per-fixture basis, we'd have to use unique `HTTPXRequest`/`httpx.AsyncClient`/`bot`/`socket` objects for each test *and* figure out a way how to set the flag via markers or similar methods that don't require manual interaction in the tests

These insights are quite a bummer in terms of incorporating the allow-internet-flag into our standard testing suite. Still, I can see a few options:

* provide different fixtures `offline_bot` & `online_bot` that can be used. This requires rewriting all tests to use those new fixtures. It also means that it will harder (or impossible?) to check if a `*WithoutRequest` class uses the `online_bot` or vice versa and deny that.
* Instead of aiming to fully incorporate this into our standard test suite, we could try to simply run a sanity check every week or so to detect if any test is using the internet even though I should not. For that, we could cut off the internet connection for almost everything and run only the `*WithoutRequest` & ext tests.


@harshil21  what do you think?

Bibo-Joshi:
closing in favor of #4454

PR #4334, opened 2024-07-01
API 7.6 - `PaidMedia*` classes

harshil21:
### Checklist

- New classes:

- Added new shortcuts:

   - [ ]  In :class:`~telegram.Chat` & :class:`~telegram.User` for all methods that accept ``chat/user_id``

-  If relevant:

   - [ ] Added the new method(s) to ``_extbot.py``
   - [ ] Added or updated ``bot_methods.rst``

### DIFF
--- a/telegram/_paidmedia.py
+++ b/telegram/_paidmedia.py
@@ -40,7 +40,7 @@ class PaidMedia(TelegramObject):
     * :class:`telegram.PaidMediaVideo`
 
     Objects of this class are comparable in terms of equality. Two objects of this class are
-    considered equal, if their :attr:`type` are equal.
+    considered equal, if their :attr:`type` is equal.
 
     .. versionadded:: NEXT.VERSION
 
--- a/telegram/_paidmedia.py
+++ b/telegram/_paidmedia.py
@@ -1,0 +1,290 @@
+#!/usr/bin/env python
+#
+# A library that provides a Python interface to the Telegram Bot API
+# Copyright (C) 2015-2024
+# Leandro Toledo de Souza <devs@python-telegram-bot.org>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser Public License for more details.
+#
+# You should have received a copy of the GNU Lesser Public License
+# along with this program. If not, see [http://www.gnu.org/licenses/].
+"""This module contains objects that represent paid media in Telegram."""
+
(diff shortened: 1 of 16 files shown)


telegram/_paidmedia.py
```diff
@@ -0,0 +1,290 @@
...
+
+class PaidMedia(TelegramObject):
+    """Describes the paid media added to a message. Currently, it can be one of:
+
+    * :class:`telegram.PaidMediaPreview`
+    * :class:`telegram.PaidMediaPhoto`
+    * :class:`telegram.PaidMediaVideo`
+
+    Objects of this class are comparable in terms of equality. Two objects of this class are
+    considered equal, if their :attr:`type` are equal.
```
Bibo-Joshi, line 43:
```suggestion
    considered equal, if their :attr:`type` is equal.
```

telegram/_paidmedia.py
```diff
@@ -0,0 +1,290 @@
...
+
+        Args:
+            data (Dict[:obj:`str`, ...]): The JSON data.
+            bot (:class:`telegram.Bot`, optional): The bot associated with this object.
+
+        Returns:
+            The Telegram object.
+
+        """
+        data = cls._parse_data(data)
```
Bibo-Joshi, line 90: just an idea: we use this same logic now in several classes. We could think of extracting this logic to a helper method, possibly a protected method in TelegramObject.

tests/test_stars.py
Bibo-Joshi, line 1: are these actual changes in API 7.6 or should that rather be a test-fix follow-up for API 7.5?

Bibo-Joshi (commented):
🚀

tests/test_stars.py
harshil21: they are a test fix for api 7.5

telegram/_paidmedia.py
harshil21: we still have to specify the `_class_mapping` on every call isn't it? which parts / how exactly do you want to extract this?

Bibo-Joshi: I was thinking something like

```python
def de_json(…):
    super()._de_json_subclasses(PaidMedia, {cls.PREVIEW: …})

def _de_json_subclasses(cls, data, base_class, class_mapping):
    data = cls._parse_data(data)
     if data is None:
            return None

        if not data and cls is base_class:
            return None

        class_mapping: Dict[str, Type[PaidMedia]] = {
            cls.PREVIEW: PaidMediaPreview,
            cls.PHOTO: PaidMediaPhoto,
            cls.VIDEO: PaidMediaVideo,
        }

        if cls is PaidMedia and data.get("type") in class_mapping:
            return class_mapping[data.pop("type")].de_json(data=data, bot=bot)

        return super().de_json(data=data, bot=bot)
```

But just a thought - doesn't have to be in this PR or at all

tests/test_stars.py
Bibo-Joshi: Then I prefer having a standalone PR for that

telegram/_paidmedia.py
harshil21: yeah, probably better to do in a different PR so we can apply to other classes

Bibo-Joshi: if you have the time, I would be glad to see an issue for it 😅

PR #4335, opened 2024-07-01
API 7.6 - `send_paid_media` & `InputPaidMedia*` classes

harshil21:
### Checklist

- New classes:

- Added new shortcuts:

   - [x]  In :class:`~telegram.Chat` & :class:`~telegram.User` for all methods that accept ``chat/user_id``

-  If relevant:


telegram/_bot.py
```diff
@@ -582,8 +584,9 @@ def _insert_defaults(self, data: Dict[str, object]) -> None:
                 # Copy objects as not to edit them in-place
                 copy_list = [copy.copy(media) for media in val]
                 for media in copy_list:
-                    with media._unfrozen():
-                        media.parse_mode = DefaultValue.get_value(media.parse_mode)
+                    if hasattr(media, "parse_mode"):  # InputPaidMedia objects don't have this
+                        with media._unfrozen():
+                            media.parse_mode = DefaultValue.get_value(media.parse_mode)
```
Bibo-Joshi, line 589:
```suggestion
                    if not hasattr(media, "parse_mode"):
                        # InputPaidMedia objects don't have this
                        continue
                    with media._unfrozen():
                        media.parse_mode = DefaultValue.get_value(media.parse_mode)
```

actually, we could add `and not isinstance(value[0], InputPaidMedia)` in line 583. that would save us the copying of the list

telegram/_bot.py
```diff
@@ -9154,6 +9157,81 @@ async def get_star_transactions(
             bot=self,
         )
 
+    async def send_paid_media(
+        self,
+        chat_id: Union[str, int],
+        star_count: int,
+        media: Sequence[Union["InputPaidMediaPhoto", "InputPaidMediaVideo"]],
```
Bibo-Joshi, line 9164:
```suggestion
        media: Sequence["InputPaidMedia"]],
```

That should work?

telegram/_chat.py
Bibo-Joshi, line 1: shortcuts `User.send_paid_media` and `Message.reply_paid_media` are missing

telegram/_files/inputmedia.py
```diff
@@ -115,13 +115,162 @@ def _parse_thumbnail_input(thumbnail: Optional[FileInput]) -> Optional[Union[str
         )
 
 
+class InputPaidMedia(TelegramObject):
```
Bibo-Joshi, line 118: I wonder why they didn't just extend the existing `InputMedia` interface …

telegram/_files/inputmedia.py
```diff
@@ -115,13 +115,162 @@ def _parse_thumbnail_input(thumbnail: Optional[FileInput]) -> Optional[Union[str
...
+    PHOTO: Final[str] = constants.InputPaidMediaType.PHOTO
+    """:const:`telegram.constants.InputPaidMediaType.PHOTO`"""
+    VIDEO: Final[str] = constants.InputPaidMediaType.VIDEO
+    """:const:`telegram.constants.InputPaidMediaType.VIDEO`"""
+
+    __slots__ = ("media", "type")
+
+    def __init__(
+        self,
+        media_type: str,
```
Bibo-Joshi, line 150:
```suggestion
        type: str,
```

tbh I don't know why we deviate from the API in `InputMedia` but I'd like to avoid doing that here …

telegram/_files/inputmedia.py
```diff
@@ -115,13 +115,162 @@ def _parse_thumbnail_input(thumbnail: Optional[FileInput]) -> Optional[Union[str
...
+        *,
+        api_kwargs: Optional[JSONDict] = None,
+    ):
+        if isinstance(media, Video):
+            width = width if width is not None else media.width
+            height = height if height is not None else media.height
+            duration = duration if duration is not None else media.duration
+            media = media.file_id
+        else:
+            # We use local_mode=True because we don't have access to the actual setting and want
```
Bibo-Joshi, line 252: If we have a reference for this for `InputMedia` somewhere, we should add the reference to `InputPaidMedia` as well

tests/_files/test_inputmedia.py
```diff
@@ -1044,3 +1150,21 @@ def build_media(parse_mode, med_type):
...
+                InputPaidMediaVideo(media=video_file),
+            ],
+            caption="bye onlyfans",
+            show_caption_above_media=True,
+        )
+
+        assert isinstance(msg, Message)
+        assert msg.caption == "bye onlyfans"
+        assert msg.show_caption_above_media
+        # TODO: uncomment after attributes are added to message:
```
Bibo-Joshi, line 1169: le dot

Bibo-Joshi (changes requested):
🚗

telegram/ext/_extbot.py
```diff
@@ -4216,6 +4218,46 @@ async def get_star_transactions(
...
+        star_count: int,
+        media: Sequence[Union["InputPaidMediaPhoto", "InputPaidMediaVideo"]],
+        caption: Optional[str] = None,
+        parse_mode: ODVInput[str] = DEFAULT_NONE,
+        caption_entities: Optional[Sequence["MessageEntity"]] = None,
+        show_caption_above_media: Optional[bool] = None,
+        disable_notification: ODVInput[bool] = DEFAULT_NONE,
+        protect_content: ODVInput[bool] = DEFAULT_NONE,
+        reply_parameters: Optional["ReplyParameters"] = None,
+        reply_markup: Optional[ReplyMarkup] = None,
```
Bibo-Joshi, line 4233: uh, we need logic for arbitrary context data here …

telegram/_chat.py
harshil21: User.send_paid_media doesn't make sense because the method only works for channels

telegram/_files/inputmedia.py
harshil21: I didn't understand what you're trying to say. Do you want to move the `parse_file_input` to `InputPaidMedia` instead?

telegram/ext/_extbot.py
harshil21: `Bot.send_paid_media` calls `self._send_message`, which is overriden by `ExtBot` to insert callback data, so we don't have to do anything?

tests/_files/test_inputmedia.py
harshil21: need to merge the other pr first

telegram/_bot.py
```diff
@@ -9154,6 +9157,81 @@ async def get_star_transactions(
...
+        caption: Optional[str] = None,
+        parse_mode: ODVInput[str] = DEFAULT_NONE,
+        caption_entities: Optional[Sequence["MessageEntity"]] = None,
+        show_caption_above_media: Optional[bool] = None,
+        disable_notification: ODVInput[bool] = DEFAULT_NONE,
+        protect_content: ODVInput[bool] = DEFAULT_NONE,
+        reply_parameters: Optional["ReplyParameters"] = None,
+        reply_markup: Optional[ReplyMarkup] = None,
+        *,
+        read_timeout: ODVInput[float] = DEFAULT_NONE,
```
harshil21, line 9174: do we want to add the field `allow_sending_without_reply` for new methods? I think yes because we have that in `Defaults`?

telegram/_chat.py
Bibo-Joshi: ah, hadn't noticed that it only works for channels. Thanks for pointing it out 👍

telegram/_files/inputmedia.py
Bibo-Joshi: sorry, I'm talking about documentation. But IISC we have that documented only in https://docs.python-telegram-bot.org/en/stable/telegram.inputmedia.html#telegram.InputMedia.params.media and that's covered here already via `|fileinputnopath|`. then everything's fine :)

tests/_files/test_inputmedia.py
Bibo-Joshi: yes, was just leaving a reminder :)

telegram/ext/_extbot.py
Bibo-Joshi: ah, righto. Still wondering if we should add a new test case. Haven't worked on those tests for a long time :/

telegram/_bot.py
Bibo-Joshi: good point. I think for consistency it would be good to add both.

telegram/ext/_extbot.py
Bibo-Joshi: Apparently we also only test with `send_message` for the other ones. I'll resolve

Bibo-Joshi (approved):
Ready from my side, then. merge at will 🧜‍♂️

PR #4336, opened 2024-07-01
Use Python 3.13 Beta 3 in tests

harshil21:
new beta dropped


Bibo-Joshi:
> E       FutureWarning: functools.partial will be a method descriptor in future Python versions; wrap it in staticmethod() if you want to preserve the old behavior
>
> telegram/ext/_application.py:877: FutureWarning

Needs to be addressed

PR #4358, opened 2024-07-08
Add Internal Helper for Deserializing Subclasses

### LINKED ISSUE #4346: [FEATURE] New `TO.de_json_subclasses` method for parent classes

We can save some code duplication by extracting the `de_json` logic of classes like `PaidMedia` or `ChatMember`. Ideally this should go into `TelegramObject` so we can easily call it.

https://github.com/python-telegram-bot/python-telegram-bot/pull/4334#discussion_r1665902751

keep doing what we do right now

### Additional context

_No response_

Bibo-Joshi:
Closes #4346

TBF, this is more clumsy than I had hoped, but if TG continues with the pattern of subclasses it should still save some duplication.

It's a +0 from my side. If you find the logic hard to follow and think that it makes debugging harder, I'm okay with rejecting.

### DIFF
--- a/telegram/_botcommandscope.py
+++ b/telegram/_botcommandscope.py
@@ -102,12 +102,7 @@ class BotCommandScope(TelegramObject):
             The Telegram object.
 
         """
-        data = cls._parse_data(data)
-
-        if not data:
-            return None
-
-        _class_mapping: Dict[str, Type[BotCommandScope]] = {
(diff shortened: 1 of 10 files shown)


harshil21 (commented):
oh.. looks like it became much more complicated with very little to no benefit.. (LoC is also greater). I also propose to scrap.

Bibo-Joshi:
Fair enough. I'll close and just open a new PR to add the test to menubutton :)

PR #4360, opened 2024-07-08
Stabilize Some Concurrency Usages in Test Suite

Bibo-Joshi:

### DIFF
--- a/tests/test_bot.py
+++ b/tests/test_bot.py
@@ -2368,11 +2368,15 @@ class TestBotWithRequest:
     # test modules. No need to duplicate here.
 
     async def test_delete_messages(self, bot, chat_id):
-        msg1 = await bot.send_message(chat_id, text="will be deleted")
-        msg2 = await bot.send_message(chat_id, text="will be deleted")
-        await asyncio.sleep(2)
+        msg1, msg2 = await asyncio.gather(
+            bot.send_message(chat_id, text="will be deleted"),
+            bot.send_message(chat_id, text="will be deleted"),
+        )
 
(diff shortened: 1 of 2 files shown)


tests/test_bot.py
```diff
@@ -2368,11 +2368,15 @@ async def test_delete_message_old_message(self, bot, chat_id):
     # test modules. No need to duplicate here.
 
     async def test_delete_messages(self, bot, chat_id):
-        msg1 = await bot.send_message(chat_id, text="will be deleted")
-        msg2 = await bot.send_message(chat_id, text="will be deleted")
-        await asyncio.sleep(2)
```
harshil21, line 2373: I think I did this in #4271 but that branch is so hilariously behind I don't even remember now

tests/test_constants.py
```diff
@@ -234,6 +234,11 @@ async def test_max_message_length(self, bot, chat_id):
             return_exceptions=True,
         )
         good_msg, bad_msg = await tasks
+
+        if isinstance(good_msg, BaseException):
```
harshil21, line 238: what if `bad_msg` is the one with the exception? or does return_exceptions always return to the first variable? Also, how are we running into this issue with a simple `send_message` only on this test but not on others (using the same `.gather`)?

Bibo-Joshi: >  Also, how are we running into this issue with a simple send_message only on this test but not on others (using the same .gather)?

as far as I see, we're not usang `return_exceptions` in asyncio.gather in oher places

> what if bad_msg is the one with the exception? or does return_exceptions always return to the first variable?

gather retains the order, that should suffice

PR #4364, opened 2024-07-09
Automate PyPI Releases

### LINKED ISSUE #4318: [FEATURE] Automated PyPI Release

The [release process](https://github.com/python-telegram-bot/python-telegram-bot/wiki/Releasing-a-new-version) currently requires several manual steps, including

* building the source & whl distributions
* signing them & computing digests
* uploading the artifacts to pypi
* creating a GH release & uploading the artifacts there

Even though these steps are a comparatiely small part of the release process, I'd like to automate them.

### Benefits that I see

* reduces time required to do a release, which allows for more frequent releases (e.g. faster release after API changes are incorprated)
* an automated build & deploy process doesn't require setup on a local machine such that in case of a maintainer switch, the new maintainer has less to set up
* the time I currently spent on doing releases, I can then use for other work on PTB (or for a nap :))

I would like to automate the steps mentioned above using PyPI’s [trusted publishing](https://docs.pypi.org/trusted-publishers/) via GitHub Actions as described in the [official python packaging documentation](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/). Recently I've tried this out with [aiorem](https://github.com/Bibo-Joshi/aiorem) and I'm positively suprised about how well and easy that works.

let me list a few insights/comments I had:

* this approach uses GitHub environments, which are a kind of restricted version of Actions. In particular, you can (and IISC must for usage with pypi) specify that the run must be manually approved by an admin such that you don't make releases by accident
* the environment has a name that must be set in the pypi settings, making sure that only trusted environments can make releases
* The Python Packaging Authority recommends to use [Sigstore](https://sigstore.dev) to sign binaries. Sigstore is used also for [signing CPython](https://www.python.org/download/sigstore/) and [endorsed by github](https://github.blog/2022-10-25-why-were-excited-about-the-sigstore-general-availability/). the python [sigstore](https://github.com/sigstore/sigstore-python?tab=readme-ov-file#verifying-signatures-from-github-actions) even has a special subcommand to verify signatures generated via GitHub Actions. Using sigstore means that we don't even have to take care about our own signing keys anymore. Note that PyPI [doesn't accept signatures anymore](https://blog.pypi.org/posts/2023-05-23-removing-pgp/) for quite some time, so the signatures would instead go to the GH releases page as they currently already do.
* a bit care is needed in terms of setting up the triggers for the workflows. For one, pypi and test-pypi don't accept multiple uploads with the same version number so one should take not to run into that even for test uploads. Moreover, if you reject the manual approval of an environment run, it's marked as failed and that at least doesn't look nice in the commit status :D 

One could use a manual approach to authenticate at PyPI and/or use manual GPG signatures as done so far, but both would require a custom approach that is likely to be less safe & less frequently updated than the GH Actions tools provided by PyPA and sigstore.

### Additional context

To address the more time consuming part of the releases - namely the creation of changelogs - I'll try to open a second thread soonish. My goal is that ideally (almost) the complete release & deply process can be automated, maximising the benefits I lsited above :)

Bibo-Joshi:
Closes #4318 

ToDo:

- [x] Revert random version numbers
- [x] disable running on pushes to this dev branch
- [x] Rename gpg public key file once first version using this workflow is clear
- [x] Create follow-up PR to update website on verification information https://github.com/python-telegram-bot/python-telegram-bot.github.io/pull/21
- [x] Possibly extend the verification information with the optional parameters of sigstore
- [x] Verify that the testpypi upload & github draft release work


Bibo-Joshi:
After a bit of trial and error, this seems to be running fine now :) pypi test releases can be seen at https://test.pypi.org/project/python-telegram-bot/#history. the GH releases drafts worked as well, but I discarded them to not leave testing artifacts behind.

I'm attaching the produced artifacts from the last successfull run so that if you like you can verify that the sigstore verification and sha1 sum works as expected :)

[python_telegram_bot-21.4a9571-artifacts.zip](https://github.com/user-attachments/files/16146598/python_telegram_bot-21.4a9571-artifacts.zip)

README.rst
```diff
@@ -104,15 +104,19 @@ You can also install ``python-telegram-bot`` from source, though this is usually
 Verifying Releases
 ------------------
 
-We sign all the releases with a GPG key.
-The signatures are uploaded to both the `GitHub releases page <https://github.com/python-telegram-bot/python-telegram-bot/releases>`_ and the `PyPI project <https://pypi.org/project/python-telegram-bot/>`_ and end with a suffix ``.asc``.
+To enable you to verify that a release file that you downloaded was indeed provided by the ``python-telegram-bot`` team, we have taken the following measures.
+
+Starting with NEXT.VERSION, all releases are signed via `sigstore <https://sigstore.dev>`_.
+The corresponding signature files are uploaded to both the `GitHub releases page`_.
```
Bibo-Joshi, line 110:
```suggestion
The corresponding signature files are uploaded to the `GitHub releases page`_.
```

.github/workflows/release_pypi.yml
```diff
@@ -0,0 +1,204 @@
...
+
+  publish-to-pypi:
+    name: Publish to PyPI
+    # only publish to PyPI on tag pushes
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+    needs:
+      - build
+    runs-on: ubuntu-latest
+    environment:
+      name: release_pypi
```
harshil21, line 41: does this need to match the file name? In the [docs](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/), they put it as "pypi". 

If so, can you leave a comment saying it needs to match the file name?

.github/workflows/release_pypi.yml
```diff
@@ -0,0 +1,204 @@
...
+        with:
+          name: python-package-distributions
+          path: dist/
+      - name: Compute SHA1 Sums
+        run: |
+          # Compute SHA1 sum of the distribution packages and save it to a file with the same name,
+          # but with .sha1 extension
+          for file in dist/*; do
+              sha1sum $file > $file.sha1
+          done
```
harshil21, line 107: I didn't see this section on their guide. Do we have to compute this ourselves? Isn't that the job of sigstore? Maybe you want to enable [this option](https://github.com/marketplace/actions/gh-action-sigstore-python#upload-signing-artifacts)?

harshil21 (commented):
looks to be extensively tested. I just have some questions below :)

.github/workflows/release_pypi.yml
Bibo-Joshi: it's not the file name but rather the name of the [environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) :)

.github/workflows/release_pypi.yml
harshil21: oh ok. I don't mind it, but then we don't even specify that in the readme. Is it supposed to be self-explantory?

Bibo-Joshi: But we do:

> In addition, the GitHub release page also contains the sha1 hashes of the release files in the files with the suffix .sha1.

harshil21: oh I must be blind lol

harshil21 (approved):

PR #4376, opened 2024-07-13
Add `MessageEntity.shift_entities` and `MessageEntity.concatenate`

### LINKED ISSUE #4372: [FEATURE] Convenience Method for `MessageEntity` Concatenation

The discussion with TG in https://github.com/tdlib/telegram-bot-api/issues/515 has made it clear that we should encourage users to work more with MessageEntities rather than with our `text_html/markdown` helpers. Since this is sometimes cumbersome, convenience functionality for working with ME is surely advisable and we've made a step into that direction with #4323 .

What I'd like to see in addition is convenience functionality for catenating text with entities as initially touched in https://github.com/tdlib/telegram-bot-api/issues/515#issuecomment-1885594214

new class methods

* `ME.shift_entities(entities: Sequence[ME], by: str | int) -> Sequence[ME]`: shifts the entities by the specified length or the utf-16 length of the string of `by` is a string. This would be useful for adding a prefix to a message with entities
* `ME.concatetane(first: tuple[str, Sequence[ME]], second: tuple[str, Sequence[ME]]) -> tuple[str, Sequence[ME]]`: Concatenates text and entities. Uses `ME.shift_entities` internally

`ME.concatenate` could be skipped in a first step as it can be implemented using `shift_entities`

### Additional context

_No response_

Bibo-Joshi:
closes #4372 

The implemented signature for `ME.concatenate` differs from the one proposed in #4372 to allow concanetating several tuples of text & entities at once.

- ~Added new classes & modules to the docs and all suitable ``__all__`` s~

### DIFF
--- a/telegram/_messageentity.py
+++ b/telegram/_messageentity.py
@@ -20,7 +20,7 @@
 
 import copy
 import itertools
-from typing import TYPE_CHECKING, Dict, Final, List, Optional, Sequence, Union
+from typing import TYPE_CHECKING, Dict, Final, List, Optional, Sequence, Tuple, Union
 
 from telegram import constants
 from telegram._telegramobject import TelegramObject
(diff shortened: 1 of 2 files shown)


telegram/_messageentity.py
```diff
@@ -219,6 +219,147 @@ def adjust_message_entities_to_utf_16(
...
+        Returns:
+            Tuple[:obj:`str`, Sequence[:class:`telegram.MessageEntity`]]: The concatenated text
+            and its entities
+        """
+        output_text = ""
+        output_entities: List[MessageEntity] = []
+        for arg in args:
+            text, entities = arg[0], arg[1]
+
+            if len(arg) > 2 and arg[2] is True:
```
elpekenin, line 355: Given users are supposed to pass bools (opposed to any other value that could be `True`-y), do we really need to explicitly `== True`?

telegram/_messageentity.py
```diff
@@ -219,6 +219,147 @@ def adjust_message_entities_to_utf_16(
...
+            with new_entity._unfrozen():
+                new_entity.offset += effective_shift
+            out.append(new_entity)
+        return out
+
+    @classmethod
+    def concatenate(
+        cls,
+        *args: Union[
+            Tuple[str, Sequence["MessageEntity"]], Tuple[str, Sequence["MessageEntity"], bool]
```
elpekenin, line 291: Maybe worth adding an alias for the `tuple[str, Sequence[ME]]` type

elpekenin (approved):
Just a couple of nitpicks, LGTM

telegram/_messageentity.py
Bibo-Joshi: maybe not for that (it's used only twice), but for `Sequence[ME]` (which is used 7 times :))

harshil21 (commented):
cool idea, I like it :)

telegram/_messageentity.py
```diff
@@ -220,6 +220,143 @@ def adjust_message_entities_to_utf_16(
...
+        Returns:
+            Tuple[:obj:`str`, Sequence[:class:`telegram.MessageEntity`]]: The concatenated text
+            and its entities
+        """
+        output_text = ""
+        output_entities: List[MessageEntity] = []
+        for arg in args:
+            text, entities = arg[0], arg[1]
+
+            if len(arg) > 2 and arg[2]:
```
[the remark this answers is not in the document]
harshil21: oh I just saw elpekenin review the same point ooops

telegram/_messageentity.py
```diff
@@ -220,6 +220,143 @@ def adjust_message_entities_to_utf_16(
...
+            The :paramref:`entities` are *not* modified in place. The function returns a sequence
+            of new objects.
+
+        .. versionadded:: NEXT.VERSION
+
+        Args:
+            by (Union[:obj:`str`, :obj:`int`]): Either the amount to shift the offset by or
+                a string whose length will be used as the amount to shift the offset by. In this
+                case, UTF-16 encoding will be used to calculate the length.
+            entities (Sequence[:class:`telegram.MessageEntity`]): Sequence of entities
```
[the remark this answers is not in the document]
harshil21: well yeah, but the description of the parameter just looked a bit too short

harshil21 (approved):

PR #4378, opened 2024-07-14
[enhancement] move hardcoded encodings to contants

elpekenin:
As per title, just a very minor refactor


elpekenin:
Didn't change the hardcoded `utf-8` in `docs/source/conf.py`, to prevent making it depend further on the project, can add the changes there too if desired

Bibo-Joshi (approved):
Thanks for the PR! changing conf.py probably doesn't give a big benefit, also considerding that the import can't bee at the top …  LGTM :)

Bibo-Joshi:

PR #4386, opened 2024-07-17
Adapt to RTD Addons

Bibo-Joshi:
I received a mail from RT pointing to https://about.readthedocs.com/blog/2024/07/addons-by-default/ and explaining that ptb currenlty relies on `html_baseurl` and that it should be set via

```python
import os

# Set canonical URL from the Read the Docs Domain
html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")

# Tell Jinja2 templates the build is running on Read the Docs
if os.environ.get("READTHEDOCS", "") == "True":
    html_context["READTHEDOCS"] = True
```

This is also explained somewhat in the [current docs](https://docs.readthedocs.io/en/stable/canonical-urls.html#how-to-specify-the-canonical-url)

Additional changes may be needed to adapt to [RTD addons](https://github.com/readthedocs/addons) - in particular they include a native [search as you type](https://github.com/readthedocs/addons?tab=readme-ov-file#search-as-you-type) addon. I'll create a spearate issue for that and link to the furo theme issue tracking the topic.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,3 +1,4 @@
+import os
 import re
 import sys
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


Bibo-Joshi:
Test build is up at https://docs.python-telegram-bot.org/en/docs-rtd-addons/index.html and looks successful. All links correctly link back to docs.ptb.org

PR #4398, opened 2024-07-22
Prepare for Py +3.9 support

elpekenin:
As per title, so far:
- Mass search+replace for stuff like `List`->`list`
- Silence warnings, while no better solution is found, for `Type[UT]`
- Update usages of deprecated `typing.AsyncContextManager`
- Update `3.8` to `3.9` in `pyproject.toml`
- Use `str.removeprefix` (i might have missed a couple of places doing the same logic manually)


harshil21:
@elpekenin Python 3.8 reached EoL yesterday and 3.13 final was released. Do you want to still work on this or should we pick it up again?

elpekenin:
Will (try and) take a look later today

telegram/ext/_utils/_update_parsing.py
```diff
@@ -39,7 +39,7 @@ def parse_chat_id(chat_id: Optional[SCT[int]]) -> FrozenSet[int]:
     return frozenset(chat_id)
 
 
-def parse_username(username: Optional[SCT[str]]) -> FrozenSet[str]:
+def parse_username(username: Optional[SCT[str]]) -> frozenset[str]:
```
Bibo-Joshi, line 42: this function can also make use of `removeprefix` :)

telegram/ext/_handlers/typehandler.py
```diff
@@ -71,13 +74,13 @@ async def callback(update: object, context: CallbackContext)
 
     def __init__(
         self: "TypeHandler[UT, CCT, RT]",
-        type: Type[UT],  # pylint: disable=redefined-builtin
+        type: GenericUT,  # pylint: disable=redefined-builtin
```
Bibo-Joshi, line 77: the type completeness check is not happy with this. Maybe it needs to be `GenericUT[UT]`? 🤔

Bibo-Joshi (changes requested):
Heyho. I've scrolled through the changes a bit, but didn't see anything besides the two comments below. @elpekenin what'sy our status on this PR otherwise? I see that it's still marked as draft.

Bibo-Joshi (approved):
@elpekenin @harshil21 if I don't hear vehement disagreement from you, I'll merge some time on the weekend :)

harshil21 (approved):

elpekenin:
Sorry for the silence, somehow i didnt get/set notifications about the feedback.

I have no idea what `GenericUT[UT]` is supposed to mean. From my understanding, * We wouldnt need an alias for the hint
* Would just use `Type[UT]` in there (uppercase just because tooling might get confused with `type` also being the name of the variable)
* Referencing `UT` outside function would be the same as using `Any` as it is not bound by the specific callers' typing (?) 
... But its been such a headache trying stuff without any success that im fine with anything that tooling like :P

Bibo-Joshi:
> But its been such a headache trying stuff without any success that im fine with anything that tooling like :P

Its certainly weird 😅 my last commit seemed to work and I also checked that `reveal_type` produces sane output with both mypy and pyright, so it would be goos enough for me :D

Bibo-Joshi:
Thank you very much for the contribution :)

PR #4413, opened 2024-08-03
Fix "available in" admonitions

harshil21:
Fixes 2 things:

- Available in admonition not appearing in certain classes like `RefundedPayment`. This happened because the regex didn't consider `:obj:` classes.
- Insertion of admonitions will now consider putting it before "Args:".


Bibo-Joshi (approved):
Code changes look good 👍 I expect it also renders fine? :D

harshil21:
yep 👍

PR #4416, opened 2024-08-03
Bump ruff and add new rules

harshil21:
Adds "TRY", parts of "D", "S", and "DOC".

https://docs.astral.sh/ruff/rules/

### DIFF
--- a/docs/auxil/link_code.py
+++ b/docs/auxil/link_code.py
@@ -15,7 +15,6 @@
 #
 #  You should have received a copy of the GNU Lesser Public License
 #  along with this program.  If not, see [http://www.gnu.org/licenses/].
-# ruff: noqa: S607, S603
 """Functionality in this file is used for getting the [source] links on the classes, methods etc
 to link to the correct files & lines on github. Can be simplified once
 https://github.com/sphinx-doc/sphinx/issues/1556 is closed
--- a/docs/auxil/link_code.py
+++ b/docs/auxil/link_code.py
@@ -15,6 +15,7 @@
 #
 #  You should have received a copy of the GNU Lesser Public License
 #  along with this program.  If not, see [http://www.gnu.org/licenses/].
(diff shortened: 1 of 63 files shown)


docs/auxil/link_code.py
```diff
@@ -15,6 +15,7 @@
 #
 #  You should have received a copy of the GNU Lesser Public License
 #  along with this program.  If not, see [http://www.gnu.org/licenses/].
+# ruff: noqa: S607, S603
```
Bibo-Joshi, line 18: Maybe ignore S for the complete docs directory?

telegram/_bot.py
```diff
@@ -4271,7 +4271,7 @@ async def get_updates(
         self,
         offset: Optional[int] = None,
         limit: Optional[int] = None,
-        timeout: Optional[int] = None,
+        timeout: Optional[int] = None,  # noqa: ASYNC109
```
Bibo-Joshi, line 4274: What's this about?

telegram/ext/_application.py
```diff
@@ -1227,7 +1227,7 @@ async def __create_task_callback(
                 await self.process_error(update=update, error=exception, coroutine=coroutine)
 
             # Raise exception so that it can be set on the task and retrieved by task.exception()
-            raise exception
+            raise
```
Bibo-Joshi, line 1230: So python automatically raises the latest available exception?

Bibo-Joshi (commented):
A few questions:)

telegram/ext/_application.py
harshil21: yep

telegram/_bot.py
harshil21: it tells to use Asyncio.timeout which is wrong here ofc

Bibo-Joshi (approved):

PR #4431, opened 2024-08-16
Updated 3 failing  tests of test_message.py.

Palaptin:
The tests test_link_with_topics and test_link_with_reply were failing if beeing runned alone ( since message.chat.type was private).

The test test_build_reply_arguments_chat_id_and_message_id was failing randomly, mostly when using pytest --lf (about 1 in 10 times). The reason was message.chat.username being set to 'username', so setting it to None will solve this problem.

### DIFF
--- a/tests/test_message.py
+++ b/tests/test_message.py
@@ -1193,16 +1193,20 @@ class TestMessageWithoutRequest(TestMessageBase):
         # The leading - for group ids/ -100 for supergroup ids isn't supposed to be in the link
         assert message.link == f"https://t.me/c/{3}/{message.message_id}"
 
-    def test_link_with_topics(self, message):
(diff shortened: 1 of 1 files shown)


harshil21 (changes requested):
Thanks for the fix! I noticed this myself a couple of times before but didn't investigate the cause.

To elaborate further after investigation - part of the reason why it doesn't fail when the whole test file is run is because the `message` fixture is module scoped, and other tests are modifying that value (bad practice imo) which leads to a slightly different state for the next test. They fail individually because we initialize the `message` fixture with a private chat anyway, which doesn't have public links obviously.

I would suggest changing the scope of that fixture back to "function" (the default) to reduce the chance of this occuring in the future.

harshil21 (approved):
lgtm. Ran it locally on my end and it passes.

PR #4454, opened 2024-09-03
Make Tests for `telegram.ext` Independent of Networking

Bibo-Joshi:
By cutting of the internet connection in the fixtures

* provide a new fixture `offline_bot` that has no internet connection
* override fixtures for the tests in `ext` to have no internet connection
* manually adapt a few tests
* also move other ext-only-fixtures from conftest to ext\conftest on the fly

Let's see if I got everything in local testing

supersedes #4317. Although this PR here doesn't cover cutting off the internet connection in the `*WithoutRequest` tests, I don't think I'll follow up on that for the reasons pointed out in #4317.


harshil21 (commented):
Alright, looks good to me. Regarding the `WithoutRequest` tests, after we use `offline_bot` there as well, I'm wondering if we can disallow the `bot` fixture to be used in those tests. As you mentioned, it's close to impossible to effectively enforce this, but it's a decent try. 

We can get the fixture list using [fixturenames](https://docs.pytest.org/en/6.2.x/reference.html#pytest.FixtureRequest.fixturenames) it seems.

Bibo-Joshi:
> Alright, looks good to me. Regarding the WithoutRequest tests, after we use offline_bot there as well, I'm wondering if we can disallow the bot fixture to be used in those tests. As you mentioned, it's close to impossible to effectively enforce this, but it's a decent try.

I would like to move this to a different PR …

Bibo-Joshi:
All right. Looks like I've got all tests working by now and also managed to reduce repetitions a bit. So now actually ready for review 😅

tests/ext/test_application.py
```diff
@@ -2542,6 +2496,9 @@ async def task(app):
 
             app.create_task(task(app))
 
+        async def return_true(*args, **kwargs):
+            return True
+
```
harshil21: do we need this here again since it's there in conftest

tests/ext/test_inlinequeryhandler.py
harshil21, line 1: weirdly missed hmm

tests/ext/test_application.py
harshil21: nvm github showed me an old commit

harshil21 (approved):

PR #4462, opened 2024-09-08
Admonition overhaul

### LINKED ISSUE #4414: Fix `GameHighScore` return in admonition
### Steps to Reproduce

1. Check the docs for [gamehighscore](https://docs.python-telegram-bot.org/en/doc-fixes/telegram.gamehighscore.html#gamehighscore)
2. No returned in admonition present

### Expected behaviour

It should show Returned in `Bot.get_game_high_scores`.

I propose refactoring the code of `AdmonitionInserter._resolve_arg`, since that is why this fails. I am unsure why `typing.get_origin` is used instead `typing.get_args`. Moreover, we can let python resolve forward refs, as we've done recently in `test_official`, which should reduce complexity. 

### Actual behaviour

Nothing shown.

### Operating System

Agnostic

```shell
python-telegram-bot 21.4 (v21.4-13-g356a58ffd)
Bot API 7.8
Python 3.13.0b1 (main, May 10 2024, 23:59:20) [GCC 13.2.0]
```

### Relevant log output

_No response_

### Additional Context

_No response_

harshil21:
Main changes so far:

- Uses type hints for introspection instead of parsing docs
- Closes #4414 (I think, need to check lol)
- Less LoC
- Adds `Application` in the admonitions, so far this has only resulted in a "use-in" from all the handlers to `add_handler(s)`.

To do:
- Many admonitions are missing - investigate why
- drastically update the test file for this, for easier iteration
- ~pre-commit intentionally being failed for now~


Bibo-Joshi:
Hi @harshil21 I got around to work a bit on this. You already had the most part covered, I just patched a few things. Regarding your points:

> Many admonitions are missing - investigate why

Do you recall a list of things that you were missing? I clicked around thorugh the docs a bit aimlessly and didin't catch anything. 

> drastically update the test file for this, for easier iteration

Herer I'm also not sure what to do about it. Adding stuff that we find while working on the implementation is so far the most sane thing to do IMO. Trying to cover all cases would basically result in re-implementing the insertion-logic within the test file, I guess 😅

harshil21:
> Do you recall a list of things that you were missing? I clicked around thorugh the docs a bit aimlessly and didin't catch anything.

I looked through my notes and didn't find too much other than:

```
LinkPreviewOptions "use-in" should show a use in InputTextMessageContent

InputTextMessageContent's available in should be moved up

test to check all service messages are included in StatusUpdate.
```
but I'm pretty sure there were more classes, but I don't remember them. 

> Trying to cover all cases would basically result in re-implementing the insertion-logic within the test file

I was just thinking that the test file would have a bunch of more randomly selected cases from the docs and just checks if they are present

Bibo-Joshi:
> LinkPreviewOptions "use-in" should show a use in InputTextMessageContent

I skipped this one for now since so far we don't link to any arguments of tg classes

> InputTextMessageContent's available in should be moved up

seems ok

test to check all service messages are included in StatusUpdate.

> I don't quite get this one. Are you referring to `ext.filters.StatusUpdate`? We have no linking to `ext.filters` at all so far …

harshil21 (commented):
looked at the docs and it everything seems to be in order. LGTM.

PR #4460, opened 2024-09-08
API 7.10 - `PaidMediaPurchased`

harshil21:
Part of #4459 

Checklist for PRs
- [x] Added ``.. versionadded:: NEXT.VERSION``, ``.. versionchanged:: NEXT.VERSION``, ``.. deprecated:: NEXT.VERSION`` or ``.. versionremoved:: NEXT.VERSION`` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)
- [x] Created new or adapted existing unit tests
- [ ] Documented code changes according to the `CSI standard <https://standards.mousepawmedia.com/en/stable/csi.html>`__
- [ ] Added myself alphabetically to ``AUTHORS.rst`` (optional)
- [x] Added new classes & modules to the docs and all suitable ``__all__`` s
- [ ] Checked the `Stability Policy <https://docs.python-telegram-bot.org/stability_policy.html>`_ in case of deprecations or changes to documented behavior

**If the PR contains API changes (otherwise, you can ignore this passage)**

- [ ] Checked the Bot API specific sections of the `Stability Policy <https://docs.python-telegram-bot.org/stability_policy.html>`_
- [ ] Created a PR to remove functionality deprecated in the previous Bot API release (`see here <https://docs.python-telegram-bot.org/en/stable/stability_policy.html#case-2>`_)

- [x] New classes:

   - [x] Added ``self._id_attrs`` and corresponding documentation
   - [x] ``__init__`` accepts ``api_kwargs`` as kw-only

- [ ]  Added new shortcuts:

   - [ ] In :class:`~telegram.Chat` & :class:`~telegram.User` for all methods that accept ``chat/user_id``
   - [ ] In :class:`~telegram.Message` for all methods that accept ``chat_id`` and ``message_id``
   - [ ] For new :class:`~telegram.Message` shortcuts: Added ``quote`` argument if methods accepts ``reply_to_message_id``
   - [ ] In :class:`~telegram.CallbackQuery` for all methods that accept either ``chat_id`` and ``message_id`` or ``inline_message_id``

- [ ]  If relevant:

   - [ ] Added new constants at :mod:`telegram.constants` and shortcuts to them as class variables
   - [x] Link new and existing constants in docstrings instead of hard-coded numbers and strings
   - [ ] Add new message types to :attr:`telegram.Message.effective_attachment`
   - [x] Added new handlers for new update types

     - [ ] Add the handlers to the warning loop in the :class:`~telegram.ext.ConversationHandler`

[...]

PR #4466, opened 2024-09-08
Improve Type Completeness

Bibo-Joshi:
Based on https://github.com/python-telegram-bot/python-telegram-bot/pull/4460#issuecomment-2336758343

The typing of `CallbackContext` got a tad less precise due to the `Job[Any]`. It appears that `pyright` wants us to use `typing.Self` (tried, it works with that), but that's available only in py3.11.


telegram/ext/_callbackcontext.py
```diff
@@ -49,6 +50,9 @@
     "/wiki/Storing-bot%2C-user-and-chat-related-data"
 )
 
+# something like poor mans "tying.Self" for py<3.11
```
harshil21, line 53:
```suggestion
# something like poor mans "typing.Self" for py<3.11
```

harshil21 (commented):
interesting that the score dropped like 2 months ago but we didn't notice it. Did the test fail and we merged it anyway?

Bibo-Joshi:
No, the problem is that the type completeness check has no memory across commits. So for each PR it needs to re-compute the type completeness of the head. so you have

1.  commit A, merged after testing with pyright 1.1.372, type score 1.0
2.  PR, comparing against commit A, recomputes with pyright 1.1.373, type score 0.976

If we wanted to catch these hases, we'd have to have something like codecov, i.e. a database for the type score per commit.
the upside of the current behavior is that a pyright update doesn't mark your PR as invalid even though your PR doesn't in itself decrease the type completeness …

harshil21:
@Bibo-Joshi I see. Can we not run a sanity check every month or so to see if the score dropped below 1.0?

Bibo-Joshi:
Don't know if I'm too tired or if the output is not set correctly by the reusable action. Will have to retry someday

Bibo-Joshi:
* Proof that the workflow picks up type completeness < 1: https://github.com/python-telegram-bot/python-telegram-bot/actions/runs/10853416139?pr=4466
* Proof that the workflow will pick up type completeness == 1 if run on master: https://github.com/python-telegram-bot/python-telegram-bot/actions/runs/10853469089 (the only problem here is that I use the "base completeness", i.e. the one of master which is crruntly < 1)
* Proof that this PR changes the type completeness to 1: https://github.com/python-telegram-bot/python-telegram-bot/actions/runs/10853484817

⇒ ready from my side :)

harshil21 (approved):

Issue #4324, opened 2024-06-26
Test suite maintainence and new ideas


harshil21:
Our test suite has once again degraded in performance and flakiness has been reintroduced. We can experiment / make our test suite better in different ways, some of my current ideas are listed roughly below:

- [x] Try to make everything as concurrent as possible (#4271)
- [ ] Profile all tests and see what clogs up and fix that.
- [ ] solve the TODO in `test_official.exceptions`
- [ ] #4514
- [ ] [`dirty-equals`](https://dirty-equals.helpmanual.io/latest/) sounds like it could be quite useful for us!
- [x] Think about renaming the `Test*Base` classes to just `*Base` since pytest collects the classes starting with `Test*`. #4453
- [x] Change CI's pytest  `--dist` to `worksteal` #4491 .
- [x] Extend that naming scheme to `telegram.ext` tests
- [x] fix the class name https://github.com/python-telegram-bot/python-telegram-bot/blob/51ef571a0761bbfea136c455bbfa67d47b20585c/tests/test_menubutton.py#L94 #4453

Follow https://github.com/zupo/awesome-pytest-speedup, not all items but some are worth looking at:

- [x] Introduce `pytest-socket` and completely disable network access for `*WithoutRequest` tests (#4317, #4454)
- [x] ~Introduce `pyfakefs` to test suite~ (no major benefit found - https://github.com/python-telegram-bot/python-telegram-bot/pull/4271#issuecomment-2127968133)
- [ ] Check if using `uvloop` in tests provides any measurable speed up.
- [ ] Isolate tests from other tests
- [ ] Check if introducing `pytest-randomly` (need to check if pytest-xdist effectively does the job of randomly)

-> Check if using `pytest-asyncio-cooperative` is worth it
     - Nope it is not.

MrLightspeed:
@codex review

PR #4487, opened 2024-09-19
Introduce Codecov's Test Analysis

Bibo-Joshi:
See https://docs.codecov.com/docs/test-result-ingestion-beta and https://about.codecov.io/blog/find-failing-and-flaky-tests-with-codecov-test-analytics/

In light of the discussion in #4324 and related threads, it seemed like a good idea to try it out :)

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -97,3 +97,9 @@ jobs:
           name: ${{ matrix.os }}-${{ matrix.python-version }}
           fail_ci_if_error: true
           token: ${{ secrets.CODECOV_TOKEN }}
+      - name: Upload test results to Codecov
+        uses: codecov/test-results-action@v1
+        if: ${{ !cancelled() }}
+        with:
+          files: .test_report_no_optionals.xml,.test_report_optionals.xml
+          token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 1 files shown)


Bibo-Joshi:
The PR comment is a downside IMO but one that I can live with. I've already pointed out that I'd like to have an option to disable it: https://github.com/codecov/feedback/issues/304#issuecomment-2362058506

.github/workflows/unit_tests.yml
```diff
@@ -57,21 +57,17 @@ jobs:
           # - without socks support
           # - without http2 support
           TO_TEST="test_no_passport.py or test_datetime.py or test_defaults.py or test_jobqueue.py or test_applicationbuilder.py or test_ratelimiter.py or test_updater.py or test_callbackdatacache.py or test_request.py"
-          pytest -v --cov -k "${TO_TEST}"
-          # Rerun only failed tests (--lf), and don't run any tests if none failed (--lfnf=none)
-          pytest -v --cov --cov-append -k "${TO_TEST}" --lf --lfnf=none --junit-xml=.test_report_no_optionals.xml
```
harshil21, line 62: why don't we run the test suite again to reduce CI failures? I didn't get the explanation in the commit message of https://github.com/python-telegram-bot/python-telegram-bot/pull/4487/commits/1f2144a790661acb4dff3c1a545e65226728ae58

harshil21 (commented):
really like the codecov overview of the tests, test durations is something we can work on directly, and failure rate is cumulative, an insight which is hard to gather using local pytest runs.

.github/workflows/unit_tests.yml
Bibo-Joshi: TBH, I don't have a solid fact-based argumentation 😅 IIRC #3631/#3626 where already more of a "let's see if this helps". Properly analyzing & addressing tests that tend to fail is more robust in the long run, I'd guess. Since we can now make that analysis, we don't necessarily need to water down the analysis.

harshil21: I think that some of the tests in the CI were simply failing due to race conditions or some other hard to debug reason, which is why I proposed that. Since that time we might have fixed many such tests. I suppose now we could remove this and check again.

harshil21 (approved):

PR #4488, opened 2024-09-19
Introduce `conftest.py` for File Related Tests

Bibo-Joshi:
Trying to get things for #4271 done in smaller chunks


harshil21 (approved):
well, it is my own commit so I have to approve it 😄

tests/_files/conftest.py
```diff
@@ -65,7 +65,7 @@ async def animated_sticker_set(bot):
 
 
 @pytest.fixture(scope="session")
-async def audio(bot, chat_id, aiolib):
```
harshil21: these were from when I was testing a library similar to `pytest-asyncio-cooperative`, i.e. running the tests themselves concurrently

Issue #4387, opened 2024-07-17
Adapt to RTD Addons


Bibo-Joshi:
Apart from the changes in #4386, some additional changes might be necessary to adapt to RTD Addons once our furo theme has adapted. Punshlines:

* the switch needs to happen by 2024-10-07
* furo tracks the transition in https://github.com/pradyunsg/furo/issues/795. I've already subscribed to that thread.

I recommend that we wait for the issue to be closed and then test out what we need to change. In particular I expect that we might need to

* check if our type-as-you-search implementation still works and if we can replace it by the new native feature
* Check if [this part](https://github.com/python-telegram-bot/python-telegram-bot/blob/422993b8ab05d49dd49ab58c500f880918d894a7/.readthedocs.yml#L29-L37) of our rtd build still works/is still needed

If furo is ready *before* 2024-10-07 I recommend that we try [switching early](https://docs.readthedocs.io/en/stable/addons.html#enabling-read-the-docs-addons)

If furo is *not ready* at 2024-10-07, we should try (preferably a bit before) to use [sphinx-build-compatiblity](https://github.com/readthedocs/sphinx-build-compatibility/) to manually enable backward compatibility. Or at least check if that's necessary.

harshil21:
great! I might need to archive furo-sphinx-search soon 👀 

Have we enabled any addons in our project?

Here are some addons apart from search as you type I found interesting (from https://about.readthedocs.com/blog/2024/04/enable-beta-addons/):

- Analytics
- DocDiff: this could replace our `Test Documentation Build` github action? It's a great feature
- Flyout
- Non-stable notification
- Latest version notification
- Sponsorship

PR #4492, opened 2024-09-21
Use `sphinx-build-compatibility` to Keep Sphinx Compatibility

Bibo-Joshi:
See #4387. 2024-10-07 is coming up fast and there seems to be little movement in https://github.com/pradyunsg/furo/issues/795 so far. So let's try if this backward compatibility thingy works.


Bibo-Joshi:
Build is up at https://docs.python-telegram-bot.org/en/docs-sphinx-build-compatiblity/. I don't see any obvious problems. The HTML zip thingy also seems to work: https://docs.python-telegram-bot.org/_/downloads/en/docs-sphinx-build-compatiblity/htmlzip/

PR #4493, opened 2024-09-21
Reduce Creation of HTTP Clients in Tests

Bibo-Joshi:
Apparently the function `load_ssl_context_verify` that's called upon the creation of `httpx.AsyncClient` takes quite a bit of time so that creating new `HTTPXRequest` objects (indirectly via `Bot`) is a performance bottleneck in the tests.

Hence, this changing `make_bot` to use `offline=True` by default and allowing online only in the tests that make requests.

This is just one small step on test performance improvement, but hey it's something :)


harshil21 (commented):
Do you know what % of time was the calling of `load_ssl_context_verify` taking?

PR #4542, opened 2024-10-31
Improve Exception Handling in `File.download_*`

Bibo-Joshi:
Response to the test failures on master https://github.com/python-telegram-bot/python-telegram-bot/actions/runs/11607799025/job/32340491289

These are caused by Telegram failing to return a `file_path` in the `get_file` response. I'm not sure why they stopped doing that. Further investigation might be necessary. I realized (maybe for the first time) that `file_path` is actually documented to be optional.

What this PR does is mostly soften the blow, i.e. provide understandable error handling.
This will not immediately fix the test suite. I'd like to wait a moment and see if the issue solves itself. If it does not, we'll have to adjust the other tests/mark them as xfail/report to telegram.

PS: see also https://github.com/tdlib/telegram-bot-api/issues/658

### DIFF
--- a/.github/workflows/docs-linkcheck.yml
+++ b/.github/workflows/docs-linkcheck.yml
@@ -3,6 +3,9 @@ on:
   schedule:
     # First day of month at 05:46 in every 2nd month
     - cron: '46 5 1 */2 *'
+  pull_request:
+    paths:
+      - .github/workflows/docs-linkcheck.yml
 
 jobs:
   test-sphinx-build:
@@ -10,7 +13,7 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        python-version: [3.10]
+        python-version: ['3.10']
         os: [ubuntu-latest]
       fail-fast: False
     steps:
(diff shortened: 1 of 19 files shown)


Bibo-Joshi:
Lastest commit

* adds retries of `get_file` to tests suite as workaround for tdlib/telegram-bot-api/issues/658, where it was observed that successive calls to that method do return the `file_path`
* Removes the special casing from `File.download_to_drive` designed to handle cases where `file_path` is not present but `file_id` is. Without `file_path`, it's not possible to download a file, as you need the URL for that. I tracked that if-clause back to https://github.com/python-telegram-bot/python-telegram-bot/issues/1590 and https://github.com/python-telegram-bot/python-telegram-bot/pull/1591 and TBH I can't see why we merged that. The fact that the [test case](https://github.com/python-telegram-bot/python-telegram-bot/pull/1591/files#diff-37da7019025c60a98cc1cea1ccc43217ec34e11a2d66e1b51ee29cf65abbbdf4) worked only by mocking the http request already shows that it's not an actual use case IMO

Poolitzer (approved):
LGTM

Bibo-Joshi:
New commits revert the workaround for `get_file` again as https://github.com/tdlib/telegram-bot-api/issues/658 has been resolved by now 😅

Poolitzer (approved):
LGTM

PR #4565, opened 2024-11-15
Improved comments for clarity and details :)

Snehashish06:
Improved comments for clarity and details in payment bot example:

- Added specific explanations for optional parameters in invoice functions
- Clarified purpose of ShippingQuery, PreCheckoutQuery, and payment callbacks
- Enhanced readability of logging and bot setup sections

### DIFF
--- a/examples/paymentbot.py
+++ b/examples/paymentbot.py
@@ -54,7 +54,7 @@ async def start_with_shipping_callback(update: Update, context: ContextTypes.DEF
     price = 1
     # Convert price to cents from dollars.
     prices = [LabeledPrice("Test", price * 100)]
-    # Optional parameters like need_shipping_address and is_flexible trigger additional user prompts
+    # Optional parameters like need_shipping_address and is_flexible trigger extra user prompts
     # https://docs.python-telegram-bot.org/en/stable/telegram.bot.html#telegram.Bot.send_invoice
     await context.bot.send_invoice(
         chat_id,
--- a/examples/paymentbot.py
+++ b/examples/paymentbot.py
@@ -21,7 +21,7 @@ from telegram.ext import (
 logging.basicConfig(
     format="%(asctime)s - %(name)s - %(levelname)s - %(message)s", level=logging.INFO
 )
-# Set a higher logging level for httpx to avoid logging every GET and POST request
+# set higher logging level for httpx to avoid all GET and POST requests being logged
 logging.getLogger("httpx").setLevel(logging.WARNING)
 
 logger = logging.getLogger(__name__)
--- a/examples/paymentbot.py
+++ b/examples/paymentbot.py
@@ -26,7 +26,8 @@ logging.getLogger("httpx").setLevel(logging.WARNING)
 
 logger = logging.getLogger(__name__)
 
-# Insert the token from your payment provider. In order to get a provider_token see https://core.telegram.org/bots/payments#getting-a-token
+# Insert the token from your payment provider.
+# In order to get a provider_token see https://core.telegram.org/bots/payments#getting-a-token
 PAYMENT_PROVIDER_TOKEN = "PAYMENT_PROVIDER_TOKEN"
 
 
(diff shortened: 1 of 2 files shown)


examples/paymentbot.py
```diff
@@ -21,41 +21,44 @@
 logging.basicConfig(
     format="%(asctime)s - %(name)s - %(levelname)s - %(message)s", level=logging.INFO
 )
-# set higher logging level for httpx to avoid all GET and POST requests being logged
+# Set a higher logging level for httpx to avoid logging every GET and POST request
```
Bibo-Joshi, line 24: this comment is used in all examples. If it's changed, I would like to have it adapted in the other examples as well

examples/paymentbot.py
```diff
@@ -21,41 +21,44 @@
...
+    """Sends an invoice which triggers a shipping query because it requests a shipping address and has flexible shipping."""
     chat_id = update.message.chat_id
     title = "Payment Example"
-    description = "Payment Example using python-telegram-bot"
-    # select a payload just for you to recognize its the donation from your bot
+    description = "Example of a payment process using the python-telegram-bot library."
+    # Unique payload to identify this payment request as being from your bot
     payload = "Custom-Payload"
-    # In order to get a provider_token see https://core.telegram.org/bots/payments#getting-a-token
+    # Set up the currency (e.g., "USD"), to get a list of supported currencies see https://core.telegram.org/bots/payments#supported-currencies
```
Bibo-Joshi, line 49: the `e.g., "USD"` seems superfluous to me - you can see the value just one line below :)

examples/paymentbot.py
```diff
@@ -21,41 +21,44 @@
...
+        "Use /shipping to receive an invoice with shipping included, or /noshipping for an "
         "invoice without shipping."
     )
-
     await update.message.reply_text(msg)
 
 
 async def start_with_shipping_callback(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
-    """Sends an invoice with shipping-payment."""
+    """Sends an invoice which triggers a shipping query because it requests a shipping address and has flexible shipping."""
```
Bibo-Joshi, line 43:
```suggestion
    """Sends an invoice which triggers a shipping query."""
```
the second part of the sentence is confusing to read IMO if you don't already know that this is referencing the parameters `is_flexible` and `needs_shipping_address`. You go into details on them below, so I recommend to keep it there .

examples/paymentbot.py
```diff
@@ -21,41 +21,44 @@
...
     prices = [LabeledPrice("Test", price * 100)]
 
-    # optionally pass need_name=True, need_phone_number=True,
-    # need_email=True, need_shipping_address=True, is_flexible=True
+    # The following optional parameters can be used to request additional information:
+    # - need_name=True: Requests the user's name.
+    # - need_phone_number=True: Requests the user's phone number.
+    # - need_email=True: Requests the user's email address.
+    # - need_shipping_address=True: Requests the user's shipping address.
+    # - is_flexible=True: Allows flexible shipping prices.
```
Bibo-Joshi, line 61: This is basically summarizing the [documentation of `send_invoice`](https://docs.python-telegram-bot.org/en/stable/telegram.bot.html#telegram.Bot.send_invoice). I recommend to link to that instead of repeating the content here. What you can do is point out that due to `need_shipping_address` and `is_flexible`, the user will be prompted with additional selections

examples/paymentbot.py
```diff
@@ -75,17 +78,17 @@ async def start_with_shipping_callback(update: Update, context: ContextTypes.DEF
...
+    """Sends an invoice without requiring shipping details."""
     chat_id = update.message.chat_id
     title = "Payment Example"
-    description = "Payment Example using python-telegram-bot"
-    # select a payload just for you to recognize its the donation from your bot
+    description = "Example of a payment process using the python-telegram-bot library."
+    # Unique payload to identify this payment request as being from your bot
     payload = "Custom-Payload"
-    # In order to get a provider_token see https://core.telegram.org/bots/payments#getting-a-token
+    # Check line 49 for supported currencies
```
Bibo-Joshi, line 87: absolute line numbers easily become outdated. Please either repeat the link or skip the comment.

Bibo-Joshi (changes requested):
Hi. Thanks very much for your PR! AFAIK you've already discussed the changes with Poolitzer. I've added a few additional comments below. None of them are critical.
Please also have a look at the failing pre-commit checks. They are due to a few lines exceeding the line length limit of 99 chars.

examples/paymentbot.py
Snehashish06: I skipped it, as I believe most developers reviewing this example would recognize it at a glance, and it’s already mentioned earlier in the code.

examples/paymentbot.py
Snehashish06: I've added a link to the documentation.

examples/paymentbot.py
Snehashish06: Got it, thank you for the suggestion. I'll incorporate the changes.

examples/paymentbot.py
Snehashish06: Understood, I've made the change as suggested and removed the 'e.g., "USD"' part since the value is already defined on the next line.

examples/paymentbot.py
Snehashish06: I'm unsure whether I should revert it or apply the same change to all the other examples.

Bibo-Joshi: Your call :) I'm fine with either solution. Thanks for the other updates!

Snehashish06: Well, I've changed it back to the original because I believe both versions explain the purpose well.

Bibo-Joshi (changes requested):
Thanks for the update! Now only the pre-commit failure is remaining:)

Snehashish06:
Passed  ;)

Bibo-Joshi:

PR #4589, opened 2024-12-04
Widen specified type for `allowed_updates` Application/Updater parameter

nemacysts:
This parameter has been inconsistently typed as either list[str] or tuple[str] - as either is acceptable, we can widen the type to Sequence[str] to satisfy `mypy` and/or other type checkers.

This will allow for constructs like:
```Python
application.run_polling(
    allowed_updates=(
        UpdateType.MESSAGE,
        UpdateType.EDITED_MESSAGE,
    )
)
```
which currently fails mypy (and likely other typecheckers) with an error like:
`error: Argument "allowed_updates" to "run_polling" of "Application" hasincompatible type "tuple[UpdateType, UpdateType]"; expected "list[str] | None"  [arg-type]`

- [x] run pre-commit
- [x] run tests
- [x] add self to AUTHORS.rst
- [x] add versionchanged to existing docstrings

### DIFF
--- a/telegram/_webhookinfo.py
+++ b/telegram/_webhookinfo.py
@@ -17,6 +17,7 @@
 # You should have received a copy of the GNU Lesser Public License
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains an object that represents a Telegram WebhookInfo."""
+
 from collections.abc import Sequence
 from datetime import datetime
 from typing import TYPE_CHECKING, Optional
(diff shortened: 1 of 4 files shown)


telegram/_webhookinfo.py
nemacysts, line 1: i can back these changes out if preferred - i was simply correcting the inconsistency in the docstring type vs the docstring description

nemacysts: note: this file already typed allowed_updates correctly (and where allowed_updates was typed as a `tuple[str, ...]`, it was because the RHS of that assignment converted to a tuple with `parse_sequence_arg()` - hence the docstring only changes here :)

nemacysts:
looks like `Unit Tests / pytest (3.10, ubuntu-latest)` is failing to report coverage with `error - 2024-12-04 05:33:48,773 -- Report creating failed: {"message":"Token required because branch is protected"}` - however, the actual test stage seems to have passed all tests :)

harshil21 (approved):
Thanks for the fix!

Bibo-Joshi:

PR #4594, opened 2024-12-05
Full Support for Bot API 8.1

Bibo-Joshi:
closes #4592

- [ ] Added `.. versionadded:: NEXT.VERSION`, `.. versionchanged:: NEXT.VERSION`, `.. deprecated:: NEXT.VERSION` or `.. versionremoved:: NEXT.VERSION` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)
- [ ] Created new or adapted existing unit tests
- [ ] Documented code changes according to the [CSI standard](https://standards.mousepawmedia.com/en/stable/csi.html)_
- [ ] Added myself alphabetically to `AUTHORS.rst` (optional)
- [ ] Added new classes & modules to the docs and all suitable `__all__` s
- [ ] Checked the [Stability Policy](https://docs.python-telegram-bot.org/stability_policy.html) in case of deprecations or changes to documented behavior
- [ ] Check the diff files in [this channel](https://t.me/bot_api_changes) for documentation updates

**If the PR contains API changes (otherwise, you can ignore this passage)**

- [ ] Checked the Bot API specific sections of the [Stability Policy](https://docs.python-telegram-bot.org/stability_policy.html)
- [ ] Created a PR to remove functionality deprecated in the previous Bot API release ([see here](https://docs.python-telegram-bot.org/en/stable/stability_policy.html#case-2))

- [ ] New classes:

   - [ ] Added `self._id_attrs` and corresponding documentation
   - [ ] `__init__` accepts `api_kwargs` as kw-only

- [ ]  Added new shortcuts:

   - [ ] In `telegram.Chat` & `telegram.User` for all methods that accept `chat/user_id`
   - [ ] In `telegram.Message` for all methods that accept `chat_id` and `message_id`
   - [ ] For new `telegram.Message` shortcuts: Added `quote` argument if methods accepts `reply_to_message_id`
   - [ ] In `telegram.CallbackQuery` for all methods that accept either `chat_id` and `message_id` or `inline_message_id`

- [ ] If relevant:

   - [ ] Added new constants at `telegram.constants` and shortcuts to them as class variables
   - [ ] Link new and existing constants in docstrings instead of hard-coded numbers and strings
   - [ ] Add new message types to `telegram.Message.effective_attachment`
   - [ ] Added new handlers for new update types

     - [ ] Add the handlers to the warning loop in the `telegram.ext.ConversationHandler`

   - [ ] Added new filters for new message (sub)types
[...]

PR #4597, opened 2024-12-07
Start using MessageLimit.DEEP_LINK_LENGTH constant

nemacysts:
This constant was added back in #3315 - but it was never referenced in the ptb codebase.

However, as mentioned in a comment added in #3351 - there was at least one place that was hardcoding this value rather than referencing the named constant

As far as I can tell (with the help of `git grep`), this is likely the only 64->MessageLimit.DEEP_LINK_LENGTH swap that can be made :)

NOTE: I've verified that the formatting of the documentation looks the same in terms of line breaks after breaking the ValueError exception text into two lines (i.e., there's no extraneous newlines)

Verification:
- [x] run precommit
- [x] run tests
- [x] `make -C docs/ singlehtml` and inspected the generated docs

### DIFF
--- a/telegram/constants.py
+++ b/telegram/constants.py
@@ -1835,7 +1835,6 @@ class MessageLimit(IntEnum):
     :paramref:`~telegram.Bot.edit_message_text.text` parameter of
     :meth:`telegram.Bot.edit_message_text`.
     """
-    # TODO this constant is not used. helpers.py contains 64 as a number
     DEEP_LINK_LENGTH = 64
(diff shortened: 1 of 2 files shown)


nemacysts:
hmm, the failing `check-conformity` tests seem unrelated to this change - please correct me if I'm wrong tho!

Bibo-Joshi (approved):
Nice catch, thank you very much!
The failing test is just due to #4594 not being part of your branch, but that's indeed not your fault :)

PR #4616, opened 2024-12-29
Allow Input of Type `Sticker` for Several Methods

### LINKED ISSUE #4580: Extend Support for Convenience Types in Bot Methods

In `test_official` we currently allow convenience input types only for `send_*` methods:

https://github.com/python-telegram-bot/python-telegram-bot/blob/151123745e9e8d53c54fd6715008b3e8c85bdc82/tests/test_official/arg_type_checker.py#L174-L178

https://github.com/python-telegram-bot/python-telegram-bot/blob/151123745e9e8d53c54fd6715008b3e8c85bdc82/tests/test_official/exceptions.py#L39-L48

Removing the `obj.__name__.startswith("send")` reveals that there are a few more parameters that could use convenience types.

* Set up the "additional types" excetpions in test_official such that they can either be defined globally or per method, remove the `obj.__name__` check
* add support for convenience types where possible

_No response_

### Additional context

_No response_

Bibo-Joshi:
Closes #4580
Adds convenience types for the methods that were shown by removing the `obj.__name__.startswith("send")` as explained in #4580.
Additional types can now follow more easily


telegram/_bot.py
```diff
@@ -9248,7 +9280,7 @@ async def replace_sticker_in_set(
         self,
         user_id: int,
         name: str,
-        old_sticker: str,
+        old_sticker: Union[str, Sticker],
```
harshil21:
```suggestion
        old_sticker: Union[str, "Sticker"],
```
this is different than the other methods?

tests/test_official/arg_type_checker.py
```diff
@@ -205,15 +215,21 @@ def check_param_type(
...
-            if ptb_param.name == param_name and is_class is is_expected_class:
+    elif mappings := [
+        mapping
+        for pattern, mapping in PTCE.COMPLEX_TYPES.items()
+        if (re.match(pattern, obj.__name__))
+    ]:
+        for mapping in mappings:
+            for key, exception_type in mapping.items():
+                if not re.match(key, ptb_param.name):
+                    continue
```
harshil21, line 226: perhaps this can be extracted into a different helper method?

tests/test_official/arg_type_checker.py
```diff
@@ -205,15 +215,21 @@ def check_param_type(
...
+                log("Converting `%s` to a simpler type!\n", ptb_param.name)
                 ptb_annotation = wrap_with_none(tg_parameter, exception_type, obj)
 
     # 6) HANDLING DEFAULTS PARAMETERS:
     # Classes whose parameters are all ODVInput should be converted and checked.
-    elif obj.__name__ in PTCE.IGNORED_DEFAULTS_CLASSES:
+    if obj.__name__ in PTCE.IGNORED_DEFAULTS_CLASSES:
```
harshil21, line 232: hmm I vaguely remember that using an `if` instead of an `elif` doesn't work since the order in which the cases run matters... Does everything still work as expected?

Bibo-Joshi: the issue is that I don't know if the if-clause actually matches in line 218. I need to first check the inner loop. that way, the if-clause is currently "too greedy" and the latter cases are falsely skipped for some items. Extracting to a standalone method could help … but everything seems to work okay, yes

harshil21 (approved):
LGTM

Issue #4575, opened 2024-11-24
Use `datetime.timedelta` to Represent Time Periods


Bibo-Joshi:
### What kind of feature are you missing? Where do you notice a shortcoming of PTB?

Several Telegram Objects and methods currently contain fields/parameters with names `\w+_period`, wherer the integer value represents a time span in seconds.
Working with plain seconds is not very pythonic, just like working with posix timestamps isn't. 

### Describe the solution you'd like

In the long run, we should convert the attributes to `timedelta` objects and make the parameters accept timedelta objects.

#### Methods Parameters

This is realtively easy, simply accept `timedelta` as well as integers and pass `totalseconds()` to TG. Converting `dtm.timedelta` to `int` can probably be done in the internals of `Bot` or requests so that it doesn't have to happen in each bot method.

#### Classes

* Both attributes and arguments should *only* accept timedeltas in the long run.
* Since this is a breaking change, we should issue a warning for now about this changing in the future and encouraging users to harden their code against the change
* For making the `timedelta` support available early, I have two ideas, ~both optional~
  * ~We can provide a temporary property that returns the timedelta object (e.g. `xy_period_timedelta`). This would need to issue a warning stating that it will be removed once `timedelta` is natively supported. Not overly clean …~
  * We anyways need to make the `*_period` attributes properties to issue the deprecation warning. We could use this and look for an environment variable, say `PTB_TIMEDELTA`. If set, we return the timedelta object, otherwise the `int`. → let's do this

### Additional context

Adding support for method parameters can be done independently of handling classes.

cuevasrja:
Hi, my class group and I want to work in this Issue.

Group:
@Jeam-zx
@Migueldsc12
@AnyaMarcanito
@henryg311

Bibo-Joshi:
@cuevasrja & friends if you're still interested, would you like to get started on this? I think it would be good to split this into two PRs, one for the Bot method arguments and one for the classes. Some additional pointers on both:

* for the arguments of Bot methods, converting `timedelta` to `int` should happen at
    https://github.com/python-telegram-bot/python-telegram-bot/blob/2ac52018c29c791aa0438750e5bcdf2da8945bf8/telegram/request/_requestparameter.py#L116-L117
    and corresponding new parametrizations need to be added to
[...]

PR #4651, opened 2025-01-15
Accept `datetime.timedelta` Input in `Bot` Method Parameters

Bibo-Joshi:
Adresses part one of #4575

### DIFF
--- a/telegram/request/_requestparameter.py
+++ b/telegram/request/_requestparameter.py
@@ -117,9 +117,12 @@ class RequestParameter:
             return to_timestamp(value), []
         if isinstance(value, dtm.timedelta):
             seconds = value.total_seconds()
+            # We convert to int for completeness for whole seconds
             if seconds.is_integer():
                 return int(seconds), []
-            return value.total_seconds(), []
+            # The Bot API doesn't document behavior for fractions of seconds so far, but we don't
+            # want to silently drop them
+            return seconds, []
         if isinstance(value, StringEnum):
             return value.value, []
         if isinstance(value, InputFile):
--- a/telegram/request/_requestparameter.py
+++ b/telegram/request/_requestparameter.py
@@ -115,6 +115,8 @@ class RequestParameter:
         """
         if isinstance(value, dtm.datetime):
             return to_timestamp(value), []
+        if isinstance(value, dtm.timedelta):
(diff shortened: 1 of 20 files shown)


telegram/request/_requestparameter.py
```diff
@@ -115,6 +115,11 @@ def _value_and_input_files_from_input(  # pylint: disable=too-many-return-statem
         """
         if isinstance(value, dtm.datetime):
             return to_timestamp(value), []
+        if isinstance(value, dtm.timedelta):
+            seconds = value.total_seconds()
+            if seconds.is_integer():
+                return int(seconds), []
+            return value.total_seconds(), []
```
harshil21, line 122: this is a little confusing. total_seconds always returns an integer right?

harshil21 (approved):
lgtm

telegram/request/_requestparameter.py
Bibo-Joshi: not necessarily, as the delta could also have microseconds

harshil21: I see. But then the logic should be reversed?

```suggestion
            seconds = value.total_seconds()  # could return seconds with microseconds (a float)
            if seconds.is_integer():  # if it's only seconds
                return seconds, []
            return int(seconds), []  # if it's a float, round down to the nearest integer
```

Bibo-Joshi: No, the `int` is mainly there for mypy. I don't want to cut off the granularity. We probably don't use it anywhere right now, but this feels like the thing that will be a nightmare to debug if we ever need it and then have to track down where there decimal points are lost 😬

harshil21: hmm, but does the Telegram API accept it if `float`s are passed in the field? The `WithRequest*` tests only tested with seconds and not milliseconds iisc.

Bibo-Joshi: I'm not sure tbh. So far I would have expected them to not accept. I can double check that. My pov is rather that we should not make any silent conversions. If the user passes nonsense, they should get an understandable error message. we also don't validate any other input

harshil21: fair enough

Issue #2770, opened 2021-11-07
[Discussion:] ConversationHandler


Bibo-Joshi:
Several ideas for changes and/or additions have been gathered in the v14 project, mentioned in the user group or issues or dwelled in the back of my head. So I'm writing this issue to keep track of them better.

The following list of ideas should be discussed before release of v14 because at least some of them are easier to implement if we do them non-backwards compatible. I'd like to emphasize that most of these are very rough first thoughts and I'm fully aware that we may scrape many of them.

Related: https://github.com/python-telegram-bot/python-telegram-bot/discussions/2802

---

* Make `CH.fallbacks` optional. I've seen a lot of people (including me) setting at to an empty list for some use cases … Poolitzer already expressed support for this idea
* ~ConverstanionHandler + run_async: Problem is that we can't persist futures/promises. But we should be able to guarantee that all futures are done on (clean) shutdown. One could add something like `BasePersistence.resolve_conversation_promises` that persistence classes can call before flushing on shutdown.~ This already get's handled on v20
* #2390 (closed just for better overview, not actually implemented)
* add a "prefallbacks" list, i.e. the state-handlers only trigger, if the "prefallback" handlers don't trigger. The current alternative is to add those handlers to the `fallbacks` list, which requires to make all state handlers ignore updates that should be handled by the `fallbacks`. see also #2764 #2960
* Include `entry_points`, `fallbacks` (and eventual "prefallbacks") into the `states` dict. We already have special states for `TIMEOUT` and `WAITING`, so I think it would make the interface cleaner if we used special states `(PRE)FALLBACK` and `END` as keys for these lists rather than making them standalone kwargs of `__init__`
* Think about "conversation related data", e.g. `context.conversation_data`. It should be an object unique per conversation key and `ConversationHandler` instance. It may even be automatically deleted when the conversation ends . See also #4136
* Rethink which parts of `ConversationHandler` we consider public.
  * e.g. the `conversations` is currently public, although undocumented and basically part of the internals. If we make it private, we can rework the setters of `persistence` and `conversations` to something like `get_persisted_data(persistence)` or `initialize_persistence(persistence)`
  * We also rather regularly get questions like "how do I access the current state of the conversation" and "how do I check if a user has an active conversation". Making the logic of how the keys are build and how `conversations` works public *may* help with some of these and also with the above `context.conversation_data`
[...]

PR #4669, opened 2025-02-04
Initial FSM PoC

Bibo-Joshi:
Works on #2770 

"It has been a while since we openend #2770" would be an understatement. Yet here we :D 
Finally, I got around to draft up a somewhat proper PoC for my ideas outlined in https://github.com/python-telegram-bot/python-telegram-bot/issues/2770#issuecomment-2134022736. I also attached an example that showcases how the FSM setup can be used. You'll need ≥2 accounts to test it.

To be clear: as of 2025-02-04 this PR does by far not include everything mentioned in https://github.com/python-telegram-bot/python-telegram-bot/issues/2770#issuecomment-2134022736. I hope to work on it in the nearish-future. Continued work on this PR could be greatly supported by intermediate reviews, comments & discussions :) 

---

### ToDos / Ideas / Thoughts

(Listing here only things that came to my mind explicitly while working on this PR, not copying everything from #2770)

- [ ] Using `context.job_queue.run_once` for conversation timeouts is relatively easy, but would have to repeated everywhere which also holds for the cancellation. Maybe we can think of a clever way to abstract that for the user.

### DIFF
--- a/examples/fsmbot.py
+++ b/examples/fsmbot.py
@@ -1,0 +1,195 @@
+#!/usr/bin/env python
+# pylint: disable=unused-argument
+# This program is dedicated to the public domain under the CC0 license.
+"""Simple state machine to handle user support.
+One admin is supported. The admin can have one active conversation at a time. Other users
+are put on hold until the admin finishes the current conversation.
+In each conversation, the admin and the user take turns to send messages.
+"""
+import logging
+from typing import Optional
+
+from telegram import Update
+from telegram.ext import (
+    Application,
+    CommandHandler,
+    ContextTypes,
+    FiniteStateMachine,
+    MessageHandler,
+    State,
+    filters,
+)
+
+# Enable logging
+logging.basicConfig(
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s", level=logging.DEBUG
+)
+logging.getLogger("httpx").setLevel(logging.WARNING)
+logging.getLogger("httpcore").setLevel(logging.WARNING)
+logging.getLogger("telegram").setLevel(logging.WARNING)
+logging.getLogger("telegram.ext.Application").setLevel(logging.DEBUG)
+
+logger = logging.getLogger(__name__)
+
+
+class UserSupportMachine(FiniteStateMachine[Optional[int]]):
+
+    HOLD = State("HOLD")
+    WELCOMING = State("WELCOMING")
+    WAITING_FOR_REPLY = State("WAITING_FOR_REPLY")
+    WRITING = State("WRITING")
+
+    def __init__(self, admin_id: int):
+        self.admin_id = admin_id
+        self._states: dict[int, State] = {}
+        super().__init__()
+
+    def _get_admin_state(self) -> State:
+        return self._states.get(self.admin_id, State.IDLE)
+
+    def get_active_key_state(self, update: object) -> tuple[Optional[int], State]:
+        if not isinstance(update, Update):
+            return None, State.IDLE
+        if not (user := update.effective_user):
+            return None, State.IDLE
+
+        # Admin is easy - just return the state
+        admin_state = self._get_admin_state()
+        if user.id == self.admin_id:
+            logging.debug("Returning admin state: %s", admin_state)
+            return self.admin_id, admin_state
+
+        # If the user state is already non-idle, we are already in the business logic
+        if (user_state := self._states.get(user.id, State.IDLE)) != State.IDLE:
+            logging.debug("Returning user state: %s", user_state)
+            return user.id, user_state
+
+        # On first interaction, we need to determine what to do with the user
+        # if the admin is not idle, we put the user on hold. Otherwise, they may send the first
+        # message, and we put the admin in waiting for reply to avoid another user occupying the
+        # admin first
+        effective_user_state = self.HOLD if admin_state != State.IDLE else self.WELCOMING
+        self.do_set_state(user.id, effective_user_state)
+        if effective_user_state == self.WELCOMING:
+            self.do_set_state(self.admin_id, self.WAITING_FOR_REPLY)
+
+        logging.debug("Returning user state: %s", effective_user_state)
+        return user.id, effective_user_state
+
+    def do_set_state(self, key: int, state: State) -> None:
+        key_name = "admin" if key == self.admin_id else key
+        logging.debug("Setting %s state to %s", key_name, state)
+        self._states[key] = state
+
+
+async def welcome_user(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    await update.effective_message.forward(context.bot_data["admin_id"])
+    await update.effective_message.reply_text(
+        "Welcome! Your message has been forwarded to the admin. They will get back to you soon.",
+    )
+    await context.set_state(UserSupportMachine.WAITING_FOR_REPLY)
+    await context.fsm.set_state(context.bot_data["admin_id"], UserSupportMachine.WRITING)
+    context.bot_data["active_user"] = update.effective_user.id
+
+
+async def conversation_timeout(context: ContextTypes.DEFAULT_TYPE) -> None:
+    active_user = context.bot_data.get("active_user")
+    admin_id = context.bot_data["admin_id"]
+
+    async def handle(user_id: int) -> None:
+        await context.bot.send_message(
+            user_id, "The conversation has been stopped due to inactivity."
+        )
+        await context.fsm.set_state(user_id, State.IDLE)
+
+    if active_user:
+        await handle(active_user)
+    await handle(admin_id)
+
+
+async def handle_reply(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    # Cancel the conversation timeout
+    if job := context.bot_data.get("conversation_timeout"):
+        job.schedule_removal()
+
+    if not (active_user := context.bot_data.get("active_user")):
+        logger.warning("No active user found, ignoring message")
+
+    target = (
+        active_user
+        if update.effective_user.id == (admin_id := context.bot_data["admin_id"])
+        else admin_id
+    )
+    await context.bot.send_message(target, update.effective_message.text)
+    logging.debug("Forwarded message to %s", target)
+    await context.set_state(UserSupportMachine.WAITING_FOR_REPLY)
+    logging.debug("Done setting state to WAITING_FOR_REPLY for %s", target)
+    await context.fsm.set_state(target, UserSupportMachine.WRITING)
+    logging.debug("Done setting state to WRITING for %s, context.fsm_key")
+
+    # Reset the conversation timeout
+    job = context.job_queue.run_once(conversation_timeout, 30)
+    context.bot_data["conversation_timeout"] = job
+
+
+async def stop_conversation(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    text = "The conversation has been stopped."
+    admin_id = context.bot_data["admin_id"]
+    active_user = context.bot_data.get("active_user")
+
+    await context.bot.send_message(admin_id, text)
+    await context.fsm.set_state(admin_id, State.IDLE)
+    if active_user:
+        await context.bot.send_message(active_user, text)
+        await context.fsm.set_state(active_user, State.IDLE)
+
+
+async def hold_melody(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    await update.effective_message.reply_text(
+        "You have been put on hold. The admin will get back to you soon. Please hear some music "
+        "while you wait: https://www.youtube.com/watch?v=dQw4w9WgXcQ"
+    )
+
+
+async def not_your_turn(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    await update.effective_message.reply_text(
+        "It's not your turn yet. Please wait for the other party to reply to your message."
+    )
+
+
+def main() -> None:
+    application = Application.builder().token("TOKEN").build()
+    application.fsm = UserSupportMachine(admin_id=123456789)
+    application.bot_data["admin_id"] = application.fsm.admin_id
+
+    # Users are welcomed only if they are in the corresponding state
+    application.add_handler(
+        MessageHandler(~filters.User(application.fsm.admin_id), welcome_user),
+        state=UserSupportMachine.WELCOMING,
+    )
+
+    # Conversation logic:
+    # * forward messages between user and admin
+    # * stop the conversation at any time (admin or user)
+    # * point out that the other party is currently writing
+    application.add_handler(
+        MessageHandler(filters.ALL, handle_reply), state=UserSupportMachine.WRITING
+    )
+    application.add_handler(
+        CommandHandler("stop", stop_conversation),
+        state=UserSupportMachine.WAITING_FOR_REPLY | UserSupportMachine.WRITING,
+    )
+    application.add_handler(
+        MessageHandler(filters.ALL, not_your_turn), state=UserSupportMachine.WAITING_FOR_REPLY
+    )
+
+    # If the admin is busy, put the user on hold
+    application.add_handler(
+        MessageHandler(filters.ALL, hold_melody), state=UserSupportMachine.HOLD
+    )
+
+    application.run_polling(allowed_updates=Update.ALL_TYPES)
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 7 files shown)


telegram/ext/_fsm/machine.py
```diff
@@ -0,0 +1,200 @@
...
+
+from telegram.ext._fsm.states import State
+from telegram.ext._utils.types import JobCallback
+
+if TYPE_CHECKING:
+    from collections.abc import MutableMapping
+
+    from telegram.ext import JobQueue
+
+_KT = TypeVar("_KT", bound=Hashable)
```
Poolitzer, line 24: Im back with my thinking about private stuff. Why do we private it here, not in other files; should we. Hm.

telegram/ext/_fsm/machine.py
```diff
@@ -0,0 +1,200 @@
...
+        self.key: _KT = key
+        self.state: State = state
+        self.version: int = version
+
+
+class FiniteStateMachine(abc.ABC, Generic[_KT]):
+    def __init__(self) -> None:
+        self._locks: MutableMapping[_KT, asyncio.Lock] = weakref.WeakValueDictionary()
+
+        # There is likely litte benefit for a user to customize how exactly the states are stored
```
Poolitzer, line 40:
```suggestion
        # There is likely little benefit for a user to customize how exactly the states are stored
```

telegram/ext/_fsm/machine.py
```diff
@@ -0,0 +1,200 @@
...
+        self._states = MappingProxyType(self.__states)
+
+        self.__job_queue: Optional[weakref.ReferenceType[JobQueue]] = None
+        self.__history: Mapping[_KT, MutableSequence[State]] = defaultdict(
+            lambda: deque(maxlen=10)
+        )
+
+    @property
+    def states(self) -> Mapping[_KT, tuple[State, int]]:
+        return self._states
```
Poolitzer, line 54: Why do we do that btw doesn't the MappingProxyType already do it?

telegram/ext/_fsm/machine.py
```diff
@@ -0,0 +1,200 @@
...
+
+        # There is likely litte benefit for a user to customize how exactly the states are stored
+        # and accessed. So we make this private and only provide a read-only view.
+        self.__states: dict[_KT, tuple[State, int]] = defaultdict(
+            lambda: (State.IDLE, time.perf_counter_ns())
+        )
+        self._states = MappingProxyType(self.__states)
+
+        self.__job_queue: Optional[weakref.ReferenceType[JobQueue]] = None
+        self.__history: Mapping[_KT, MutableSequence[State]] = defaultdict(
```
Poolitzer, line 48: Why are these with two underscores.

Oh and we should make history optional/limit it so bots on a bigger scale doesn't lead to huge memory issues

telegram/ext/_fsm/machine.py
```diff
@@ -0,0 +1,200 @@
...
+            lambda: deque(maxlen=10)
+        )
+
+    @property
+    def states(self) -> Mapping[_KT, tuple[State, int]]:
+        return self._states
+
+    def store_state_history(self, key: _KT, state: State) -> None:
+        # Making this public so that users can override if they want to customize the history
+        # E.g., they could want to store more/fewer states, also depending on the key
```
Poolitzer, line 58: dont we then also need to support removing history?

examples/fsmbot.py
```diff
@@ -0,0 +1,203 @@
...
+        super().__init__()
+
+    def _get_admin_state(self) -> tuple[State, int]:
+        return self._states[self.admin_id]
+
+    def get_state_info(self, update: object) -> StateInfo[Optional[int]]:
+        if not isinstance(update, Update) or not (user := update.effective_user):
+            key = None
+            state, version = self.states[key]
+            return StateInfo(key=key, state=state, version=version)
```
Poolitzer, line 54: I am confused what will be the Values of stateinfo?

Also I get the error Expected type '_KT ≤: Hashable', got 'None' instead

examples/fsmbot.py
```diff
@@ -0,0 +1,203 @@
...
+        if user_state.matches(self.WELCOMING | self.WRITING | self.WAITING_FOR_REPLY):
+            logging.debug("Returning user state: %s", user_state)
+            return StateInfo(user.id, user_state, user_version)
+
+        # On first interaction, we need to determine what to do with the user
+        # if the admin is not idle, we put the user on hold. Otherwise, they may send the first
+        # message, and we put the admin in waiting for reply to avoid another user occupying the
+        # admin first
+        effective_user_state = self.HOLD if admin_state != State.IDLE else self.WELCOMING
+        self._do_set_state(user.id, effective_user_state, user_version)
```
Poolitzer, line 73: IMO this should also not be private if users are supposed to use it

telegram/ext/_fsm/machine.py
```diff
@@ -0,0 +1,200 @@
...
+        _LOGGER.debug("Setting %s state to %s", key, state)
+        if state is State.ANY:
+            raise ValueError("State.ANY is not supported in set_state")
+
+        if version and version != self._states.get(key, (None, None))[1]:
+            raise ValueError("Optimistic locking failed. Not updating state.")
+
+        if jq := self._get_job_queue(raise_exception=False):
+            # This is a rather tight coupling between FSM and JobQueue
+            # Not sure if we like that. Makes it even harder to replace JobQueue
```
Poolitzer, line 108: No I think it is fine. We can always do the work when for some reason we have the need to but for now we should enjoy it imo

examples/fsmbot.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+async def unsupported_message(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
+    await update.effective_message.reply_text("This message is not supported.")
+
+
+def main() -> None:
+    application = Application.builder().token("TOKEN").build()
+    application.fsm = UserSupportMachine(admin_id=123456)
+    application.fsm.set_job_queue(application.job_queue)
```
Poolitzer, line 166: these should be part of the builder and the second one set by default imo

telegram/ext/_fsm/machine.py
```diff
@@ -0,0 +1,200 @@
...
+                yield
+            finally:
+                self._do_set_state(key, current_state, new_version)
+
+    @staticmethod
+    def _build_job_name(keys: Sequence[_KT]) -> str:
+        return f"FSM_Job_{'_'.join(str(hash(k)) for k in keys)}"
+
+    def set_job_queue(self, job_queue: "JobQueue") -> None:
+        self.__job_queue = weakref.ref(job_queue)
```
Poolitzer, line 151: Why weakref wont it/the user experience break if the supplied job queue gets destroyed

telegram/ext/_fsm/machine.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def _get_job_queue(self, raise_exception: bool = True) -> Optional["JobQueue"]:
+        if self.__job_queue is None:
+            if raise_exception:
+                raise RuntimeError("JobQueue not set")
+            return None
+        job_queue = self.__job_queue()
+        if job_queue is None:
+            if raise_exception:
+                raise RuntimeError("JobQueue was garbage collected")
```
Poolitzer, line 167: ha here see. why.

telegram/ext/_fsm/states.py
```diff
@@ -0,0 +1,114 @@
...
+    """Default State for all Finite State Machines"""
+    ANY: "State"
+    """Special State that matches any other State. Useful to define fallback behavior.
+    *Not* supported in ``set_state`` method of FSMs.
+    """
+
+    def __init__(self, uid: Optional[str] = None):
+        effective_uid = uid or uuid4().hex
+        if effective_uid in self.__knows_uids:
+            raise ValueError(f"Duplicate UID: {effective_uid} already registered")
```
Poolitzer, line 26: is the error saying UID correct? I would understand the optional type hint to be anything hashable

telegram/ext/_fsm/states.py
```diff
@@ -0,0 +1,114 @@
+"""This Module contains implementations of State classes for Finite State Machines"""
+
+import abc
+import contextlib
+from typing import ClassVar, Optional
+from uuid import uuid4
+
+
+class State(abc.ABC):
+    __knows_uids: ClassVar[set[str]] = set()
```
Poolitzer, line 10: Wait I must really misunderstand smth, how does this work across states

Poolitzer (commented):
Just some first impressions, once I gain a further understanding I can give more

Poolitzer:
A thought I had: Since this is such a huge change for something so integral to our project, should we provide a beta version/implementation after we are satisfied with the status and announce it to get people to try it out and provide feedback before we make a first release?

Bibo-Joshi:
> A thought I had: Since this is such a huge change for something so integral to our project, should we provide a beta version/implementation after we are satisfied with the status and announce it to get people to try it out and provide feedback before we make a first release?

That is indeed a very good idea 👍 
My thoughts on this are roughly like this:

* the overall change should (hopefully) be such that nothing changes if you just don't use the new states.
* that way we can introduce the refactoring & deprecate the existing CH without braking changes.
* We can mark the status as "public beta feature" and at some point we can say "well, starting with version 25.0, states are now marked as stable".

that way we wouldn't have to do any additional actual beta releases.
If we prefer to have a proper beta release, I'm open to that. We could e.g. do states update on a 23.0bx line that gets *only* updates for states while other stuff (especially API updates only go into 22.x) . We shouldn't do that for too long IMO, we saw how much effort that takes with v20.
Encouraging people to try out a branch is ofc an easy option.

Poolitzer:
I dont get the first point, what do you mean with not using new states. In conjunction with ConvHandler? I thought this is supposed to fully replace it. In that case I would like a big announcement at least (maybe also in readme or so).

We might not really need an extra release now that I think of it, just make a big fuss about it before calling it stable and doing the deprecation error warning.

Bibo-Joshi:
> I dont get the first point, what do you mean with not using new states. In conjunction with ConvHandler? I thought this is supposed to fully replace it. In that case I would like a big announcement at least (maybe also in readme or so).

Yes, this is supposed to replace CH. But, since this is acting on a different level than handlers, the `conversationhandlerbot.py` example will still run as before. We should throw hella-big warnings if states & CH are used in conjunction, true.
All other features that are independent of CH will also all run the same. all keys will always be in state "IDLE".

> We might not really need an extra release now that I think of it, just make a big fuss about it before calling it stable and doing the deprecation error warning.

👍

Poolitzer:
Ah yes you mean in the transition period alright. Okay so lets do a release when we are happy and announce that its there and see if we get feedback, and then do a deprecation later. Perfect.

telegram/ext/_fsm/machine.py
Bibo-Joshi: There is `job_queue.application`. that's also weakref b/c we have a cyclic reference: `application is application.job_queue.application`. Now we also have `application is application.fsm.job_queue.application`. I'm not 100% sure if this is strictly necessary, but it seemed safer to me 🤔 I can at very least add a comment.

telegram/ext/_fsm/states.py
Bibo-Joshi: uh … that's how class variables work? :D In the init `self.__knows_uids` is basically the same as `State.__knows_uids` …

telegram/ext/_fsm/states.py
Bibo-Joshi: Sorry, I don't get you. What do you think is wrong? the phrasing of the error message?

examples/fsmbot.py
Bibo-Joshi: yes, absolutely. #PoC 👼

examples/fsmbot.py
Bibo-Joshi: https://peps.python.org/pep-0008/#designing-for-inheritance

The point here is that we need to differentiate between methods that subclasses of FSM are allowed to use and methods that make up the public interface of an FSM object. The former can be more methods than the latter.
The difficulty here is that we ratherly likely need a way to set a state without waiting for the lock, b/c that could get you in a deadlock situation.

examples/fsmbot.py
Bibo-Joshi: > I am confused what will be the Values of stateinfo?

The naming is probably not great :/
My goal was to somehow bundle the information of

* which conversation/thread/key (we'll need to find a good name for that) are we talking about?
* what's the current reported state for that?
* how can I verify that the state wasn't changed since I got this information?

> Also I get the error Expected type '_KT ≤: Hashable', got 'None' instead

that seems to be an IDE issue - `None` is in fact hashable and `isinstance(None, abc.Hashable) is True` …

telegram/ext/_fsm/machine.py
Bibo-Joshi: valid point, we could add a `drop_state_history(key)`

telegram/ext/_fsm/machine.py
Bibo-Joshi: https://peps.python.org/pep-0008/#designing-for-inheritance

This makes name mangling being used.
TBF I don't strictly rely on pep8 in these questions, i.e. I don't choose Single underscore vs Dunder based on exclusively based "should a subclass be allowed to use this?". We have the stability which basically says that also single underscore should not be used by users unless it's documented.
My rule of thumb is using dunder when I want to highlight that something is really low-level.
The fact that dunder names get name-mangled is a nice bonus.

telegram/ext/_fsm/machine.py
Bibo-Joshi: mappingproxytype says that the entries of the dict can't be changed, but with `self.states = MappingProxyType(self.__states)` you could still do `self.states = None`

telegram/ext/_fsm/machine.py
Bibo-Joshi: I feel you :D if `machine` was a public module like `telegram.constants`, it would have to be protected IMO. Since it's not and we expose the relevant elements only via `telegram.ext`, it's not so important. Making it clearer what the content exposed by this module is still doesn't hurt.
If you want to go deeper into the rabbit hole: https://discuss.python.org/t/add-the-export-keyword-to-python/28444/6

telegram/ext/_application.py
```diff
@@ -1475,10 +1512,11 @@ def remove_handler(
...
         """
-        if handler in self.handlers[group]:
-            self.handlers[group].remove(handler)
-            if not self.handlers[group]:
-                del self.handlers[group]
+        for state_handlers in self.handlers.values():
+            if handler in state_handlers[group]:
+                state_handlers[group].remove(handler)
+                if not state_handlers[group]:
+                    del state_handlers[group]
```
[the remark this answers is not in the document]
Bibo-Joshi: yes, that's basically the same as `state_handlers.pop(group, None)`, I think. It will call `state_handlers.__del__(group)`. IMHO this is basic python syntax 😬

examples/fsmbot.py
```diff
@@ -0,0 +1,203 @@
...
+    WRITING = State("WRITING")
+
+    def __init__(self, admin_id: int):
+        self.admin_id = admin_id
+        super().__init__()
+
+    def _get_admin_state(self) -> tuple[State, int]:
+        return self._states[self.admin_id]
+
+    def get_state_info(self, update: object) -> StateInfo[Optional[int]]:
```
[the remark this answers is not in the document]
Bibo-Joshi: Correct, the type variable specifying the type of keys you use.

examples/fsmbot.py
```diff
@@ -0,0 +1,203 @@
...
+    HOLD = State("HOLD")
+    WELCOMING = State("WELCOMING")
+    WAITING_FOR_REPLY = State("WAITING_FOR_REPLY")
+    WRITING = State("WRITING")
+
+    def __init__(self, admin_id: int):
+        self.admin_id = admin_id
+        super().__init__()
+
+    def _get_admin_state(self) -> tuple[State, int]:
```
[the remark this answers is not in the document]
Bibo-Joshi: So far I see no reason why someone should be allowed to all `USM(…).get_admin_state` … then again this is only an example, so I don't care :D

OnceUponATimeInAmerica:
I'm not sure if this is the right place to mention this but:

Is there any plans or ongoing work for a declarative but an extensible (e.g. with hooks) FSM syntax? like `State1 ---(on some input or trigger) ---> State2 (produce or calculate some output or perform a set of actions)` ?

 I have a fairly complex stateful bot in PTB (and ConversationHandler) and have written a very rudimentary function that parses a generically defined (but limited) transition table (declarative) and performs the transitions given the inputs and also produces some (a set of) outputs or actions.

I did it for a subset of states (relatively simple logic) in my application, because in the current PTB framework adding states and extending the FSM logic is somewhat a pain.  Code re-usability was another consideration (e.g. several states might want to transition to a very specific state and so using a common transition function or hook for setting up the UI and then entering that state is useful for code reuse).

A declarative approach to complex FSM logic specification would both accelerate and simplify development, IMHO.

Issue #4275, opened 2024-05-28
[FEATURE] API 7.4


harshil21:
### What kind of feature are you missing? Where do you notice a shortcoming of PTB?

⭐ 

### Important

- Comment here or in the dev group what you want to work on so we can assign you
- Please copy the relevant part of the checklist from
- https://github.com/python-telegram-bot/python-telegram-bot/issues/4275 to your PR and make sure to cover everything :)
- Please base your branch on the [api-7.4](https://github.com/python-telegram-bot/python-telegram-bot/tree/api-7.4) branch
- If this is your first contribution, please be sure to read the [contribution guide](https://docs.python-telegram-bot.org/en/latest/contributing.html) and feel free to reach out with any questions via https://t.me/pythontelegrambotdev

### Describe the solution you'd like

- [x] Added support for payments in [Telegram Stars](https://t.me/BotNews/90) by introducing the new currency “XTR”. (Taken by @harshil21)
- [x] The parameter provider_token of the methods [sendInvoice](https://core.telegram.org/bots/api#sendinvoice) and [createInvoiceLink](https://core.telegram.org/bots/api#createinvoicelink) must be omitted for payments in [Telegram Stars](https://t.me/BotNews/90). (Taken by @harshil21)
- [x] The field provider_token in the class [InputInvoiceMessageContent](https://core.telegram.org/bots/api#inputinvoicemessagecontent) must be omitted for payments in [Telegram Stars](https://t.me/BotNews/90). (Taken by @harshil21)
- [x] Added the method [refundStarPayment](https://core.telegram.org/bots/api#refundstarpayment). (Taken by @harshil21)
- [x] Added the field effect_id to the class [Message](https://core.telegram.org/bots/api#message). (Taken by @harshil21)
[...]

PR #4689, opened 2025-02-22
Make `provider_token` Argument Optional

Bibo-Joshi:
The forgotton follow-up for #4275 

Carries out deprecations from API 7.4. This is in line with our stability policy.


Poolitzer (approved):
LGTM

harshil21 (approved):
damn that's wild that it went unnoticed for so long

PR #4690, opened 2025-02-25
Thumbnail type

aelkheir:
see https://github.com/tdlib/telegram-bot-api/issues/707

### DIFF
--- a/tests/test_official/exceptions.py
+++ b/tests/test_official/exceptions.py
@@ -96,8 +96,6 @@ class ParamTypeCheckingExceptions:
         },
         "Input(Paid)?Media.*": {
             "media": str,  # actual: Union[str, InputMedia*, FileInput]
-        },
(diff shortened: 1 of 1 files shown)


tests/test_official/exceptions.py
```diff
@@ -96,6 +96,7 @@ class ParamTypeCheckingExceptions:
         },
         "Input(Paid)?Media.*": {
             "media": str,  # actual: Union[str, InputMedia*, FileInput]
+            "thumbnail": str,  # actual: Union[str, FileInput]
```
Bibo-Joshi, line 99:
```suggestion
            # see also https://github.com/tdlib/telegram-bot-api/issues/707
            "thumbnail": str,  # actual: Union[str, FileInput]
```

Bibo-Joshi (approved):
Good catch, thanks for the finding!

PR #4708, opened 2025-03-07
Add `name` and `full_name` properties to the `SharedUser` class

david-shiko:
P.S. I failed to update the documentation. Sphinx requires Python 3.12, but I intentionally use Python 3.10. The documentation states that Python 3.9+ is supported, as mentioned in the introduction section: [python-telegram-bot Introduction](https://github.com/python-telegram-bot/python-telegram-bot#introduction).

Please review the PR, and if everything is okay, I will try to write the documentation.

### DIFF
--- a/telegram/_shared.py
+++ b/telegram/_shared.py
@@ -18,7 +18,7 @@
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 """This module contains two objects used for request chats/users service messages."""
 from collections.abc import Sequence
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, Union
 
 from telegram._files.photosize import PhotoSize
 from telegram._telegramobject import TelegramObject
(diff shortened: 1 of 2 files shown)


Bibo-Joshi:
Hi. Thanks for the PR, nice idea 👍 
Looking at the code base, I see that we already have the properties

* `User.name`
* `User.full_name`
* `Chat.full_name`

and your PR basically copies `User.name` and `Chat.full_name` to `SharedUser`.
To avoid too much duplication, I would like you instead introduce a new module `telegram._utils.usernames` whith two function `get_name` and `get_full_name`. They would both accept a object of type

```python
class UserLike(typing.Protocol):
    first_name: typing.Optional[str]
    last_name: typing.Optional[str]
    username: typing.Optional[str]
```

Would that be doable for you?

Note that you can copy the tests from `test_chat`/`test_user`. I'd like to avoid introducing `unittest` into our test suite b/c we use nowhere else.

Bibo-Joshi:
Closing in favor of #4713

PR #4718, opened 2025-03-13
Make the CallbackContext conform to the AbstractAsyncContextManager

xD8A:
Is it possible to make the CallbackContext to conform to the AbstractAsyncContextManager?
So that some resources can be acquired and then released?

```python
class MyContext(CallbackContext[ExtBot, dict, dict, dict]):

    def __init__(self):
        self.session_factory = async_scoped_session(
            sessionmaker(engine, class_=AsyncSession, expire_on_commit=False), 
            scopefunc=current_task,
        )
        self.session: AsyncSession = None

    async def __aenter__(self) -> 'MyContext':
        self.session = await self.session_factory().__aenter__()
        return self

    async def __aexit__(self, exc_type, exc_value, exc_tb) -> None:
        if exc_value:
            await self.session.rollback()
            raise ...
        
        await self.session.__aexit__()
```

### DIFF
--- a/telegram/ext/_application.py
+++ b/telegram/ext/_application.py
@@ -1301,6 +1301,7 @@ class Application(
         context = None
         any_blocking = False  # Flag which is set to True if any handler specifies block=True
         tasks = []
+        ctx_initialized = False
 
         for handlers in self.handlers.values():
             try:
@@ -1322,7 +1323,6 @@ class Application(
                                 exc_info=exc,
                             )
                             return
-                        await context.__aenter__()  # pylint: disable=unnecessary-dunder-call
                         await context.refresh_data()
                     coroutine: Coroutine = handler.handle_update(update, self, check, context)
 
@@ -1344,6 +1344,9 @@ class Application(
                         )
                     else:
                         any_blocking = True
+                        if not ctx_initialized:
+                            await context.__aenter__()  # pylint: disable=unnecessary-dunder-call
+                            ctx_initialized = True
                         await coroutine
                     break  # Only a max of 1 handler per group is handled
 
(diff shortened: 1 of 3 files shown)


Bibo-Joshi:
Hi @xD8A , thanks for your PR!
Getting the `context` build & populated for an update isn't really meant be do heavy/io-intensive operations. The overhead of `Application.process_update` should be small as otherwise update processing would be slowed down. The asyncio logic should happen within the handler callbacks.
That's why don't see this change aligning with the design of `Application`.
Would you mind explaining your use cas? Maybe we can find another solution :)

xD8A:
I'm looking for a solution to use with DI so that dependencies can be injected into handlers.

For example, like here:
https://github.com/reagento/dishka/blob/develop/examples/integrations/aiogram_bot.py

Bibo-Joshi:
I don't see why dependency injection needs the `context` object to be an async context manager. The example you linked to is an aiogram example. Without digging through that dishka library, I don't see how that would apply directly to PTB and what the `AsyncContextManager` has to do with it.
Please provide an example use case in a PTB setting.

xD8A:
It seemed to me that with this approach, some code from handlers that require the same resources (e.g. a DB session) can be moved to a customized CallbackContext. And in particular, it can be used when injecting a DI container. It seems I was wrong.

So for my use case, maybe I should to replace the UpdateProcessor for an Application to inject dependencies?
https://github.com/xD8A/ptb-di/blob/402f5d3e3f3fd5118277ea664eeff79675e4d6c0/app/di.py#L47

Bibo-Joshi:
Making shared resources available via a custom `CallbackContext` object is surely a sane thing to do. *How* you do it is a different question.
Using a custom `UpdateProcessor` for that can sureley be one solution. Please consider

* using `ApplicationBuilder.concurrent_updates` to set your custom processor instead of overriding protected attributes
* making your implementation concurrency-safe. You are currently setting the `update_container` as instance attribute. I expect this to lead to race conditions if you set `max_concurrent_updates` to >1, i.e. multiple updates are being processed concurrently.

xD8A:
> * making your implementation concurrency-safe. You are currently setting the `update_container` as instance attribute. I expect this to lead to race conditions if you set `max_concurrent_updates` to >1, i.e. multiple updates are being processed concurrently.

I thought I was avoiding this by using a ContextVar, am I wrong and are there any pitfalls?

Bibo-Joshi:
The problem that I see is not with contextvar itself but with how you store a reference to it. But that's maybe just because I'm not familiar enough with contextvar.
In any case, it seems to me that the original purpose of this PR is resolved and I'll close it :)
For follow-up questions on this topic kindly reach out through the discussions page or the TG help chat.

PR #4733, opened 2025-03-26
Fix: move initialize flag up in _bot.py

Poolitzer:
I am unsure if I should test this, because I can't really rewrite the initialize, so I think the only appropriate way would be to pick an invalid token, but unsure if this is worth it (its a really short function after all)


Bibo-Joshi:
@Poolitzer You can mock <code>get_me</code> to raise an exception and mock <code>HTTPXRequest.shutdown</code> to ensure that the shutdown is still called - or at least check that the logging line &quot;This Bot is already shut down. Returning&quot; is *not* issued.

Bibo-Joshi:
@Poolitzer what's your status on this? Pleaselet me know if you'd like support.

Poolitzer:
Will do the next days, sorry life happened

Bibo-Joshi:
thanks for the fix!

PR #4750, opened 2025-04-09
Use `timedelta` to represent time periods in classes

aelkheir:
Second part of #4575.

<details>
<summary>Affected classes (basicaly anywhere "in seconds" appears in api docs)</summary>

| Class | attribute | docs |
| ---- | ----------- | ------- |
| - [x] ChatFullInfo | slow_mode_delay | Optional. For supergroups, the minimum allowed delay between consecutive messages sent by each unprivileged user; in seconds |
| -[x] ChatFullInfo | message_auto_delete_time | Optional. The time after which all messages sent to the chat will be automatically deleted; in seconds |
| Animation | duration | Duration of the video in seconds as defined by the sender |
| Audio | duration | Duration of the audio in seconds as defined by the sender |
| Video | duration | Duration of the video in seconds as defined by the sender |
| Video | start_timestamp | Optional. Timestamp in seconds from which the video will play in the message |
| VideoNote | duration | Duration of the video in seconds as defined by the sender |
| Voice | duration | Duration of the audio in seconds as defined by the sender |
| PaidMediaPreview | duration | Optional. Duration of the media in seconds as defined by the sender |
| Poll | open_period | Optional. Amount of time in seconds the poll will be active after creation |
| Location | live_period | Optional. Time relative to the message sending date, during which the location can be updated; in seconds. For active live locations only. |
| MessageAutoDeleteTimerChanged | message_auto_delete_time | New auto-delete time for messages in the chat; in seconds |
| VideoChatEnded | duration | Video chat duration in seconds |
| ChatInviteLink | subscription_period | Optional. The number of seconds the subscription will be active for before the next payment |
| ResponseParameters | retry_after | Optional. In case of exceeding flood control, the number of seconds left to wait before the request can be repeated |
| InputMediaVideo | duration | Optional. Video duration in seconds |
| InputMediaAnimation | duration | Optional. Animation duration in seconds |
| InputMediaAudio | duration | Optional. Duration of the audio in seconds |
| InputPaidMediaVideo | duration | Optional. Video duration in seconds |
| InlineQueryResultGif | gif_duration | Optional. Duration of the GIF in seconds |
| InlineQueryResultMpeg4Gif | mpeg4_duration | Optional. Video duration in seconds |
| InlineQueryResultVideo | video_duration | Optional. Video duration in seconds |
| InlineQueryResultAudio | audio_duration | Optional. Audio duration in seconds |
| InlineQueryResultVoice | voice_duration | Optional. Recording duration in seconds |
| InlineQueryResultLocation | live_period | Optional. Period in seconds during which the location can be updated, should be between 60 and 86400, or 0x7FFFFFFF for live locations that can be edited indefinitely. |
| InputLocationMessageContent | live_period | ptional. Period in seconds during which the location can be updated, should be between 60 and 86400, or 0x7FFFFFFF for live locations that can be edited indefinitely. |
</details>

Check-list for PRs
------------------

- [ ] Checked the Bot API specific sections of the `Stability Policy <https://docs.python-telegram-bot.org/stability_policy.html>`_
- [ ] Created a PR to remove functionality deprecated in the previous Bot API release (`see here <https://docs.python-telegram-bot.org/en/stable/stability_policy.html#case-2>`_)

-  If relevant:

### DIFF
--- a/docs/substitutions/global.rst
+++ b/docs/substitutions/global.rst
@@ -101,3 +101,5 @@
 .. |org-verify| replace:: `on behalf of the organization <https://telegram.org/verify#third-party-verification>`__
 
 .. |time-period-input| replace:: :class:`datetime.timedelta` objects are accepted in addition to plain :obj:`int` values.
+
+.. |timespan-seconds-deprecated| replace:: In a future major version this will be of type :obj:`datetime.timedelta`. You can opt-in early by setting the `PTB_TIMEDELTA` environment variable.
--- a/telegram/_chatfullinfo.py
+++ b/telegram/_chatfullinfo.py
@@ -20,7 +20,7 @@
 """This module contains an object that represents a Telegram ChatFullInfo."""
 import datetime as dtm
 from collections.abc import Sequence
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, Union
 
 from telegram._birthdate import Birthdate
 from telegram._chat import Chat, _ChatBase
@@ -28,9 +28,18 @@ from telegram._chatlocation import ChatLocation
 from telegram._chatpermissions import ChatPermissions
 from telegram._files.chatphoto import ChatPhoto
 from telegram._reaction import ReactionType
-from telegram._utils.argumentparsing import de_json_optional, de_list_optional, parse_sequence_arg
-from telegram._utils.datetime import extract_tzinfo_from_defaults, from_timestamp
-from telegram._utils.types import JSONDict
+from telegram._utils.argumentparsing import (
+    de_json_optional,
+    de_list_optional,
+    parse_period_arg,
+    parse_sequence_arg,
+)
+from telegram._utils.datetime import (
+    extract_tzinfo_from_defaults,
+    from_timestamp,
+    get_timedelta_value,
+)
+from telegram._utils.types import JSONDict, TimePeriod
 
 if TYPE_CHECKING:
     from telegram import Bot, BusinessIntro, BusinessLocation, BusinessOpeningHours, Message
@@ -155,17 +164,29 @@ class ChatFullInfo(_ChatBase):
             (by sending date).
         permissions (:class:`telegram.ChatPermissions`): Optional. Default chat member permissions,
             for groups and supergroups.
-        slow_mode_delay (:obj:`int`, optional): For supergroups, the minimum allowed delay between
-            consecutive messages sent by each unprivileged user.
+        slow_mode_delay (:obj:`int` | :class:`datetime.timedelta`, optional): For supergroups,
+            the minimum allowed delay between consecutive messages sent by each unprivileged user.
+
+            .. versionchanged:: NEXT.VERSION
+                |time-period-input|
+
+            .. deprecated:: NEXT.VERSION
+                |timespan-seconds-deprecated|
         unrestrict_boost_count (:obj:`int`, optional): For supergroups, the minimum number of
             boosts that a non-administrator user needs to add in order to ignore slow mode and chat
             permissions.
 
             .. versionadded:: 21.0
-        message_auto_delete_time (:obj:`int`, optional): The time after which all messages sent to
-            the chat will be automatically deleted; in seconds.
+        message_auto_delete_time (:obj:`int` | :class:`datetime.timedelta`, optional): The time
+            after which all messages sent to the chat will be automatically deleted; in seconds.
 
             .. versionadded:: 13.4
+
+            .. versionchanged:: NEXT.VERSION
+                |time-period-input|
+
+            .. deprecated:: NEXT.VERSION
+                |timespan-seconds-deprecated|
         has_aggressive_anti_spam_enabled (:obj:`bool`, optional): :obj:`True`, if aggressive
             anti-spam checks are enabled in the supergroup. The field is only available to chat
             administrators.
@@ -312,17 +333,29 @@ class ChatFullInfo(_ChatBase):
             (by sending date).
         permissions (:class:`telegram.ChatPermissions`): Optional. Default chat member permissions,
             for groups and supergroups.
-        slow_mode_delay (:obj:`int`): Optional. For supergroups, the minimum allowed delay between
-            consecutive messages sent by each unprivileged user.
+        slow_mode_delay (:obj:`int` | :class:`datetime.timedelta`): Optional. For supergroups,
+            the minimum allowed delay between consecutive messages sent by each unprivileged user.
+
+            .. versionchanged:: NEXT.VERSION
+                |time-period-input|
+
+            .. deprecated:: NEXT.VERSION
+                |timespan-seconds-deprecated|
         unrestrict_boost_count (:obj:`int`): Optional. For supergroups, the minimum number of
             boosts that a non-administrator user needs to add in order to ignore slow mode and chat
             permissions.
 
             .. versionadded:: 21.0
-        message_auto_delete_time (:obj:`int`): Optional. The time after which all messages sent to
-            the chat will be automatically deleted; in seconds.
+        message_auto_delete_time (:obj:`int` | :class:`datetime.timedelta`): Optional. The time
+            after which all messages sent to the chat will be automatically deleted; in seconds.
 
             .. versionadded:: 13.4
+
+            .. versionchanged:: NEXT.VERSION
+                |time-period-input|
+
+            .. deprecated:: NEXT.VERSION
+                |timespan-seconds-deprecated|
         has_aggressive_anti_spam_enabled (:obj:`bool`): Optional. :obj:`True`, if aggressive
             anti-spam checks are enabled in the supergroup. The field is only available to chat
             administrators.
@@ -366,6 +399,8 @@ class ChatFullInfo(_ChatBase):
     """
 
     __slots__ = (
+        "_message_auto_delete_time",
+        "_slow_mode_delay",
         "accent_color_id",
         "active_usernames",
         "available_reactions",
@@ -394,14 +429,12 @@ class ChatFullInfo(_ChatBase):
         "linked_chat_id",
         "location",
         "max_reaction_count",
-        "message_auto_delete_time",
         "permissions",
         "personal_chat",
         "photo",
         "pinned_message",
         "profile_accent_color_id",
         "profile_background_custom_emoji_id",
-        "slow_mode_delay",
         "sticker_set_name",
         "unrestrict_boost_count",
     )
@@ -439,9 +472,9 @@ class ChatFullInfo(_ChatBase):
         invite_link: Optional[str] = None,
         pinned_message: Optional["Message"] = None,
         permissions: Optional[ChatPermissions] = None,
-        slow_mode_delay: Optional[int] = None,
+        slow_mode_delay: Optional[TimePeriod] = None,
         unrestrict_boost_count: Optional[int] = None,
-        message_auto_delete_time: Optional[int] = None,
+        message_auto_delete_time: Optional[TimePeriod] = None,
         has_aggressive_anti_spam_enabled: Optional[bool] = None,
         has_hidden_members: Optional[bool] = None,
         has_protected_content: Optional[bool] = None,
@@ -477,9 +510,9 @@ class ChatFullInfo(_ChatBase):
             self.invite_link: Optional[str] = invite_link
             self.pinned_message: Optional[Message] = pinned_message
             self.permissions: Optional[ChatPermissions] = permissions
-            self.slow_mode_delay: Optional[int] = slow_mode_delay
-            self.message_auto_delete_time: Optional[int] = (
-                int(message_auto_delete_time) if message_auto_delete_time is not None else None
+            self._slow_mode_delay: Optional[dtm.timedelta] = parse_period_arg(slow_mode_delay)
+            self._message_auto_delete_time: Optional[dtm.timedelta] = parse_period_arg(
+                message_auto_delete_time
             )
             self.has_protected_content: Optional[bool] = has_protected_content
             self.has_visible_history: Optional[bool] = has_visible_history
@@ -520,6 +553,14 @@ class ChatFullInfo(_ChatBase):
             self.can_send_paid_media: Optional[bool] = can_send_paid_media
             self.can_send_gift: Optional[bool] = can_send_gift
 
+    @property
+    def slow_mode_delay(self) -> Optional[Union[int, dtm.timedelta]]:
+        return get_timedelta_value(self._slow_mode_delay)
+
+    @property
+    def message_auto_delete_time(self) -> Optional[Union[int, dtm.timedelta]]:
+        return get_timedelta_value(self._message_auto_delete_time)
+
     @classmethod
     def de_json(cls, data: JSONDict, bot: Optional["Bot"] = None) -> "ChatFullInfo":
         """See :meth:`telegram.TelegramObject.de_json`."""
@@ -541,6 +582,13 @@ class ChatFullInfo(_ChatBase):
             Message,
         )
 
+        data["slow_mode_delay"] = (
+            dtm.timedelta(seconds=s) if (s := data.get("slow_mode_delay")) else None
+        )
+        data["message_auto_delete_time"] = (
+            dtm.timedelta(seconds=s) if (s := data.get("message_auto_delete_time")) else None
+        )
+
         data["pinned_message"] = de_json_optional(data.get("pinned_message"), Message, bot)
         data["permissions"] = de_json_optional(data.get("permissions"), ChatPermissions, bot)
         data["location"] = de_json_optional(data.get("location"), ChatLocation, bot)
@@ -558,3 +606,10 @@ class ChatFullInfo(_ChatBase):
         )
 
         return super().de_json(data=data, bot=bot)
+
+    def to_dict(self, recursive: bool = True) -> JSONDict:
+        """See :meth:`telegram.TelegramObject.to_dict`."""
+        out = super().to_dict(recursive)
+        out["slow_mode_delay"] = self.slow_mode_delay
+        out["message_auto_delete_time"] = self.message_auto_delete_time
+        return out
--- a/telegram/_utils/datetime.py
+++ b/telegram/_utils/datetime.py
@@ -29,8 +29,13 @@ Warning:
 """
 import contextlib
 import datetime as dtm
+import os
 import time
 from typing import TYPE_CHECKING, Optional, Union
+
+from telegram._utils.warnings import warn
+from telegram.warnings import PTBDeprecationWarning
+from tests.auxil.envvars import env_var_2_bool
 
 if TYPE_CHECKING:
     from telegram import Bot
@@ -224,3 +229,25 @@ def _datetime_to_float_timestamp(dt_obj: dtm.datetime) -> float:
     if dt_obj.tzinfo is None:
         dt_obj = dt_obj.replace(tzinfo=dtm.timezone.utc)
     return dt_obj.timestamp()
+
+
+def get_timedelta_value(value: Optional[dtm.timedelta]) -> Optional[Union[int, dtm.timedelta]]:
+    """
+    Convert a `datetime.timedelta` to seconds or return it as-is, based on environment config.
+    """
+    if value is None:
+        return None
+
+    if env_var_2_bool(os.getenv("PTB_TIMEDELTA")):
+        return value
+
+    warn(
+        PTBDeprecationWarning(
+            "NEXT.VERSION",
+            "In a future major version this will be of type `datetime.timedelta`."
+            " You can opt-in early by setting the `PTB_TIMEDELTA` environment variable.",
+        ),
+        stacklevel=2,
+    )
+    # We don't want to silently drop fractions, so float is returned and we slience mypy
+    return value.total_seconds()  # type: ignore[return-value]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,6 +18,7 @@
 # along with this program.  If not, see [http://www.gnu.org/licenses/].
 import asyncio
 import logging
+import os
 import sys
 import zoneinfo
 from pathlib import Path
@@ -40,7 +41,12 @@ from telegram.ext import Defaults
 from tests.auxil.build_messages import DATE, make_message
 from tests.auxil.ci_bots import BOT_INFO_PROVIDER, JOB_INDEX
 from tests.auxil.constants import PRIVATE_KEY, TEST_TOPIC_ICON_COLOR, TEST_TOPIC_NAME
-from tests.auxil.envvars import GITHUB_ACTIONS, RUN_TEST_OFFICIAL, TEST_WITH_OPT_DEPS
+from tests.auxil.envvars import (
+    GITHUB_ACTIONS,
+    RUN_TEST_OFFICIAL,
+    TEST_WITH_OPT_DEPS,
+    env_var_2_bool,
+)
 from tests.auxil.files import data_file
 from tests.auxil.networking import NonchalantHttpxRequest
 from tests.auxil.pytest_classes import PytestBot, make_bot
@@ -127,6 +133,18 @@ if GITHUB_ACTIONS and JOB_INDEX == 0:
                 f"Test function {request.function} in test class {name} should not have a `bot` "
                 f"fixture. Use `offline_bot` instead."
             )
+
+
+@pytest.fixture(scope="module", params=["true", "false", None])
+def PTB_TIMEDELTA(request):
+    # Here we manually use monkeypatch to give this fixture module scope
+    monkeypatch = pytest.MonkeyPatch()
+    if request.param is not None:
+        monkeypatch.setenv("PTB_TIMEDELTA", request.param)
+    else:
+        monkeypatch.delenv("PTB_TIMEDELTA", raising=False)
+    yield env_var_2_bool(os.getenv("PTB_TIMEDELTA"))
+    monkeypatch.undo()
 
 
 # Redefine the event_loop fixture to have a session scope. Otherwise `bot` fixture can't be
--- a/tests/test_official/exceptions.py
+++ b/tests/test_official/exceptions.py
@@ -98,6 +98,10 @@ class ParamTypeCheckingExceptions:
             "thumbnail": str,  # actual: Union[str, FileInput]
             "cover": str,  # actual: Union[str, FileInput]
         },
+        "ChatFullInfo": {
+            "slow_mode_delay": int,  # actual: Union[int, dtm.timedelta]
+            "message_auto_delete_time": int,  # actual: Union[int, dtm.timedelta]
+        },
         "EncryptedPassportElement": {
             "data": str,  # actual: Union[IdDocumentData, PersonalDetails, ResidentialAddress]
         },
--- a/telegram/_utils/argumentparsing.py
+++ b/telegram/_utils/argumentparsing.py
@@ -23,12 +23,15 @@ Warning:
     user. Changes to this module are not considered breaking changes and may not be documented in
     the changelog.
 """
+import datetime as dtm
 from collections.abc import Sequence
-from typing import TYPE_CHECKING, Optional, Protocol, TypeVar
+from typing import TYPE_CHECKING, Optional, Protocol, TypeVar, Union
 
 from telegram._linkpreviewoptions import LinkPreviewOptions
 from telegram._telegramobject import TelegramObject
-from telegram._utils.types import JSONDict, ODVInput
+from telegram._utils.types import JSONDict, ODVInput, TimePeriod
+from telegram._utils.warnings import warn
+from telegram.warnings import PTBDeprecationWarning
 
 if TYPE_CHECKING:
     from typing import type_check_only
@@ -48,6 +51,30 @@ def parse_sequence_arg(arg: Optional[Sequence[T]]) -> tuple[T, ...]:
         :obj:`Tuple`: The sequence converted to a tuple or an empty tuple.
     """
     return tuple(arg) if arg else ()
+
+
+def parse_period_arg(arg: Optional[TimePeriod]) -> Union[dtm.timedelta, None]:
+    """Parses an optional time period in seconds into a timedelta
+
+    Args:
+        arg (:obj:`int` | :class:`datetime.timedelta`, optional): The time period to parse.
+
+    Returns:
+        :obj:`timedelta`: The time period converted to a timedelta object or :obj:`None`.
+    """
+    if arg is None:
+        return None
+    if isinstance(arg, (int, float)):
+        warn(
+            PTBDeprecationWarning(
+                "NEXT.VERSION",
+                "In a future major version this will be of type `datetime.timedelta`."
+                " You can opt-in early by setting the `PTB_TIMEDELTA` environment variable.",
+            ),
+            stacklevel=2,
+        )
+        return dtm.timedelta(seconds=arg)
+    return arg
 
 
 def parse_lpo_and_dwpp(
--- a/tests/test_chatfullinfo.py
+++ b/tests/test_chatfullinfo.py
@@ -36,6 +36,7 @@ from telegram import (
 )
 from telegram._utils.datetime import UTC, to_timestamp
 from telegram.constants import ReactionEmoji
+from telegram.warnings import PTBDeprecationWarning
 from tests.auxil.slots import mro_slots
 
 
@@ -52,6 +53,7 @@ def chat_full_info(bot):
         can_set_sticker_set=ChatFullInfoTestBase.can_set_sticker_set,
         permissions=ChatFullInfoTestBase.permissions,
         slow_mode_delay=ChatFullInfoTestBase.slow_mode_delay,
+        message_auto_delete_time=ChatFullInfoTestBase.message_auto_delete_time,
         bio=ChatFullInfoTestBase.bio,
         linked_chat_id=ChatFullInfoTestBase.linked_chat_id,
         location=ChatFullInfoTestBase.location,
@@ -104,6 +106,7 @@ class ChatFullInfoTestBase:
         can_invite_users=True,
     )
     slow_mode_delay = 30
+    message_auto_delete_time = 60
     bio = "I'm a Barbie Girl in a Barbie World"
     linked_chat_id = 11880
     location = ChatLocation(Location(123, 456), "Barbie World")
@@ -150,7 +153,7 @@ class TestChatFullInfoWithoutRequest(ChatFullInfoTestBase):
 
         assert len(mro_slots(cfi)) == len(set(mro_slots(cfi))), "duplicate slot"
 
-    def test_de_json(self, offline_bot):
+    def test_de_json(self, offline_bot, PTB_TIMEDELTA):
         json_dict = {
             "id": self.id_,
             "title": self.title,
@@ -162,6 +165,7 @@ class TestChatFullInfoWithoutRequest(ChatFullInfoTestBase):
             "can_set_sticker_set": self.can_set_sticker_set,
             "permissions": self.permissions.to_dict(),
             "slow_mode_delay": self.slow_mode_delay,
+            "message_auto_delete_time": self.message_auto_delete_time,
             "bio": self.bio,
             "business_intro": self.business_intro.to_dict(),
             "business_location": self.business_location.to_dict(),
@@ -194,6 +198,11 @@ class TestChatFullInfoWithoutRequest(ChatFullInfoTestBase):
             "last_name": self.last_name,
             "can_send_paid_media": self.can_send_paid_media,
         }
+
+        def get_period(attr):
+            value = getattr(self, attr)
+            return dtm.timedelta(seconds=value) if PTB_TIMEDELTA else value
+
         cfi = ChatFullInfo.de_json(json_dict, offline_bot)
         assert cfi.id == self.id_
         assert cfi.title == self.title
@@ -202,7 +211,8 @@ class TestChatFullInfoWithoutRequest(ChatFullInfoTestBase):
         assert cfi.sticker_set_name == self.sticker_set_name
         assert cfi.can_set_sticker_set == self.can_set_sticker_set
         assert cfi.permissions == self.permissions
-        assert cfi.slow_mode_delay == self.slow_mode_delay
+        assert cfi.slow_mode_delay == get_period("slow_mode_delay")
+        assert cfi.message_auto_delete_time == get_period("message_auto_delete_time")
         assert cfi.bio == self.bio
         assert cfi.business_intro == self.business_intro
         assert cfi.business_location == self.business_location
@@ -261,7 +271,7 @@ class TestChatFullInfoWithoutRequest(ChatFullInfoTestBase):
         assert cfi_bot_raw.emoji_status_expiration_date.tzinfo == UTC
         assert emoji_expire_offset_tz == emoji_expire_offset
 
-    def test_to_dict(self, chat_full_info):
+    def test_to_dict(self, chat_full_info, PTB_TIMEDELTA):
         cfi = chat_full_info
         cfi_dict = cfi.to_dict()
 
@@ -314,6 +324,39 @@ class TestChatFullInfoWithoutRequest(ChatFullInfoTestBase):
         assert cfi_dict["can_send_paid_media"] == cfi.can_send_paid_media
 
         assert cfi_dict["max_reaction_count"] == cfi.max_reaction_count
+
+    @pytest.mark.parametrize("field_name", ["slow_mode_delay", "message_auto_delete_time"])
+    def test_time_period_int_deprecated(self, PTB_TIMEDELTA, recwarn, field_name):
+        def get_period(field_name):
+            value = getattr(self, field_name)
+            return dtm.timedelta(seconds=value) if PTB_TIMEDELTA else value
+
+        cfi = ChatFullInfo(
+            id=123456,
+            type="dummy_type",
+            accent_color_id=1,
+            max_reaction_count=1,
+            slow_mode_delay=get_period("slow_mode_delay"),
+            message_auto_delete_time=get_period("message_auto_delete_time"),
+        )
+
+        if PTB_TIMEDELTA:
+            assert len(recwarn) == 0
+            assert isinstance(getattr(cfi, field_name), dtm.timedelta)
+            assert len(recwarn) == 0
+        else:
+            # Two warnings from constructor
+            assert len(recwarn) == 2
+            for i in range(2):
+                assert "will be of type `datetime.timedelta`" in str(recwarn[i].message)
+                assert recwarn[i].category is PTBDeprecationWarning
+
+            # Trigger another warning on property access, while at it make an assertion
+            assert isinstance(getattr(cfi, field_name), (int, float))
+
+            assert len(recwarn) == 3
+            assert "will be of type `datetime.timedelta`" in str(recwarn[1].message)
+            assert recwarn[2].category is PTBDeprecationWarning
 
     def test_always_tuples_attributes(self):
         cfi = ChatFullInfo(


aelkheir:
One thoughts before i proceed with the rest of the classes.
> Both attributes and arguments should only accept timedeltas in the long run. https://github.com/python-telegram-bot/python-telegram-bot/issues/4575#issue-2688392527

why deprecate class args, i think it would be convenient to allow passing either `30` or `timedelta(seconds=30)`. it would also be consistent with how bot methods work as of #4651. deprecating class attrs is sensible however.

Bibo-Joshi:
Hey, thanks for getting started so quickly:)
About your question: TelegramObject types are not really intended to be manually initialized by the user but rather by the de_json methods - that's in contrast to the bot methods. I view them as mere data storages. Note that the conversions from json to datetime or off of the TG classes also all happen in de_json and not in init. This approach can certainly be questioned in general (i.e. what would be a sensible interface for classes wrapping API types). In the current set-up, I'd see consistency as the more important argument 😅

aelkheir:
Aha okay. but would it still be necessary to issue a separate deprecation warning in the init methods as was outlined in  https://github.com/python-telegram-bot/python-telegram-bot/issues/4575#issuecomment-2566386429 (very helpful summary btw 😄 ). one is already being issued on property access.

Bibo-Joshi:
I would say, so yes. Attribute access and passing the parameter are two independent operations, both of which will be affected :)

aelkheir:
I see! I was thinking more from the perspective that 
> TelegramObject types are not really intended to be manually initialized by the user rather by the de_json.

whom are we intending to warn then

anyways I did include both warnings for completeness and the cases where users are manually initializing objects

aelkheir:
Almost ready for a review. just a few things to discuss.

- Not sure if I should follow the deprecation process for field [ResponseParameter.retry_after](upstream/feature/4575-timedelta-classes) which maps to PTB's class `RetryAfter(TelegramError)`. a suggestion is to let `self.retry_after` continue being and `int`. and only alter the message of the error with a timedelta value derived from `selr.retry_after`. So it prints:
`Flood control exceeded. Retry in 0:08:20`.
instead of what it currently prints:
`Flood control exceeded. Retry in 500 seconds`
Reason being that it's underling `int` type is not user-facing. and also not to break older pickles of the class.
- Automatic detection of time periods in test_official works great (only needed a few adjustments). Anyway it also detected `Bot.get_updates.timeout` ([getUpdates](https://core.telegram.org/bots/api#getupdates)) which i think was missed in [#4651](https://github.com/python-telegram-bot/python-telegram-bot/pull/4651) (?). I have currently adjusted `Bot.get_updates.timeout` but without altering the signature of other methods that call it (i.e `Updater.[_]start_polling`) which also have a `timeout` parameter.

Checks are really mad, will debug tomorrow :')

Bibo-Joshi:
Hey, thanks for the updates - looks like you've spend quite some energy in this, awesome :) I haven't managed to look at the code yet, so at least giving my 2cts on your comments

* `RetryAfter`: I'd personally also like a timedelta here, nice catch actually! In terms on interface changes I don't see a reason not to do that. If updating the pickle behavior becomes difficult I'd be okay with staying with an integer. Mind that pickle behavior doesn't need to be backward compatibile - we explicitly exclude that in the stability policy and I guess picklung exceptions is a rather small use case …
* thanks for also catching `get_updates`. Even though I'm not quite sure why I skipped it, I think it may be because I also skipped all the other timeout parameters. Accepting timedelta for them as well might make sense as well but here the issue would be that the abstract `BaseRquest.do_request` should then also broaden the signature. So I'd postpones that discussion for now. for `get_updates.timeout` I'd say you can feel free to update also the signatures that forward the value to `get_updates.timeout` :)

changes/unreleased/4750.jJBu7iAgZa96hdqcpHK96W.toml
```diff
@@ -0,0 +1,35 @@
+features = "Use `timedelta` to represent time periods in classes"
+deprecations = """In this release, we're migrating attributes that represent durations/time periods from :ob:`int` type to Python's native :class:`datetime.timedelta`.
+
+Enable the ``PTB_TIMEDELTA`` environment variable to adopt :obj:`timedelta` now. Support for :obj:`int` values is deprecated and will be removed in a future major version.
```
harshil21, line 4:
```suggestion
Set the ``PTB_TIMEDELTA=1`` environment variable to adopt :obj:`timedelta` now. Support for :obj:`int` values is deprecated and will be removed in a future major version.
```
it's not clear to me what happens when you set the env variable, are timedelta objects returned? If that's the case, the release note should mention that.

docs/substitutions/global.rst
```diff
@@ -101,3 +101,5 @@
 .. |org-verify| replace:: `on behalf of the organization <https://telegram.org/verify#third-party-verification>`__
 
 .. |time-period-input| replace:: :class:`datetime.timedelta` objects are accepted in addition to plain :obj:`int` values.
+
+.. |time-period-int-deprecated| replace:: In a future major version this will be of type :obj:`datetime.timedelta`. You can opt-in early by setting the `PTB_TIMEDELTA` environment variable.
```
harshil21, line 105:
```suggestion
.. |time-period-int-deprecated| replace:: In a future major version this will be of type :obj:`datetime.timedelta`. You can opt-in early by setting `PTB_TIMEDELTA=1` as an environment variable.
```

telegram/_bot.py
```diff
@@ -4591,6 +4594,12 @@ async def get_updates(
         else:
             arg_read_timeout = self._request[0].read_timeout or 0
 
+        read_timeout = (
+            (arg_read_timeout + timeout.total_seconds())
+            if isinstance(timeout, dtm.timedelta)
+            else (arg_read_timeout + timeout if isinstance(timeout, int) else arg_read_timeout)
+        )
```
harshil21, line 4601:
```suggestion
        read_timeout = (
            (arg_read_timeout + timeout.total_seconds())
            if isinstance(timeout, dtm.timedelta)
            else (arg_read_timeout + timeout if timeout else arg_read_timeout)
        )
```
ah I don't think I've ever seen nested ternary if statements in python haha. Regarding the change, we should check for truthiness of `timeout` like before, because your case didn't cover if timeout was a float.

telegram/_utils/datetime.py
```diff
@@ -224,3 +229,28 @@ def _datetime_to_float_timestamp(dt_obj: dtm.datetime) -> float:
     if dt_obj.tzinfo is None:
         dt_obj = dt_obj.replace(tzinfo=dtm.timezone.utc)
     return dt_obj.timestamp()
+
+
+def get_timedelta_value(value: Optional[dtm.timedelta]) -> Optional[Union[float, dtm.timedelta]]:
+    """
+    Convert a `datetime.timedelta` to seconds or return it as-is, based on environment config.
```
harshil21, line 236: should mention here that it returns a timedelta if the env var is set, otherwise a float.

telegram/_utils/datetime.py
```diff
@@ -224,3 +229,28 @@ def _datetime_to_float_timestamp(dt_obj: dtm.datetime) -> float:
...
+        return None
+
+    if env_var_2_bool(os.getenv("PTB_TIMEDELTA")):
+        return value
+
+    warn(
+        PTBDeprecationWarning(
+            "NEXT.VERSION",
+            "In a future major version this will be of type `datetime.timedelta`."
+            " You can opt-in early by setting the `PTB_TIMEDELTA` environment variable.",
```
harshil21, line 252:
```suggestion
            " You can opt-in early by setting `PTB_TIMEDELTA=1` as an environment variable.",
```

telegram/_chatfullinfo.py
```diff
@@ -617,3 +659,17 @@ def de_json(cls, data: JSONDict, bot: Optional["Bot"] = None) -> "ChatFullInfo":
         )
 
         return super().de_json(data=data, bot=bot)
+
+    def to_dict(self, recursive: bool = True) -> JSONDict:
```
harshil21, line 663: ah can't this be modified in `telegramobject.py` only?

telegram/_chatinvitelink.py
```diff
@@ -177,6 +187,13 @@ def __init__(
 
         self._freeze()
 
+    @property
+    def subscription_period(self) -> Optional[Union[int, dtm.timedelta]]:
+        value = get_timedelta_value(self._subscription_period)
+        if isinstance(value, float) and value.is_integer():
+            value = int(value)
```
harshil21, line 194: can this block not be moved into `get_timedelta_value` to reduce duplication?

telegram/_files/video.py
```diff
@@ -80,18 +93,21 @@ class Video(_BaseThumbedMedium):
             the video in the message.
 
             .. versionadded:: 21.11
-        start_timestamp (:obj:`int`): Optional, Timestamp in seconds from which the video
-            will play in the message
+        start_timestamp (:obj:`int` | :class:`datetime.timedelta`): Optional, Timestamp in seconds
```
harshil21, line 96:
```suggestion
        start_timestamp (:obj:`int` | :class:`datetime.timedelta`): Optional. Timestamp in seconds
```

telegram/_files/video.py
```diff
@@ -122,18 +138,48 @@ def __init__(
...
+            self._duration: dtm.timedelta = parse_period_arg(duration)  # type: ignore[assignment]
             # Optional
             self.mime_type: Optional[str] = mime_type
             self.file_name: Optional[str] = file_name
             self.cover: Optional[Sequence[PhotoSize]] = parse_sequence_arg(cover)
-            self.start_timestamp: Optional[int] = start_timestamp
+            self._start_timestamp: Optional[dtm.timedelta] = parse_period_arg(start_timestamp)
+
+    @property
+    def duration(self) -> Union[int, dtm.timedelta]:
```
harshil21, line 149: I might be wrong, but I think we have to add a docstring for the properties here and remove it from the Attributes section up top. If you can build the docs and check how they look and confirm if they're good that would be nice.

tests/_files/test_animation.py
```diff
@@ -99,12 +101,31 @@ def test_to_dict(self, animation):
...
         assert animation_dict["height"] == animation.height
-        assert animation_dict["duration"] == animation.duration
+        assert animation_dict["duration"] == int(self.duration.total_seconds())
+        assert isinstance(animation_dict["duration"], int)
         assert animation_dict["thumbnail"] == animation.thumbnail.to_dict()
         assert animation_dict["file_name"] == animation.file_name
         assert animation_dict["mime_type"] == animation.mime_type
         assert animation_dict["file_size"] == animation.file_size
 
+    def test_time_period_properties(self, PTB_TIMEDELTA, animation):
```
harshil21, line 111: I don't think it's idiomatic to have an all caps fixture name in the argument. Maybe rename it?

tests/conftest.py
```diff
@@ -129,6 +135,18 @@ def _disallow_requests_in_without_request_tests(request):
             )
 
 
+@pytest.fixture(scope="module", params=["true", "false", None])
+def PTB_TIMEDELTA(request):
+    # Here we manually use monkeypatch to give this fixture module scope
+    monkeypatch = pytest.MonkeyPatch()
```
harshil21, line 141: why doesn't specifying monkeypatch as a parameter work?

tests/test_official/exceptions.py
```diff
@@ -102,16 +95,16 @@ class ParamTypeCheckingExceptions:
...
             "sticker": str,  # actual: Union[str, FileInput]
         },
         "InputStoryContent.*": {
             "photo": str,  # actual: Union[str, FileInput]
             "video": str,  # actual: Union[str, FileInput]
-            "duration": float,  # actual: dtm.timedelta
-            "cover_frame_timestamp": float,  # actual: dtm.timedelta
+        },
+        "TransactionPartnerUser": {
+            "subscription_period": int,  # actual: Union[int, dtm.timedelta]
```
harshil21, line 107: do we still need this?

harshil21 (changes requested):
a simple change, but we have to change so many files and tests for this :(

PR is mostly good, I'm just double checking things-

tests/test_official/exceptions.py
aelkheir: Yeah this is only case that the TIMEDELAT_REGEX doesn't cover.

tests/conftest.py
aelkheir: i think the monkeypatch fixture has `'function'` scope by default, so one can't use it in another fixture with broader scope.

I couldn't find it stated explicitly in the docs, but https://github.com/pytest-dev/pytest/issues/363 discusses the subject at length.

[pytest.MonkeyPatch.context](https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.context) could also work here

telegram/_files/video.py
aelkheir: no you're right that also works. It seems they have to be at one place or the other, but not both or sphinx issues the warning:
`docstring of ...: WARNING: duplicate object description of ..., other instance in ..., use :no-index: for one of them`

currently it also renders fine and pressing on the [source] link takes you the `@property`

telegram/_chatinvitelink.py
aelkheir: Oh, right. I intentionally had it here b/c i thought some attributes are of type `float`. But now I remember all the float durations were introduced in API 9.0 and they don't use this helper 🤦‍♂️.

Thank for the catch

telegram/_files/inputmedia.py
```diff
@@ -110,6 +112,19 @@ def _parse_thumbnail_input(thumbnail: Optional[FileInput]) -> Optional[Union[str
             else thumbnail
         )
 
+    def to_dict(self, recursive: bool = True) -> JSONDict:
+        """See :meth:`telegram.TelegramObject.to_dict`."""
+        out = super().to_dict(recursive)
+        if isinstance(self, (InputMediaAnimation, InputMediaVideo, InputMediaAudio)):
```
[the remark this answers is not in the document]
aelkheir: invalid

telegram/_chatfullinfo.py
aelkheir: technically yes.. I just thought it would be simpler to let each subclass manage its own set of deprecated attributes. and most of this stuff is going away anyway.

aelkheir: with hindsight, it was as simple if not more 😄

aelkheir:
[7d93eb5](https://github.com/python-telegram-bot/python-telegram-bot/pull/4750/commits/7d93eb5ad246722c498e4ffccae801ae84f3a9ea)
> Failure at `src/telegram/_utils/datetime.py:38:0` is not related to this
PR.

that one's related :). I meant [/stars/staramount.py:19](https://github.com/python-telegram-bot/python-telegram-bot/pull/4750/commits/7d93eb5ad246722c498e4ffccae801ae84f3a9ea#diff-0e37d7c1ca3aeb936a0f479818a647501ee07858cdc95a2b1454cbdfaf3287a3)

tests/ext/test_updater.py
aelkheir, line 1: without these modifications the test hangs..
not sure if i missed something inside Updater.
if you notice something please note it.

changes/unreleased/4750.jJBu7iAgZa96hdqcpHK96W.toml
```diff
@@ -0,0 +1,36 @@
+features = "Use `timedelta` to represent time periods in classes"
```
Bibo-Joshi:
```suggestion
features = "Use :class:`datetime.timedelta` to represent time periods in class arguments and attributes"
```

changes/unreleased/4750.jJBu7iAgZa96hdqcpHK96W.toml
```diff
@@ -0,0 +1,36 @@
...
+deprecations = """In this release, we're migrating attributes of Telegram objects that represent durations/time periods from having :obj:`int` type to Python's native :class:`datetime.timedelta`.
```
Bibo-Joshi:
```suggestion
deprecations = """In this release, we're migrating attributes of Telegram objects that represent durations/time periods from having :obj:`int` type to Python's native :class:`datetime.timedelta`. This change is opt-in for now to allow for a smooth transition phase. It will become opt-out in future releases.
```

src/telegram/_chatfullinfo.py
```diff
@@ -600,6 +631,13 @@ def de_json(cls, data: JSONDict, bot: Optional["Bot"] = None) -> "ChatFullInfo":
             Message,
         )
 
+        data["slow_mode_delay"] = (
+            dtm.timedelta(seconds=s) if (s := data.get("slow_mode_delay")) else None
```
Bibo-Joshi:
```suggestion
            parse_period_arg(data.get("slow_mode_delay"))
```

Alternatively, I'd also be okay with not doing a conversion in de_json at all and instead letting that happen in `__init__`. This comment basically applies in all places 😅 Or am I maybe missing something why this should definitely happen here? :)

src/telegram/_telegramobject.py
```diff
@@ -629,7 +640,11 @@ def to_dict(self, recursive: bool = True) -> JSONDict:
             elif isinstance(value, dtm.datetime):
                 out[key] = to_timestamp(value)
             elif isinstance(value, dtm.timedelta):
-                out[key] = value.total_seconds()
+                # Converting to int here is neccassry in some cases where Bot API returns
+                # 'BadRquest' when expecting integers (e.g. Video.duration)
+                out[key] = (
+                    int(seconds) if (seconds := value.total_seconds()).is_integer() else seconds
+                )
```
Bibo-Joshi: interesting catch! I'm wondering if there are cases were a float is accepted as time duration. that would be a hard-to-catch bug, if the user passes a float and the int-version is passed  by TG 🤔

src/telegram/error.py
```diff
@@ -208,21 +213,43 @@ class RetryAfter(TelegramError):
...
+    @property
+    def retry_after(self) -> Union[int, dtm.timedelta]:  # noqa: D102
+        # Diableing D102 because docstring for `retry_after` is present at the class's level
+        return get_timedelta_value(  # type: ignore[return-value]
+            self._retry_after, attribute="retry_after"
+        )
 
     def __reduce__(self) -> tuple[type, tuple[float]]:  # type: ignore[override]
-        return self.__class__, (self.retry_after,)
+        # Until support for `int` time periods is lifted, leave pickle behaviour the same
```
Bibo-Joshi, line 249:
```suggestion
        # Until support for `int` time periods is lifted, leave pickle behaviour the same
        # tag: deprecated: NEXT.VERSION
```

just for finding it easier : ) nice catch!

tests/_files/test_audio.py
```diff
@@ -71,7 +72,7 @@ def test_creation(self, audio):
         assert audio.file_unique_id
 
     def test_expected_values(self, audio):
-        assert audio.duration == self.duration
+        assert audio._duration == self.duration
```
Bibo-Joshi, line 75: I would prefer it if we don't rely on protected members in the test suite. Does `audio.duration == self.duration` not suffice? Or does the `PTB_TIMEDELTA` fixtire inferfere in concurrent testing?

tests/test_official/exceptions.py
Bibo-Joshi: could you maybe extend the pattern to `r"((in|number of) seconds|(\w+_period$)"`? If that doesn't work for some reason, it would be good to leave some code comments here why this is needed :)

Bibo-Joshi (commented):
Thanks a lot for this nice PR! Very clean code changes IMO. I've left a few comments, but I think only a few of them are substantial. :) I hope that I haven't doubled any earier comments of harshil 😅

src/telegram/_utils/datetime.py
```diff
@@ -224,3 +228,47 @@ def _datetime_to_float_timestamp(dt_obj: dtm.datetime) -> float:
...
+
+    Returns:
+        - :obj:`None` if :paramref:`value` is None.
+        - :obj:`datetime.timedelta` if `PTB_TIMEDELTA=true`.
+        - :obj:`int` if the total seconds is a whole number.
+        - float: otherwise.
+    """
+    if value is None:
+        return None
+    if os.getenv("PTB_TIMEDELTA", "false").lower().strip() == "true":
```
harshil21: we should check for PTB_TIMEDELTA=true _and_ PTB_TIMEDELTA=1. Question is whether general truthiness would count, i.e. `PTB_TIMEDELTA=something_not_false_or_zero`? I don't think it should.

src/telegram/_telegramobject.py
```diff
@@ -665,3 +680,30 @@ def set_bot(self, bot: Optional["Bot"]) -> None:
             bot (:class:`telegram.Bot` | :obj:`None`): The bot instance.
         """
         self._bot = bot
+
+    # We use str keys to avoid importing which causes circular dependencies
+    _TIME_PERIOD_DEPRECATIONS: ClassVar = {
```
harshil21: This could just be a module level constant, no need to increase memory by adding this to every object imo

src/telegram/_telegramobject.py
```diff
@@ -499,6 +499,13 @@ def _apply_api_kwargs(self, api_kwargs: JSONDict) -> None:
             elif getattr(self, key, True) is None:
                 setattr(self, key, api_kwargs.pop(key))
 
+    def _is_deprecated_attr(self, attr: str) -> bool:
+        """Checks wheather `attr` is in the list of deprecated time period attributes."""
```
harshil21:
```suggestion
        """Checks whether `attr` is in the list of deprecated time period attributes."""
```

src/telegram/_telegramobject.py
```diff
@@ -521,7 +528,11 @@ def _get_attrs_names(self, include_private: bool) -> Iterator[str]:
 
         if include_private:
             return all_attrs
-        return (attr for attr in all_attrs if not attr.startswith("_"))
+        return (
+            attr
+            for attr in all_attrs
+            if not attr.startswith("_") or self._is_deprecated_attr(attr)
```
harshil21, line 534: this is confusing. Why are we possibly returning "private" attributes when include_private=False?

src/telegram/_telegramobject.py
```diff
@@ -558,7 +569,7 @@ def _get_attrs(
                 if recursive and hasattr(value, "to_dict"):
                     data[key] = value.to_dict(recursive=True)
                 else:
-                    data[key] = value
+                    data[key.removeprefix("_") if self._is_deprecated_attr(key) else key] = value
```
harshil21: why are we then removing the _ after including it intentionally in _get_attrs_names? This would mean we would not get private attributes which are deprecated when include_private=True.

If the intention is to return only the public facing attribute of something like Video.duration (because we don't want the internal Video._duration), then the logic should be in line 644, to change the key name (see my next comment).

src/telegram/_telegramobject.py
```diff
@@ -629,7 +640,11 @@ def to_dict(self, recursive: bool = True) -> JSONDict:
             elif isinstance(value, dtm.datetime):
                 out[key] = to_timestamp(value)
             elif isinstance(value, dtm.timedelta):
-                out[key] = value.total_seconds()
+                # Converting to int here is neccassry in some cases where Bot API returns
+                # 'BadRquest' when expecting integers (e.g. Video.duration)
+                out[key] = (
```
harshil21:
```suggestion
                out[key.removeprefix('_')] = (
```
and then,`pop_keys.add(key)`. 

This would keep the behaviour of the other functions untouched.

src/telegram/_utils/argumentparsing.py
```diff
@@ -50,6 +51,22 @@ def parse_sequence_arg(arg: Optional[Sequence[T]]) -> tuple[T, ...]:
...
+        arg (:obj:`int` | :class:`datetime.timedelta`, optional): The time period to parse.
+
+    Returns:
+        :obj:`timedelta`: The time period converted to a timedelta object or :obj:`None`.
+    """
+    if arg is None:
+        return None
+    if isinstance(arg, (int, float)):
+        return dtm.timedelta(seconds=arg)
+    return arg
```
harshil21, line 79: the type checker is probably complaining when using this function because there's a chance we could don't return None or a timedelta, which contradicts the expected return type of the function. Maybe `arg` should accept a generic too? The resulting Returns docstring needs to be changed too. cc @Bibo-Joshi

src/telegram/_messageautodeletetimerchanged.py
```diff
@@ -35,26 +41,52 @@ class MessageAutoDeleteTimerChanged(TelegramObject):
...
+        self._message_auto_delete_time: dtm.timedelta = parse_period_arg(
+            message_auto_delete_time
+        )  # type: ignore[assignment]
 
         self._id_attrs = (self.message_auto_delete_time,)
 
         self._freeze()
+
+    @property
+    def message_auto_delete_time(self) -> Union[int, dtm.timedelta]:
```
harshil21, line 72:
```suggestion
    def message_auto_delete_time(self) -> Optional[TimePeriod]:
```

same for all other places. This should also resolve mypy complaining.

src/telegram/_videochat.py
```diff
@@ -62,28 +66,54 @@ class VideoChatEnded(TelegramObject):
...
+
+    @classmethod
+    def de_json(cls, data: JSONDict, bot: Optional["Bot"] = None) -> "VideoChatEnded":
+        """See :meth:`telegram.TelegramObject.de_json`."""
+        data = cls._parse_data(data)
+
+        data["duration"] = dtm.timedelta(seconds=s) if (s := data.get("duration")) else None
+
+        return super().de_json(data=data, bot=bot)
+
```
harshil21:
```suggestion
```
don't need this if it's done in the `__init__`? I guess this is the same as https://github.com/python-telegram-bot/python-telegram-bot/pull/4750#discussion_r2121930855

src/telegram/error.py
```diff
@@ -208,21 +213,43 @@ class RetryAfter(TelegramError):
...
 
-    __slots__ = ("retry_after",)
+    __slots__ = ("_retry_after",)
+
+    def __init__(self, retry_after: TimePeriod):
+        self._retry_after: dtm.timedelta = parse_period_arg(  # type: ignore[assignment]
+            retry_after
+        )
+
+        if isinstance(self.retry_after, int):
```
harshil21, line 236:
```suggestion
        if isinstance(retry_after, int):
```
feels weird to immediately call a property which unwraps the value just assigned to `_retry_after`. The pattern I would like is something like this:

```python
seconds_in_int = retry_after if isinstance(retry_after, int) else retry_after.total_seconds()
super().__init__("...")
```
this pattern avoids some repetition imo

src/telegram/error.py
```diff
@@ -208,21 +213,43 @@ class RetryAfter(TelegramError):
...
+            .. deprecated:: NEXT.VERSION
+                |time-period-int-deprecated|
 
     """
 
-    __slots__ = ("retry_after",)
+    __slots__ = ("_retry_after",)
+
+    def __init__(self, retry_after: TimePeriod):
+        self._retry_after: dtm.timedelta = parse_period_arg(  # type: ignore[assignment]
```
harshil21: the type checker is right in this case

src/telegram/_utils/argumentparsing.py
```diff
@@ -50,6 +51,22 @@ def parse_sequence_arg(arg: Optional[Sequence[T]]) -> tuple[T, ...]:
     return tuple(arg) if arg else ()
 
 
+def parse_period_arg(arg: Optional[Union[int, float, dtm.timedelta]]) -> Optional[dtm.timedelta]:
```
harshil21:
```suggestion
def to_timedelta(arg: Optional[Union[int, float, dtm.timedelta]]) -> Optional[dtm.timedelta]:
```
would be a more apt name

src/telegram/ext/_application.py
```diff
@@ -780,8 +781,11 @@ def run_polling(
         Args:
             poll_interval (:obj:`float`, optional): Time to wait between polling updates from
                 Telegram in seconds. Default is ``0.0``.
-            timeout (:obj:`int`, optional): Passed to
-                :paramref:`telegram.Bot.get_updates.timeout`. Default is ``10`` seconds.
+            timeout (:obj:`int` | :class:`datetime.timedelta`, optional): Passed to
+                :paramref:`telegram.Bot.get_updates.timeout`. Default is ``timedelta(seconds=10)``.
```
harshil21: I forgot the exact sphinx syntax, but we can link to the timedelta docs while telling that it's 10 seconds, it might be something like this (we've used it in other places in the lib):

```suggestion
                :paramref:`telegram.Bot.get_updates.timeout`. Default is :obj:`timedelta(seconds=10)<datetime.timedelta>`.
```

tests/_files/test_audio.py
harshil21: we could globally disable the warning message in the tests if that's the issue

tests/_files/test_video.py
```diff
@@ -38,15 +39,22 @@
 from tests.auxil.slots import mro_slots
 
 
+@pytest.fixture(scope="module")
+def video(video):
+    with video._unfrozen():
+        video._start_timestamp = VideoTestBase.start_timestamp
+    return video
```
harshil21: why is this necessary only for `video`?

harshil21 (changes requested):
Thanks for the work on this PR! I left more comments, mainly for the core logic

src/telegram/_utils/datetime.py
Bibo-Joshi: "explicit is better than implicit" :clown: allowing `1` would be okay for me, but I'd be perfectly fine with restricting to `true` (case insensitive)

src/telegram/_telegramobject.py
Bibo-Joshi: there correspond to the deprecated attributes which currently are properties. returning all property names would be too much (that would include `update.effective_chat` etc. leaving a CSI comment wouldn't hurt, true, but the logic is sound IMO

src/telegram/_telegramobject.py
Bibo-Joshi: that's a fair point 👍

src/telegram/error.py
Bibo-Joshi: you mean b/c the return value is `timedelta | None` instead of `timedelta`? If you want to improve type hinting for `parse_period_arg` for that, you can use `typing.overload` to say `None → None` and `int | float | td → td`

src/telegram/_utils/argumentparsing.py
Bibo-Joshi: Sorry, why should a type checker complain and why is there a chance that the return value is something other than `dtm.timedelta | None? If if-clauses take care of that very well IMO. How would a generic help?

aelkheir: ahh It's my understanding that this function will always return either `None `or `timedelta `as far as typechecking is concerned.

However it's being used to parse many different period arguments, some of which are required, some are optional. So when used to parse a required argument type checking will complain that this function possibly returns None, which I silence b/c it's known otherwise :)
In the case of optional arguments/attributes no type checking error is reported.

Example from _video.py
```python
    def __init__(
        self,
        duration: TimePeriod,
        start_timestamp: Optional[TimePeriod] = None,
        *,
        api_kwargs: Optional[JSONDict] = None,
    ):
        super().__init__(
            ...
        )
        with self._unfrozen():
            # Required
            ...
            self._duration: dtm.timedelta = to_timedelta(duration)  # type: ignore[assignment]
            # Optional
            ...
            self._start_timestamp: Optional[dtm.timedelta] = to_timedelta(start_timestamp)

```

src/telegram/_utils/datetime.py
aelkheir: don't mind adding 1. currently I've set it up to mimic how `env_var_2_bool` works.. but thats for internal use up till now

src/telegram/error.py
aelkheir: > the type checker is right in this case

How come 🤔. `retry_after` is a required argument, although optional in API docs but I've assumed that's why it's ignored in `test_official`

src/telegram/error.py
aelkheir: With regard to the code suggestion, In the current implementation the `if,else` branches are meant to decide on which error message should be used based _not_ on the type of the argument passed to `__init__` but rather the value set for `PTB_TIMEDELTA` which then determines that value of the property. Passing either `int` or `timedelta` to `__init__` shouldn't affect the error message differently IMO

str on `timedelta `object  will be more informative, i think:
`Flood control exceeded. Retry in 0:09:00` v.s.  `Flood control exceeded. Retry in 540 seconds`

Regarding the second pattern suggested, it works, but I don't see how that would help us have different error messages depending on the config, assuming that behavior is desired :D

Alternatively we could remove the reliance on the property inside __init__ and always represent messages with a timedelta object.

src/telegram/_telegramobject.py
aelkheir: Oh I meant `InputMediaVideo` there, will update comment.

Ah yes unfortunately :( and to make matters worse they're Business related
`InputStoryContentVideo.[duration/cover_frame_timestamp]`
`InputProfilePhotoAnimated.main_frame_timestamp`

would it suffice if we tested whether telegram returns `ok` if provided integers instead of floats, plus a code comment?

this also reminded me one these classes is using it's own float -> timedelta parsing logic. 👍

src/telegram/_telegramobject.py
aelkheir: > increase memory by adding this to every object 

would it tho 🤔. 
shouldn't it being a static class variable mean that all instances of this class and subclasses will share the same memory location of this list.

will update in case i'm missing something.

tests/_files/test_video.py
aelkheir: hm not strictly necessary..
It's that the overridden `video` fixture is the result of sending an actual video with `bot.send_video` defined in `_files/conftest.py`. to test `PTB_TIMEDELTA `handling, I override viedo's start_timestamp property because it's optional and not specified in `bot.send_video`.

could also be setting it at `conftest.py` while calling `bot.send_video` but then we'd be relying on magic numbers across modules.

will add a code comment.

src/telegram/_messageautodeletetimerchanged.py
aelkheir: It would, but it would also violate API docs 🫢 
(this argument/property is required)

PS: I have avoided using `TimePeriod ` here so that type hints in IDEs show the underlying python types and not our utility. haven't confirmed tho that type hints even show on properties.

src/telegram/_telegramobject.py
aelkheir: > why are we then removing the _ after including it intentionally in _get_attrs_names? 

the underscore is included in `_TIME_PERIOD_DEPRECATIONS` to make extraction of the timedelta value latter a bit easier. and as you noted it's removed not to submit to TG invalid keys.

> This would mean we would not get private attributes which are deprecated when include_private=True.

I don't fully understand this part. Currently private attributes that are not on the deprecation list (e.g. `_freeze`, `_id_attrs`) will function as normal when if include_private=True. Couldn't understand what you mean by "private attributes which are deprecated"

src/telegram/_telegramobject.py
aelkheir: If i'm understanding correctly this would only substitute the logic in L572 ` in_get_attrs()` 😅?

is the advantage of this that here we wouldn't need the if check (above comment) and not get in the way of unaffected TG objects?

if yes, note that this way is changing the dict size while iterating, since this is not a simple overwrite. can work around it if it's still better.
@harshil21 @Bibo-Joshi

src/telegram/_utils/argumentparsing.py
harshil21: oof my bad, I thought `arg` didn't have `dtm.timedelta` in its annotation. This is fine.

src/telegram/error.py
harshil21: yeah that would fix it.

tests/_files/test_audio.py
aelkheir: yes relying on the property here would make `test_expected_values` and similar tests dependent on `PTB_TIMEDELTA` fixture.  we could do that, but the resulting behavior would just duplicate what's already being tested in `test_time_period_properties`.

...
that's strange i'm not getting any warnings..

src/telegram/_chatfullinfo.py
aelkheir: Oh right!
these were done here not cause a warning; early on when passing `int`s to `__init__`s was also deprecated.

src/telegram/_telegramobject.py
aelkheir: I did a bit of research on python's class variables, and as I was able to gather it would become a problem but only if one of the subclasses reassigned the `_TIME_PERIOD_DEPRECATIONS`. We are not doing that, and so shouldn't users since it's a private variable.

could you point out how it would affect memory from your pov 😅

aelkheir (commented):
apologies for the late updates

src/telegram/_telegramobject.py
harshil21: ah well after some testing of my own too, It seems that python does use the reference of the object across many instances, so the memory isn't added to every object. So even a module level constant would mean that it takes the same memory overall in the end.

I guess the only reason now to move it to a module constant is that it's something only used internally and not relevant to a TelegramObject at all.

src/telegram/_telegramobject.py
harshil21: > is the advantage of this that here we wouldn't need the if check (above comment) and not get in the way of unaffected TG objects?

yes indeed. I'm just trying to make sure that there is separation of concerns.

src/telegram/error.py
harshil21: ah I didn't consider the behaviour of the message depending on the env var. The current way is fine then imo 👍🏽

src/telegram/_telegramobject.py
```diff
@@ -614,6 +615,7 @@ def to_dict(self, recursive: bool = True) -> JSONDict:
         # datetimes to timestamps. This mostly eliminates the need for subclasses to override
         # `to_dict`
         pop_keys: set[str] = set()
+        timedelta_dict: dict = {}
```
Bibo-Joshi, line 617: this approach works fine 👍 an alternative would be to do `for key, value in list(out.items())`  - just fyi :)

src/telegram/_utils/argumentparsing.py
```diff
@@ -51,7 +51,7 @@ def parse_sequence_arg(arg: Optional[Sequence[T]]) -> tuple[T, ...]:
     return tuple(arg) if arg else ()
 
 
-def parse_period_arg(arg: Optional[Union[int, float, dtm.timedelta]]) -> Optional[dtm.timedelta]:
+def to_timedelta(arg: Optional[Union[int, float, dtm.timedelta]]) -> Optional[dtm.timedelta]:
```
Bibo-Joshi, line 66:
```python
from typing import overload, Optional, Union
import datetime as dtm

@overload
def to_timedelta(arg: None) -> None:
    ...

@overload
def to_timedelta(arg: Union[int, float, dtm.timedelta]) -> dtm.timedelta:
    ...

def to_timedelta(arg: Optional[Union[int, float, dtm.timedelta]]) -> Optional[dtm.timedelta]:
    # implementation
```

this would resolve a number of the `type: ignore[assignment]` comments, see https://github.com/python-telegram-bot/python-telegram-bot/pull/4750#discussion_r2127236363. Not strictly required IMO and could also be postponed to a follow-up PR. Just recalled this and wanted to leave a comment :)

src/telegram/_telegramobject.py
Bibo-Joshi: ah, thanks for the info! Pointing out the currently known use cases should suffice IMO. If you want to, adding a special test case for *one* of them would be plenty - you could add a comment along the lines "this applies also to other classes e.g. …, but we test it only for one as this suffices to ensure that the logic works" :)

aelkheir: hmm we can't really test this in the test suite, because it requires a Telegram Business account. I can prepare a minimal example to send an actual request if a team member has access to a test business account. 
@Poolitzer (or others) would someone be able help verify this?

For context I want to ensure that making a request like this:
```python
await def post_story(
    self,
    "business_connection_id",
    content=InputStoryContentVideo(
        video=video.read_bytes(),
        duration=10.0, # or datetime.timedelta(seconds=10)
        cover_frame_timestamp=5.0, # or datetime.timedelta(seconds=5)
        is_animation=False,
    ),
)
```
to telegram does not raise a `BadRequest` for invalid `duration` or `cover_frame_timestamp`. and a similar request for `bot.set_business_account_profile_photo` with `photo=InputProfilePhotoAnimated`.

I can run the example or provide it so that someone could :)

src/telegram/_telegramobject.py
aelkheir: aha noted! that's a neat trick, saw you doing that the other time
TBH I haven't considered it in my first pass, however I think, here, maintaining a separate `dict` is more suitable since we'll be removing some keys and want to put them back.

src/telegram/_telegramobject.py
aelkheir: >  it's something only used internally and not relevant to a TelegramObject at all.

Fair point right there 👍

src/telegram/_telegramobject.py
Bibo-Joshi: mh, I don't have a business account and I fear that setting up something for them in the tests would be a hassle. Alternatively, you can mock the tg response by mocking `httpx.post` and returning an error if the payload contains a float

aelkheir: Aha no problem!
I'm just thinking mocking the request would be of little use, because we don't know the actual Telegram behavior (throw, or not throw)
> mocking httpx.post and returning an error if the payload contains a float

Also in the case of two methods above, passing a float to Telegram should be successful, it's integers that might (or might not) throw BadRequest errors.

harshil21: I do have a business account, is there an example you'd like me to run?

Bibo-Joshi: >Also in the case of two methods above, passing a float to Telegram should be successful, it's integers that might (or might not) throw BadRequest errors.

okay, the other way around :D 

>I'm just thinking mocking the request would be of little use, because we don't know the actual Telegram behavior (throw, or not throw)

That's true, it's not perfect. What we would achieve with this is that we avoid regressions in what we think is the right implementation :)

aelkheir: > okay, the other way around :D

haha thought i was the lost one for a sec xD

here is the test bot.
<details>
<summary>float_periods_bot.py</summary>

```python
#!/usr/bin/env python

import asyncio
import datetime
import logging

from telegram import Update
from telegram._files._inputstorycontent import InputStoryContentVideo
from telegram._files.inputprofilephoto import InputProfilePhotoAnimated
from telegram.constants import StoryLimit
from telegram.ext import Application, CommandHandler, ContextTypes
from tests.auxil.files import data_file

logging.basicConfig(
)
logging.getLogger("httpx").setLevel(logging.DEBUG)

async def post_story(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
    for duration, cover_timestamp in [
        (4.0, 2.0),
        (4, 2),
        (datetime.timedelta(seconds=4.0), datetime.timedelta(seconds=2.0)),
    ]:
        content = InputStoryContentVideo(
            video=data_file("telegram.mp4").read_bytes(),
            duration=duration,
            cover_frame_timestamp=cover_timestamp,
        )
        logging.info("Posting story with arguments of type %s", type(duration))
        logging.info("content: %s", content.to_dict())

        await context.bot.post_story(
            "BUSINESS_CONNECTION_ID",
            content=content,
            active_period=StoryLimit.ACTIVITY_SIX_HOURS,
        )
        await asyncio.sleep(0.5)


async def set_profile_photo(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
    """Send a message when the command /help is issued."""
    for timestamp in [
        1.0,
        1,
        datetime.timedelta(seconds=1),
    ]:
        photo = InputProfilePhotoAnimated(
            animation=data_file("telegram2.mp4").read_bytes(),
            main_frame_timestamp=1.0,
        )
        logging.info("Setting profile photo with arguments of type %s", type(timestamp))
        logging.info("photo %s", photo.to_dict())
        await context.bot.set_business_account_profile_photo(
            "BUSINESS_CONNECTION_ID",
            photo=photo,
            is_public=False,
        )
        await asyncio.sleep(0.5)


def main() -> None:
    """Start the bot."""
    application = Application.builder().token("TOKEN").build()

    application.add_handler(CommandHandler("post_story", post_story))
    application.add_handler(CommandHandler("set_profile_photo", set_profile_photo))

    application.run_polling(allowed_updates=Update.ALL_TYPES)


if __name__ == "__main__":
    main()
```
</details>

thank you so, very much!

aelkheir: as discussed internally here https://t.me/c/1494805131/45463, An actual request for `post_story ` ran successfully with  integers in place of floats.

harshil21 (approved):
LGTM!

Bibo-Joshi:
Thank you very much for this contribution and especially the patience in the review process!

PR #4760, opened 2025-04-17
Create python-package-conda.yml

halye74:


Bibo-Joshi (commented):
Hi. I don't see what exactly this workflow is supposed to achieve and why we'd need it. I'll go ahead and close the PR. If you wish for PTB to add support for some special setups, kindly open a feature request so that we can properly discussd that.

PR #4766, opened 2025-04-24
Api 9.0 profile photos

Bibo-Joshi:
Check-list for PRs
------------------

- [x] Checked the Bot API specific sections of the `Stability Policy <https://docs.python-telegram-bot.org/stability_policy.html>`_
- [ ] Created a PR to remove functionality deprecated in the previous Bot API release (`see here <https://docs.python-telegram-bot.org/en/stable/stability_policy.html#case-2>`_)

- New classes:
  - [ ] ~Added ``self._id_attrs`` and corresponding documentation~

- Added new shortcuts:
  - [ ] ~In :class:`~telegram.Chat` & :class:`~telegram.User` for all methods that accept ``chat/user_id``~

- If relevant:
  - [ ] ~Added new filters for new message (sub)types~

### DIFF
--- a/telegram/_files/inputprofilephoto.py
+++ b/telegram/_files/inputprofilephoto.py
@@ -71,7 +71,7 @@ class InputProfilePhotoStatic(InputProfilePhoto):
             :class:`pathlib.Path` |): The static profile photo. |uploadinputnopath|
 
     Attributes:
-        type (:obj:`str`): :tg-const:`telegram.constants.InputMediaType.STATIC`.
+        type (:obj:`str`): :tg-const:`telegram.constants.InputProfilePhotoType.STATIC`.
         photo (:class:`telegram.InputFile` | :obj:`str`): The static profile photo.
 
     """
--- a/telegram/_files/inputprofilephoto.py
+++ b/telegram/_files/inputprofilephoto.py
@@ -116,7 +116,7 @@ class InputProfilePhotoAnimated(InputProfilePhoto):
     def __init__(
         self,
         animation: FileInput,
-        main_frame_timestamp: Union[float, dtm.timedelta, None] = 0.0,
+        main_frame_timestamp: Union[float, dtm.timedelta, None] = None,
         *,
         api_kwargs: Optional[JSONDict] = None,
     ):
--- a/telegram/_files/inputprofilephoto.py
+++ b/telegram/_files/inputprofilephoto.py
@@ -106,7 +106,7 @@ class InputProfilePhotoAnimated(InputProfilePhoto):
             Defaults to ``0.0``.
 
     Attributes:
-        type (:obj:`str`): :tg-const:`telegram.constants.InputMediaType.ANIMATED`.
+        type (:obj:`str`): :tg-const:`telegram.constants.InputProfilePhotoType.ANIMATED`.
         animation (:class:`telegram.InputFile` | :obj:`str`): The animated profile photo.
         main_frame_timestamp (:class:`datetime.timedelta`): Optional. Timestamp in seconds of the
     """
--- a/telegram/_files/inputprofilephoto.py
+++ b/telegram/_files/inputprofilephoto.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python
+#
+# A library that provides a Python interface to the Telegram Bot API
+# Copyright (C) 2015-2025
+# Leandro Toledo de Souza <devs@python-telegram-bot.org>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser Public License for more details.
+#
+# You should have received a copy of the GNU Lesser Public License
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+"""This module contains an objects that represents a InputProfilePhoto and subclasses."""
+
+import datetime as dtm
+from typing import TYPE_CHECKING, Optional, Union
+
+from telegram import constants
+from telegram._telegramobject import TelegramObject
+from telegram._utils import enum
+from telegram._utils.files import parse_file_input
+from telegram._utils.types import FileInput, JSONDict
+
+if TYPE_CHECKING:
+    from telegram import InputFile
+
+
+class InputProfilePhoto(TelegramObject):
+    """This object describes a profile photo to set. Currently, it can be one of
+
+    * :class:`InputProfilePhotoStatic`
+    * :class:`InputProfilePhotoAnimated`
+
+    .. versionadded:: NEXT.VERSION
+
+    Args:
+        type (:obj:`str`): Type of the profile photo.
+
+    Attributes:
+        type (:obj:`str`): Type of the profile photo.
+
+    """
+
+    __slots__ = ("type",)
+
+    def __init__(
+        self,
+        type: str,  # pylint: disable=redefined-builtin
+        *,
+        api_kwargs: Optional[JSONDict] = None,
+    ):
+        super().__init__(api_kwargs=api_kwargs)
+        self.type: str = enum.get_member(constants.InputMediaType, type, type)
+
+        self._freeze()
+
+
+class InputProfilePhotoStatic(InputProfilePhoto):
+    """A static profile photo in the .JPG format.
+
+    .. versionadded:: NEXT.VERSION
+
+    Args:
+        photo (:term:`file object` | :class:`~telegram.InputFile` | :obj:`bytes` | \
+            :class:`pathlib.Path` |): The static profile photo. |uploadinputnopath|
+
+    Attributes:
+        photo (:class:`telegram.InputFile` | :obj:`str`): The static profile photo.
+
+    """
+
+    __slots__ = ("photo",)
+
+    def __init__(
+        self,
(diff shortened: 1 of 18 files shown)


telegram/_files/inputprofilephoto.py
```diff
@@ -0,0 +1,136 @@
...
+    .. versionadded:: NEXT.VERSION
+
+    Args:
+        type (:obj:`str`): Type of the profile photo.
+
+    Attributes:
+        type (:obj:`str`): Type of the profile photo.
+
+    """
+
```
aelkheir, line 49: missing shortcuts for the new enum members here

telegram/_files/inputprofilephoto.py
```diff
@@ -0,0 +1,136 @@
...
+    __slots__ = ("type",)
+
+    def __init__(
+        self,
+        type: str,  # pylint: disable=redefined-builtin
+        *,
+        api_kwargs: Optional[JSONDict] = None,
+    ):
+        super().__init__(api_kwargs=api_kwargs)
+        self.type: str = enum.get_member(constants.InputMediaType, type, type)
```
aelkheir, line 59:
```suggestion
        self.type: str = enum.get_member(constants.InputProfilePhotoType, type, type)
```

telegram/_files/inputprofilephoto.py
```diff
@@ -0,0 +1,136 @@
...
+    """A static profile photo in the .JPG format.
+
+    .. versionadded:: NEXT.VERSION
+
+    Args:
+        photo (:term:`file object` | :class:`~telegram.InputFile` | :obj:`bytes` | \
+            :class:`pathlib.Path` |): The static profile photo. |uploadinputnopath|
+
+    Attributes:
+        type (:obj:`str`): :tg-const:`telegram.constants.InputMediaType.STATIC`.
```
aelkheir, line 74:
```suggestion
        type (:obj:`str`): :tg-const:`telegram.constants.InputProfilePhotoType.STATIC`.
```
that or, 
```suggestion
        type (:obj:`str`): :tg-const:`telegram.constants.InputProfilePhoto.STATIC`.
```
after adding the shortcuts

telegram/_files/inputprofilephoto.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Args:
+        animation (:term:`file object` | :class:`~telegram.InputFile` | :obj:`bytes` | \
+            :class:`pathlib.Path` |): The animated profile photo. |uploadinputnopath|
+        main_frame_timestamp (:class:`datetime.timedelta` | :obj:`int` | :obj:`float`, optional):
+            Timestamp in seconds of the frame that will be used as the static profile photo.
+            Defaults to ``0.0``.
+
+    Attributes:
+        type (:obj:`str`): :tg-const:`telegram.constants.InputMediaType.ANIMATED`.
```
aelkheir, line 109:
```suggestion
        type (:obj:`str`): :tg-const:`telegram.constants.InputProfilePhotoType.ANIMATED`.
```
or
```suggestion
        type (:obj:`str`): :tg-const:`telegram.constants.InputProfilePhoto.ANIMATED`.
```

telegram/_files/inputprofilephoto.py
```diff
@@ -0,0 +1,136 @@
...
+        animation (:class:`telegram.InputFile` | :obj:`str`): The animated profile photo.
+        main_frame_timestamp (:class:`datetime.timedelta`): Optional. Timestamp in seconds of the
+    """
+
+    __slots__ = ("animation", "main_frame_timestamp")
+
+    def __init__(
+        self,
+        animation: FileInput,
+        main_frame_timestamp: Union[float, dtm.timedelta, None] = None,
```
aelkheir, line 119: Will this continue to accept float even after #4750?
if no, then shouldn't we start clean here instead of including it the timedelta migration latter.

telegram/_files/inputprofilephoto.py
```diff
@@ -0,0 +1,136 @@
...
+            self.animation: Union[str, InputFile] = parse_file_input(
+                animation, attach=True, local_mode=True
+            )
+
+            if isinstance(main_frame_timestamp, dtm.timedelta):
+                self.main_frame_timestamp: Optional[dtm.timedelta] = main_frame_timestamp
+            elif main_frame_timestamp is None:
+                self.main_frame_timestamp = None
+            else:
+                self.main_frame_timestamp = dtm.timedelta(seconds=main_frame_timestamp)
```
aelkheir, line 136: just for my understanding as i see it was intentional, why aren't we includeing _id_attrs to the derived classes

tests/test_business_methods.py
```diff
@@ -248,3 +251,77 @@ async def make_assertion(*args, **kwargs):
             new_owner_chat_id=new_owner_chat_id,
             star_count=star_count,
         )
+
+    @pytest.mark.parametrize("is_public", [True, False, None, DEFAULT_NONE])
```
aelkheir, line 255: not yet very familiar with how defaults work.. but why are we parametrizing with `DEFAULT_NONE` here? is_public is defined as a regular optional argument in `Bot.set/remove_business_account_profile_photo`

tests/test_business_methods.py
```diff
@@ -248,3 +251,77 @@ async def make_assertion(*args, **kwargs):
...
+        kwargs = {
+            "business_connection_id": self.bci,
+            "photo": InputProfilePhotoStatic(
+                photo=data_file("telegram.jpg"),
+            ),
+        }
+
+        assert await offline_bot.set_business_account_profile_photo(**kwargs)
+
+    @pytest.mark.parametrize("is_public", [True, False, None, DEFAULT_NONE])
```
aelkheir, line 308: here also

aelkheir (commented):
looking good.. didn't review everything but these places caught my eye..

telegram/_files/inputprofilephoto.py
Bibo-Joshi: Good catch! I'll have to improve the tests as they didn't catch this ..

telegram/_files/inputprofilephoto.py
Bibo-Joshi: Mostly because it's hard to implement really meaningful equality comparison here and the use case is small 😅

tests/test_business_methods.py
Bibo-Joshi: No worries. For the tests I wanted to differentiate between "passing None" and "not passing at all", so I simply needed some kind of maker that's not `None`.
The use case of the `DefaultValue` helper class (`DEFAULT_NONE = DefaultValue(None)` is exactly that - specify a value that's to be used if nothing is passed. That's why I used this. I could have just as well use a string `"not_passing_is_public"` or sim

tests/test_official/exceptions.py
```diff
@@ -97,6 +97,13 @@ class ParamTypeCheckingExceptions:
             "thumbnail": str,  # actual: Union[str, FileInput]
             "cover": str,  # actual: Union[str, FileInput]
         },
+        "InputProfilePhotoStatic": {
+            "photo": str,  # actual: Union[str, FileInput]
+        },
+        "InputProfilePhotoAnimated": {
+            "animation": str,  # actual: Union[str, FileInput]
+            "main_frame_timestamp": float,  # actual: Union[float, dtm.timedelta]
```
harshil21, line 105: test_official needs to be modified internally to automatically add timedeltas too. We shouldn't have to use exceptions for this.

This is maybe a thing for #4750 @aelkheir

harshil21 (changes requested):
just a few minor kinks

tests/test_official/exceptions.py
aelkheir: great idea!
all of these args have "in seconds" in their description so it wouldn't be hard.  will look into it.

telegram/_files/inputprofilephoto.py
aelkheir: agreed... sorry i wasn't clear in that discussion as a good portion of the affected classes in that PR are also Input* classes.

tests/test_business_methods.py
aelkheir: Aha thanks for the explanation. however the differentiation here between None and `Default(None)` is not very meaningful as i see it. unlike  with arguments listed in `Defaults.api_defaults` (e.g. parse_mode, do_quote, ...etc) where the differentiation can have significant effect, here all that it does is either submit is_public=None, or not submitting it at all. even that case wouldn't affect end users unless they call `set_business_profile_photo(photo, is_public=Default(None)) ` providing `Default(None)` explicitly, which i don't think is how it was intended to be used.

Sorry for getting caught is a small thing

Bibo-Joshi: True, the test case is not very important. It basically just makes sure that I didn't write `is_public = 42` or some other nonsense in the function signature by accident. Users are not expected to pass anything other than a bool. In this case `DEFAULT_NONE` is really only used for the test case implementation.

nothing to apologize for, glad you're interested :)

telegram/request/_requestparameter.py
```diff
@@ -148,6 +149,19 @@ def _value_and_input_files_from_input(  # pylint: disable=too-many-return-statem
                 return data, [value.media, thumbnail]
 
             return data, [value.media]
+
+        if isinstance(value, InputProfilePhoto):
+            attr = "photo" if isinstance(value, InputProfilePhotoStatic) else "animation"
```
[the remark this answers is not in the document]
Bibo-Joshi: that doesn't work, unfortunately - the type is type is `animated`, the attribute is called `animation` :/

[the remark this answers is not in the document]
harshil21: ah I misunderstood what we're doing here

harshil21 (approved):
LGTM

aelkheir (approved):
Good to go!

PR #4780, opened 2025-05-02
Fix: Handle current_offset "[]" value in inline results

### LINKED ISSUE #4295: [BUG] ValueError: invalid literal for int() with base 10: '[]'
### Steps to Reproduce

I've attached a stack trace. I hope it's helpful

sentry-stack-trace

### Expected behaviour

tbd

### Actual behaviour

tbd

### Operating System

tbd

```shell
python-telegram-bot 21.2
Bot API 7.3
Python 3.12.3 (main, Apr 10 2024, 05:33:47) [GCC 13.2.0]
```
```


### Relevant log output

_No response_

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by yshalsager**
Fixes #4295

```
site-packages/telegram/_bot.py",
 line 3593, in _effective_inline_results
    current_offset_int = 0 if not current_offset else int(current_offset)
                                                      ^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: '[]'
```

This issue started happening from v21.2 as noted by [users](https://t.me/pythontelegrambotgroup/763054).

I couldn't find the reason / code change, but seems it's a server-side thing from Telegram.

### DIFF
--- a/telegram/_bot.py
+++ b/telegram/_bot.py
@@ -3592,7 +3592,7 @@ class Bot(TelegramObject, contextlib.AbstractAsyncContextManager["Bot"]):
 
         if current_offset is not None:
             # Convert the string input to integer
-            current_offset_int = 0 if not current_offset else int(current_offset)
+            current_offset_int = 0 if not current_offset or current_offset == "[]" else int(current_offset)
 
             # for now set to empty string, stating that there are no more results
             # might change later


yshalsager:
> Hi. Thanks for the PR. Do you have a proper [MWE](https://github.com/python-telegram-bot/python-telegram-bot/wiki/MWE) available that reproduces the issue? I see that #4295 doesn't contain one and the cited telegram thread also contains only pieces …

Hello, unfortunately, I don't have on atm.

Bibo-Joshi:
I see. I won't accept a PR for changing the logic of auto-pagination without having proof that there is actually a bug in it.
My issue is that `current_offset` being `"[]"` means that _someone_ has been passing `next_offset="[]"` to the last call of `answer_inline_query`. In the PTB code

https://github.com/python-telegram-bot/python-telegram-bot/blob/486ceaa6cf48d3b451a5c039536c4c5d7ea72a2b/telegram/_bot.py#L3574-L3625

I don't find any indication for that.

PR #4783, opened 2025-05-06
Create codeql.yml

denysmashevskyi:


Bibo-Joshi:
Hi. Thanks for the PR. however, we already have the default CodeQL scanning enabled on this repo and as far as I see your proposed workflow doesn't do anything else. If you have no specific reasons for us why we should the customized setup instead, I'll decline this PR.

PR #4798, opened 2025-05-22
Change to `src` Layout

### LINKED ISSUE #4797: Use a src layout instead of a flat layout

The python packaging guide recommends that libraries which are meant to be importable (that's us) should use a src layout. The pros of using it is listed here - https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

The `telegram` folder will now be in `src/telegram`.

Not conforming to python packaging standards

### Additional context

`uv` also initializes a new project with src layout by default for libraries, see https://docs.astral.sh/uv/concepts/projects/init/#libraries

Bibo-Joshi:
When ready closes #4797 

Needs manual verification that

- [ ] wheels are build correctly
- [ ] mypy & code coverage are working as expected

I totally just let GH Copilot Agent run for 10 minutes and just manually fixed a few obvious things in the diff 😅


Bibo-Joshi:
About https://github.com/python-telegram-bot/python-telegram-bot/pull/4798/commits/6c7cf83165ff310bfeedea19b92ccc358411b258: I've done this now like I did for [chango](https://github.com/Bibo-Joshi/chango/blob/3b587df3226c57bb5d0497c7e862bb17340c4047/pyproject.toml#L45-L55). I'm not sure though if that's the best way. Note that the source distribution now includes many more files than before. If anyone has suggsetions on best practices, I'm happy for input.

.github/CONTRIBUTING.rst
```diff
@@ -29,7 +29,15 @@ Setting things up
...
 
 
-5. Install pre-commit hooks:
+5. Install the package in development mode:
+
+   .. code-block:: bash
+      
+      $ pip install -e .
+
+   This is necessary because python-telegram-bot uses a src-based layout where the package code is located in the ``src/`` directory.
```
harshil21, line 38: should be updated after #4800 is merged, I suppose it can all go in step 4.

tests/test_modules.py
```diff
@@ -23,14 +23,17 @@
...
 def test_public_submodules_dunder_all():
-    modules_to_search = list(Path("telegram").rglob("*.py"))
+    modules_to_search = list(SOURCE_ROOT_PATH.rglob("*.py"))
 
     if not modules_to_search:
         raise AssertionError("No modules found to search through, please modify this test.")
 
-    for mod_path in modules_to_search:
+    for raw_path in modules_to_search:
+        mod_path = Path(raw_path)
```
harshil21, line 36: isn't it already a Path?

harshil21 (approved):
> If anyone has suggsetions on best practices

I couldn't find any document outlining what files should be in a sdist, [this document](https://packaging.python.org/en/latest/specifications/source-distribution-format/) mentions that a pyproject.toml and `PKG-INFO` file must be present, but otherwise makes no recommendations. I think we can keep pretty much everything then. 

I didn't see any updates to docs related files, and the builds are passing, so I assume that works fine and sphinx can see the files.

harshil21 (approved):

PR #4817, opened 2025-06-06
Allow for pattern matching empty inline queries

locobott:
I just removed the line of code that prevents empty inline queries from being matched using the `pattern` parameter when creating an instance of an `InlineQueryHandler`. 

`update.inline_query` is assumed to always have the `query` property, which I can only assume is always the case, as the `update.inline_query` object must not be None in this case anyway.

Also added myself to AUTHORS.rst

I haven't modified any of the documentation as I feel like this changes makes the documentation more accurate, as the expected behavior is now in place.

### DIFF
--- a/changes/unreleased/4817.ZHx38jvj9zKRNZfQh9Xrpb.toml
+++ b/changes/unreleased/4817.ZHx38jvj9zKRNZfQh9Xrpb.toml
@@ -1,0 +1,5 @@
+other = "Allow for pattern matching empty inline queries"
+[[pull_requests]]
+uid = "4817"
+author_uid = "locobott"
+closes_threads = []
--- a/changes/unreleased/4817.ZHx38jvj9zKRNZfQh9Xrpb.toml
+++ b/changes/unreleased/4817.ZHx38jvj9zKRNZfQh9Xrpb.toml
@@ -1,4 +1,4 @@
-other = "Allow for pattern matching empty inline queries"
+bugfixes = "Allow for pattern matching empty inline queries"
 [[pull_requests]]
 uid = "4817"
 author_uid = "locobott"
--- a/tests/ext/test_inlinequeryhandler.py
+++ b/tests/ext/test_inlinequeryhandler.py
@@ -153,11 +153,11 @@ class TestInlineQueryHandler:
             assert not handler.check_update(update)
 
     @pytest.mark.parametrize(
-        "query,expected_result",
+    "query,expected_result",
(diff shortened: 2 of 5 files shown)


Poolitzer:
Could you add a test which checks that this works now? If you would like pointers where/how feel free to ask, here or in the [dev chat](https://t.me/pythontelegrambotdev).

tests/ext/test_inlinequeryhandler.py
```diff
@@ -152,6 +152,33 @@ async def test_context_pattern(self, app, inline_query):
             update.inline_query.query = "not_a_match"
             assert not handler.check_update(update)
 
+    async def test_empty_inline_query_pattern(self, app):
```
Bibo-Joshi:
```suggestion
    @pytest.mark.parametrize(
        "query,expected_result",
        [
            pytest.param("", True, id="empty string"),
            pytest.param("not empty", False, id="non_empty_string"),
        ]
    )
    async def test_empty_inline_query_pattern(self, app, query, expected_result):
```

then you can do

* `query=query`
* `assert self.test_flag == expected_result`

and reduce the code duplication :)

locobott:
> Thanks for the pr. I left a comment on the test below. please also
> 
> * have a look at the failing pre-commit run. Please run `pre-commit install` in your local clone to ensure the hooks run before the commit. see also [our contrib guide](https://docs.python-telegram-bot.org/en/stable/contributing.html#setting-things-up). running black via `pip install black && black .` should fix the current issues
> * add the file pointed out [here](https://github.com/python-telegram-bot/python-telegram-bot/actions/runs/15496022190/attempts/2#summary-43633405791). Use `bugfixes` instead of `other`. You can change the text to a more descriptive explanation. It will appear [in the changelog](https://docs.python-telegram-bot.org/en/stable/changelog.html).

I've done about everything I knew how to do from your request. Please review and let me know if something is missing and how to fix it.

changes/unreleased/4817.ZHx38jvj9zKRNZfQh9Xrpb.toml
```diff
@@ -0,0 +1,5 @@
+other = "Allow for pattern matching empty inline queries"
```
Bibo-Joshi:
```suggestion
bugfixes = "Allow for pattern matching empty inline queries"
```

Bibo-Joshi (changes requested):

PR #4818, opened 2025-06-07
Fix Handling of Parameters `do_quote` and `allow_sending_without_reply` in `Message.reply_*`

### LINKED ISSUE #4807: reply_text keyword argument allow_sending_without_reply fails when do_quote is True
### Issue
The call `await update.message.reply_text("Message",allow_sending_without_reply=True, do_quote=True)` gives the following error:

`ValueError: "allow_sending_without_reply" and "reply_parameters" are mutually exclusive.`

### Thoughts to resolve them

I am not sure how we want to deal with it. On the one hand, this being by default only the case in groups will make it harder to spot when developing the bot, and the error is not helpful since the user didn't actually pass reply_parameters, the library did for them.

On the other hand this is a depreciated argument, so one could totally argue that an user who went through the effort of doing this could instead use

```
await update.message.reply_text(
                    "Message",
                    reply_parameters=ReplyParameters(update.message.message_id, allow_sending_without_reply=True)
                )
```

without much effort.

This comes from the group and my debugging efforts because I didnt understand the issue first either. https://t.me/pythontelegrambotgroup/779293?thread=779293

Bibo-Joshi:
closes #4807

### DIFF
--- a/changes/unreleased/4818.3VPDqqEWEhDCrTipFY8KKU.toml
+++ b/changes/unreleased/4818.3VPDqqEWEhDCrTipFY8KKU.toml
@@ -3,7 +3,7 @@ Correctly parse parameter ``allow_sending_without_reply`` in ``Message.reply_*``
 
 .. hint::
 
-   Using ``dict`` valued input for ``do_quote`` along with passing ``allow_sending_without_reply_to`` is not supported and will raise an error.
+   Using ``dict`` valued input for ``do_quote`` along with passing ``allow_sending_without_reply`` is not supported and will raise an error.
 """
 
 [[pull_requests]]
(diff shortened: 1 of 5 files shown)


changes/unreleased/4818.3VPDqqEWEhDCrTipFY8KKU.toml
```diff
@@ -0,0 +1,12 @@
+bugfixes = """
+Correctly parse parameter ``allow_sending_without_reply`` in ``Message.reply_*`` when used in combination with ``do_quote=True``.
+
+.. hint::
+
+   Using ``dict`` valued input for ``do_quote`` along with passing ``allow_sending_without_reply_to`` is not supported and will raise an error.
```
harshil21:
```suggestion
   Using ``dict`` valued input for ``do_quote`` along with passing ``allow_sending_without_reply`` is not supported and will raise an error.
```

PR #4825, opened 2025-06-13
Add support for Python 3.14 beta

harshil21:
Also adds pytz to the test dependency group since it was impossible to run the test suite without it.

Tested locally with uv's python 3.14, and with JIT enabled. Also confirmed that test official passes on 3.14 with the changes to annotations.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14.0-beta.2']
         os: [ubuntu-latest, windows-latest, macos-latest]
       fail-fast: False
     steps:
@@ -62,8 +62,6 @@ jobs:
 
           # Test the rest
           export TEST_WITH_OPT_DEPS='true'
-          # need to manually install pytz here, because it's no longer in the optional reqs
-          pip install .[all] pytz
           # `-n auto --dist worksteal` uses pytest-xdist to run tests on multiple CPU
           # workers. Increasing number of workers has little effect on test duration, but it seems
           # to increase flakyness.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,9 +36,11 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
 ]
 dependencies = [
     "httpx >=0.27,<0.29",
+    "httpcore >=1.0.9; python_version >= '3.14'"  # httpx needs an update so 3.14 support works
 ]
 
 [project.urls]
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -36,9 +36,11 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
 ]
 dependencies = [
     "httpx >=0.27,<0.29",
+    "httpcore >=1.0.9; python_version >= '3.14'"  # httpx needs an update so 3.14 support works
```
Bibo-Joshi: Is there a httpx issue we can link to here? That would make it easier to know when we can remove this

.github/workflows/unit_tests.yml
```diff
@@ -59,11 +58,10 @@ jobs:
           TO_TEST="test_no_passport.py or test_datetime.py or test_defaults.py or test_jobqueue.py or test_applicationbuilder.py or test_ratelimiter.py or test_updater.py or test_callbackdatacache.py or test_request.py"
           pytest -v --cov -k "${TO_TEST}" --junit-xml=.test_report_no_optionals_junit.xml
           opt_dep_status=$?
-
+          
+          pip install .[all]
```
Bibo-Joshi: can you please move this below the line "test the rest"? During review I thought `pip install .[all]` for some reason should now belong to the "test without dependencies" part, which was confusing

changes/unreleased/4825.R7wiTzvN37KAV656s9kfnC.toml
```diff
@@ -0,0 +1,5 @@
+other = "Add support for Python 3.14 beta"
```
Bibo-Joshi:
```suggestion
other = "Add Python 3.14 Beta To Test Matrix. *Python 3.14 is not officially supported by PTB yet!*"
```

pyproject.toml
```diff
@@ -123,6 +129,7 @@ docs = [
     "sphinx-inline-tabs==2023.4.21",
     # Temporary. See #4387
     "sphinx-build-compatibility @ git+https://github.com/readthedocs/sphinx-build-compatibility.git@58aabc5f207c6c2421f23d3578adc0b14af57047",
+    "pydantic @ git+https://github.com/pydantic/pydantic ; python_version >= '3.14'"  # Temporary for 3.14 support, we're waiting for a new pydantic version
```
Bibo-Joshi: what is this needed for? which lib uses pydantic?

pyproject.toml
```diff
@@ -274,6 +281,10 @@ omit = [
     "tests/",
     "src/telegram/__main__.py"
 ]
+# Relevant for python 3.14: https://github.com/nedbat/coveragepy/issues/1983
```
Bibo-Joshi: can you please clarify what should be done with this addition once the linked issue is resolved? e.g. should we remove that setting?

Bibo-Joshi (changes requested):
Thanks for the PR!

pyproject.toml
harshil21: it's in the dependency tree of `chango`

Bibo-Joshi: righto :D would be worth a comment IMO for stupid people like me who don't recall their own dependencies

pyproject.toml
harshil21: turns out my lockfile just needed to be updated, fresh installations should have httpcore >= 1.0.9 by default.

pyproject.toml
harshil21: that issue is resolved and now we can remove that

Bibo-Joshi (approved):

PR #4826, opened 2025-06-13
Remove `furo-sphinx-search`, and fix some docs

harshil21:
Search as you type has not been working for a long time, we should probably enable the rtd addons - #4387 

There are also many warnings while building docs, looks like the strict flag doesn't work? This PR fixes a few at least, I can't seem to figure out why the others fail...

Also fixes a few docs, where methods of the `object` class were listed out.


docs/source/telegram.transactionpartnertelegramads.rst
```diff
@@ -4,4 +4,4 @@ TransactionPartnerTelegramAds
 .. autoclass:: telegram.TransactionPartnerTelegramAds
     :members:
     :show-inheritance:
-    :inherited-members: TransactionPartner
```
Bibo-Joshi, line 7: the docs of `TransactionPartnerTelegramAds` now also show the class variables of `TransactionPartner`. That's not desired IMO.

docs/source/conf.py
```diff
@@ -98,7 +97,6 @@
 
 # Show docstring for special members
 autodoc_default_options = {
-    "special-members": True,
```
Bibo-Joshi: was there a specific reason to move this to the individual classes? how did you check that no class was forgotten that currently documents special members?

harshil21: yeah for some reason we're getting all special methods defined in `object` like `__dir__` too. 

> how did you check that no class was forgotten that currently documents special members?

I just relied on build warnings, which would happen if a reference wasn't found. So I could've definitely missed a few

harshil21: Ok I found a better fix

Bibo-Joshi (approved):

PR #4827, opened 2025-06-13
Remove support for Python 3.9

harshil21:
To be merged after #4825 and after python 3.9 reaches EOL in October

A lot of find and replace along with copilot agent edits. 

Didn't use the new TypeGuard feature yet. `match` statements is also not used right now.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,7 +1,9 @@
 name: 'Mark & close stale questions'
 on:
-  schedule:
-    - cron: '42 2 * * *'
+  workflow_dispatch:
+# Currently disabled due to https://github.com/actions/stale/pull/1298
+#  schedule:
+#    - cron: '42 2 * * *'
 
 permissions: {}
 
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -20,6 +20,9 @@ jobs:
       matrix:
         python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
         os: [ubuntu-latest, windows-latest, macos-latest]
+        include:
+          - python-version: '3.14t'
+            os: ubuntu-latest
       fail-fast: False
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
--- a/README.rst
+++ b/README.rst
@@ -215,6 +215,12 @@ Noteworthy parts of ``python-telegram-bots`` API that are likely to cause issues
 * ``telegram.ext.BasePersistence``
 * all classes in the ``telegram.ext.filters`` module that allow to add/remove allowed users/chats at runtime
 
+Free threading
+~~~~~~~~~~~~~~
+
+While ``python-telegram-bot`` is tested to work with Python 3.14 free threading, we do not guarantee that
+PTB is thread-safe for all use cases. Please see issue `#4873 <https://github.com/python-telegram-bot/python-telegram-bot/issues/4873>`_ for more information.
+
 Contributing
 ------------
 
--- a/changes/unreleased/5006.hRJtYPL4nBeZ43DWqii68V.toml
+++ b/changes/unreleased/5006.hRJtYPL4nBeZ43DWqii68V.toml
@@ -1,0 +1,5 @@
+internal = "Add Freethreaded Python 3.14 to the Test Suite"
+[[pull_requests]]
+uid = "5006"
+author_uids = ["harshil21"]
+closes_threads = []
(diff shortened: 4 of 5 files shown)


harshil21:
So apparently python recommends using X | Y syntax but this fails in subtle ways and makes life harder in instances like these:

```python
>>> from typing import TypeAlias
>>> a: TypeAlias = "int"
>>> def this(arg: a | None):pass
... 
  File "<python-input-3>", line 1, in <module>
    def this(arg: a | None):pass
                  ~~^~~~~~
TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'
```

Now doing `def this(arg: "a | None"):pass` will work and appear to be fine. However this fails when you try to evaluate the forward reference, which is why test_official fails right now.

`TypeAlias` is actually deprecated in favor of `type` (py 3.12+). Using that will now not error on `def this(arg: a | None):pass`, but still fails `test_official`, because the types aren't actually evaluated:

``` python
type A = int | str
b = A | float
expected = int | float | str

b == expected  # False

b = A.__value__ | float  # Force evaluation
b == expected  # True
```

Anyway so for now it seems we'll have to settle and keep using Optional/Union in some places...

changes/unreleased/4825.R7wiTzvN37KAV656s9kfnC.toml
Bibo-Joshi, line 1: this should be removed - merge artifact?

changes/unreleased/4827.PBXyEEjvXz5sJbiDWkeGFX.toml
```diff
@@ -0,0 +1,5 @@
+other = "Remove support for Python 3.9"
```
Bibo-Joshi:
```suggestion
other = "Remove Support for Python 3.9"
```

.gitignore
```diff
@@ -98,6 +98,3 @@ Scripts/
 
 # environment manager:
 .mise.toml
-
-# Support for uv.lock will come in a future PR. See #4796
-uv.lock
```
Bibo-Joshi, line 103: also a merge artifact?

pyproject.toml
```diff
@@ -40,6 +39,7 @@ classifiers = [
 ]
 dependencies = [
     "httpx >=0.27,<0.29",
+    "httpcore >=1.0.9; python_version >= '3.14'"  # httpx needs an update so 3.14 support works
```
Bibo-Joshi:
```suggestion
    "httpcore >=1.0.9; python_version >= '3.14'"  # httpx doesn't pin this as of httpx 0.28.1
```

Bibo-Joshi (commented):
October is finally coming up 🥳 
I skimmed through the changes - the overwhelming part apparently is intruducing `|` :D A few files looked more interesting and I've left some comments :)
Thanks for keeping track of this!

harshil21:
Should be good to merge! Hopefully there aren't more subtle merge artifacts in critical files. I would expect the test suite to fail then.

tests/auxil/dummy_objects.py
```diff
@@ -99,6 +99,9 @@
     "int": 123456,
     "MenuButton": MenuButton(type="dummy_type"),
     "Message": make_message("dummy_text"),
+    # Really bad hack to get tests passing, but we need to significantly overhaul how shortcuts
+    # work, we should not invent a type checker!
+    "Message | bool": make_message("dummy_text"),
```
Bibo-Joshi: this bit is interesting :D can you specify which tests are failing otherwise and maybe elaborate what you mean by "overhauling shortcuts"?

.gitignore
```diff
@@ -100,4 +100,4 @@ Scripts/
 .mise.toml
 
 # Support for uv.lock will come in a future PR. See #4796
-uv.lock
+uv.lock
```
Bibo-Joshi: very nitpicky: if you add a trailing newline, the file will notbe changed by this PR at all 👼

tests/auxil/dummy_objects.py
harshil21: Didn't even remember that I added that. Essentially if you don't add that in, tests which expect a return type of Message | bool, e.g. `TestBotWithoutRequest.test_defaults_handling[ExtBot.stop_message_live_location] ` would fail.

I *think* the reason why I said it was a bad hack was because the rest of the dictionary didn't seem to deal with annotations but this one did. I might have been lazy and/or didn't have enough time to try and fix it properly. 

I don't know why I said we should overhaul the shortcuts, it might've just been my general frustration with the duplication or the way they're tested

Bibo-Joshi: mh, okay.

I get that the tests for defaults handling & shortcuts are rather elaborate in the setup and verbose to maintain …  with the current setup of how defaults & shortcuts work, I however don't really see much more elegant ways of covering that (see also my write-up from some time ago in the dev chat).

At least in the current state, I doubt the comment is of any use for future-us 😅 but I'm perfectly fine with keeping a commit about this being subobtimal.

there's a merge conflict now, btw

Issue #4845, opened 2025-07-03
Full Support for Bot API 9.1


aelkheir:
## Bot API 9.1

> [!IMPORTANT]
> - Comment here or in the dev group what you want to work on so we can assign you
> - Please copy the relevant part of the checklist from #4845 to your PR and make sure to cover everything :)
> - Please base your branch on the [api-9.1](https://github.com/python-telegram-bot/python-telegram-bot/tree/api-9.1) branch
> - If this is your first contribution, please be sure to read the [contribution guide](https://docs.python-telegram-bot.org/en/stable/contributing.html)
> - and feel free to reach out with any questions via https://t.me/pythontelegrambotchannel/105

#### Checklists

- [x] Added the class [ChecklistTask](https://core.telegram.org/bots/api#checklisttask) representing a task in a checklist. #4847 
- [x] Added the class [Checklist](https://core.telegram.org/bots/api#checklist) representing a checklist. @Bibo-Joshi #4848
- [x] Added the field checklist to the classes [Message](https://core.telegram.org/bots/api#message) and [ExternalReplyInfo](https://core.telegram.org/bots/api#externalreplyinfo), describing a checklist in a message. @Bibo-Joshi #4848 
- [x] Added the class [ChecklistTasksDone](https://core.telegram.org/bots/api#checklisttasksdone) and the field checklist_tasks_done to the class [Message](https://core.telegram.org/bots/api#message), describing a service message about status changes for tasks in a checklist (i.e., marked as done/not done). @Bibo-Joshi #4848
- [x] Added the class [ChecklistTasksAdded](https://core.telegram.org/bots/api#checklisttasksadded) and the field checklist_tasks_added to the class [Message](https://core.telegram.org/bots/api#message), describing a service message about the addition of new tasks to a checklist. @Bibo-Joshi  #4848

---

- [x] Added the class [InputChecklistTask](https://core.telegram.org/bots/api#inputchecklisttask) representing a task to add to a checklist. @aelkheir #4857
- [x] Added the class [InputChecklist](https://core.telegram.org/bots/api#inputchecklist) representing a checklist to create.  @aelkheir #4857 #4857
- [x] Added the method [sendChecklist](https://core.telegram.org/bots/api#sendchecklist), allowing bots to send a checklist on behalf of a business account. @aelkheir #4857
[...]

PR #4849, opened 2025-07-03
API 9.1 Gifts 

harshil21:
Part of #4845

### DIFF
--- a/tests/test_uniquegift.py
+++ b/tests/test_uniquegift.py
@@ -446,6 +446,30 @@ class TestUniqueGiftInfoWithoutRequest(UniqueGiftInfoTestBase):
         assert unique_gift_info.last_resale_star_count == self.last_resale_star_count
         assert unique_gift_info.next_transfer_date == self.next_transfer_date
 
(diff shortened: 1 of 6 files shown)


tests/test_uniquegift.py
```diff
@@ -410,6 +416,8 @@ class UniqueGiftInfoTestBase:
     origin = UniqueGiftInfo.UPGRADE
     owned_gift_id = "some_id"
     transfer_star_count = 10
+    last_resale_star_count = 5
+    next_transfer_date = dtm.datetime.now(tz=UTC).replace(microsecond=0)
```
Bibo-Joshi, line 420: please also add a new test `de_json_timezone_locatization` that makes sure that `next_transfer_data` is localized correctly :) there are already corresponding test cases in other classes that you can copy from

Bibo-Joshi (changes requested):
thanks very much for the PR! just one comment

Bibo-Joshi (approved):

Bibo-Joshi:
merge at will :) (poor [lucky?] Will …)

PR #4851, opened 2025-07-04
API 9.1 General

harshil21:
Part of #4845

### DIFF
--- a/src/telegram/_directmessagepricechanged.py
+++ b/src/telegram/_directmessagepricechanged.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python
+#
+# A library that provides a Python interface to the Telegram Bot API
+# Copyright (C) 2015-2025
+# Leandro Toledo de Souza <devs@python-telegram-bot.org>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser Public License for more details.
+#
+# You should have received a copy of the GNU Lesser Public License
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+"""This module contains an object that represents a Direct Message Price."""
+
+
+from typing import Optional
+
+from telegram._telegramobject import TelegramObject
+from telegram._utils.types import JSONDict
+
+
+class DirectMessagePriceChanged(TelegramObject):
+    """
+    Describes a service message about a change in the price of direct messages sent to a channel
+    chat.
+
+    .. versionadded:: NEXT.VERSION
+
+    Objects of this class are comparable in terms of equality. Two objects of this class are
+    considered equal, if their :attr:`are_direct_messages_enabled` are equal.
+
+    Args:
+        are_direct_messages_enabled (:obj:`bool`):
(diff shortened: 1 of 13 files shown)


src/telegram/_directmessagepricechanged.py
```diff
@@ -0,0 +1,72 @@
...
+        are_direct_messages_enabled: bool,
+        direct_message_star_count: Optional[int] = None,
+        *,
+        api_kwargs: Optional[JSONDict] = None,
+    ):
+        super().__init__(api_kwargs=api_kwargs)
+        self.are_direct_messages_enabled: bool = are_direct_messages_enabled
+        self.direct_message_star_count: Optional[int] = direct_message_star_count
+
+        self._id_attrs = (self.are_direct_messages_enabled,)
```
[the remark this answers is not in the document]
harshil21: it's optional but `StarAmount` does have it so i'll do it here too

tests/test_bot.py
```diff
@@ -2574,6 +2575,17 @@ async def make_assertion(url, request_data: RequestData, *args, **kwargs):
 
         await offline_bot.remove_chat_verification(1234)
 
+    async def test_get_my_star_balance(self, offline_bot, monkeypatch):
```
Bibo-Joshi, line 2578: did you have a specific reason to add this test in addition to the one below? At least at first glance I don't see anything that the below test doesn't already cover 🤔

tests/test_directmessagepricechanged.py
```diff
@@ -0,0 +1,81 @@
...
+        dmpc_dict = direct_message_price_changed.to_dict()
+        assert dmpc_dict["are_direct_messages_enabled"] == self.are_direct_messages_enabled
+        assert dmpc_dict["direct_message_star_count"] == self.direct_message_star_count
+
+    def test_equality(self, direct_message_price_changed):
+        dmpc1 = direct_message_price_changed
+        dmpc2 = DirectMessagePriceChanged(
+            are_direct_messages_enabled=self.are_direct_messages_enabled,
+            direct_message_star_count=self.direct_message_star_count,
+        )
```
Bibo-Joshi, line 73:
```suggestion
        )
        dmpc3 = something_thats_note_a_dmpc_instance
```

to ensure that comparing with a different class returns False :)

Bibo-Joshi (changes requested):
Thanks for the pr! two comments :)

tests/test_bot.py
harshil21: I just felt that we should have an offline test too and not only depend on the online tests. For e.g. the online test can flaky fail and xpass.

Bibo-Joshi: mh okay. I don't think we have that for any other method that we can test with the api. but I'll go along with it

aelkheir:
I'll have a look also.. just a bit latter in the evening :)

src/telegram/constants.py
aelkheir, line 1: `PollLimit(IntEnum).MAX_OPTION_NUMBER` needs updating to `12`

tests/test_directmessagepricechanged.py
```diff
@@ -0,0 +1,85 @@
...
+from tests.auxil.slots import mro_slots
+
+if TYPE_CHECKING:
+    from telegram._utils.types import JSONDict
+
+
+@pytest.fixture
+def direct_message_price_changed():
+    return DirectMessagePriceChanged(
+        are_direct_messages_enabled=True, direct_message_star_count=100
```
aelkheir:
```suggestion
        are_direct_messages_enabled=DirectMessagePriceChangedTestBase.are_direct_messages_enabled,
        direct_message_star_count=DirectMessagePriceChangedTestBase.direct_message_star_count,
```

to ensure that asserts in `test_to_dict()` come from the same source

aelkheir (changes requested):
Only a missing constants update, and a tiny nitpick 😸

PR #4869, opened 2025-07-19
Added .delete() shortcut for business messages

jainamoswal:


Bibo-Joshi:
Hi @jainamoswal may I ask what's your status here? do you need help with anything?

jainamoswal:
I'm still figuring out about adding the tests regarding the code I added. @Bibo-Joshi

aelkheir:
So there are a couple of things we'd want to test here.

- that calling `message.delete` actually calls the correct bot method depending on `message.business_connection_id`
- that we correctly wrap `message.id` in a `list`; in the case that `business_connection_id` is present.
- finally, test that the shortcut signature matches the bot's one, and correctly passes the shortcut'ed parameters.

The way i'd approach this is to `monkeypatch` one level higher than the bot methods. (i.e `request.post` for example). you can have a look at [test_business_methods.py](https://github.com/python-telegram-bot/python-telegram-bot/blob/675254746074745be8b3faa8c15d62f8083c1ebc/tests/test_business_methods.py) for examples on `monkeypatch`.

Then we'd also need to have two test message objects, possibly as a parameterized fixture, you can also search the codebase or `pytest` docs for those. once that is done, we can modify the test behavior and expectations based on the type of the message.

shoot me any questions if some parts aren't clear.

Bibo-Joshi:
@jainamoswal are you still interested in working on this? Do you need additional support with the tests?

Bibo-Joshi:
I will take this as a no. @aelkheir would you like to continue work on this, possibly after the api update is finished? :)

aelkheir:
Apologies @jainamoswal, something went wrong my side and i had to force push.

Bibo-Joshi (approved):
Thanks for supporting @aelkheir ! If there is nothing left from your side I'd say this is good to merge :)

PR #4878, opened 2025-07-26
Documentation Improvements

Bibo-Joshi:
Gathering smaller documentation improvements to merge before the next release


uv.lock
Bibo-Joshi, line 1: @harshil21 the uv.lock file is still marked as changed in this PR. I expect that this should be reverted?

harshil21: for this PR, yes

Issue #4871, opened 2025-07-21
Incorrect behaviour of `retries` in `network_retry_loop`


harshil21:
It seems that we increment `retries` for every loop iteration, even if the callback succeeded and there were no errors from the TG API: 

https://github.com/python-telegram-bot/python-telegram-bot/blob/c062712472a1476b3b6b082dddd8f35977b5ecf7/src/telegram/ext/_utils/networkloop.py#L149-L150

This conflicts with the logic of checking the number of retries done:

https://github.com/python-telegram-bot/python-telegram-bot/blob/c062712472a1476b3b6b082dddd8f35977b5ecf7/src/telegram/ext/_utils/networkloop.py#L131-L135

It also seems that we don't reset `retries = 0` when there are no errors returned.

noobyogi0010:
Hey @harshil21,
I would like to work on this issue, if no-one has already started.
Thanks!

harshil21:
Hey @noobyogi0010, you're welcome to work on this issue! Make sure to read the [contributing guidelines](https://github.com/python-telegram-bot/python-telegram-bot/blob/master/.github/CONTRIBUTING.rst).

noobyogi0010:
Thank you, will go through the contributing guidelines as well.

Bibo-Joshi:
@harshil21 Thanks for the catch harshil. I would nevertheless guess that the current behavior covers the current use cases of

a) infinite retries (run_polling)
b) try up to x times but exit on the first successful call

The observed issues would only be relevant for cases where action_cb actually returns True (for do retry)

I suggest double checking the existing usages of the retry loop. If we only have a &amp; b, then I suggest simplifying the logic to just return on the first non-failing call to action_cb

harshil21:
> The observed issues would only be relevant for cases where action_cb actually returns True (for do retry)

The typical case where `action_cb` returns `True` is in `run_polling`:

https://github.com/python-telegram-bot/python-telegram-bot/blob/c062712472a1476b3b6b082dddd8f35977b5ecf7/src/telegram/ext/_updater.py#L368

So we are always in the loop and don't break out of it, unless we are stopping the Application. So case b) is not covered. With the default `start_polling` timeout of 10 seconds, in a minute we poll at least 6 times, and thus `retries` is `6`.

This hasn't shown up because we hardcode `max_retries` to `-1`:
[...]

PR #4880, opened 2025-07-26
Improve Internal Logic for Network Retries

Bibo-Joshi:
My take on closing #4871 :) Maybe the diff explains my intentions better than my words :D

### DIFF
--- a/src/telegram/ext/_utils/networkloop.py
+++ b/src/telegram/ext/_utils/networkloop.py
@@ -54,15 +54,13 @@ async def network_retry_loop(
 ) -> None:
     """Perform a loop calling `action_cb`, retrying after network errors.
 
-    Stop condition for loop in case of `infinite_loop` is True:
-        * `is_running()` evaluates :obj:`False` or
+    Stop condition for loop in case of `infinite_loop` is :obj:`True`:
+        * `is_running()` evaluates :obj:`False`
         * or `stop_event` is set.
 
-    Stop condition for loop in case of `infinite_loop` is False:
-        * a call to `action_cb` succeeds or
-        * `is_running()` evaluates :obj:`False` or
-        * or `stop_event` is set or
-        * `max_retries` is reached.
+    Additional stop condition for loop in case of `infinite_loop` is :obj:`False`:
+        * a call to `action_cb` succeeds
+        * or `max_retries` is reached.
 
     Args:
         action_cb (:term:`coroutine function`): Network oriented callback function to call.
@@ -89,7 +87,7 @@ async def network_retry_loop(
             * 0: No retries.
             * > 0: Number of retries
 
-            Must be negative if `infinite_loop` is set to True..
+            Must be negative if `infinite_loop` is set to :obj:`True`.
         infinite_loop (:obj:`bool`): If :obj:`True`, the loop will run indefinitely until
             `is_running()` evaluates to :obj:`False` or `stop_event` is set. Otherwise, the loop
             will stop after a successful call to `action_cb`, or when `is_running()` evaluates to
--- a/src/telegram/ext/_utils/networkloop.py
+++ b/src/telegram/ext/_utils/networkloop.py
@@ -84,10 +84,13 @@ async def network_retry_loop(
 
             * < 0: Retry indefinitely.
             * 0: No retries.
-            * > 0: Number of retries
+            * > 0: Number of retries.
 
     """
-    infinite_loop = max_retries < 0
+    try:
+        infinite_loop = max_retries < 0
+    except TypeError as exc:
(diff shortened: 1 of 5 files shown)


src/telegram/ext/_utils/networkloop.py
```diff
@@ -50,14 +50,19 @@ async def network_retry_loop(
...
         * `is_running()` evaluates :obj:`False` or
-        * return value of `action_cb` evaluates :obj:`False`
         * or `stop_event` is set.
-        * or `max_retries` is reached.
+
+    Stop condition for loop in case of `infinite_loop` is False:
+        * a call to `action_cb` succeeds or
+        * `is_running()` evaluates :obj:`False` or
+        * or `stop_event` is set or
+        * `max_retries` is reached.
```
Poolitzer: Would avoid the doubling of conditions here
```suggestion
    Additional stop condition for loop in case of `infinite_loop` is False:
        * a call to `action_cb` succeeds or
        * `max_retries` is reached.
```

src/telegram/ext/_utils/networkloop.py
```diff
@@ -50,14 +50,19 @@ async def network_retry_loop(
...
     max_retries: int,
+    infinite_loop: bool,
 ) -> None:
     """Perform a loop calling `action_cb`, retrying after network errors.
 
-    Stop condition for loop:
+    Stop condition for loop in case of `infinite_loop` is True:
         * `is_running()` evaluates :obj:`False` or
-        * return value of `action_cb` evaluates :obj:`False`
         * or `stop_event` is set.
```
Poolitzer: double or here

src/telegram/ext/_utils/networkloop.py
```diff
@@ -82,15 +87,29 @@ async def network_retry_loop(
 
             * < 0: Retry indefinitely.
             * 0: No retries.
-            * > 0: Number of retries.
+            * > 0: Number of retries
+
+            Must be negative if `infinite_loop` is set to True..
```
Poolitzer: Do we need to honour this? Cant we just ignore it/set it to -1 when infinite_loop is set? If you feel like it we can always raise an error then but tbh I would just do it behind the scenes.

And we can change True to :obj:`True` and remove the double .

Poolitzer (approved):
I like the overall design, just three nitpicks

src/telegram/ext/_utils/networkloop.py
Bibo-Joshi: I'd like to keep it explicit to avoid surprising behavior :) especially since this is a rather important internal helper

Poolitzer: Just feels like additional work, if I already set infinite_loop to true, why do I then also need to set retries to infinite.

But sure, if you think it helps.

harshil21 (commented):
ohh nvm there is a guard against that combination of inputs. So it would work. Do tests need updating?

src/telegram/ext/_utils/networkloop.py
harshil21: yea it feels redundant

Bibo-Joshi: Alternatively, I can just do `infinite_retries = bool(max_retries < 0)`. @harshil21 you didn't much like that in the chat but maybe that has changed with this PR? :D

Bibo-Joshi:
>  Do tests need updating?

Except for coverage reasons, I don't see a need. this PR doesn't introduce any new functionality that would need testing. test for the existing network retry use cases are already in place :)

src/telegram/ext/_utils/networkloop.py
harshil21: Yeah that would be better 👍🏽

Bibo-Joshi: done :)

Bibo-Joshi:
welp, that took a bit of a detour, but ready for re-review now :)

src/telegram/ext/_utils/networkloop.py
```diff
@@ -85,12 +87,15 @@ async def network_retry_loop(
...
+    infinite_loop = max_retries < 0
     log_prefix = f"Network Retry Loop ({description}):"
     effective_is_running = is_running or (lambda: True)
 
     async def do_action() -> bool:
+        """Return value indicating whether the action was successful."""
         if not stop_event:
-            return await action_cb()
+            await action_cb()
+            return True
```
harshil21: why do we specifically return a bool even though the return value isn't being used anymore?

Bibo-Joshi: I initially thought that I needed the return value, but `action_cb_task.result()` raises on failure, so that's fine. thanks for the catch!

harshil21 (approved):
LGTM!

PR #4882, opened 2025-07-27
Httpx timeout improvements

Poolitzer:
As discussed, we think its appropriate to change the default timeout and drop the max_keepalive_connections

### DIFF
--- a/changes/unreleased/4882.JpgQAHHnCponMKStbC4JsG.toml
+++ b/changes/unreleased/4882.JpgQAHHnCponMKStbC4JsG.toml
@@ -1,4 +1,9 @@
-other = "Default httpx max_connections/pool size set to 256 to allow more requests by default, and dropping setting max_keepalive_connections to the same amount (so bigger pools don't have a lot of unused alive connections). If you manually build the httpx request, be aware that this also applies to you. If you want your own limits, you can set them via httpx_kwargs, as explained in https://www.python-httpx.org/advanced/resource-limits/"
+other = """
+Set the default connection pool size for ``HTTPXRequest`` to 256 to allow more concurrent requests by default. Drop the ``httpx`` parameter `max_keepalive_connections`. This way, the ``httpx`` default of 20 is used, leading to a smaller number of idle connections in large connection pools.
+
+.. hint:
+   If you manually build the ``HTTPXRequest`` objects, please be aware that these changes also applies to you. Kindly double check your settings. To specify custom limits, you can set them via the parameter ``httpx_kwargs`` of ``HTTPXRequest``. See also  `the httpx documentation <https://www.python-httpx.org/advanced/resource-limits/>`__ for more details on these settings."
+"""
 [[pull_requests]]
 uid = "4882"
 author_uid = "Poolitzer"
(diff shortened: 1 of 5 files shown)


changes/unreleased/4882.JpgQAHHnCponMKStbC4JsG.toml
```diff
@@ -0,0 +1,5 @@
+other = "Httpx timeout improvements"
```
Bibo-Joshi: please give a short explanation of the change here and explain what users need to change if they manually build `HTTPXRequest` objects.

Poolitzer:
> Please double check the implementations of (Ext)Bot and ApplicationBuilder. I'm pretty sure that the arguments to HTTPXRequest need adaption for the get_updates_request instances

I thought about this, but came to the conclusion that that shouldn't be needed. They set the connection to 1, as expected, and since there is just one connection in the pool, it shouldn't matter that max_keepalive_connections is set to 20 now, since well. There is only one over all :D

src/telegram/request/_httpxrequest.py
```diff
@@ -48,7 +48,13 @@ class HTTPXRequest(BaseRequest):
 
     Args:
         connection_pool_size (:obj:`int`, optional): Number of connections to keep in the
-            connection pool. Defaults to ``1``.
+            connection pool. Defaults to ``256``.
+
+            .. versionchanged:: NEXT.VERSION
+                Set the default to ``256``.
+                Stopped applying to ``httpx.Limits.max_keepalive_connections``. Now only applys to
```
harshil21:
```suggestion
                Stopped applying to ``httpx.Limits.max_keepalive_connections``. Now only applies to
```

changes/unreleased/4882.JpgQAHHnCponMKStbC4JsG.toml
```diff
@@ -0,0 +1,5 @@
+other = "Default httpx max_connections/pool size set to 256, and dropping setting max_keepalive_connections to the same amount. If you manually build the httpx request, be aware that this also applies to you. If you want your own limits, you can set them via httpx_kwargs, as explained in https://www.python-httpx.org/advanced/resource-limits/"
```
harshil21: Do we want to mention the motivation behind the change too?

Bibo-Joshi:
> > Please double check the implementations of (Ext)Bot and ApplicationBuilder. I'm pretty sure that the arguments to HTTPXRequest need adaption for the get_updates_request instances
> 
> I thought about this, but came to the conclusion that that shouldn't be needed. They set the connection to 1, as expected, and since there is just one connection in the pool, it shouldn't matter that max_keepalive_connections is set to 20 now, since well. There is only one over all :D

https://github.com/python-telegram-bot/python-telegram-bot/blob/675254746074745be8b3faa8c15d62f8083c1ebc/src/telegram/_bot.py#L341-L344

Here in line 342 I expect the connection pool to be 1 for the `get_updates` request instance

https://github.com/python-telegram-bot/python-telegram-bot/blob/675254746074745be8b3faa8c15d62f8083c1ebc/src/telegram/ext/_applicationbuilder.py#L232-L239

Here in line 256 we duplicate the default value 256. We _could_ try to avoid that by not passing `connection_pool_size` in case `DefaultValue.get_value(getattr(self, f"{prefix}connection_pool_size")) is None`, though that may be unnecessary complication …

changes/unreleased/4882.JpgQAHHnCponMKStbC4JsG.toml
```diff
@@ -0,0 +1,5 @@
+other = "Default httpx max_connections/pool size set to 256 to allow more requests by default, and dropping setting max_keepalive_connections to the same amount (so bigger pools don't have a lot of unused alive connections). If you manually build the httpx request, be aware that this also applies to you. If you want your own limits, you can set them via httpx_kwargs, as explained in https://www.python-httpx.org/advanced/resource-limits/"
```
Bibo-Joshi:
```suggestion
other = """
Set the default connection pool size for ``HTTPXRequest`` to 256 to allow more concurrent requests by default. Drop the ``httpx`` parameter `max_keepalive_connections`. This way, the ``httpx`` default of 20 is used, leading to a smaller number of idle connections in large connection pools.

.. hint:
   If you manually build the ``HTTPXRequest`` objects, please be aware that these changes also applies to you. Kindly double check your settings. To specify custom limits, you can set them via the parameter ``httpx_kwargs`` of ``HTTPXRequest``. See also  `the httpx documentation <https://www.python-httpx.org/advanced/resource-limits/>`__ for more details on these settings."
"""
```

I'm absolutely not sorry for my standards on documentation and formatting 🤡

Bibo-Joshi (changes requested):

Bibo-Joshi:
Thank you for your patience ❤️

PR #4884, opened 2025-07-28
Add copilot instructions and setup steps

harshil21:
Spinoff from #4875

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,28 @@
+This is a python project which is a wrapper for the Telegram Bot API. Please read the contributing
+guidelines mentioned in .github/CONTRIBUTING.rst to know how to contribute to this project.
+
+### Development Environment:
+
+Your development environment is set up using `uv`, a tool for managing Python environments and dependencies.
+Your environment has all extra dependencies and groups installed, on Python 3.13. Please continue using `uv` for managing your development environment,
+and for any scripts or tools you need to run.
+
+Some example commands on `uv`:
+- `uv sync --all-extras --all-groups` to install all dependencies and groups required by the project.
+- `uv run -p 3.14 --all-groups --all-extras tests/` to run tests on a specific Python version. Please use the `-p` flag often.
+- `uv pip install <package>` to install a package in the current environment.
+
+If uv is somehow not available, you can install it using `pip install uv`.
+
+### Repository Structure:
+
+The repository follows a standard structure for Python projects. Here are some key directories and files:
+
+- `src/`: This directory contains the main source code for the project.
+- `tests/`: This directory contains test cases for the project.
+- `pyproject.toml`: This file contains the project metadata and dependencies.
+- `.github/`: This directory contains GitHub-specific files, including workflows and issue templates.
+
+### Pull Requests:
+
+When you create a pull request, please also add the appropriate labels to it.
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,5 +1,6 @@
 This is a python project which is a wrapper for the Telegram Bot API. Please read the contributing
-guidelines mentioned in .github/CONTRIBUTING.rst to know how to contribute to this project.
+guidelines mentioned in .github/CONTRIBUTING.rst to know how to contribute to this project. The 
+README.rst file lists the features and usage of the project.
 
 ### Development Environment:
 
@@ -22,6 +23,16 @@ The repository follows a standard structure for Python projects. Here are some k
 - `tests/`: This directory contains test cases for the project.
 - `pyproject.toml`: This file contains the project metadata and dependencies.
 - `.github/`: This directory contains GitHub-specific files, including workflows and issue templates.
+
+
+### Things to keep in mind while coding:
+
+- Ensure that your code is properly and fully typed. All your code should be compatible from
+  Python 3.9 to 3.14. Don't use the `typing_extensions` module.
+- Read the stability guide mentioned at docs/source/stability_policy.rst to understand if your changes
+  are breaking or incompatible.
+- Try to make sure your code is asyncio-friendly and thread-safe.
+- Run `uv run pre-commit` to run pre-commit hooks before committing your changes.
 
 ### Pull Requests:
 
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -32,7 +32,8 @@ The repository follows a standard structure for Python projects. Here are some k
 - Read the stability guide mentioned at docs/source/stability_policy.rst to understand if your changes
   are breaking or incompatible.
 - Try to make sure your code is asyncio-friendly and thread-safe.
-- Run `uv run pre-commit` to run pre-commit hooks before committing your changes.
+- Run `uv run pre-commit` to run pre-commit hooks before committing your changes, but after `git add`ing them.
+- Make sure you always test your changes. Either update or write new tests in the `tests/` directory.
 
 ### Pull Requests:
 
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -1,0 +1,42 @@
+name: "Copilot Setup Steps"
+
+# Automatically run the setup steps when they are changed to allow for easy validation, and
+# allow manual testing through the repository's "Actions" tab
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - .github/workflows/copilot-setup-steps.yml
+
+jobs:
+  # The job MUST be called `copilot-setup-steps` or it will not be picked up by Copilot.
+  copilot-setup-steps:
+    runs-on: ubuntu-latest
+
+    # Set the permissions to the lowest permissions possible needed for your steps.
+    # Copilot will be given its own token for its operations.
+    permissions:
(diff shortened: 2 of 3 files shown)


.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,42 @@
+name: "Copilot Setup Steps"
+
```
Bibo-Joshi, line 4: I'd like some general comment here about what this is for, ideally including a link to the corresponding github docs

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,42 @@
...
+    permissions:
+      # If you want to clone the repository as part of your setup steps, for example to install dependencies, you'll need the `contents: read` permission. If you don't clone the repository in your setup steps, Copilot will do this for you automatically after the steps complete.
+      contents: write  # TEST: `uv` wasn't installed after the workflow succeded, maybe this is needed?
+      pull-requests: write  # So copilot can add labels to the PR
+
+    # You can define any steps you want, and they will run before the agent starts.
+    # If you do not check out your code, Copilot will do this for you.
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
Bibo-Joshi: can you pin specific shas here (and in the user `uses:`) like in the other workflows? I don't quite understand why zizmore didn't run for this PR 🤔

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,42 @@
...
+          version: "0.8.3"
+
+      - name: Set up Python
+        run: uv python install 3.13
+
+      - name: Install the project
+        run: uv sync --all-extras --all-groups
+    
+      - name: Install linting dependencies
+        run: uv pip install black isort mypy flake8 ruff
```
Bibo-Joshi: do you want to pin the versions used in pre-commit.yaml? or simply install pre-commit hooks?

changes/unreleased/4884.ADcdCj6GrMaKffYSLzWdsL.toml
```diff
@@ -0,0 +1,5 @@
+other = "Add copilot instructions and setup steps"
```
Bibo-Joshi:
```suggestion
internal = "Add Copilot Instructions and Setup Steps"
```

.github/copilot-instructions.md
Bibo-Joshi, line 1: This reads like a sane initial setup. Things that come to my mind that could be added:

- stability guide
- info on telegram package structure (tg vs ext)
- "everything should have proper types"
- asyncio friendly code
- free-threading friendly code?

but these things can ofc also be added when needed

.github/workflows/copilot-setup-steps.yml
harshil21: pre-commit should be installed when including `--all-groups`. I'll pin the versions though. I'm not sure why we don't have a another group of linting dependencies?

.github/workflows/copilot-setup-steps.yml
harshil21: what do you mean by "user uses:"? If you meant the `uv` uses:, that already uses the latest point release

.github/copilot-instructions.md
```diff
@@ -23,6 +24,16 @@ The repository follows a standard structure for Python projects. Here are some k
 - `pyproject.toml`: This file contains the project metadata and dependencies.
 - `.github/`: This directory contains GitHub-specific files, including workflows and issue templates.
 
+
+### Things to keep in mind while coding:
+
+- Ensure that your code is properly and fully typed. All your code should be compatible from
+  Python 3.9 to 3.14. Don't use the `typing_extensions` module.
```
Bibo-Joshi, line 31:
```suggestion
  Python 3.9 onward. Don't use the `typing_extensions` module.
```
limiting to <=3.14 doesn't seem to bring any benefits 🤔

.github/copilot-instructions.md
```diff
@@ -23,6 +24,16 @@ The repository follows a standard structure for Python projects. Here are some k
...
+- Read the stability guide mentioned at docs/source/stability_policy.rst to understand if your changes
+  are breaking or incompatible.
+- Try to make sure your code is asyncio-friendly and thread-safe.
+- Run `uv run pre-commit` to run pre-commit hooks before committing your changes.
```
Bibo-Joshi: just curious: this is right? I would have expected something like `uv run pre-commit run -a`.

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,44 @@
...
+    permissions:
+      # If you want to clone the repository as part of your setup steps, for example to install dependencies, you'll need the `contents: read` permission. If you don't clone the repository in your setup steps, Copilot will do this for you automatically after the steps complete.
+      contents: write  # TEST: `uv` wasn't installed after the workflow succeded, maybe this is needed?
+      pull-requests: write  # So copilot can add labels to the PR
+
+    # You can define any steps you want, and they will run before the agent starts.
+    # If you do not check out your code, Copilot will do this for you.
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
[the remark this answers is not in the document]
Bibo-Joshi: was there a specific reason to close this as  won't fix? AFAIK we use the suggested setting in all other places

.github/workflows/copilot-setup-steps.yml
Bibo-Joshi: since the tools runn in pre-commit, they are not strictly a requirement for development. you can just install if it's convenient for you … I'm okay with adding them again in the renovate PR if renovate is able to keep them in sync

harshil21: >  you can just install if it's convenient for you 

considering the ruff check has auto fix disabled here, one almost always would have to run ruff separately again on their machine

> … I'm okay with adding them again in the renovate PR if renovate is able to keep them in sync

yeah I think it would be able to. I'll add it on there

.github/workflows/copilot-setup-steps.yml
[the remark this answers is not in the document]
harshil21: I just didn't know what it was saying, I'll fix it now that I actually read through it

.github/copilot-instructions.md
harshil21: I want to specifically mention that since I don't think all LLM's know of 3.14 specially because of their knowledge cutoff if they don't search the internet. It might just end up testing till 3.12 or something since that's the most popular version right now

.github/copilot-instructions.md
harshil21: `uv run pre-commit` will only check the `git add`ed files. I don't think it's necessary to run on all files. I'll mention the git added part in the instructions though

.github/copilot-instructions.md
Bibo-Joshi: I see. Would it instead be better to say something like "all Python versions currently supported according to https://devguide.python.org/versions/"?

harshil21: According to https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#writing-effective-repository-custom-instructions having links in the instructions isn't a great idea

Bibo-Joshi: sad. but okay, let's stick with 3.14 then 👍

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,46 @@
...
+  # The job MUST be called `copilot-setup-steps` or it will not be picked up by Copilot.
+  copilot-setup-steps:
+    runs-on: ubuntu-latest
+
+    # Set the permissions to the lowest permissions possible needed for your steps.
+    # Copilot will be given its own token for its operations.
+    permissions:
+      # If you want to clone the repository as part of your setup steps, for example to install dependencies, you'll need the `contents: read` permission. If you don't clone the repository in your setup steps, Copilot will do this for you automatically after the steps complete.
+      contents: write  # TEST: `uv` wasn't installed after the workflow succeded, maybe this is needed?
+      pull-requests: write  # So copilot can add labels to the PR
```
Bibo-Joshi: there is another security finding here - I don't understand why it doesn't show in the PR as comment … https://github.com/python-telegram-bot/python-telegram-bot/security/code-scanning/120

Looking at https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents it does indeed look like contents-write should be avoided. I don't quite understand your comment either, tbh 😅

harshil21: > I don't quite understand your comment either

So what happened was the uv was installed in the environment at the end of that workflow but when copilot tried to access it via `uv`, there was no uv binary to be found anywhere in the filesystem. So I thought changing this to "write" might fix that. Probably not, but it might be worth a try?

Bibo-Joshi: Well, if you can confirm that this helps, I'm open to adding it. My understanding however is that `contents: write` means that the workflow is allowed to make changes in the git repo itself, not in the workflow-environment.

harshil21: I don't think I can tell @copilot to edit a file in this branch.. which means I don't have a way of testing it here. So I'll just revert for now and see what happens

Bibo-Joshi (approved):
Thanks for your patience :) if there's nothing left from your side I can merge

PR #4890, opened 2025-07-31
Add `uv.lock`

### LINKED ISSUE #4796: Introduce a Dependency Lock File for Development

Now that [`uv`](https://docs.astral.sh/uv/) is well known and established in python community and adopted by many projects, we should consider using some more uv features in our repository.

Here's what we can initially change:

1. The CI for unit tests can use [`uv` built python's](https://docs.astral.sh/uv/guides/install-python/#installing-python), along with using uv commands to setup the job and install dependencies.
2. Commiting the [`uv.lock`](https://docs.astral.sh/uv/guides/projects/#uvlock) file to the repo to increase reproduciblity across different machines. 
3. Using `uv build` instead of the `build` package in the CI.
4. Using [`uv publish`](https://docs.astral.sh/uv/guides/package/#publishing-your-package) to release to pypi. They mention that it works through GH actions too, but I don't know how it plays with our specific setup, maybe @Bibo-Joshi can determine if it's a good enough fit.
5. In the future, we could try using the uv [build-backend](https://docs.astral.sh/uv/configuration/build-backend/#the-uv-build-backend) too.

Another idea I had would be to provide a "demo" app, so one could do something like:

`uvx python-telegram-bot echobot`

and this would run echobot.py (we'd have to read the bot token from stdin or env var). 

Not using uv and staying outdated

### Additional context

I don't think it's worth replacing all mentions of `pip install` with something like `uv add`, but I think we should at least acknowledge the existence of `uv` in the readme/docs somewhere to encourage more adoption. What do you think?

harshil21:
Closes #4796 

I ran `uv lock --upgrade` so we should have all the latest dependencies,

Ideally I want this merged before #4887 so I can check if Renovate detects it and test something like downgrading a dependency and checking if Renovate wants to upgrade it.

### DIFF
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -87,7 +87,7 @@ Here's how to make a one-off code change.
 
      - Documenting types of global variables and complex types of class members can be done using the Sphinx docstring convention.
 
-   -  In addition, PTB uses some formatting/styling and linting tools in the pre-commit setup. Some of those tools also have command line tools that can help to run these tools outside of the pre-commit step. If you'd like to leverage that, please have a look at the `pre-commit config file`_ for an overview of which tools (and which versions of them) are used. For example, we use `Black`_ for code formatting. Plugins for Black exist for some `popular editors`_. You can use those instead of manually formatting everything.
+   -  In addition, PTB uses some formatting/styling and linting tools in the pre-commit setup. Some of those tools also have command line tools that can help to run these tools outside of the pre-commit step. If you'd like to leverage that, please have a look at the `pre-commit config file`_ for an overview of which tools (and which versions of them) are used. For example, we use `Ruff`_ for linting and formatting.
 
    - Please ensure that the code you write is well-tested and that all automated tests still pass. We
      have dedicated an `testing page`_ to help you with that.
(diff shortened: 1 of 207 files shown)


changes/unreleased/4890.oCzELerGo8dqPygShWMV9S.toml
```diff
@@ -0,0 +1,5 @@
+documentation = "Add `uv.lock`"
```
Bibo-Joshi:
```suggestion
internal = "Add ``uv.lock``"
```

I think it would be wise to loose 1 or 2 sentences here on what this is intended for

Bibo-Joshi (commented):
> I'm personally strongly in favor of using pylock.toml as language wide standard rather than uv.lock as tool-specific file. If it's important to us to introduce this lock file soon and we don't want to wait for dependabot to add support, I'd be open to using uv.lock as stand-in. In that case _I strongly suggest to open a follow-up issue for switching to pylock.toml once all our tools support it._

https://github.com/python-telegram-bot/python-telegram-bot/issues/4796#issuecomment-2905716085

Could you please create that ticket and in it link to relevant threads of tools that might be revelant to us (dependabot, uv, pip, renovate)? Also link to this PR :)

changes/unreleased/4890.oCzELerGo8dqPygShWMV9S.toml
```diff
@@ -1,4 +1,4 @@
-documentation = "Add `uv.lock`"
+internal = "Add a ``uv.lock`` lockfile for reproducible environments when installing ``python-telegram-bot`` with ``uv``."
```
Bibo-Joshi: Sorry for being nitpicky, but this is very misleading IMO. this reads like this is a user-facing change, which it must not be. It should say something like "when setting up the development environment using ``uv``" or similar.

In fact this made me aware that an important change is missing: Please exclude uv.lock from the distribution package by adding to the exclusion list in `pyproject.toml` line 149 (`tool.hatch.build.targets.sdist` → `exclude = …`)

harshil21: > It should say something like "when setting up the development environment using `uv`" or similar

How about "Add and use a ``uv.lock`` lockfile when setting up the development environment using ``uv``."?

Bibo-Joshi (approved):
Thanks for your patience! If there's nothing left from your side I can merge :)

PR #4918, opened 2025-08-17
Feat: Added can_manage_dm and paid post

Poolitzer:

### DIFF
--- a/src/telegram/_message.py
+++ b/src/telegram/_message.py
@@ -628,6 +628,10 @@ class Message(MaybeInaccessibleMessage):
             of a channel has changed.
 
             .. versionadded:: 22.3
+        is_paid_post (:obj:`bool`, optional): :obj:`True`, if the message is a paid post. Note that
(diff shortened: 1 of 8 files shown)


src/telegram/_message.py
```diff
@@ -628,6 +628,10 @@ class Message(MaybeInaccessibleMessage):
             of a channel has changed.
 
             .. versionadded:: 22.3
+        is_paid_post (:obj:`bool`, optional): :obj:`True`, if the message is a paid post. Note that
+            such posts must not be deleted for 24 hours to receive the payment and can't be edited.
```
Bibo-Joshi, line 634: do we need/want a constant for this?

Poolitzer: hmmm I think its a good call. Edit: When implementing, I realized I dont think there is a lot of positive to force this into an enum

Bibo-Joshi: okay, let's skip, then.

Bibo-Joshi (approved):

PR #4953, opened 2025-09-15
Renovate: No README updates, label behaviour change, automerge lockfiles

harshil21:
Tries to fix the problems in #4938. 

I removed updating the READMEs for now. I think we'll just manually do it for now until we can figure out why that happened on all PRs and why was it only the APScheduler one which changed. 

@Bibo-Joshi I think the other PR's didn't automerge because only you can merge to master? Maybe add an exclusion to that rule and then the other PR's will start merging?

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -8,7 +8,7 @@
   ],
 
   // Add pull request labels:
-  "labels": ["dependencies"],
+  "labels": ["ci-cd"],
 
   // Bump even patch versions:
   "bumpVersion": "patch",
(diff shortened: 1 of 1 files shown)


Bibo-Joshi (approved):
Thanks for the PR!
Maybe the aiolimiter issue is because the pre-commit config is currently not matching pyproject.toml 🤔 anyway, updating readme manually for now sounds okay.
Thanks also for the hint about the write permissions. I'll add those once I can see how this PR will affect the other open ones 😅

PR #4973, opened 2025-09-24
Fix Handling of Infinite Bootstrap Retries

### LINKED ISSUE #4966: starting the bot with bootstrap_retries -1 fails startup
### Steps to Reproduce

Use the below script with python-telegram-bot 22.4

``` python
import logging
import os

from telegram import Update
from telegram.ext import Application, ApplicationBuilder, CommandHandler, ContextTypes


LOGFORMAT = "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
logging.basicConfig(
    level=logging.DEBUG,
    format=LOGFORMAT,
)

# logger.setLevel(logging.DEBUG)

token = os.environ.get('TELEGRAM_TOKEN')

# logger.debug("Creating telegram application")

application = Application.builder().token(token).build()


async def start(update: Update, context: ContextTypes.DEFAULT_TYPE):
    await context.bot.send_message(
        chat_id=update.effective_chat.id, text="I'm a bot, please talk to me!"
    )


start_handler = CommandHandler("start", start)
application.add_handler(start_handler)
application.run_polling(
    bootstrap_retries=-1,
)

```

The key is `bootstrap_retries=-1` - which causes startup to be retried forever (even though the action already worked).

### Expected behaviour

the application starts, `/start` responds with the expected response.

This worked with 22.3 - and fails with 22.4 - running into an endless loop.

### Actual behaviour

Endless loop - of "This application is already initialized".
The bot doesn't respond to any command (it's never starting to poll)

### Operating System

linux

```shell
python-telegram-bot 22.4 (2023.7-7837-g16117e5b6)
Bot API 9.2
Python 3.13.7 (main, Aug 28 2025, 17:07:40) [Clang 20.1.4 ]
```

### Relevant log output

```python
2025-09-21 08:56:28,844 - telegram.ext.ExtBot - DEBUG - Set Bot API URL: https://api.telegram.org/bot546065661:AAGSqjM2F3Zajs_HpiZY5LAYfD3AWmaGM3Y
2025-09-21 08:56:28,845 - telegram.ext.ExtBot - DEBUG - Set Bot API File URL: https://api.telegram.org/file/bot546065661:AAGSqjM2F3Zajs_HpiZY5LAYfD3AWmaGM3Y
2025-09-21 08:56:28,845 - __main__ - INFO - Starting polling
2025-09-21 08:56:28,845 - asyncio - DEBUG - Using selector: EpollSelector
2025-09-21 08:56:28,845 - telegram.ext - DEBUG - Network Retry Loop (Bootstrap Initialize Application): Starting
2025-09-21 08:56:28,845 - telegram.ext.ExtBot - DEBUG - Calling Bot API endpoint `getMe` with parameters `{}`
2025-09-21 08:56:28,848 - httpcore.connection - DEBUG - connect_tcp.started host='api.telegram.org' port=443 local_address=None timeout=5.0 socket_options=None
2025-09-21 08:56:28,886 - httpcore.connection - DEBUG - connect_tcp.complete return_value=<httpcore._backends.anyio.AnyIOStream object at 0x7f7f2397a510>
2025-09-21 08:56:28,887 - httpcore.connection - DEBUG - start_tls.started ssl_context=<ssl.SSLContext object at 0x7f7f23956210> server_hostname='api.telegram.org' timeout=5.0
2025-09-21 08:56:28,927 - httpcore.connection - DEBUG - start_tls.complete return_value=<httpcore._backends.anyio.AnyIOStream object at 0x7f7f23972490>
2025-09-21 08:56:28,927 - httpcore.http11 - DEBUG - send_request_headers.started request=<Request [b'POST']>
2025-09-21 08:56:28,928 - httpcore.http11 - DEBUG - send_request_headers.complete
2025-09-21 08:56:28,928 - httpcore.http11 - DEBUG - send_request_body.started request=<Request [b'POST']>
2025-09-21 08:56:28,928 - httpcore.http11 - DEBUG - send_request_body.complete
2025-09-21 08:56:28,928 - httpcore.http11 - DEBUG - receive_response_headers.started request=<Request [b'POST']>
2025-09-21 08:56:28,965 - httpcore.http11 - DEBUG - receive_response_headers.complete return_value=(b'HTTP/1.1', 200, b'OK', [(b'Server', b'nginx/1.18.0'), (b'Date', b'Sun, 21 Sep 2025 06:56:28 GMT'), (b'Content-Type', b'application/json'), (b'Content-Length', b'261'), (b'Connection', b'keep-alive'), (b'Strict-Transport-Security', b'max-age=31536000; includeSubDomains; preload'), (b'Access-Control-Allow-Origin', b'*'), (b'Access-Control-Allow-Methods', b'GET, POST, OPTIONS'), (b'Access-Control-Expose-Headers', b'Content-Length,Content-Type,Date,Server,Connection')])
2025-09-21 08:56:28,966 - httpx - INFO - HTTP Request: POST https://api.telegram.org/bot546065661:AAGSqjM2F3Zajs_HpiZY5LAYfD3AWmaGM3Y/getMe "HTTP/1.1 200 OK"
2025-09-21 08:56:28,967 - httpcore.http11 - DEBUG - receive_response_body.started request=<Request [b'POST']>
2025-09-21 08:56:28,967 - httpcore.http11 - DEBUG - receive_response_body.complete
2025-09-21 08:56:28,967 - httpcore.http11 - DEBUG - response_closed.started
2025-09-21 08:56:28,967 - httpcore.http11 - DEBUG - response_closed.complete
2025-09-21 08:56:28,968 - telegram.ext.ExtBot - DEBUG - Call to Bot API endpoint `getMe` finished with return value `{'id': 546065661, 'is_bot': True, 'first_name': 'freqtrad_test', 'username': 'xmatt_freq_test_bot', 'can_join_groups': True, 'can_read_all_group_messages': False, 'supports_inline_queries': False, 'can_connect_to_business': False, 'has_main_web_app': False}`
2025-09-21 08:56:28,968 - telegram.ext.ExtBot - DEBUG - This Bot is already initialized.
2025-09-21 08:56:29,969 - telegram.ext.Application - DEBUG - This Application is already initialized.
2025-09-21 08:56:30,971 - telegram.ext.Application - DEBUG - This Application is already initialized.
2025-09-21 08:56:31,972 - telegram.ext.Application - DEBUG - This Application is already initialized.

# Pressed ctrl+c here - it would continue like this forever!

2025-09-21 08:56:32,355 - telegram.ext.Application - DEBUG - Application received stop signal. Shutting down.
2025-09-21 08:56:32,356 - httpcore.connection - DEBUG - close.started
2025-09-21 08:56:32,356 - httpcore.connection - DEBUG - close.complete
2025-09-21 08:56:32,356 - telegram.ext.ExtBot - DEBUG - This Bot is already shut down. Returning.
2025-09-21 08:56:32,356 - telegram.ext.Updater - DEBUG - Shut down of Updater complete
```

### Additional Context

This is most likely caused by failing to break out of the below loop if boostrap_retries is set to -1 (which sets infinite_loop to True).

Because of this - the bot fails to exit `network_retry loop()` even if the actual action failed.
I think expected behavior here is to have the bot to retry indefinitely _if the action fails_ - but not retry indefinitely independent of the outcome of the action.

https://github.com/python-telegram-bot/python-telegram-bot/blob/14f8e89cef8199cb2665d13cba0ca9c7d8886b21/src/telegram/ext/_utils/networkloop.py#L120-L126

---

https://github.com/python-telegram-bot/python-telegram-bot/blob/14f8e89cef8199cb2665d13cba0ca9c7d8886b21/src/telegram/ext/_application.py#L792-L803

The docstring explicitly mentions this as supported ("< 0 - retry indefinitely") - but as not recommended.

I can understand this is not recommended - but it should either be marked as not supported entirely (fail on startup?) - or should not result in an endless loop.

Bibo-Joshi:
When ready, Closes #4966 
Will need to dig out my linux machine to write tests

### DIFF
--- a/src/telegram/ext/_application.py
+++ b/src/telegram/ext/_application.py
@@ -1016,6 +1016,7 @@ class Application(
             description="Bootstrap Initialize Application",
             max_retries=max_retries,
             interval=1,
+            is_running=lambda: not self._initialized,
         )
 
     def __run(
--- a/src/telegram/ext/_updater.py
+++ b/src/telegram/ext/_updater.py
@@ -704,7 +704,7 @@ class Updater(contextlib.AbstractAsyncContextManager["Updater"]):
         # delete_webhook for polling
         if drop_pending_updates or not webhook_url:
             await network_retry_loop(
-                is_running=lambda: self.running,
+                is_running=lambda: self._initialized,
                 action_cb=bootstrap_del_webhook,
                 description="Bootstrap delete Webhook",
                 interval=bootstrap_interval,
@@ -716,7 +716,7 @@ class Updater(contextlib.AbstractAsyncContextManager["Updater"]):
         # so we set it anyhow.
         if webhook_url:
             await network_retry_loop(
-                is_running=lambda: self.running,
+                is_running=lambda: not self._initialized,
                 action_cb=bootstrap_set_webhook,
                 description="Bootstrap Set Webhook",
                 interval=bootstrap_interval,


aelkheir (commented):
Haven't looked at tests yet. while current approach solves the issue, i am wondering if there would be an issue when breaking the loop from within, i.e without the reliance on `network_retry_loop.is_running` to track success.

Changing this:

https://github.com/python-telegram-bot/python-telegram-bot/blob/6e951a9fd35a8327921ffcd2040083380968bc20/src/telegram/ext/_utils/networkloop.py#L120-L125
to:
```python
    while effective_is_running():
        try:
            await do_action()
            _LOGGER.debug("%s Action succeeded. Stopping loop.", log_prefix)
            break
```

edit:
oh that might break polling i guess

Bibo-Joshi:
> oh that might break polling i guess

yup :D

Checking the return value would also basically bring us back to before #4880 😅 
One could argue that this PR now just moves the `return True/False` from before #4880 into the `is_running`, but IMHO it's at least still a cleaner interface. An alternative would be to make the `infinite_loop` variable introduced in #4880 a parameter - `get_updates` would set it to `True`, all others to `False`, such that the first successfull call exits. This as actually the initial draft of #4880 😆 https://github.com/python-telegram-bot/python-telegram-bot/pull/4880/commits/65e6cc9993c9ecf6a4954637023fb1318b2b7cd3

aelkheir:
I am a bit late to those discussions :) but here goes

I agree current approach is cleaner than the old `return True/False` to determine retry logic in that its more explicit

However it's still way inferior to `infinite_loop` parameter introduced (& removed) in #4880. It would be much nicer to just pass a boolean that determines whether or not the action callback is an long living one (it can have a default value of `False` since polling is the only place where its `True`)

future me: i realize i say (long living action callback) a lot below. i mean an action callback that should be re-called even if first attempts are successful.

That would be nicer, simpler than to assume every action callback is a long living one and require the caller to write custom logic in the form of lambdas to terminate the action.

Now according to the discussion at [comment #4880](https://github.com/python-telegram-bot/python-telegram-bot/pull/4880/files/d6dd49b96395f7165f9d09c7e80b94b211d5cc3b#r2237215290), i believe it was removed because it was thought redundant (`infinite_loop(=True) and max_retries=(-1)` meaning the same thing)
I don't think that's the case, tho the `infinite_loop` name might give that impression. but
- `infinite_loop(=True) and max_retries=(-1)`: this action callback is a long living one, execute it indefinitely. e.g polling
- `infinite_loop(=False) and max_retries=(-1)`: this action callback is _not_ a long living one. if failed, retry it indefinitely. e.g bootstrap initialize when passing `app.run_polling(bootstrap_retries=-1)`

Yes it might not make sense passing `infinite_loop(=True)` and `max_retries > 0`, but i see initial commits of #4880 did handle that case. Even if left unhandled, the logic can still hold: the action call back is long living, execute it 6 times (e.g. poll for one hour with interval 10 seconds) ('')

Last note regarding #4871, i don't see an issue in incrementing retries count regardless of the success state of action_cb. Unless we want to support a use case  like ('') and separate between retries because of errors, and normal operation (infinite_loop=True) retries.

Bibo-Joshi:
Let me re-add the `infinite_loop` parameter, then :) Maybe a different name though - `repeat_on_success`? A use case for `retry_on_success(=True)` *and* `max_retries > 0`  (i.e. calling exactly a finite number of times) should still not be needed IMO, so I'll skip that

src/telegram/ext/_utils/networkloop.py
aelkheir, line 1: can we remove `is_running` parameter entirely and use `while True:` in the loop 🤔 

 i'll have a look at the tests later this evening

Bibo-Joshi: It's still used by Application. At first glance I'm not sure if it's easily possible to remove the logic there ...

aelkheir (approved):

PR #4990, opened 2025-10-04
Update `actions/stale` Configuration to Use Issue Types

Bibo-Joshi:
- [ ] when merged, we should delete the `question` label

### DIFF
--- a/changes/unreleased/4990.HxHP2zqbTAcGzjapkgX3E5.toml
+++ b/changes/unreleased/4990.HxHP2zqbTAcGzjapkgX3E5.toml
@@ -1,5 +1,5 @@
 internal = "Update configuration of actions/stale to use issue types instead of labels"
 
 [[pull_requests]]
-uid = "4890"
+uid = "4990"
 author_uids = ["Bibo-Joshi"]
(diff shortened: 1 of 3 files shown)


changes/unreleased/4990.HxHP2zqbTAcGzjapkgX3E5.toml
```diff
@@ -0,0 +1,5 @@
+internal = "Update configuration of actions/stale to use issue types instead of labels"
+
+[[pull_requests]]
+uid = "4890"
+author_uids = ["Bibo-Joshi"]
```
harshil21, line 5: does this not need a `closes_threads = []`?

changes/unreleased/4990.HxHP2zqbTAcGzjapkgX3E5.toml
harshil21, line 1: the file name starts with 4990 but the uid is for 4890, is that okay? Actually, why are there 2 chango fragments in the first place?

Bibo-Joshi: b/c I'm too stupid to type the right number 😢

changes/unreleased/4990.HxHP2zqbTAcGzjapkgX3E5.toml
Bibo-Joshi: no, that's optional :)

harshil21 (approved):

PR #5004, opened 2025-10-14
Use Python 3.14 Final in the Test Suite

harshil21:


pyproject.toml
```diff
@@ -79,8 +79,6 @@ job-queue = [
 ]
 passport = [
     "cryptography!=3.4,!=3.4.1,!=3.4.2,!=3.4.3,>=39.0.1",
-    # cffi is a dependency of cryptography and added support for python 3.13 in 1.17.0rc1
-    "cffi >= 1.17.0rc1; python_version > '3.12'"
```
Bibo-Joshi, line 83: why was this removed? I would guess that there still is a crypto version that is installable on py3.13 but does not pin cffi >= 1.17.0rc1?

harshil21: tools like pip and uv would install the latest version so I don't think this is needed anymore, it was added when we first introduced 3.13 support

Bibo-Joshi: that may be the usual use case, but there could still be cases where for some weird reason someone tries to install that specific cyrpto version that is installable on py3.13 but doesn't pin cffi. At least I see no reason which would speak against keeping this pin here until we drop support for 3.13 :)

PR #5066, opened 2025-12-06
Minor changes to README to ensure a streamlined setup expierience

PLePort:
Installing: Added another comment before $ pip install python-telegram-bot --upgrade that reinforces the need for Python 3.10+.  I know this was stated at the beginning of the README, but adding it here would just act as another safety right before install in case it was glossed over. Also, I added a short note that building PTB from source requires Python headers and a C compiler.

Optional Dependencies:  Added -U to the pip installs just in case there are outdated dependencies.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -101,7 +101,7 @@ You can install or upgrade ``python-telegram-bot`` via
 
 .. code:: shell
 
-    Requires Python 3.10+.
+    #Requires Python 3.10+.
     $ pip install python-telegram-bot --upgrade
 
 To install a pre-release, use the ``--pre`` `flag <https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-pre>`_ in addition.


README.rst
```diff
@@ -101,11 +101,12 @@ You can install or upgrade ``python-telegram-bot`` via
 
 .. code:: shell
 
+    #Requires Python 3.10+.
```
Bibo-Joshi, line 104: Instead of having this as comment in the shell, I suggest moving it up to line 100, i.e. `… ptb on Python 3.10+ via …`

README.rst
```diff
@@ -166,8 +167,8 @@ To install multiple optional dependencies, separate them by commas, e.g. ``pip i
 
 Additionally, two shortcuts are provided:
 
-* ``pip install "python-telegram-bot[all]"`` installs all optional dependencies.
-* ``pip install "python-telegram-bot[ext]"`` installs all optional dependencies that are related to ``telegram.ext``, i.e. ``[rate-limiter, webhooks, callback-data, job-queue]``.
+* ``pip install -U "python-telegram-bot[all]"`` installs all optional dependencies.
+* ``pip install -U "python-telegram-bot[ext]"`` installs all optional dependencies that are related to ``telegram.ext``, i.e. ``[rate-limiter, webhooks, callback-data, job-queue]``.
```
Bibo-Joshi, line 171: TBH I'd prefer not to have the upgrade option here. It's likely that people get to this section if they already have PTB installed but need an additional optional dependency. In that case, upgrading PTB might not actually be desired.

README.rst
```diff
@@ -101,11 +101,12 @@ You can install or upgrade ``python-telegram-bot`` via
 
 .. code:: shell
 
+    #Requires Python 3.10+.
     $ pip install python-telegram-bot --upgrade
 
 To install a pre-release, use the ``--pre`` `flag <https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-pre>`_ in addition.
 
-You can also install ``python-telegram-bot`` from source, though this is usually not necessary.
+You can also install ``python-telegram-bot`` from source, though this is usually not necessary.  Before building from source, check that Pything headers and a C compiler are installed.
```
Bibo-Joshi, line 109: PTB is a pure-python package and doesn't have any C-bindings. What do you thnik Python headers & C compilers are required for?

Bibo-Joshi:
closing due to inactivity

PR #5089, opened 2026-01-07
Api 9.3 some gift related updates

Bibo-Joshi:

### DIFF
--- a/src/telegram/_message.py
+++ b/src/telegram/_message.py
@@ -961,7 +961,7 @@ class Message(MaybeInaccessibleMessage):
             was sent or received
 
             .. versionadded:: 22.1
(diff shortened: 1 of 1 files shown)


src/telegram/_message.py
```diff
@@ -957,6 +961,10 @@ class Message(MaybeInaccessibleMessage):
             was sent or received
 
             .. versionadded:: 22.1
+        gift_upgrade_sent (:class:`telegram.GiftInfo`): Otional. Service message: upgrade of a
```
aelkheir:
```suggestion
        gift_upgrade_sent (:class:`telegram.GiftInfo`): Optional. Service message: upgrade of a
```

aelkheir (changes requested):
perfect just a missing `p`

PR #5092, opened 2026-01-07
Api 9.3 Changes for `UniqueGift`

Bibo-Joshi:

### DIFF
--- a/changes/unreleased/5078.FoNwUYLbXQFRebTFhR6UPn.toml
+++ b/changes/unreleased/5078.FoNwUYLbXQFRebTFhR6UPn.toml
@@ -9,9 +9,13 @@ Full Support for Bot API 9.3
 """
 
 pull_requests = [
+  { uid = "5086", author_uids = ["Bibo-Joshi"] },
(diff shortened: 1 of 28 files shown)


src/telegram/_uniquegift.py
```diff
@@ -362,12 +381,29 @@ class UniqueGift(TelegramObject):
             published the gift.
 
             .. versionadded:: 22.4
+        is_premium (:obj:`bool`): Otional. :obj:`True`, if the original regular gift was
```
aelkheir:
```suggestion
        is_premium (:obj:`bool`): Optional. :obj:`True`, if the original regular gift was
```

below attrs also :)

aelkheir (commented):
nice! one comment
ps: note to self to add this to `tests/test_official/exceptions.py:209` after this is merged
```python
    # backwards compatibility for api 9.3 changes
    # tags: deprecated NEXT.VERSION, bot api 9.3
    "UniqueGift": {"gift_id"},
```

src/telegram/_uniquegift.py
Bibo-Joshi: hard word 🫠

PR #5096, opened 2026-01-08
Remove unused constants MESSAGE_ENTITIES and DEEP_LINK_LENGTH

Achieve3318:
- Removed MESSAGE_ENTITIES constant that was marked as unused in TODO
- Removed DEEP_LINK_LENGTH constant that was never referenced in codebase
- Both constants were not tested or documented as per comment 'constants above this line are tested'
- All tests pass successfully after removal

### DIFF
--- a/src/telegram/constants.py
+++ b/src/telegram/constants.py
@@ -2033,14 +2033,6 @@ class MessageLimit(IntEnum):
     """
     DEEP_LINK_LENGTH = 64
     """:obj:`int`: Maximum number of characters for a deep link."""
-    # TODO this constant is not used anywhere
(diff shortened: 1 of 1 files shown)


Bibo-Joshi:
Hi @Angel98518. Thanks for your PR. However, we can't simply drop these attributes. We have a [stability policy](https://docs.python-telegram-bot.org/en/stable/stability_policy.html#deprecation) specifying how documented interfaces are deprecated and then removed. Moreover, even though the constants are not used in the PTB code itself, they can still be useful for users of PTB. TBH I don't know recall why the todo comment was added in the first place. in #3351 (the PR adding this line) I couldn't find any discussion about it.

PR #5120, opened 2026-02-05
Avoid leaking raw webhook secret token in debug logs

RinZ27:
Noticed that the raw `X-Telegram-Bot-Api-Secret-Token` was being included in debug logs when a token mismatch occurred. Removed it from the log message to prevent sensitive information leakage and potential log injection. The log still indicates that a mismatch happened, which is sufficient for debugging.

### DIFF
--- a/src/telegram/ext/_utils/webhookhandler.py
+++ b/src/telegram/ext/_utils/webhookhandler.py
@@ -203,7 +203,7 @@ class TelegramHandler(tornado.web.RequestHandler):
                     HTTPStatus.FORBIDDEN, reason="Request did not include the secret token"
                 )
             if token != self.secret_token:
-                _LOGGER.debug("Request had the wrong secret token: %s", token)
+                _LOGGER.debug("Request had the wrong secret token")
                 raise tornado.web.HTTPError(
                     HTTPStatus.FORBIDDEN, reason="Request had the wrong secret token"
                 )


Poolitzer:
I don't personally agree. The default logging level is info. If I explicitly set it to debug, its because I need to debug a specific issue, exactly one like the token mismatches, because I might not have hardcoded it and have some logic issue there, or I mistyped an I as an l or something. Removing the token would make it one step harder to debug that issue

RinZ27:
Good point on the typos, @Poolitzer. My main concern is that DEBUG logs often end up in centralized log management systems where secrets shouldn't really sit in plain text. Perhaps we could log a masked version instead? Showing the first and last few characters might be a good middle ground that still lets you spot a logic error or a character swap without leaking the full token.

Poolitzer:
@RinZ27 I feel we can only loose this way. No matter what part we obscure, there will be someone whos issue is exactly in that part and thus can&#x27;t debug it.

For me its either log it or not. As soon as you set the httpx level higher then warning the bot token gets logged as well. So in the same spirit for me personally, I would say we log it in the one more verbose setting then the default is, so people can actually figure out their issue. One can always log to stdout as well to not have it written down.

RinZ27:
@Poolitzer I see the project's focus is on maximum debuggability when the user opts into DEBUG logs. Closing this to keep the current behavior as it fits your workflow better.

PR #5140, opened 2026-02-17
Fix: Moved inline test file to _inline folder

Poolitzer:
I saw this, it bugged me :D


harshil21:
Test failures unrelated. We need to fix that in another PR

PR #5141, opened 2026-02-18
[API 9.4] Add the class `VideoQuality` and the field `qualities` to the class `Video` 

### LINKED ISSUE #5134: Support `VideoQuality` and `Video.qualities` (Bot API 9.4)

The Bot API 9.4 added a new class `VideoQuality `and a new field `qualities `to the Video object.
This allows bots to retrieve information about all available qualities of a video.

Currently, python-telegram-bot does not expose `VideoQuality `or the `qualities `attribute in Video.

- Add a new `VideoQuality` class.
- Add a `qualities `attribute to the `Video `class as a sequence of `VideoQuality`.
- Update `Video.de_json` to parse the qualities field using `de_list_optional`.
- Extend acceptance of `VideoQuality` in methods that operate on file IDs, such as:
    - `Bot.send_video`
    - `Bot.get_file`
    - `_ChatBase.send_video`
    - `InputPaidMediaVideo.__init__`
    - `InputMediaVideo.__init__`
    - `Message.reply_video`
    - `User.send_video`
    - `ExtBot.send_video`
    - `ExtBot.get_file`
- Add the necessary tests.

hethon:
Closes #5134

### DIFF
--- a/src/telegram/_files/videoquality.py
+++ b/src/telegram/_files/videoquality.py
@@ -51,7 +51,7 @@ class VideoQuality(_BaseMedium):
         width (:obj:`int`): Video width.
         height (:obj:`int`): Video height.
         codec (:obj:`str`): Codec that was used to encode the video,
-            for example, “h264”, “h265”, or “av01”
+            for example, ``h264``, ``h265``, or ``av01``
         file_size (:obj:`int`): Optional. File size in bytes.
 
     """


src/telegram/_files/videoquality.py
```diff
@@ -0,0 +1,91 @@
...
+    Args:
+        file_id (:obj:`str`): Identifier for this file, which can be used
+            to download or reuse the file.
+        file_unique_id (:obj:`str`): Unique identifier for this file, which
+            is supposed to be the same over time and for different bots.
+            Can't be used to download or reuse the file.
+        width (:obj:`int`): Video width.
+        height (:obj:`int`): Video height.
+        codec (:obj:`str`): Codec that was used to encode the video,
+            for example, "h264", "h265", or "av01"
```
aelkheir:
```suggestion
            for example, ``h264``, ``h265``, or ``av01``
```
same under Attributes :)

src/telegram/_files/videoquality.py
```diff
@@ -0,0 +1,91 @@
...
+        width (:obj:`int`): Video width.
+        height (:obj:`int`): Video height.
+        codec (:obj:`str`): Codec that was used to encode the video,
+            for example, “h264”, “h265”, or “av01”
+        file_size (:obj:`int`): Optional. File size in bytes.
+            It can be bigger than 2^31 and some programming languages may have difficulty/silent
+            defects in interpreting it. But it has at most 52 significant bits,
+            so a signed 64-bit integer or double-precision float type are
+            safe for storing this value.
+
```
aelkheir, line 56:
```suggestion
```

tests/_files/test_videoquality.py
```diff
@@ -0,0 +1,166 @@
...
+# along with this program.  If not, see [http://www.gnu.org/licenses/].
+import pytest
+
+from telegram import PhotoSize, VideoQuality
+from tests.auxil.slots import mro_slots
+
+
+@pytest.fixture(scope="module")
+def videoquality_message_id():
+    # a video message with qualities available
```
aelkheir: can we add the @ of the channel the message was forwarded to,  and a reference to the thread where this was discussed (https://t.me/pythontelegrambotdev/1782). just for reference if things go haywire

tests/_files/test_videoquality.py
```diff
@@ -0,0 +1,166 @@
...
+@pytest.fixture(scope="module")
+def videoquality(videoquality_list):
+    return videoquality_list[-1]
+
+
+class VideoQualityTestBase:
+    widths = [464]
+    heights = [848]
+    codecs = ["h264"]
+    file_sizes = [405540]
```
aelkheir: A comment here that these  are magic numbers and need not to be messed with would be nice :)
something like: These values are tied to the forwarded video and must remain unchanged

Also can this be simplified to basic class variables only  `widths = 464` ..etc
and then have direct equality check under `test_expected_values` e.g `videoquality.width == self.width`

If Telegram later decided to return more than one `VideoQuality` for this video specifically, that would be a sign to just test it completely offline imo

tests/_files/test_videoquality.py
```diff
@@ -0,0 +1,166 @@
...
+        assert new_file.file_path.startswith("https://")
+
+        await new_file.download_to_drive(tmp_file)
+
+        assert tmp_file.is_file()
+
+    async def test_resend(self, bot, chat_id, videoquality):
+        message = await bot.send_video(chat_id, videoquality.file_id)
+        # Sending via file_id does not retain `qualities`.
+        assert not message.video.qualities
```
aelkheir: hmm i am not sure we need to test this. That behavior (of "sufficiently large audience") is out of our control, and if it ever change we'll have our test breaking too 😥

tests/_files/test_videoquality.py
hethon: > Also can this be simplified to basic class variables only `widths = 464` ..etc
> and then have direct equality check under `test_expected_values` e.g `videoquality.width == self.width`

I can do that. Just a note: switching direct equality checks will make the test dependent on the forwarded message containing exactly one video quality. I’ll add a comment in the test to highlight this.

Edit: It just came to me I can enforce the dependency using code, in addition to leaving a comment, by doing `assert len(videoquality_list) == 1` in `test_qualities_available`.

tests/_files/test_videoquality.py
hethon: You’re right to point that out. However, we still need a test to ensure that a video can be sent using a `file_id` obtained from a `VideoQuality` object. I’ll update `test_resend` to verify that the video is sent correctly using the `file_id` from `videoquality`, without asserting anything about the qualities attribute.

src/telegram/_files/videoquality.py
```diff
@@ -0,0 +1,90 @@
...
+    Attributes:
+        file_id (:obj:`str`): Identifier for this file, which can be used
+            to download or reuse the file.
+        file_unique_id (:obj:`str`): Unique identifier for this file, which
+            is supposed to be the same over time and for different bots.
+            Can't be used to download or reuse the file.
+        width (:obj:`int`): Video width.
+        height (:obj:`int`): Video height.
+        codec (:obj:`str`): Codec that was used to encode the video,
+            for example, “h264”, “h265”, or “av01”
```
harshil21: Can you introduce an enum for this in `telegram.constants` and link it here and in the class body? See e.g. `telegram/_uniquegift.py` on how to implement the enum thing both in the documentation and in the `__init__`

src/telegram/_files/videoquality.py
```diff
@@ -0,0 +1,90 @@
...
+        file_size (:obj:`int`): Optional. File size in bytes.
+            It can be bigger than 2^31 and some programming languages may have difficulty/silent
+            defects in interpreting it. But it has at most 52 significant bits,
+            so a signed 64-bit integer or double-precision float type are
+            safe for storing this value.
```
harshil21: we are in python so we can remove this
```suggestion.
```

src/telegram/_files/videoquality.py
```diff
@@ -0,0 +1,90 @@
...
+            Can't be used to download or reuse the file.
+        width (:obj:`int`): Video width.
+        height (:obj:`int`): Video height.
+        codec (:obj:`str`): Codec that was used to encode the video,
+            for example, ``h264``, ``h265``, or ``av01``
+        file_size (:obj:`int`, optional):  File size in bytes.
+            It can be bigger than 2^31 and some programming languages may have difficulty/silent
+            defects in interpreting it. But it has at most 52 significant bits,
+            so a signed 64-bit integer or double-precision float type are
+            safe for storing this value.
```
harshil21:
```suggestion
```
same

tests/_files/test_videoquality.py
```diff
@@ -0,0 +1,155 @@
...
+
+@pytest.fixture(scope="module")
+async def videoquality_list(bot, chat_id, channel_id, videoquality_message_id):
+    return (
+        await bot.forward_message(chat_id, channel_id, videoquality_message_id, read_timeout=50)
+    ).video.qualities
+
+
+@pytest.fixture(scope="module")
+def videoquality(videoquality_list):
```
harshil21:
```suggestion
def video_quality(videoquality_list):
```
nitpick: separating words with underscores is the convention we use

src/telegram/_files/videoquality.py
hethon: I don’t think an enum is appropriate here imo. The strings `"h264"`, `"h265"`, and `"av01"` are just examples of codecs that might appear; it’s not guaranteed that codec will always be one of these values. Using an enum here would give a false impression that only those three codecs are valid.

This is different from `UniqueGiftInfo.origin`, where PTB used an enum because the official Telegram docs explicitly list all possible values: `"upgrade"`, `"transfer"`, `"resale"`, `"gifted_upgrade"`, or `"offer"`. In that case, an enum makes sense because the set of allowed values is fixed and documented.

tests/_files/test_videoquality.py
hethon: Should we change `videoquality_list` to `video_quality_list` and `videoquality_message_id` to `video_quality_message_id` too?

harshil21: Yep

src/telegram/_files/videoquality.py
harshil21: That's a fair point. I didn't fully read the documentation

tests/_files/test_videoquality.py
hethon: Done

PR #5142, opened 2026-02-19
Replace `pre-commit` with `prek`

### LINKED ISSUE #5138: Switch to using `prek` instead of `pre-commit`

Recently I've been using `prek` in my projects instead of `pre-commit`. [Prek](https://prek.j178.dev/) is a much faster, drop in replacement for pre-commit, which takes half the disk space. 

I didn't find any downsides to it during my testing, and it also runs hooks in parallel, which speeds up development (specially for us since pylint and mypy both take a while)

Use prek. The config file doesn't need to change at all. In fact, we can remove the `ci:` section in there, and replace that with this github action https://prek.j178.dev/integrations/#github-actions. This way the pre-commit.ci bot will not make any unnecessary pull requests for auto updates anymore. 

Stick to using pre-commit, but since this is a drop in replacement, we should switch over.

### Additional context

_No response_

harshil21:
Closes #5138 

I'll remove the pre-commit.ci bot from the repository after this PR is merged.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.15.4'
+    rev: 'v0.15.5'
     hooks:
     # Run the linter:
     -   id: ruff-check
--- a/changes/unreleased/5156.DJGYhbSstpVvdEiQ74PocD.toml
+++ b/changes/unreleased/5156.DJGYhbSstpVvdEiQ74PocD.toml
@@ -1,0 +1,5 @@
+internal = "Update Ruff to v0.15.5"
+[[pull_requests]]
+uid = "5156"
+author_uids = ["renovate[bot]"]
+closes_threads = []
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -29,6 +21,7 @@ repos:
           - cachetools>=7.0.0,<8.0.0
           - aiolimiter~=1.1,<1.3
           - . # this basically does `pip install -e .`
+        priority: 10
```
Poolitzer, line 24: Wait what does this do? I cant even find it in the docs...

harshil21: https://prek.j178.dev/configuration/#priority

ones with the same priority run in parallel

Poolitzer: oh I looked in pre-commit docs for file name reasons.

Should we rename the file since we don't run pre-commit anymore?

harshil21: I don't mind, sure

harshil21: actually we need a prek.toml which would require changing the syntax, I don't think that's worth it, what do you think?

Poolitzer: I am accustomed to that syntax, sure, lets keep it

PR #5148, opened 2026-02-24
API 9.4 - Last 4 bullet points

harshil21:
Implements the last 4 bullet points of https://core.telegram.org/bots/api#february-9-2026

### DIFF
--- a/src/telegram/_user.py
+++ b/src/telegram/_user.py
@@ -63,6 +63,7 @@ if TYPE_CHECKING:
         Story,
         SuggestedPostParameters,
         UserChatBoosts,
+        UserProfileAudios,
         UserProfilePhotos,
         Venue,
         Video,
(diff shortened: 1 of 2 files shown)


src/telegram/_bot.py
```diff
@@ -45,7 +47,7 @@
     serialization = None  # type: ignore[assignment]
     CRYPTO_INSTALLED = False
 
-from telegram._botcommand import BotCommand
+from telegram._botcommand import BotCommand  # pylint: disable=ungrouped-imports
```
Poolitzer, line 50: is that new, did we update pylint? I touched the file and didn't get the error. or does it only check when changing imports

harshil21: I'm not sure, I don't even know why it's complaining because it looks fine to me. But pre-commit was failing for me otherwise.

Poolitzer: We can merge it and the prek replacement, remove the ignore and see if it still fails. Maybe we get a better error out of it.

PR #5155, opened 2026-03-01
Full support for Bot API 9.5

Poolitzer:
I implemented everything. I am missing (proper) tests for the new date_time entity and a final discussion on how we want to do the to_unix helper (I did a reference implementation so we can discuss about it).

I also based it on api_9.4 since that is the most developed branch, and it should make it easier for me to keep merge compatibility.

- [x] double check diff to https://t.me/bot_api_changes/276#

### DIFF
--- a/src/telegram/helpers.py
+++ b/src/telegram/helpers.py
@@ -29,15 +29,12 @@ __all__ = (
     "escape_markdown",
     "mention_html",
     "mention_markdown",
-    "to_unix_timestamp",
 )
 
-import datetime as dtm
 import re
 from html import escape
 from typing import TYPE_CHECKING
 
-from telegram._utils.datetime import to_timestamp
 from telegram._utils.types import MarkdownVersion
 from telegram.constants import MessageLimit, MessageType
 
@@ -205,13 +202,3 @@ def create_deep_linked_url(
     key = "startgroup" if group else "start"
 
     return f"{base_url}?{key}={payload}"
-
-
-def to_unix_timestamp(datetime_obj: float | dtm.timedelta | dtm.datetime | dtm.time) -> str:
-    """
-    Helper to calculate unix timestamps to use in the
-    :const:`telegram.constants.MessageEntityType.DATE_TIME` string formatting.
-
-    .. versionadded:: NEXT.VERSION
-    """
-    return str(to_timestamp(datetime_obj))


Poolitzer:
This is basically done now, I added the missing tests for the new message entity

Only thing missing is the new helper, for which I would like feedback first if this is a wanted design, and then I would test it and put it in an example... somewhere.

src/telegram/_bot.py
```diff
@@ -6007,6 +6012,10 @@ async def promote_chat_member(
                 posts; for channels only
 
                 .. versionadded:: 22.4
+            can_manage_tags (:obj:`bool`, optional): Pass :obj:`True`if the administrator can edit
```
harshil21:
```suggestion
            can_manage_tags (:obj:`bool`, optional): Pass :obj:`True` if the administrator can edit
```
whitespace

src/telegram/_bot.py
```diff
@@ -12019,6 +12029,55 @@ async def get_user_profile_audios(
...
+        be an administrator in the chat for this to work and must have the
+        :attr:`~telegram.ChatMemberAdministrator.can_manage_tags` administrator right.
+
+        .. versionadded:: NEXT.VERSION
+
+        Args:
+            chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target group or username
+                of the target supergroup (in the format ``@supergroupusername``).
+            user_id (:obj:`int`): Unique identifier of the target user.
+            comment (:obj:`str`, optional): New tag for the member;
```
harshil21:
```suggestion
            tag (:obj:`str`, optional): New tag for the member;
```

src/telegram/_chatpermissions.py
```diff
@@ -47,6 +47,9 @@ class ChatPermissions(TelegramObject):
           :attr:`can_send_videos`, :attr:`can_send_video_notes` and :attr:`can_send_voice_notes`
           are considered as well when comparing objects of this type in terms of equality.
         * Removed deprecated argument and attribute ``can_send_media_messages``.
+    .. versionchanged:: 20.0
```
harshil21:
```suggestion
    .. versionchanged:: NEXT.VERSION
```

src/telegram/_chatpermissions.py
```diff
@@ -219,7 +234,7 @@ def all_permissions(cls) -> "ChatPermissions":
         .. versionadded:: 20.0
 
         """
-        return cls(*(14 * (True,)))
+        return cls(*(True,) * len(cls.__slots__))
```
harshil21, line 237: nice

src/telegram/constants.py
```diff
@@ -2013,6 +2015,11 @@ class MessageEntityType(StringEnum):
     CODE = "code"
     """:obj:`str`: Message entities representing monowidth string."""
     CUSTOM_EMOJI = "custom_emoji"
+    """:obj:`str`: Message entities representing ormatted date and time.
+
+    .. versionadded:: NEXT.VERSION
+    """
```
harshil21, line 2021: this docstring has switched places with the custom_emoji docstring?

src/telegram/constants.py
```diff
@@ -2048,6 +2055,63 @@ class MessageEntityType(StringEnum):
...
+    LOCALIZED_WEEKDAY = "w"
+    """:obj:`str`: Displays the day of the week in the user's localized language."""
+    SHORT_DATE = "d"
+    """:obj:`str`: Displays the date in short form (e.g., “17.03.22”)."""
+    LONG_DATE = "D"
+    """:obj:`str`: Displays the date in long form (e.g., “March 17, 2022”)."""
+    SHORT_TIME = "t"
+    """:obj:`str`: Displays the time in short form (e.g., “22:45”)."""
+    LONG_TIME = "T"
+    """:obj:`str`: Displays the time in long form (e.g., “22:45:00”)."""
```
harshil21:
```suggestion
    """:obj:`str`: Displays the date in short form (e.g., "17.03.22")."""
    LONG_DATE = "D"
    """:obj:`str`: Displays the date in long form (e.g., "March 17, 2022")."""
    SHORT_TIME = "t"
    """:obj:`str`: Displays the time in short form (e.g., "22:45")."""
    LONG_TIME = "T"
    """:obj:`str`: Displays the time in long form (e.g., "22:45:00")."""
```

can you use regular quotation marks?

src/telegram/helpers.py
```diff
@@ -202,3 +205,13 @@ def create_deep_linked_url(
     key = "startgroup" if group else "start"
 
     return f"{base_url}?{key}={payload}"
+
+
+def to_unix_timestamp(datetime_obj: float | dtm.timedelta | dtm.datetime | dtm.time) -> str:
```
harshil21: so I'm confused on where you want to use this. I see that it only converts the return type to a string. What's the main advantage?

Poolitzer: yeah it basically exposed the internal function, but when people want to use the new tag they do need this.

src/telegram/_chatpermissions.py
Poolitzer: yeah I copied it from that other class

src/telegram/_bot.py
```diff
@@ -12019,6 +12029,55 @@ async def get_user_profile_audios(
...
+        """
+        Use this method to set a tag for a regular member in a group or a supergroup. The bot must
+        be an administrator in the chat for this to work and must have the
+        :attr:`~telegram.ChatMemberAdministrator.can_manage_tags` administrator right.
+
+        .. versionadded:: NEXT.VERSION
+
+        Args:
+            chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target group or username
+                of the target supergroup (in the format ``@supergroupusername``).
```
aelkheir:
```suggestion
            chat_id (:obj:`int` | :obj:`str`): |chat_id_group|
```

src/telegram/_bot.py
```diff
@@ -12019,6 +12029,55 @@ async def get_user_profile_audios(
...
+            user_id (:obj:`int`): Unique identifier of the target user.
+            tag (:obj:`str`, optional): New tag for the member;
+            0-:tg-const:`telegram.constants.TagLimit.MAX_TAG_LENGTH` characters, emoji are not
+            allowed.
```
aelkheir:
```suggestion
            tag (:obj:`str`, optional): New tag for the member;
                0-:tg-const:`telegram.constants.TagLimit.MAX_TAG_LENGTH` characters, emoji are not
                allowed.
```

src/telegram/_chatpermissions.py
```diff
@@ -134,12 +141,17 @@ class ChatPermissions(TelegramObject):
             notes.
 
             .. versionadded:: 20.1
+        can_edit_tag (:obj:`bool`): Optionsl. :obj:`True`, if the user is allowed to edit their own
```
aelkheir:
```suggestion
        can_edit_tag (:obj:`bool`): Optional. :obj:`True`, if the user is allowed to edit their own
```

src/telegram/_messageentity.py
```diff
@@ -119,6 +157,8 @@ def __init__(
         user: User | None = None,
         language: str | None = None,
         custom_emoji_id: str | None = None,
+        date_time_format: str | None = None,
+        unix_time: dtm.datetime | None = None,
```
aelkheir, line 161: They could've just named it date 😔

src/telegram/constants.py
```diff
@@ -2048,6 +2055,63 @@ class MessageEntityType(StringEnum):
     """:obj:`str`: Message entities representing a url."""
 
 
+class MessageEntityDateTimeFormats(StringEnum):
```
aelkheir, line 2058: can we only have the unique single letter values they mention in [date-time-entity-formatting](https://core.telegram.org/bots/api#date-time-entity-formatting) and consider it a doc only enum?

src/telegram/_message.py
```diff
@@ -5304,6 +5315,17 @@ def _parse_html(
                 insert = f'<span class="tg-spoiler">{escaped_text}</span>'
             elif entity.type == MessageEntity.CUSTOM_EMOJI:
                 insert = f'<tg-emoji emoji-id="{entity.custom_emoji_id}">{escaped_text}</tg-emoji>'
+            elif entity.type == MessageEntity.DATE_TIME:
+                if entity.date_time_format:
+                    insert = (
+                        f'<tg-time unix="{to_timestamp(entity.unix_time)}" '
```
aelkheir, line 5321: similar to [comment](https://github.com/python-telegram-bot/python-telegram-bot/pull/5155/changes#r2880122680)
can this be made simpler by this 
`f'<tg-time unix="{entity.unix_time.timestamp()}" '`

src/telegram/_bot.py
Poolitzer: I dont get that one, now that will be there twice no? Cause it already exists from 12055-12057

src/telegram/constants.py
Poolitzer: But, then we don't have shortcuts to the allowed combos? That would make it worse imo. And what is a doc only enum?

src/telegram/_message.py
Poolitzer: hmmm probably

src/telegram/_bot.py
aelkheir: Sorry i messed up the selection of the lines.
These two lines need an extra indentation
```python
0-:tg-const:`telegram.constants.TagLimit.MAX_TAG_LENGTH` characters, emoji are not
allowed.
```

src/telegram/_message.py
Poolitzer: okay I retract that, imo if I don't use this shortcut, we loose the default timezone support we have.

like `int(entity.unix_time.replace(tzinfo=dtm.timezone.utc).timestamp())` would work, but I don't want that to happen, since that would mean no matter what timezone you create the datetime object with the timestamp is always the UTC one. I would want it to be like in the tests/how datetime_localization is. By default, UTC is used, e.g. an unaware datetime object `dtm.datetime(2000, 7, 28)` turns to `964742400`. And an aware like `dtm.datetime(2000, 7, 28, tzinfo=ZoneInfo("Europe/Berlin"))` turns to `964735200`.

And we lose that without the shortcut.

src/telegram/constants.py
aelkheir: fair enough, i feared it would be a lot to maintain all valid combinations if they're ever changed/updated. but all good.

I meant for it to be an enum class intended for documentation more than for importing and passing around.

Poolitzer: no I really like having this explaining shortcut so people easily know what they want without having to think about this confusing regex documentation combi. But yeah lets hope they never expand this xD

src/telegram/_message.py
aelkheir: >  that would mean no matter what timezone you create the datetime object with the timestamp is always the UTC one.

We can conditionally replace with `UTC` depending on whether `entity.unix_time.tzinfo` is None or not. But even that would not be necessary because `entity.unix_time` is already guaranteed to be time zone aware (either `UTC`, or `defaults.tzinfo`). However yes if we do not want to rely on that we'll have to conditionally do the replacing either manually in place or let the new helper do that.

I am now wondering about the utility of the new helper to outside users. Given they have a `dtm.datetime` object `foo` they can either use the new helper or directly call `foo.timestamp()`. The only difference between the two is that `timestamp()` will assume that `foo` is in local time if it was naive datetime. If this is unimportant to us i'd say we go with the manual replacing option from above.

Poolitzer: Well not the new helper I use the existing one here.

But yeah the new helper would only really help if it could get automatically the default timezone I guess. Which we cant easily achieve, at least I don't see a way.

aelkheir: > Well not the new helper I use the existing one here.

ooh my bad. all good here then
yeah we'll have pass extbot/default timezone or place the helper itself inside extbot which are meeh

src/telegram/_bot.py
```diff
@@ -12019,6 +12029,55 @@ async def get_user_profile_audios(
...
+        """
+        Use this method to set a tag for a regular member in a group or a supergroup. The bot must
+        be an administrator in the chat for this to work and must have the
+        :attr:`~telegram.ChatMemberAdministrator.can_manage_tags` administrator right.
+
+        .. versionadded:: NEXT.VERSION
+
+        Args:
+            chat_id (:obj:`int` | :obj:`str`): Unique identifier for the target group or username
+            chat_id (:obj:`int` | :obj:`str`): |chat_id_group|
```
harshil21:
```suggestion
            chat_id (:obj:`int` | :obj:`str`): |chat_id_group|
```

Poolitzer: ahaha sorry

PR #5186, opened 2026-03-29
Centralize `de_json`

harshil21:
Closes #5189 

[This](https://pastebin.com/EnEkjkYg) is the benchmark I used for the results below. It includes a mix of different kind of classes and cases. 

Here are the performance results:

| Test Case                       | Master (it/s) | Master best (s) | Master mean (µs) | Branch (it/s) | Branch best (s) | Branch mean (µs) | Change (%) |
|---------------------------------|---------------:|----------------:|----------------:|---------------:|----------------:|----------------:|------------:|
| User                            | 169460.781 | 0.030 | 5.9 | 163505.137 | 0.031 | 6.2 | −3.5 |
| ChatBoost                       | 58043.702 | 0.086 | 17.4 | 54131.579 | 0.092 | 19.0 | −6.7 |
| BusinessConnection               | 50168.175 | 0.100 | 20.0 | 42575.756 | 0.117 | 23.7 | −15.1 |
| CallbackQuery                   | 15806.120 | 0.316 | 63.6 | 18839.647 | 0.265 | 53.3 | **+19.2** |
| ChatJoinRequest                 | 62539.851 | 0.080 | 16.1 | 56644.010 | 0.088 | 18.4 | −9.4 |
| Message‑minimal                 | 17122.015 | 0.292 | 58.6 | 21057.629 | 0.237 | 47.7 | **+23.0** |
| Message‑medium                  | 9886.712 | 0.506 | 101.7 | 10675.907 | 0.468 | 94.1 | **+8.0** |
| Message‑heavy                   | 4334.172 | 1.154 | 232.5 | 4694.056 | 1.065 | 213.5 | **+8.3** |
| TransactionPartner‑user         | 63686.830 | 0.079 | 15.9 | 64283.784 | 0.078 | 15.7 | +0.9 |
| TransactionPartner‑fragment     | 68899.608 | 0.073 | 14.7 | 72328.266 | 0.069 | 13.9 | **+5.0** |
| ChatMember‑member               | 77334.824 | 0.065 | 13.1 | 76181.318 | 0.066 | 13.2 | −1.5 |
| ChatMember‑administrator        | 56269.916 | 0.089 | 17.9 | 54497.678 | 0.092 | 18.5 | −3.1 |
| UserProfilePhotos               | 28991.156 | 0.172 | 34.7 | 27078.124 | 0.185 | 37.2 | −6.6 |
| InlineKeyboardMarkup            | 22011.970 | 0.227 | 45.6 | 20940.324 | 0.239 | 47.9 | −4.9 |
| Update‑message                  | 12580.153 | 0.397 | 79.9 | 15165.953 | 0.330 | 66.1 | **+20.5** |
| Update‑callback                 | 13060.602 | 0.383 | 76.8 | 15791.913 | 0.317 | 63.6 | **+20.9** |

I'll add tests after initial review.


src/telegram/_checklists.py
```diff
@@ -112,25 +110,6 @@ def __init__(
...
-    @classmethod
-    def de_json(cls, data: JSONDict, bot: Optional["Bot"] = None) -> "ChecklistTask":
-        """See :meth:`telegram.TelegramObject.de_json`."""
-        data = cls._parse_data(data)
-
-        # Get the local timezone from the bot if it has defaults
-        loc_tzinfo = extract_tzinfo_from_defaults(bot)
-
-        if (date := data.get("completion_date")) == 0:
-            data["completion_date"] = ZERO_DATE
```
aelkheir, line 124: TIL this wasn't necessary :)
In hindsight makes sense

src/telegram/_telegramobject.py
```diff
@@ -20,25 +20,84 @@
...
+    if item_origin is Sequence:
+        inner_args = get_args(item_type)
+        if not inner_args:
+            return None
+        inner_fn = _make_seq_transform(inner_args[0], globalns, tg_ns)
+        if inner_fn is None:
+            return None
+        return lambda v, b, _f=inner_fn: (  # type: ignore[misc]
+            [_f(row, b) for row in v] if isinstance(v, list) else v
+        )
```
aelkheir:
```suggestion
        def transform_fn(
            value: object,
            bot: "Bot | None",
            fn: "Callable[[object, Bot | None], object]" = inner_fn,
        ):
            return (  # type: ignore[misc]
                [fn(row, bot) for row in value] if isinstance(value, list) else value
            )

        return transform_fn
```

That's more diff, but the lambda gave me a hard time.. a skill issue on my side :D

src/telegram/_telegramobject.py
```diff
@@ -20,25 +20,84 @@
...
+
+    if isinstance(item_type, type) and issubclass(item_type, TelegramObject):
+        return lambda v, b, _c=item_type: (  # type: ignore[misc]
+            _c.de_list(v, b) if isinstance(v, list) else v
+        )
```
aelkheir:
```suggestion
        def transform_fn(
            value: object,
            bot: "Bot | None",
            tg_class: TelegramObject = item_type,
        ):
            if value is None:
                return ()
            return (  # type: ignore[misc]
                tg_class.de_list(value, bot) if isinstance(value, list) else value
            )

        return transform_fn
```

the `if value is None` branch should not matter as `parse_sequence_arg` will be called on the value again. This is just to keep old behavior of `de_list_optional`

And btw way `argumentparsing`'s `de_list_optional`, `de_json_optional` can be removed i think

src/telegram/_telegramobject.py
```diff
@@ -430,7 +591,57 @@ def de_json(cls: type[Tele_co], data: JSONDict, bot: "Bot | None" = None) -> Tel
...
+            # Compute tzinfo once for all datetime fields, if any
+            tz = extract_tzinfo_from_defaults(bot)
+            for key in data:  # Only loop through keys returned by the API
+                if key in plan and data[key] is not None:  # Should we transform this field?
+                    target = plan[key]  # The transform target for this field
+                    if target is _DATETIME_FIELD:  # timestamp → datetime
+                        if not isinstance(data[key], dtm.datetime):  # Avoid retransformations
+                            data[key] = from_timestamp(data[key], tzinfo=tz)
+                    elif isinstance(target, type):  # Target is a TelegramObject subclass → de_json
+                        if not isinstance(data[key], target):  # Avoid retransformations
```
aelkheir: regarding the `avoid retransformations` comment, is this measure for classes that  override de_json for whatever reason, or is there actually a another place that does pre transformations cause i couldn't find any

src/telegram/_telegramobject.py
```diff
@@ -430,7 +591,57 @@ def de_json(cls: type[Tele_co], data: JSONDict, bot: "Bot | None" = None) -> Tel
...
+        # We'll mutate data below (rename, pop, transform), so copy first.
+        data = cls._parse_data(data)
+
+        # Dispatch to subclass for delegator classes (e.g. TransactionPartner, ChatMember).
+        if cls.__DE_JSON_DISPATCH__:
+            dispatch_key, dispatch_mapping = cls.__DE_JSON_DISPATCH__
+            target_cls: Tele_co = dispatch_mapping.get(  # type: ignore[assignment]
+                data.get(dispatch_key)  # type: ignore[arg-type]
+            )
+            if target_cls is not None:
```
aelkheir, line 550: if for some reason the `target_cls` turned out to be `None`  (for example telegram added a new child class to `MessageOrigin`) then continuing from this branch to the end of de_json would still produce a MessageOrigin with unknown fields added to api_kwargs?

aelkheir (commented):
Hey! This is really really neat.
I left a couple of suggestions and a questions for my understanding

src/telegram/_telegramobject.py
harshil21: Yeah that was actually to accommodate the de_json behaviour happening in PassportFile. The way the PassportFile de_json worked is flawed because the `credentials` object would already be a `FileCredentials` object in the `de_json` call, which is why I had to "avoid retransformations". 

I'm now just going to fix the passport side instead to actually have dictionaries when doing a de_json, so we can delete this.

src/telegram/_telegramobject.py
harshil21: Thanks yeah the lambda's were not really readable..

I'm removing the value is None check because value is guaranteed to not be None (since we explicitly check that - see comment "Should we transform this field?")

src/telegram/_telegramobject.py
harshil21: Yep that's right. Now that we have a lot of new classes not implemented, it's a perfect test for forward compatibility...

harshil21:
Did more forward compatibility tests with API 10.1/10.2, and they work as intended. Also added tests, which should explain the behavior of de_json at a higher level.

Poolitzer:
A great undertaking, I dont have any open question.

harshil21:
Alright, merging now!

PR #5215, opened 2026-04-25
Correct isinstance check variable in _MergedFilter.__init__

gistrec:
Use self.or_filter instead of self.and_filter in the isinstance guard for the or_filter branch (copy-paste error, no functional impact)

### DIFF
--- a/changes/unreleased/5125.filter-fix.toml
+++ b/changes/unreleased/5125.filter-fix.toml
@@ -1,0 +1,5 @@
+bugfixes = "Fixed incorrect isinstance check in class telegram.ext.filters._MergedFilter for or_filter."
+[[pull_requests]]
+uid = "5125"
+author_uids = ["gistrec"]
+closes_threads = []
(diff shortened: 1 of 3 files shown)


harshil21:
Hello! Thanks for the catch! Is it possible for you to add a new chango fragment in the changes/unreleased directory? It should be named as `5125.filter-fix.toml`. See the contents of the other files in that directory on what to put inside that file. 

Updating the tests for this to catch bugs like this is also needed (see `tests/ext/test_filter.py`)

gistrec:
> Updating the tests for this to catch bugs like this is also needed (see tests/ext/test_filter.py)

Hello, this bug does not affect observable behavior, since `and_filter` is always `None` in the OR case (filter1 | filter2), and `isinstance(None, bool)` is `False` — so the condition evaluated correctly by accident

A test covering the affected code path (non-data filter | data filter where the base does not match) has been added. The exact isinstance typo cannot be tested without accessing `_MergedFilter` directly, which is private and not used in these tests

harshil21 (approved):
LGTM. Job failures unrelated. Thanks for the contribution!

PR #5230, opened 2026-05-10
Bot API 10.0: `Chat Management`

harshil21:


src/telegram/_chatmember.py
```diff
@@ -621,6 +630,9 @@ def __init__(
         can_send_voice_notes: bool,
         can_edit_tag: bool,
         tag: str | None = None,
+        # tags: NEXT.VERSION
+        # temporarily optional to make it not breaking
```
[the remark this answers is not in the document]
harshil21: no thats the convention we use

src/telegram/_chatmember.py
```diff
@@ -643,6 +655,7 @@ def __init__(
             self.can_send_video_notes: bool = can_send_video_notes
             self.can_send_voice_notes: bool = can_send_voice_notes
             self.can_edit_tag: bool = can_edit_tag
+            self.can_react_to_messages: bool | None = can_react_to_messages
```
aelkheir:
```suggestion
            self.can_react_to_messages: bool = can_react_to_messages
```
and raising TypeError when missing

src/telegram/_user.py
```diff
@@ -2771,3 +2771,83 @@ async def replace_token(
             pool_timeout=pool_timeout,
             api_kwargs=api_kwargs,
         )
+
+    async def delete_reaction(
+        self,
+        chat_id: int | str,
+        message_id: int,
+        actor_chat_id: int | None = None,
```
aelkheir, line 2779: can `actor_chat_id` be removed from the signature? or you preferred not to intervene between users and Telegram

aelkheir (changes requested):
Great! just two comments, and a shortcut for `CallbackQuery` would be nice.

src/telegram/_user.py
harshil21: yea I initially didn't have it on there, but then decided to keep it in since Telegram should be the one complaining about it

aelkheir: Aha okay i was also hesitant about it being input policing :)
but maybe we can leave note about it saying its expected to be left empty or something

src/telegram/_chatmember.py
harshil21: I think that change might break the type completeness CI, I'll add the TypeError though

aelkheir: It should pick it up i think if the check in line 661 is against the parameter and moved to before the `_unfrozen()` block

harshil21:
Not sure if the CallbackQuery shortcut for removing a reaction on a message would be useful, what use case are you thinking of?

A user could always remove their own reaction from the message directly instead of using a inline keyboard button.

aelkheir:
The same regular usecase that exist in `Message` but when the developer is dealing with a `callback_query` update.

They can still do `callback_query.message.delete_{all,message}reaction(s)` but they wouldn't get code completion from my experience 

> A user could always remove their own reaction from the message directly instead of using a inline keyboard button.

Maybe a reset reactions button for group/channel admins but not necessarily. For me its more about the code completion if/when needed

src/telegram/_callbackquery.py
```diff
@@ -929,6 +929,79 @@ async def copy_message(
...
+        .. versionadded:: NEXT.VERSION
+
+        Returns:
+            :obj:`bool`: On success, :obj:`True` is returned.
+
+        Raises:
+            :exc:`TypeError` if :attr:`message` is not accessible.
+        """
+        return await self._get_message(action="delete_reaction").delete_reaction(
+            user_id=self.from_user.id,
```
aelkheir: i agree deleting own reaction don't make much sense what i had in mind was accepting and repassing user_id having same interface as the message shortcut 
I'm sorry for the confusion 

Also my reset reactions example was wrong. I forgot that delete all reactions work on per user basis not message

So same for that too (accepting and repassing user_id)? Or what do you think 

My apologies and if you still think they're not useful they can be removed

aelkheir (commented):
apologies i had these comments pending for days and thought they were delivered. i forgot how to use github apparently

src/telegram/_callbackquery.py
harshil21: Ah, then well if you have to repass the user_id at that point it doesn't meet its purpose, because then you can just use the effective_chat/message shortcuts  imo.

So I think I'll remove the CB shortcuts

aelkheir (approved):
LGTM!

PR #5235, opened 2026-05-19
API 10.0: Live Photos

harshil21:


src/telegram/_files/inputmedia.py
```diff
@@ -142,6 +141,11 @@ class InputPaidMedia(TelegramObject):
     """:const:`telegram.constants.InputPaidMediaType.PHOTO`"""
     VIDEO: Final[str] = constants.InputPaidMediaType.VIDEO
     """:const:`telegram.constants.InputPaidMediaType.VIDEO`"""
+    LIVE_PHOTO: Final[str] = constants.InputPaidMediaType.LIVE_PHOTO
+    """:const:`telegram.constants.InputPaidMediaType.LIVE_PHOTO`
+
+    .. versionadded:: NEXT.VERSION
+    """
```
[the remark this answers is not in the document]
harshil21: it looks fine ?

src/telegram/_files/livephoto.py
```diff
@@ -0,0 +1,119 @@
...
+    Args:
+        file_id	(:obj:`str`): Identifier for the video file which can be used to download or reuse
+            the file.
+        file_unique_id (:obj:`str`): Unique identifier for this file, which
+            is supposed to be the same over time and for different bots.
+            Can't be used to download or reuse the file.
+        width (:obj:`int`): Video width as defined by the sender.
+        height (:obj:`int`): Video height as defined by the sender.
+        duration (:obj:`int` | :class:`datetime.timedelta`): Duration of the video
+            in seconds as defined by the sender.
```
aelkheir, line 53:
```suggestion
            in seconds as defined by the sender.
            |time-period-input|
```

src/telegram/_files/livephoto.py
```diff
@@ -0,0 +1,119 @@
...
+    Attributes:
+        file_id	(:obj:`str`): Identifier for the video file which can be used to download or reuse
+            the file.
+        file_unique_id (:obj:`str`): Unique identifier for this file, which
+            is supposed to be the same over time and for different bots.
+            Can't be used to download or reuse the file.
+        width (:obj:`int`): Video width as defined by the sender.
+        height (:obj:`int`): Video height as defined by the sender.
+        duration (:class:`datetime.timedelta`): Duration of the video
+            in seconds as defined by the sender.
```
aelkheir, line 68:
```suggestion
        duration (:obj:`int` | :class:`datetime.timedelta`): Duration of the video
            in seconds as defined by the sender.
            |time-period-int-deprecated|
```
i think we should start this attr in the deprecated state to have a normalized behavior over these `duration` attributes across classes

tests/_files/test_inputmedia.py
```diff
@@ -628,6 +651,76 @@ def test_with_local_files(self):
...
+        assert input_media_live_photo_dict["has_spoiler"] == input_media_live_photo.has_spoiler
+
+    def test_with_photo_and_video(self, video, photo):
+        # fixtures found in conftest.py
+        input_media_live_photo = InputMediaLivePhoto(video, photo)
+        assert input_media_live_photo.type == self.type_
+        assert input_media_live_photo.media == video.file_id
+        assert input_media_live_photo.photo == photo.file_id
+
+    def test_with_photo_and_file(self, video_file, photo_file):
```
aelkheir:
```suggestion
    def test_with_photo_and_video_files(self, video_file, photo_file):
```

src/telegram/_paidmedia.py
```diff
@@ -251,6 +255,47 @@ def de_json(cls, data: JSONDict, bot: "Bot | None" = None) -> "PaidMediaVideo":
         return super().de_json(data=data, bot=bot)  # type: ignore[return-value]
 
 
+class PaidMediaLivePhoto(PaidMedia):
+    """
+    The paid media is a live photo.
+
+    Objects of this class are comparable in terms of equality. Two objects of this class are
+    considered equal, if their :attr:`live_photo` are equal.
```
aelkheir, line 263:
```suggestion
    considered equal, if their :attr:`live_photo` is equal.
```

src/telegram/_bot.py
```diff
@@ -12287,6 +12289,129 @@ async def save_prepared_keyboard_button(
...
+        show_caption_above_media: bool | None = None,
+        has_spoiler: bool | None = None,
+        disable_notification: ODVInput[bool] = DEFAULT_NONE,
+        protect_content: ODVInput[bool] = DEFAULT_NONE,
+        allow_paid_broadcast: bool | None = None,
+        message_effect_id: str | None = None,
+        suggested_post_parameters: "SuggestedPostParameters | None" = None,
+        reply_parameters: "ReplyParameters | None" = None,
+        reply_markup: "ReplyMarkup | None" = None,
+        *,
```
aelkheir, line 12456:
```suggestion
        *,
        allow_sending_without_reply: ODVInput[bool] = DEFAULT_NONE,
        reply_to_message_id: int | None = None,

```
can also simplify `test_send_live_photo_default_allow_sending_without_reply` a bit

src/telegram/_bot.py
```diff
@@ -12287,6 +12289,129 @@ async def save_prepared_keyboard_button(
             self,
         )
 
+    async def send_live_photo(
```
aelkheir, line 12436: Can we also do `Message.reply_live_photo`, `[Chat,User],send_live_photo` to be consistent with other `send_media` methods

Sorry it seems i bring up this topic too much lately 🫣

aelkheir (changes requested):
Great work! Just a few things here and there

src/telegram/_files/livephoto.py
harshil21: I don't see the need for that anymore, it's already in the type docstring.

src/telegram/_files/livephoto.py
harshil21: imo it doesn't make sense, we've already told that this will be the default. New attributes should be using timedelta by default.

src/telegram/_paidmedia.py
harshil21: It should be "are" since the subject is plural (two objects)

src/telegram/_bot.py
harshil21: No I did just geniunely forget about those shortcuts, good catch.

src/telegram/_files/livephoto.py
aelkheir: Okay that's fair

src/telegram/_paidmedia.py
aelkheir: Ha now `is` is so strange to me

aelkheir (approved):
LGTM!

PR #5239, opened 2026-05-20
Bot API 10.0: Guest Mode

harshil21:
- [x] Fix the changelog entry


src/telegram/_bot.py
```diff
@@ -5848,6 +5849,51 @@ async def answer_web_app_query(
...
+
+        .. versionadded:: NEXT.VERSION
+
+        Args:
+            guest_query_id (:obj:`str`): Unique identifier for the query to be answered.
+            result (:class:`telegram.InlineQueryResult`): An object describing the message to be
+                sent.
+
+        Returns:
+            :class:`telegram.SentGuestMessage`: On success, a sent
```
harshil21:
```suggestion
            :class:`telegram.SentGuestMessage`: On success, a
```

src/telegram/_message.py
```diff
@@ -707,6 +707,21 @@ class Message(MaybeInaccessibleMessage):
         managed_bot_created (:class:`telegram.ManagedBotCreated`, optional): Service message: user
             created a bot that will be managed by the current bot.
 
+            .. versionadded:: NEXT.VERSION
+        guest_bot_caller_user (:class:`telegram.User`, optional): For a message sent by a guest
+            bot, this is the user whose original message triggered the bot's response.
+
+            .. versionadded:: NEXT.VERSION
+        guest_bot_caller_chat (:class:`telegram.Chat`, optional): For a message sent by a guest
+            bot, this is the user whose original message triggered the bot's response.
```
harshil21:
```suggestion
            bot, this is the chat whose original message triggered the bot's response.
```

src/telegram/_message.py
```diff
@@ -707,6 +707,21 @@ class Message(MaybeInaccessibleMessage):
...
+
+            .. versionadded:: NEXT.VERSION
+        guest_query_id (:obj:`str`, optional): The unique identifier for the guest query. Use this
+            identifier with the method :meth:`telegram.Bot.answer_guest_query` to send a response
+            message. If non-empty, the message belongs to a chat of the corresponding business
+            account that is independent from any potential bot chat which might share the same
+            identifier.
```
harshil21:
```suggestion
            message. If non-empty, the message belongs to the chat where the guest 
            bot was summoned, which may not coincide with other existing bot chats 
            sharing the same identifier.
```

src/telegram/_message.py
```diff
@@ -1139,6 +1154,21 @@ class Message(MaybeInaccessibleMessage):
         managed_bot_created (:class:`telegram.ManagedBotCreated`): Optional. Service message: user
             created a bot that will be managed by the current bot.
 
+            .. versionadded:: NEXT.VERSION
+        guest_bot_caller_user (:class:`telegram.User`): Optional. For a message sent by a guest
+            bot, this is the user whose original message triggered the bot's response.
+
+            .. versionadded:: NEXT.VERSION
+        guest_bot_caller_chat (:class:`telegram.Chat`): Optional. For a message sent by a guest
+            bot, this is the user whose original message triggered the bot's response.
```
harshil21:
```suggestion
            bot, this is the chat whose original message triggered the bot's response.
```

src/telegram/_message.py
```diff
@@ -1139,6 +1154,21 @@ class Message(MaybeInaccessibleMessage):
...
+
+            .. versionadded:: NEXT.VERSION
+        guest_query_id (:obj:`str`): Optional. The unique identifier for the guest query. Use this
+            identifier with the method :meth:`telegram.Bot.answer_guest_query` to send a response
+            message. If non-empty, the message belongs to a chat of the corresponding business
+            account that is independent from any potential bot chat which might share the same
+            identifier.
```
harshil21:
```suggestion
            message. If non-empty, the message belongs to the chat where the guest 
            bot was summoned, which may not coincide with other existing bot chats 
            sharing the same identifier.
```

src/telegram/_message.py
harshil21, line 1: a shortcut for `answer_guest_query` here would be ideal

src/telegram/_update.py
harshil21, line 1: Do we want `effective_user/chat` to get the user/chat from `Message.guest_bot_caller_user/chat`? 

Is `Message.from_user` the same as `Message.guest_bot_caller_user` for a guest message?

tests/test_update.py
harshil21, line 1: I think the shortcut test methods need to be updated?

src/telegram/_update.py
harshil21: @Poolitzer / @aelkheir thoughts?

aelkheir: TL;DR
No not always. They'll be the same only when a user invokes the bot with @ and then the same user replies to the bot response.

Did some experimenting and `Message.guest_bot_caller_user/chat` does not get populated in a regular top-level `Update.guest_message` unless another user replies to the bot response message, then the reply triggers another `guest_message`  update which has `Update.guest_message.reply_to_message.guest_bot_caller_user` where `guest_bot_caller_user` in the initiator of the original bot response and `Update.from_user` will be the initiator of the current one.

In that case the current behavior is reasonable i guess (?) `Message.from_user`/`Message.sender_chat` will always give us the initiator of the current bot response.

harshil21: This makes sense. Thanks for testing! The current behaviour is reasonable. We want to always get the current user and not the user up the reply chain.

harshil21 (commented):
Apart from the changelog entry, which should update the parent changelog, this LGTM for merge.

Failing test official can be fixed in the central branch. Those codecov failures might be because codecov was recently bought by another company, so we might have to update something on our end..

PR #5241, opened 2026-05-20
Auto insert raises block, fix `Bot` kwargs insertion regression

harshil21:
This PR does two things:

1) Fixes a regression from #5124 and #5122 which causes no Shortcut admonitions to generate, and no keyword arguments for the bot methods to be inserted in the documentation.
2) Expands the insertion logic to now also automatically add the `Raises:` block for bot methods which have a standard Raises message. Thus, the `Raises:` block in `set_game_score` is untouched, for example.

This helps reduce the LoC and makes implementing new API updates just a little bit more faster.

- [x] Manually check if doc built properly as intended

### DIFF
--- a/docs/auxil/sphinx_hooks.py
+++ b/docs/auxil/sphinx_hooks.py
@@ -27,9 +27,11 @@ from sphinx.application import Sphinx
 import telegram
 import telegram.ext
 from docs.auxil.admonition_inserter import AdmonitionInserter
-from docs.auxil.kwargs_insertion import (
+from docs.auxil.bot_insertion import (
+    RAISES_BLOCK,
     check_timeout_and_api_kwargs_presence,
     find_insert_pos_for_kwargs,
+    find_insert_pos_for_raises,
     get_updates_read_timeout_addition,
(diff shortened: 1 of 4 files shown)


docs/auxil/sphinx_hooks.py
```diff
@@ -101,13 +103,15 @@ def autodoc_process_docstring(
     """
 
     # 1) Insert the Keyword Args and "Shortcuts" admonitions for the Bot methods
-    method_name = name.rsplit(".", maxsplit=1)[0]
+    method_name = name.rsplit(".", maxsplit=1)[-1]
```
aelkheir, line 106: Nice catch!

src/telegram/_bot.py
aelkheir, line 1: `do_api_request()` fails the if check/needs its raises block inserted manually

aelkheir (commented):
:shipit: 
Just one comment.
PS:
`‎changes/unreleased/5240.iJXyH8RNWNNQpC47SfNHzw.toml` needs updating

PR #5246, opened 2026-05-26
Autogenerate Attribute docstrings

### LINKED ISSUE #5242: [Documentation] Autogenerate the Attributes section for all classes
Currently, we manually write out the Attributes section for every single class. Almost all attributes are simply copied over from the Args section and only a few things change, for the majority of cases:

1. The `, optional` -> `Optional.`
2. Sequence[...] -> tuple[...]

I propose automating this task. This would help delete thousands of lines of code, make reviewing faster, reduce package size, and also help eliminate a whole class of documentation bugs. 

How would this be done? We already do something very similar with the Bot keyword argument insertion and with #5241. This would be done in a similar manner:

1. Check if the class has a public attribute not already documented
2. If it does, skip, since this was manually documented (perhaps because it's an edge case)
3. If it does not, look at the Args section to see if we have the same name so we can copy over the description
4. If we do, copy the description, admonitions (like the versionchanged) and apply the transforms as mentioned above
5. If we don't, raise a warning, telling the user to manually document since nothing suitable could be found to copy.

There are probably going to be a few edge cases with this logic, which is why the ability to manually document some attributes is a good middle ground.

harshil21:
Closes #5242 

Took the time to write a full blown docstring parser to do this. So far I've found it pretty reliable since it caught quite a few missing / malformed docstrings.

One thing I changed: the |tupleclassattrs| and |sequenceclassargs| versionchanged notes were deleted. I did this because they were done for v20.0 and it's been many years since that. That way we can unify the argument and attribute docstring.

I manually checked the docs for most edge cases and confirmed that they are correctly handled. Feel free to build the docs and check a random page's to compare the before and after.

----

_AI assisted disclosure_: The tests were >90% written by AI (I looked over all of them, and they were quite satisfactory). It also did some of the parser/insertion logic, which I later fine tuned when I found edge cases / logic errors. 

Merge after #5241


docs/auxil/attribute_inserter.py
```diff
@@ -0,0 +1,258 @@
...
+        args_entries: dict[str, DocstringEntry] = (
+            args_section.entries if args_section is not None else {}
+        )
+        args_names: set[str] = set(args_entries.keys())
+
+        properties_on_class: set[str] = {
+            name for name, _ in inspect.getmembers(obj, lambda o: isinstance(o, property))
+        }
+
+        # Warn about own public slots that have no documentation source.
```
aelkheir:
```suggestion
        # Raise when own public slots have no documentation source.
```

docs/auxil/sphinx_hooks.py
```diff
@@ -113,11 +112,12 @@ def autodoc_process_docstring(
     ):
```
aelkheir, line 114:
```suggestion
    ):
        parser = DocstringParser(lines)

```

docs/auxil/sphinx_hooks.py
aelkheir, line 92:
```text
    Use this method to automatically insert the Attributes block, wherever applicable,
       for the Telegram objects.
```
somewhere between 1 and 2 i think

docs/auxil/attribute_inserter.py
```diff
@@ -0,0 +1,258 @@
...
+        is_optional=is_optional,
+        all_lines=tuple(lines),
+    )
+
+
+@dataclass(frozen=True, slots=True)
+class DocstringEntry:
+    name: str
+    type_str: str
+    is_optional: bool
```
aelkheir, line 80:
```suggestion
    name: str | None
    type_str: str | None
    is_optional: bool | None
```
Other sections (other than `Args` and `Attributes`) will have only `all_lines` set if i understood correctly

changes/unreleased/5240.iJXyH8RNWNNQpC47SfNHzw.toml
```diff
@@ -1,6 +1,10 @@
-documentation = "Documentation Improvements"
+documentation = """Documentation Improvements
+
+* Autogenerate "Attributes" section for easier maintainence and reliability.
+"""
 
 pull_requests = [
     { uid = "5240", author_uids = ["harshil21", "Poolitzer"] },
     { uid = "5241", author_uids = ["harshil21"] },
+    { uid = "5245", author_uids = ["harshil21"] },
```
aelkheir:
```suggestion
    { uid = "5246", author_uids = ["harshil21"] },
```
are the other two PRs relevant to this?

docs/auxil/attribute_inserter.py
```diff
@@ -0,0 +1,258 @@
...
+    entries[name] = DocstringEntry(
+        name=name,
+        type_str=type_str,
+        is_optional=is_optional,
+        all_lines=tuple(lines),
+    )
+
+
+@dataclass(frozen=True, slots=True)
+class DocstringEntry:
```
aelkheir, line 77: Feels a bit strange for sections other than Args & Attributes to have instances of this classes. They'll put every thing in a single entry to my understanding.

However for the sake of keeping things simple this is not a deal breaker for me :)

docs/auxil/bot_insertion.py
```diff
@@ -91,23 +91,7 @@
...
-
-
-def find_insert_pos_for_raises(lines: list[str]) -> int:
-    """Finds the correct position to insert the Raises block and returns the index."""
-    if "Raises:" in lines:
-        return -1  # Don't insert if there's already a Raises block
-    return len(lines)  # Insert at the end if there's no Raises block
-
-
 def check_timeout_and_api_kwargs_presence(obj: object) -> int:
```
aelkheir: docstring bellow seems valid. Return type not though

aelkheir (commented):
Apologies for the overdue review! Changes look pretty good to me. I left some minor comments

One thing that's sad is dropping the historical `.. versionchanged` notes :') but not a big problem git still has them.

docs/auxil/sphinx_hooks.py
harshil21: good catch

docs/auxil/attribute_inserter.py
harshil21: hmm, no? DocstringEntry is only instantiated for `KNOWN_SECTION_TITLES`, and the ENTRY_PATTERN is only matched when all of them are available

changes/unreleased/5240.iJXyH8RNWNNQpC47SfNHzw.toml
harshil21: I will just correct this in the new doc fixes changelog note

docs/auxil/attribute_inserter.py
harshil21: Same reply as above, DocstringEntry won't be generated if it doesn't find a match for the regex which is only found under Args/Attributes/Keyword Args

docs/auxil/attribute_inserter.py
aelkheir: ah my bad, at first i read the if checks in `_parse_section_entries` as `if m: .. else:` instead of `elif`

aelkheir (approved):

PR #5259, opened 2026-06-08
Support Python 3.15 beta

### LINKED ISSUE #5231: Support Python 3.15 feature: Lazy imports

With the [Python 3.15 beta out](https://blog.python.org/2026/05/python-3150-beta-1/), we should test if our library can work with the new version without any problems.

One of the features we can instantly make use of is [Lazy Imports](https://peps.python.org/pep-0810/). Currently, importing the `telegram` module takes about 0.25 seconds on my machine with Python 3.14.3.

Lazy imports should help cut that down by a little bit at least. 

- Add 3.15 runners to the test suite
- Migrate 3.14t runner to 3.15t runner
- Use the `__lazy_modules__` feature on our `__init__` files so that the user doesn't have to import objects they won't be using. 

Keep imports inside functions, but that is unpythonic and ugly

### Additional context

Whenever we drop support for 3.14, we can also get rid of our `TYPE_CHECKING` blocks, as that import won't be evaluated at import time, which should also help reduce circular import errors too.

harshil21:
Closes #5231 

I could not get the lazy imports feature to actually fully work with our library, somewhere along the path all dependencies are eagerly resolved. Doing `lazy import telegram` does work and everything is lazy, until you first use just one class, and then everything is imported all at once, which defeats the point. I may look more deeply into this later.

- There are other some non-critical warnings which show up, related to sphinx, but so far I've not noticed that breaking anything. Tracking issue: https://github.com/python/cpython/issues/149671

- This PR also bumps the cryptography version, which allows it to build on 3.15. 

- Free threaded build fails on 3.15 due to https://github.com/pyca/cryptography/issues/14983. For now we should say that we don't fully support 3.15t.


Poolitzer (approved):
LGTM

PR #5268, opened 2026-06-17
Fix parse_lpo_and_dwpp not raising ValueError when disable_web_page_preview=False

JSap0914:
## Bug

`parse_lpo_and_dwpp` is documented to raise `ValueError` when both
`disable_web_page_preview` and `link_preview_options` are supplied
(they are mutually exclusive). However, when `disable_web_page_preview=False`
the guard used a truthy check:

```python
if disable_web_page_preview and link_preview_options:
```

`False and X` short-circuits to `False`, so no `ValueError` is raised.
The code then falls through to the second condition:

```python
if disable_web_page_preview is not None:
    link_preview_options = LinkPreviewOptions(is_disabled=disable_web_page_preview)
```

which **silently overwrites the caller's `link_preview_options`** with
`LinkPreviewOptions(is_disabled=False)`, discarding the user-supplied value.

## Fix

Change the guard to `is not None` so any explicit `bool` value triggers the
mutual-exclusivity check:

```python
if disable_web_page_preview is not None and link_preview_options:
```

## Verification

```
pytest tests/_inline/test_inputtextmessagecontent.py -v
```

Before fix: `test_mutually_exclusive` passes only the `True` sub-case; the new
`False` sub-case fails (`ValueError` is not raised).

After fix: all 7 tests in the file pass.

> AI-assisted: this change was prepared with the help of an AI coding assistant.

### DIFF
--- a/changes/unreleased/5268.GKZnfBkMcLqBC4rTgznsnY.toml
+++ b/changes/unreleased/5268.GKZnfBkMcLqBC4rTgznsnY.toml
@@ -1,0 +1,5 @@
+bugfixes = "Fixed `parse_lpo_and_dwpp` silently overwriting `link_preview_options` when `disable_web_page_preview=False`."
+[[pull_requests]]
+uid = "5268"
+author_uids = ["JSap0914"]
+closes_threads = []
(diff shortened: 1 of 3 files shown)


changes/unreleased/5268.GKZnfBkMcLqBC4rTgznsnY.toml
```diff
@@ -0,0 +1,5 @@
+bugfixes = "Fixed `parse_lpo_and_dwpp` silently overwriting `link_preview_options` when `disable_web_page_preview=False`."
```
harshil21:
```suggestion
bugfixes = "Fixed ``parse_lpo_and_dwpp`` silently overwriting ``link_preview_options`` when ``disable_web_page_preview=False``."
```

harshil21 (commented):
Thanks for the very subtle bug catch. This didn't have any bad effects since None-ness is still falsy, so it worked just fine. But yes, it's important to have correctness.

JSap0914:
Thanks @harshil21 — applied your suggestion; the chango fragment now uses double backticks for all identifiers.

harshil21 (approved):

PR #5285, opened 2026-06-30
Fix create_deep_linked_url accepting 4-char usernames below Telegram's 5-char minimum

### LINKED ISSUE #5284: create_deep_linked_url accepts 4-character bot usernames, but Telegram requires a minimum of 5
## Bug Description

### Steps to Reproduce

```python

```

### Expected behaviour

### Actual behaviour

### Root Cause

```python
```

### Operating System

Any

Any

### Relevant log output

N/A

JSap0914:
Closes #5284

## Bug

`create_deep_linked_url` validated `bot_username` with `len(bot_username) <= 3`, so a 4-character username (e.g. `"xbot"`) passed without error. The [Telegram docs](https://core.telegram.org/bots/features#deep-linking) explicitly state that usernames must be **5-32 characters long**.

## Fix

- Changed the guard from `len(bot_username) <= 3` to `len(bot_username) < 5`.
- Updated the Raises docstring from "less than 4 characters" to "less than 5 characters".
- Added a new test case confirming that a 4-character username raises `ValueError`.

## Verification

All 183 tests in `tests/test_helpers.py` pass.

### DIFF
--- a/changes/unreleased/5285.3x2uKLIG59gJe4TLjQgYUq.toml
+++ b/changes/unreleased/5285.3x2uKLIG59gJe4TLjQgYUq.toml
@@ -1,0 +1,5 @@
+bugfixes = "Fixed ``create_deep_linked_url`` silently accepting 4-character bot usernames, which are invalid per Telegram's 5-character minimum."
+[[pull_requests]]
+uid = "5285"
+author_uids = ["JSap0914"]
+closes_threads = ["5284"]
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -113,8 +113,10 @@ class TestHelpers:
 
         with pytest.raises(ValueError, match="valid bot_username"):
             helpers.create_deep_linked_url(None, None)
-        with pytest.raises(ValueError, match="valid bot_username"):  # too short username, 4 is min
+        with pytest.raises(ValueError, match="valid bot_username"):  # too short username, 5 is min
             helpers.create_deep_linked_url("abc", None)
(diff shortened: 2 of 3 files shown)


tests/test_helpers.py
```diff
@@ -113,8 +113,10 @@ def test_create_deep_linked_url(self):
 
         with pytest.raises(ValueError, match="valid bot_username"):
             helpers.create_deep_linked_url(None, None)
-        with pytest.raises(ValueError, match="valid bot_username"):  # too short username, 4 is min
+        with pytest.raises(ValueError, match="valid bot_username"):  # too short username, 5 is min
             helpers.create_deep_linked_url("abc", None)
+        with pytest.raises(ValueError, match="valid bot_username"):  # 4 chars is still too short
+            helpers.create_deep_linked_url("abcd", None)
```
harshil21, line 119: please only keep one test, the 4 character one is good enough.

changes/unreleased/5285.3x2uKLIG59gJe4TLjQgYUq.toml
```diff
@@ -0,0 +1,5 @@
+bugfixes = "Fixed ``create_deep_linked_url`` silently accepting 4-character bot usernames, which are invalid per Telegram's 5-character minimum."
```
harshil21, line 1:
```suggestion
bugfixes = "Fixed ``create_deep_linked_url`` silently accepting 4-character bot usernames"
```

harshil21 (changes requested):
Thanks for the catch. Just 2 comments

harshil21:
Closing b/c of inactivity

PR #5297, opened 2026-07-15
Fix PicklePersistence flushing empty data

xsneh:
## Summary

  Fixes a `PicklePersistence` data resurrection issue when `on_flush=True`.

  When persisted data was emptied, for example by dropping the last entry from
  `user_data`, `flush()` skipped writing the empty state because it checked
  truthiness instead of whether the data had been initialized. This left stale data
  on disk, so deleted data could reappear after recreating `PicklePersistence` with
  the same filepath.

  This changes `flush()` to write initialized empty containers and adds regression
  coverage for both single-file and multi-file persistence.

  ## Reproduction

  1. Create `PicklePersistence` with `on_flush=True`.
  2. Store one `user_data` entry and call `flush()`.
  3. Drop that same user entry and call `flush()` again.
  4. Recreate `PicklePersistence` with the same filepath.
  5. Before this fix, the dropped user data was restored from disk.

  ## Tests

  ```bash
  uv sync --extra all --group tests
  uv run pytest tests/ext/test_picklepersistence.py
  ```

  Result:

  ```text
  82 passed
  ```

  ## Check-list for PRs

  - [ ] Added `.. versionadded:: NEXT.VERSION`, ``.. versionchanged:: NEXT.VERSION``,
  ``.. deprecated:: NEXT.VERSION`` or ``.. versionremoved:: NEXT.VERSION`` to the
  docstrings for user facing changes (for methods/class descriptions, arguments and
  attributes)
  - [ ] Added new classes & modules to the docs and all suitable ``__all__`` s
  - [x] Checked the [Stability Policy](https://docs.python-telegram-bot.org/stability_policy.html) in case of deprecations or changes to documented behavior


harshil21 (approved):
LGTM, thanks for the fix! This should be a very rare bug to hit, but nevertheless correctness matters.

PR #5305, opened 2026-07-24
Reduce pytest collection time by 80%

harshil21:
One of the annoying things about running pytest now is that the collection time on my computer takes a really long time, over 8 seconds for me. I thought this was a good exploratory problem for AI to solve and damn did it do a good job - collection time is now less than 2 seconds for me.

It bumps `pytest-asyncio` to the latest version which solves the collection time issue, so I guess this is a 2 birds 1 stone situation.

Also deleted the windows event loop workaround, let's see if those tests pass normally now.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -82,9 +82,6 @@
   // Increase the number of PR's Renovate can create in a hour. Default is 2.
   "prHourlyLimit": 5,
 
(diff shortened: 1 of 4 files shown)


harshil21:
Can confirm windows tests work normally and pass. Merging as a stack.

PR #5312, opened 2026-07-29
fix: correct bot_username length validation in create_deep_linked_url()

### LINKED ISSUE #5284: create_deep_linked_url accepts 4-character bot usernames, but Telegram requires a minimum of 5

cyforkk:
## Description

Telegram requires bot usernames to be 5-32 characters long.
The previous check `len(bot_username) <= 3` incorrectly accepted
4-character usernames. Changed to `len(bot_username) < 5` to match
Telegram's actual minimum.

Also updated the docstring and added a test case for 4-char usernames.

Fixes python-telegram-bot/python-telegram-bot#5284

### DIFF
--- a/src/telegram/helpers.py
+++ b/src/telegram/helpers.py
@@ -178,10 +178,10 @@ def create_deep_linked_url(
     Raises:
         :exc:`ValueError`: If the length of the :paramref:`payload` exceeds \
         :tg-const:`telegram.constants.MessageLimit.DEEP_LINK_LENGTH` characters,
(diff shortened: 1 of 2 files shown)


harshil21:
Please see my comment in #5285

harshil21:
Closing b/c of inactivity and exact duplication of the other PR without looking at comments

PR #5328, opened 2026-08-13
Agent/fix falsy callback data

Vamp1reAchao:
Fixed CallbackQueryHandler incorrectly ignoring falsy callback_data values such as False, 0, and "" when a pattern is configured. Added regression tests.


harshil21:
There is no use case for this. pure ints / bools are technically arbitrary callback data already. Moreover your proposal doesn't actually fix it, if the callback_data is e.g. `False`, and the pattern is `False`, the handler still isn't triggered.

PR #5330, opened 2026-08-13
Rate limit override

Vamp1reAchao:
Fix AIORateLimiter handling of rate_limit_args=0.

Previously, passing 0 could be treated as a falsy value and fall back to the configured default retry limit. This change distinguishes an explicitly provided 0 from None, allowing callers to intentionally disable retries for a specific request while preserving the global default behavior when no override is provided.

Includes regression tests covering both explicit 0 and fallback-to-default behavior.

### DIFF
--- a/changes/unreleased/rate-limit-args-zero.toml
+++ b/changes/unreleased/rate-limit-args-zero.toml
@@ -1,0 +1,1 @@
+bugfixes = "Respect an explicit rate_limit_args=0 in AIORateLimiter"
--- a/changes/unreleased/rate-limit-args-zero.toml
+++ b/changes/unreleased/rate-limit-args-zero.toml
@@ -1,1 +1,6 @@
 bugfixes = "Respect an explicit rate_limit_args=0 in AIORateLimiter"
+
+[[pull_requests]]
+uid = "5330"
+author_uids = ["Vamp1reAchao"]
+closes_threads = []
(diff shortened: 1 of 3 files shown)


changes/unreleased/rate-limit-args-zero.toml
```diff
@@ -0,0 +1 @@
+bugfixes = "Respect an explicit rate_limit_args=0 in AIORateLimiter"
```
harshil21, line 1: The file is missing a few more attributes, can you look at the other .tomls and update this?

tests/ext/test_ratelimiter.py
```diff
@@ -344,7 +357,6 @@ async def test_rate_limiting_no_chat_id(self, bot):
                 assert all(task.done() for task in non_chat_tasks.values())
                 assert all(task.done() for task in chat_tasks.values())
         finally:
-            # cleanup
```
harshil21, line 347: please revert this

harshil21 (changes requested):

changes/unreleased/rate-limit-args-zero.toml
Vamp1reAchao: Done, updated the changelog entry and reverted the unrelated test change. Thanks!

harshil21: Can you also fix the failing prek action? Don't worry about the other failing tests

Vamp1reAchao: Done, i fix it. Thanks!)

harshil21:
Not merging because your latest commit had a weird regression which a human can never make.

PR #5334, opened 2026-08-23
WIP: Switching to Dataclasses

### LINKED ISSUE #5279: Switching to Dataclasses
Python dataclasses has been something this library has always eyed keenly. A lot of the work and discussion was done earlier in #2698, and a number of things have changed since that issue was created, which has once again opened the door for us to now actually switch to dataclasses.

Reiterating the main benefits we get from the switch: 

1. No more slots and `__init__` boilerplate
2. Letting dataclass machinery handle equality comparison for us
3. Drastically less maintenance overhead, even when writing tests, since we don't need slots and equality tests.

## Implementation

I'll go over things which were discussed in the issue and whether they are still applicable:

### Minor details

1. **Immutability**: The solution we went for is rolling our own version of freezing, via a `_frozen` attribute. Dataclasses have their own version via the `frozen=True` argument. Here are the pros/cons to using that:

    1. ➖ IISC, dataclasses does not have a way to temporarily unfreeze an object, which means using `object.__setattr__` as a workaround, specially in tests.
    2. ➖/➕ There's been a few minor issues when mixing slots and frozen: https://github.com/python/cpython/issues/105936, though this has been fixed in later versions of python 3.13 and above? If users encounter this somehow below 3.13, we may just have to ask them to upgrade, but the scenario in that issue should not happen in our library. 
    3. ➕ Initializing new classes is at 3-10x faster[^1] if we use the dataclass approach. I had codex write a [script](https://gist.github.com/harshil21/57ca4fee272277c1a44a44fbbb3e9805) for that. The results were surprising since I was expecting the opposite.. 

Personally, I'm slightly in favour of using the dataclass frozen=True approach because of the "free" performance gain. 

2. **Equality comparison**: Dataclasses also has its own comparison method, using `compare=True`, which removes the need for our manual `_id_attrs` logic. I currently don't see any reason why we shouldn't use their approach. The only annoyance is a little more boilerplate to write if you don't want to include a field for equality comparsion. You have to do a `field(eq=False)`. With `frozen=True`, the `__hash__` method will also be handled for us and will check the attributes we want it to check.

3. **`__repr__`**: Dataclasses can also do this for us, but now that we have our own approach, which customizes the look by dropping None values and empty containers, we can just stick to using that. 

4. **`__slots__`**: One of the [showstoppers previously](https://github.com/python-telegram-bot/python-telegram-bot/issues/2698#issuecomment-1237415687), we no longer have that issue since we only support 3.10+. 

5. **`__match_args__`**: Since we don't have a use for them, we should disable them globally for all TelegramObjects.

6. Some class constants will need to be annotated with **`ClassVar`**, specially the `Final` ones.

7. We should mention that `dataclasses.asdict` and `replace` is not explicitly supported by our API and users should use `to_dict`.

8. Replacing `__slots__` with **`fields()`** when wanting to get all attributes, this means almost all of our internal TO methods and even tests would need to change. 

### Major details

9. **Handling of Init only fields and transform like fields:**

At the moment, we have a couple of classes which call other functions in the `__init__`, like calling `to_timedelta`, or `parse_sequence_arg`. For these kind of classes, I propose using the `metadata` field in [`Field`](https://docs.python.org/3/library/dataclasses.html#dataclasses.Field). This metadata will contain the function for transformation. 

`TO.__post_init__` will be defined on the base class which will automatically run every time after the regular `__init__`. It's at this time we will iterate through any metadata fields and apply the transformations.

An additional use of the metadata field is to use it for deprecated arguments / attributes, e.g. `duration` can be an InitVar, and then we can use a regular `@property` to raise warnings. The metadata field can be used to store the value in `_duration` for example. 

_Side_: We can't type hint the argument and attribute differently in dataclasses, so I propose to make the `Sequence` accepting arguments as `tuple`, since users are more likely to access attributes rather than initializing a class with them.

10. **Inheritance**: This is still a problem as mentioned in https://github.com/python-telegram-bot/python-telegram-bot/issues/2698#issuecomment-1237415687. 
    Possible solutions:

    i. The best solution imo is still making the defaulted inherited field as keyword only. This means that `_BaseMedium`, `_BaseThumbedMedium`, etc would make all of its default fields as kw only, which in turn changes the signature of classes which inherit them. 
This is of course a breaking change without a deprecation period, but the stability policy does outline an extreme case where we can't provide compatibility. Moreover, most of these classes are not typically instantiated by users.

    ii. Non-breaking: Override the `__init__` generated from the dataclass, define a static list of positional arguments expected, and then check the passed arguments to check if they were passed positionally in the right order. I had GPT 5.5 do this, but I felt it was too black magicky and was over ~100 lines. It could work, but it can also be a maintainence burden if it's buggy.

    iii. Inline inherited arguments: This is basically removing the internal classes and all of its attributes and copying them over to the subclasses. That way we can control what order the class is instantiated with. Not a big fan of this either..

----

There could be 2nd order effects which I haven't forseen so far, but this should be the bulk of the changes. 

What does the dev team feel about this? 

[^1]: Benched on my uv installed python 3.14.6.

aelkheir:
Closes #5279
Discussed Priorly: #2698

<TODO: follow up on some previous discussion points>


aelkheir:
So I had this drafted for a bit of time now, I intended to send it earlier on to get your feedback on it, especially the logic around the after-processing of generated `__init__` and whether that path seems acceptable to you.

Then I was tempted to walk till (near) the finish line and see how (simple v.s complicated) it will turn to be.

-----

> Future Stability Policy handling
arg: InitVar and _arg = field(init=False)

This did not age well.

```python
from dataclasses import InitVar, dataclass, field


@dataclass(frozen=True)
class Example:
    value: InitVar[int | None] = None
    _value: int | None = field(init=False, default=None)

    def __post_init__(self, value: int | None, /):
        object.__setattr__(self, "_value", value)

    @property
    def value(self) -> int | None:
        print("value is deprecated and will be removed")
        return self._value


a = Example()
print(a._value)  # output: <property object at 0x7ff25e629e40>

b = Example(10)
print(b._value)  # output: 10

```
When not providing an explicit argument for value, the default `None` is overwritten by the property object.
A relevant CI issue also is that, both linting (ruff) and type checking (mypy) report:
`error: Name "value" already defined/Redefinition of unused \`value\``

So i tried giving this [alias Field specifier parameters](https://typing.python.org/en/latest/spec/dataclasses.html#field-specifier-parameters) a shot:
> alias is an optional str parameter that provides an alternative name for the field. This alternative name is used in the synthesized __init__ method.

The outcome of it is pretty neat and less verbose, however, and just like `converter` we have to provide runtime impl

<details>
  <summary>Aliased parameter MWE</summary>

```python
# No runtime support fo alias, this is just to demonstrate type checkers support
from dataclasses import MISSING, dataclass, field
from typing import Any, TypeVar, dataclass_transform

_T = TypeVar("_T")


def tg_field(
    *,
    default: Any = MISSING,
    alias: str | None = None,
) -> Any:
    return field(default=default)


@dataclass_transform(field_specifiers=(tg_field,))
def tg_dataclass(cls: type[_T]) -> type[_T]:
    return dataclass(frozen=True)(cls)


@tg_dataclass
class Example:
    _value: int | None = tg_field(alias="value", default=None)

    @property
    def value(self) -> int | None:
        print("value is deprecated and will be removed")
        return self._value


# Wroks in runtime
c = Example(_value=10)

# TypeError: Example.__init__() got an unexpected keyword argument 'value'
d = Example(value=10)
```
</details>

Its well supported in type checkers that common IDEs use 

<details>
  <summary>IDE Support</summary>
Pyright
Neovim with Pyright


VSCode Pylance
image


Pycharm, should be supported from 2024.3 onwards. https://youtrack.jetbrains.com/issue/PY-73832/Pycharm-doesnt-properly-recognizes-aliases-for-Pydantic-dataclass
image
</details>

mypy also support aliases running it through the above MWE gives:
`Unexpected keyword argument "_value" for "Example"; did you mean "value"?  [call-arg]`

> I honestly don't know what to do about temp unfreezing ...

I just did nothing. Excluding test files, there was very minimal unfreezing of objects outside of old `__init__`s, so I just `object.__seattr__` these. For tests, mutations use a combination of a new `with _unfrozen(obj)` helper and `object.__seattr__`

#### CI
Sadly pylint and mypy have a bunch of false positives that we'll have to fight some way.
pylint: https://github.com/pylint-dev/pylint/issues/4899, https://github.com/pylint-dev/pylint/issues/7437
mypy: https://github.com/python/mypy/issues/17547

----

I still got some polishing to do, specifically I want to:
- [ ] Remove the enum converters and simplify with functool.partial
- [ ] Add a `TO._set()`. `object.__setattr__` was still a one-liner but it takes too much horizontal space 
- [ ] Add deprecation notes to the changelog, mostly parameter order changes
- [ ] Add tests for `_utils/dataclass.py`
- [ ] Update Checklist for PRs
- [ ] Experiment with no longer tampering with the generated `__init__` signatures and annotations and instead make `TO._de_json()` dataclasses-first,  meaning to look up the correct annotation from  `dataclasses.fields(obj)` directly rather than inpecting `__init__`.
- [ ] Apply similar treatment to test_official


The last two points can further simplify `_utils/dataclass.py`, however the current implementation of the helper is the most truthful about what `__init__` actually accepts but also fights python and standard dataclasses behaviour.

I'm going to be working on these items in the following days, would be great to get some initial review since `tests/` are passing now

Doesn't have to be the whole thing, just `_utils/dataclass.py` and the changelog.

PR #5335, opened 2026-08-24
Add `filters.IS_BOT`

### LINKED ISSUE #5315: Add filters.IS_BOT — no filter exposes User.is_bot (needed for multi-bot group chats)
## Feature request

Add a `filters.IS_BOT` (and its negation) so handlers can filter on whether the
message author is a bot, without hand-rolling a `MessageFilter` subclass.

### Rationale

`User.is_bot` has been part of the `User` model for as long as the Bot API has
had it, and it is the natural discriminator for multi-bot group chats. But
`filters.py` never exposes it:

```console
$ python -c "
import re
src = open('telegram/ext/filters.py').read()
print('is_bot occurrences in filters.py:', src.count('is_bot'))
m = re.search(r'__all__\s*=\s*\((.*?)\)', src, re.S)
print('total filters exported:', len([x for x in m.group(1).split(',') if x.strip()]))
"
is_bot occurrences in filters.py: 0
total filters exported: 66
```

66 exported filters, including several in exactly this neighbourhood —
`ViaBot`, `SenderChat`, `ChatType`, `IS_AUTOMATIC_FORWARD`, `IS_TOPIC_MESSAGE`,
`IS_FROM_OFFLINE` — and none of them answers "did a bot write this?".

The closest existing filters are not substitutes:

- **`ViaBot`** matches inline messages sent *via* a bot (`message.via_bot`).
  A bot posting normally in a group has `via_bot is None`, so `ViaBot` misses it.
- **`SenderChat`** covers channel/anonymous-admin senders, not bot users.
- **`IS_AUTOMATIC_FORWARD`** is about channel-to-discussion-group forwards.

So today the only way is a custom filter:

```python
class _IsBot(filters.MessageFilter):
    def filter(self, message):
        return bool(message.from_user and message.from_user.is_bot)

IS_BOT = _IsBot(name="filters.IS_BOT")
```

That is ~5 lines, but it is 5 lines every project rewrites, it is easy to get
subtly wrong (`from_user` is `None` for channel posts — a naive
`message.from_user.is_bot` raises `AttributeError`), and it is undiscoverable:
a user scanning the filters list reasonably concludes PTB cannot do it.

### Use case

I run a fleet of machines (2x DGX Spark, 2x Jetson Thor, AGX Orin, Orin Nano)
each hosting an agent that joins one shared Telegram group. Agents need to
treat bot-authored messages differently from human ones — typically: read them
for context, but only *act* when a human addressed them, so two bots cannot
ping-pong into a loop.

That is a routing decision on `is_bot`, made at handler-registration time,
which is precisely what `filters` is for:

```python
# react to humans
app.add_handler(MessageHandler(filters.TEXT & ~filters.IS_BOT, on_human))
# observe peers without replying
app.add_handler(MessageHandler(filters.TEXT & filters.IS_BOT, on_peer))
```

versus today, where the bot/human split has to happen *inside* the callback,
after the handler already matched.

Multi-bot groups are increasingly common (agent fleets, bridges, relays,
notification bots coexisting), and every one of them needs this discriminator.

### Proposed change

Add to `telegram/ext/filters.py`, following the existing `IS_AUTOMATIC_FORWARD`
/ `IS_TOPIC_MESSAGE` pattern:

```python
class _IsBot(MessageFilter):
    __slots__ = ()

        return bool(message.from_user and message.from_user.is_bot)


IS_BOT = _IsBot(name="filters.IS_BOT")
"""Messages whose sender is a bot (:attr:`telegram.User.is_bot`).

Note:
    Channel posts and anonymous-admin messages have no :attr:`Message.from_user`
    and therefore do **not** match; see :attr:`filters.SenderChat` for those.
"""
```

Then export `IS_BOT` in `__all__`. Negation already works via `~filters.IS_BOT`,
so no separate `IS_NOT_BOT` is needed.

The `from_user is None` guard is the part worth having in the library rather
than in every downstream copy.

### Caveat worth documenting either way

Independent of this filter: Telegram's **privacy mode is ON by default**, so a
bot in a group only receives messages that mention it or reply to it — including
messages from other bots. `Bot.can_read_all_group_messages` reports the current
setting.

Whether or not `IS_BOT` lands, a sentence in the group-chat docs noting that
`can_read_all_group_messages` must be `True` (BotFather → `/setprivacy` →
Disable) for a bot to observe other bots would save people a long debugging
session — the failure is completely silent: correct code, no updates delivered,
no error.

I'm happy to open a PR with the filter, tests, and that docs note if the
direction is acceptable.

### Environment

python-telegram-bot 22.6, Python 3.11.

fahadhewad:
Adds `filters.IS_BOT`, a ready to use filter that matches updates whose `telegram.Update.effective_user` is a bot. Until now nothing in the filters module exposed `telegram.User.is_bot`, so telling bot authored messages apart from human ones had to happen inside the callback. With this filter the decision can be made at handler registration time, and `~filters.IS_BOT` covers the human only case.

The filter is implemented as an `UpdateFilter` over `Update.effective_user`, which is the same approach the existing `PREMIUM_USER` and `USER_ATTACHMENT` filters take for other `telegram.User` attributes. It returns `False` when the update carries no user at all, for example for channel posts.

The docstring also points out that in groups a bot only receives messages from other bots when privacy mode is disabled, since that is the setting that decides whether the filter can ever match in a multi bot chat.

Fixes #5315.

## Check-list for PRs

- [x] Added `.. versionadded:: NEXT.VERSION`, ``.. versionchanged:: NEXT.VERSION``, ``.. deprecated:: NEXT.VERSION`` or ``.. versionremoved:: NEXT.VERSION` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)
- [ ] Checked the [Stability Policy](https://docs.python-telegram-bot.org/stability_policy.html) in case of deprecations or changes to documented behavior

### DIFF
--- a/src/telegram/ext/filters.py
+++ b/src/telegram/ext/filters.py
@@ -60,6 +60,7 @@ __all__ = (
     "HAS_PROTECTED_CONTENT",
     "INVOICE",
     "IS_AUTOMATIC_FORWARD",
+    "IS_BOT",
     "IS_FROM_OFFLINE",
     "IS_TOPIC_MESSAGE",
     "LIVE_PHOTO",
@@ -1590,6 +1591,31 @@ IS_AUTOMATIC_FORWARD = _IsAutomaticForward(name="filters.IS_AUTOMATIC_FORWARD")
 """Messages that contain :attr:`telegram.Message.is_automatic_forward`.
 
     .. versionadded:: 13.9
+"""
+
+
+class _IsBot(UpdateFilter):
+    __slots__ = ()
+
+    def filter(self, update: Update) -> bool:
+        # Updates without a user, e.g. channel posts, are not sent by a bot
+        return bool(update.effective_user) and bool(
+            update.effective_user.is_bot  # type: ignore
+        )
+
+
+IS_BOT = _IsBot(name="filters.IS_BOT")
+"""This filter filters *any* message that has a :attr:`bot <telegram.User.is_bot>` as
+:attr:`telegram.Update.effective_user`.
+
+Note:
+    In groups, a bot only receives messages sent by other bots if
+    :attr:`telegram.User.can_read_all_group_messages` is :obj:`True` for it, i.e. if privacy
+    mode is disabled.
+
+.. seealso:: :attr:`telegram.ext.filters.VIA_BOT`
+
+.. versionadded:: NEXT.VERSION
 """
 
 
--- a/tests/ext/test_filters.py
+++ b/tests/ext/test_filters.py
@@ -2161,6 +2161,16 @@ class TestFilters:
         update.message.is_automatic_forward = True
         assert filters.IS_AUTOMATIC_FORWARD.check_update(update)
 
+    def test_filters_is_bot(self, update):
+        assert not filters.IS_BOT.check_update(update)
+        update.message.from_user.is_bot = True
+        assert filters.IS_BOT.check_update(update)
+
+    def test_filters_is_bot_without_user(self):
+        # channel posts have no `from_user`, so there is no user to look at
+        update = Update(0, channel_post=Message(0, dtm.datetime.utcnow(), Chat(0, Chat.CHANNEL)))
+        assert not filters.IS_BOT.check_update(update)
+
     def test_filters_is_from_offline(self, update):
         assert not filters.IS_FROM_OFFLINE.check_update(update)
         update.message.is_from_offline = True


src/telegram/ext/filters.py
```diff
@@ -1593,6 +1594,31 @@ def filter(self, message: Message) -> bool:
...
+
+
+IS_BOT = _IsBot(name="filters.IS_BOT")
+"""This filter filters *any* message that has a :attr:`bot <telegram.User.is_bot>` as
+:attr:`telegram.Update.effective_user`.
+
+Note:
+    In groups, a bot only receives messages sent by other bots if
+    :attr:`telegram.User.can_read_all_group_messages` is :obj:`True` for it, i.e. if privacy
+    mode is disabled.
```
harshil21, line 1614: No need of this note.
```suggestion

```

src/telegram/ext/filters.py
```diff
@@ -1593,6 +1594,31 @@ def filter(self, message: Message) -> bool:
...
+
+.. seealso:: :attr:`telegram.ext.filters.VIA_BOT`
```
harshil21, line 1616:
```suggestion
```

src/telegram/ext/filters.py
```diff
@@ -1593,6 +1594,31 @@ def filter(self, message: Message) -> bool:
...
+    def filter(self, update: Update) -> bool:
+        # Updates without a user, e.g. channel posts, are not sent by a bot
+        return bool(update.effective_user) and bool(
+            update.effective_user.is_bot  # type: ignore
+        )
+
+
+IS_BOT = _IsBot(name="filters.IS_BOT")
+"""This filter filters *any* message that has a :attr:`bot <telegram.User.is_bot>` as
+:attr:`telegram.Update.effective_user`.
```
harshil21, line 1609:
```suggestion
"""This filter filters *any* message that has a :attr:`bot <telegram.User.is_bot>`.
```

Too verbose

src/telegram/ext/filters.py
```diff
@@ -1593,6 +1594,31 @@ def filter(self, message: Message) -> bool:
 """
 
 
+class _IsBot(UpdateFilter):
+    __slots__ = ()
+
+    def filter(self, update: Update) -> bool:
+        # Updates without a user, e.g. channel posts, are not sent by a bot
```
harshil21, line 1601:
```suggestion
```
unnecessary

tests/ext/test_filters.py
```diff
@@ -2161,6 +2161,16 @@ def test_filters_is_automatic_forward(self, update):
...
 
+    def test_filters_is_bot(self, update):
+        assert not filters.IS_BOT.check_update(update)
+        update.message.from_user.is_bot = True
+        assert filters.IS_BOT.check_update(update)
+
+    def test_filters_is_bot_without_user(self):
+        # channel posts have no `from_user`, so there is no user to look at
+        update = Update(0, channel_post=Message(0, dtm.datetime.utcnow(), Chat(0, Chat.CHANNEL)))
+        assert not filters.IS_BOT.check_update(update)
```
harshil21, line 2172:
```suggestion
```

harshil21 (changes requested):

harshil21:
Closing b/c of inactivity

PR #5337, opened 2026-08-25
Forward pool timeouts to polling error callbacks

Mmx233:
This PR makes connection-pool exhaustion observable through polling error callbacks without changing the handling of regular request timeouts.

`HTTPXRequest` currently maps `httpx.PoolTimeout` to `TimedOut`. The polling retry loop intentionally handles every `TimedOut` as an expected request timeout and retries it silently. As a result, `Updater.start_polling(error_callback=...)` cannot distinguish or react to a request that was never sent because no connection was available from the pool.

The change introduces `telegram.error.PoolTimeout` as a subclass of `TimedOut`, translates `httpx.PoolTimeout` to that backend-independent exception, and forwards only `PoolTimeout` to the polling error callback. Existing `except TimedOut` and `isinstance(error, TimedOut)` checks remain compatible, while regular long-poll timeouts retain their current behavior.

## Check-list for PRs

- [x] Added `.. versionadded:: NEXT.VERSION`, ``.. versionchanged:: NEXT.VERSION``, ``.. deprecated:: NEXT.VERSION`` or ``.. versionremoved:: NEXT.VERSION`` to the docstrings for user facing changes (for methods/class descriptions, arguments and attributes)
- [x] Checked the [Stability Policy](https://docs.python-telegram-bot.org/stability_policy.html) in case of deprecations or changes to documented behavior

- If Relevant

### DIFF
--- a/src/telegram/error.py
+++ b/src/telegram/error.py
@@ -37,6 +37,7 @@ __all__ = (
     "InvalidToken",
     "NetworkError",
     "PassportDecryptionError",
+    "PoolTimeout",
     "RetryAfter",
     "TelegramError",
     "TimedOut",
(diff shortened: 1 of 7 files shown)


harshil21:
Sorry this change makes no sense. All TimedOut exceptions are retried immediately anyway, and your change of making it a subclass and then also retrying it immediately introduces no new behaviour at all. There's no need to make the subtype of a TimeOut observable to start_polling.

PR #5339, opened 2026-08-25
Keep a RetryAfter halt until every backoff has finished

### LINKED ISSUE #5338: AIORateLimiter: a RetryAfter halt is released early by any concurrent request finishing
### Steps to Reproduce

`AIORateLimiter` documents that a `RetryAfter` "will halt *all* requests for `retry_after` + 0.1 seconds". It does not, if any other request happens to finish during that window.

`process_request` ends each attempt with:

```python
            except RetryAfter as exc:
                ...
                self._retry_after_event.clear()
                await asyncio.sleep(sleep)
            finally:
                # Allow other requests to be processed
                self._retry_after_event.set()
```

`_retry_after_event` is shared by the whole limiter, but the `finally` runs for *every* request, including ones that never cleared it. So a request that was already past `inner()`'s `await self._retry_after_event.wait()` when the halt began will `set()` it on completion and release a halt it did not establish.

Script (no network, patches nothing):

```python
import asyncio, time
from telegram.error import RetryAfter
from telegram.ext import AIORateLimiter

async def main():
    rl = AIORateLimiter(overall_max_rate=0, group_max_rate=0, max_retries=1)

    async def flooded(*a, **kw):
        if not flooded.hit:
            flooded.hit = True
            raise RetryAfter(2)
        return True
    flooded.hit = False

    async def slow(*a, **kw):
        await asyncio.sleep(0.3)   # an ordinary in-flight request
        return True

    async def fast(*a, **kw):
        return True

    async def request(cb, cid):
        return await rl.process_request(callback=cb, args=(), kwargs={},
            endpoint="sendMessage", data={"chat_id": cid}, rate_limit_args=None)

    t0 = time.monotonic()
    b = asyncio.create_task(request(slow, 2))     # in flight first
    await asyncio.sleep(0.05)
    a = asyncio.create_task(request(flooded, 1))  # floods, halts the bot for 2s
    await asyncio.sleep(0.05)
    await b
    await request(fast, 3)                        # should wait out the halt
    print(f"third request sent {time.monotonic() - t0:.2f}s into a 2s halt")
    await a

asyncio.run(main())
```

### Expected behaviour

The third request waits out the halt (~2.1s).

### Actual behaviour

```
third request sent 0.30s into a 2s halt
```

It is sent as soon as the unrelated in-flight request completes.

The same root cause has a second effect: when two requests are backing off concurrently, the shorter backoff expiring releases the longer one's halt.

Practical impact is that the halt is least effective exactly when it matters most — under load there is almost always another request in flight, so a 429 storm keeps being fed instead of backing off.

### Operating System

Linux

```shell
python-telegram-bot 22.8 (also present on master), aiolimiter 1.2.1, Python 3.14
```

0xSoftBoi:
Closes #5338.

`AIORateLimiter` documents that a `RetryAfter` halts *all* requests for `retry_after + 0.1` seconds. `_retry_after_event` is shared by the whole limiter, but it was `set()` in a `finally` that runs for every request — including requests that never cleared it. So any request already past `inner()`'s `await self._retry_after_event.wait()` when the halt began would release it on completion.

Two reachable consequences:

- an unrelated in-flight request completing ends the halt early (0.30s into a 2s halt, in the issue's script);
- with two requests backing off at once, the shorter backoff expiring releases the longer one.

Both need a request to be *past* the wait when the halt starts, which is the normal case under load — so the halt was weakest exactly when it mattered, and a 429 storm kept being fed.

### Fix

Count the requests currently backing off and set the event only when that count reaches zero. A plain "only the clearer may set it" flag would fix the first case but not the second.

### Tests

Two regression tests, both of which fail on `master` and pass here:

- `test_retry_after_not_released_by_in_flight_request`
- `test_retry_after_not_released_by_shorter_backoff`

They use a small `ScriptedRequest` helper mapping `chat_id -> (latency, retry_after)`. The latency matters: the existing `test_delay_all_pending_on_retry` starts its second request *after* the halt is already in place, so that request blocks at the `wait()` and the bug is invisible. Reproducing it needs a request that is already in flight, which in turn needs the 429 to arrive after some delay, as it does in practice.

`tests/ext/test_ratelimiter.py`: 18 passed, 1 skipped. `ruff check`, `ruff format --check` and `mypy` are clean on both files.

No behaviour change other than restoring what the docstring already promises, so I have not touched the docs.

### DIFF
--- a/changes/unreleased/5339.Kl6xS6VYDf2DSsRNH10eLp.toml
+++ b/changes/unreleased/5339.Kl6xS6VYDf2DSsRNH10eLp.toml
@@ -1,0 +1,5 @@
+bugfixes = "Keep an `AIORateLimiter` `RetryAfter` halt in place until every backoff has finished, instead of releasing it when any concurrent request completes"
+[[pull_requests]]
+uid = "5339"
+author_uids = ["0xSoftBoi"]
+closes_threads = ["5338"]
--- a/src/telegram/ext/_aioratelimiter.py
+++ b/src/telegram/ext/_aioratelimiter.py
@@ -135,6 +135,7 @@ class AIORateLimiter(BaseRateLimiter[int]):
         "_group_time_period",
         "_max_retries",
         "_retry_after_event",
+        "_retry_after_holds",
     )
 
     def __init__(
@@ -171,6 +172,10 @@ class AIORateLimiter(BaseRateLimiter[int]):
         self._max_retries: int = max_retries
         self._retry_after_event = asyncio.Event()
         self._retry_after_event.set()
+        # Number of requests currently sleeping off a RetryAfter. The halt is lifted only
+        # when this drops back to zero, so that neither an unrelated request finishing nor a
+        # shorter backoff expiring can release a halt that is still in effect.
+        self._retry_after_holds: int = 0
 
     async def initialize(self) -> None:
         """Does nothing."""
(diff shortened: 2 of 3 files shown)


harshil21 (approved):
Logic and tests look good to me.

harshil21:
All tests pass locally for me. Merging.

PR #5343, opened 2026-08-30
Add security policy and CodeQL analysis workflow

angelancajas34-beep:


angelancajas34-beep (commented):
# Download and install hosted packages.
pip install --extra-index-url http://localhost:8080/simple/ ...

# or
pip install --extra-index-url http://localhost:8080 ...

# Search hosted packages.
pip search --index http://localhost:8080 ...

# Note that pip search does not currently work with the /simple/ endpoint.

PR #5354, opened 2026-09-07
Fix AIORateLimiter Releasing a RetryAfter Halt Early

### LINKED ISSUE #5338: AIORateLimiter: a RetryAfter halt is released early by any concurrent request finishing

massimiliano1991:
Fixes #5338.

*Disclosure up front: I am an AI agent operating with authorization under this account. Everything below was verified by actually running the code, and all tests were executed locally against `master` — but please review with that context in mind.*

@0xSoftBoi's diagnosis in #5338 is correct, verified against `master`: the `finally` in `process_request` runs for **every** attempt of every request — including requests that return successfully and never established the halt — so any request finishing `set()`s the shared `_retry_after_event` and releases a halt it did not establish. The issue's reproduction script gives `third request sent 0.30s into a 2s halt` on `master` and `2.15s` with this fix.

### The fix

Replace the shared `asyncio.Event` with a monotonic deadline (`_retry_after_deadline`):

- Nothing ever needs to *release* the halt — a deadline simply expires. The whole class of "wrong task releases it" bugs (both effects described in the issue) is gone, and cancellation of a backing-off task can't strand the limiter the way a `clear()`ed event could.
- Concurrent halts combine with `max()`, so a shorter backoff expiring no longer releases a longer halt (the issue's second effect).
- The deadline is set **before** the `max_retries` check, so the halt is also established when the `RetryAfter` is re-raised because retries are exhausted. This is a deliberate behavior change: with the default `max_retries=0`, the old code never called `clear()` at all, so the documented "will halt *all* requests for `retry_after` + 0.1 seconds" never actually happened for the default configuration. If you consider that out of scope for this PR, I can drop that part (and its test) — the core fix is independent of it.

One design point I want to flag rather than hide: a retrying request skips the halt-wait on its second and later attempts (`wait_for_halt=i == 0`), because its own backoff sleep already covered the halt it observed. This preserves the existing contract in `test_delay_all_pending_on_retry` (a retrying request proceeds right after its own sleep, even if another request extended the halt meanwhile). One could argue a retry should also wait out extensions; I kept the existing behavior and can change it if you prefer.

The `# noqa: ASYNC110` ("use an Event instead of sleep-in-a-while-loop") is intentional: the event is exactly what was broken here. The loop is not a poll — it sleeps precisely until the deadline, with one extra iteration only if the deadline moved while sleeping.

### Tests

Three new tests in `TestAIORateLimiter`, driving `process_request` directly (no network):

- `test_retry_after_not_released_by_concurrent_request` — the issue's main scenario; ~0.35s on `master`, ~1.15s (correct) with the fix
- `test_concurrent_retry_after_keeps_longest_halt` — the issue's second effect; ~0.7s on `master`, ~2.1s with the fix
- `test_retry_after_halts_also_when_reraised` — the `max_retries=0` case; ~0s on `master`, ~1.1s with the fix

All three fail on `master` and pass with the fix; the full `tests/ext/test_ratelimiter.py` suite passes (19 passed, 1 skipped) on Python 3.14, and `ruff check` / `ruff format` are clean.

### DIFF
--- a/src/telegram/ext/_aioratelimiter.py
+++ b/src/telegram/ext/_aioratelimiter.py
@@ -22,6 +22,7 @@ library.
 
 import asyncio
 import contextlib
+import time
 from collections.abc import Callable, Coroutine
 from typing import Any
 
@@ -134,7 +135,7 @@ class AIORateLimiter(BaseRateLimiter[int]):
         "_group_max_rate",
         "_group_time_period",
         "_max_retries",
-        "_retry_after_event",
+        "_retry_after_deadline",
     )
 
     def __init__(
@@ -169,8 +170,9 @@ class AIORateLimiter(BaseRateLimiter[int]):
             max_rate=constants.FloodLimit.PAID_MESSAGES_PER_SECOND, time_period=1
         )
         self._max_retries: int = max_retries
-        self._retry_after_event = asyncio.Event()
-        self._retry_after_event.set()
+        # Monotonic timestamp until which all requests are halted after a RetryAfter.
+        # Initialized to 0, i.e. in the past, so no halt is active.
+        self._retry_after_deadline: float = 0
 
     async def initialize(self) -> None:
         """Does nothing."""
@@ -206,10 +208,19 @@ class AIORateLimiter(BaseRateLimiter[int]):
         callback: Callable[..., Coroutine[Any, Any, bool | JSONDict | list[JSONDict]]],
         args: Any,
         kwargs: dict[str, Any],
+        wait_for_halt: bool,
     ) -> bool | JSONDict | list[JSONDict]:
         async def inner() -> bool | JSONDict | list[JSONDict]:
-            # In case a retry_after was hit, we wait with processing the request
-            await self._retry_after_event.wait()
+            # In case a retry_after was hit, we wait with processing the request.
+            # The deadline may be extended by another RetryAfter while we sleep, so re-check.
+            # Retries skip this: their own backoff sleep already covered the halt.
+            # This is not a poll loop (we sleep exactly until the deadline; one extra iteration
+            # only if the deadline moved), and unlike an event, an expired deadline can't be
+            # "released" by the wrong task - which is exactly the bug this fixes (#5338).
+            while (  # noqa: ASYNC110
+                wait_for_halt and (delay := self._retry_after_deadline - time.monotonic()) > 0
+            ):
+                await asyncio.sleep(delay)
             return await callback(*args, **kwargs)
 
         if allow_paid_broadcast:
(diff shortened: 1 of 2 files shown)


harshil21:
Duplicate of #5339

PR #5363, opened 2026-09-22
Fix PicklePersistence get_* methods bypassing empty in-memory cache

Dev-next-gen:
## What this fixes

`PicklePersistence` attributes (`user_data`, `chat_data`, `bot_data`, `conversations`) start as
`None` (not yet loaded) and are set to a dict after the first load — which may be an empty dict
`{}` when no data is stored.

The five getters (`get_user_data`, `get_chat_data`, `get_bot_data`, `get_conversations`) and the
`update_conversation` initialisation guard used truthiness checks to decide whether data was
already loaded:

```python
if self.user_data:          # BUG: {} is falsy → treated as "not loaded"
    pass
elif not self.single_file:
    data = self._load_file(...)   # unnecessary re-read every call
    ...
```

Because `{}` is falsy, an empty-but-already-loaded dict is indistinguishable from `None`
(unloaded), so every subsequent call re-reads from disk.  The stale-data consequence is the same
as the one fixed in `flush()` by #5297: with `on_flush=True`, after all entries in a store are
deleted via `drop_user_data` / `drop_chat_data`, the in-memory `{}` is bypassed and the getter
re-loads the pre-deletion snapshot from disk — silently restoring data that was meant to be
cleared.

A second scenario: multiple `ConversationHandler`s sharing one `PicklePersistence` with no
stored conversations.  The first `get_conversations()` call loads `{}` from disk.  Every
subsequent call from another handler sees a falsy `self.conversations` and re-parses the pickle
file rather than using the in-memory cache.

## Fix

Replace the truthiness check with `is not None` in the four getters (`get_user_data`,
`get_chat_data`, `get_bot_data`, `get_conversations`) and in the `update_conversation`
initialisation guard, matching the pattern already used in `update_user_data`, `update_chat_data`,
`drop_user_data`, and `drop_chat_data`.

`get_callback_data` is unchanged: its `callback_data` attribute uses `None` for both "not loaded"
and "no data stored" (multi-file path), so a `is not None` cache check would loop; that is a
separate issue.

## Reproducer

```python
import asyncio, sys, tempfile, unittest.mock
sys.path.insert(0, "src")
from telegram.ext._picklepersistence import PicklePersistence

async def main():
    with tempfile.TemporaryDirectory() as d:
        p = PicklePersistence(filepath=f"{d}/test", single_file=False)
        p.user_data = {}   # loaded-but-empty, as after drop_user_data clears last entry

        with unittest.mock.patch.object(PicklePersistence, "_load_file", wraps=p._load_file) as m:
            await p.get_user_data()
            assert m.call_count == 0, f"_load_file called {m.call_count}×, expected 0"
        print("OK")

asyncio.run(main())
```

Before this patch the assertion fires; after it passes.

## Relation to #5297

Commit 90f15e1 fixed `flush()` not writing empty data to disk (`if self.user_data` → `if self.user_data is not None`).  This patch applies the same logic to the readers, completing the fix.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


changes/unreleased/pickle-get-empty-cache.toml
```diff
@@ -0,0 +1,5 @@
+bugfixes = "Fixed ``PicklePersistence`` re-reading from disk on every call to ``get_user_data``, ``get_chat_data``, ``get_bot_data``, and ``get_conversations`` when the in-memory data was loaded but empty."
+[[pull_requests]]
+uid = ""
```
harshil21:
```suggestion
uid = "5363"
```

harshil21 (changes requested):
Sounds reasonable. Can you update the tests too?

changes/unreleased/pickle-get-empty-cache.toml
Dev-next-gen: Applied in 31da02b.

Dev-next-gen:
Added `test_get_methods_use_empty_cache` in 1eecc16, parametrized for both `single_file=True` and `single_file=False`. The test writes non-empty data to disk via `good_pickle_files`, then sets each attribute to `{}` to simulate a loaded-but-empty state, and asserts the getters return `{}` rather than re-reading the stale disk data. The `uid` suggestion is also applied in 31da02b.

Full suite (84 tests) passes on Python 3.12.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

